PR #19, opened 2025-12-20
Add Rust extension for high-performance block allocation with O(1) op…

ericcurtin:
…erations

Falling back to Python implementation when unavailable.


reneleonhardt:
Python 3.15 is around the corner, PyO3 0.29.2 contains many features and fixes since 0.24 (Mar 9, 2025):
https://github.com/PyO3/pyo3/releases

PR #25, opened 2025-12-21
Add per-request sampling parameters support

WindChimeRan:
Add per-request sampling parameters support

  Description:

  Summary

  - Add support for per-request sampling parameters (temperature, top_k, top_p, penalties)
  - Integrate vLLM's Sampler class directly instead of reimplementing sampling logic

  Approach

  Converts MLX logits → torch tensors → vLLM's Sampler, reusing SamplingMetadata pattern.

Reference: 
* https://github.com/vllm-project/vllm/blob/main/vllm/v1/sample/tpu/metadata.py
* https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/gpu_input_batch.py#L755

  Pros:
  - Per-request params in batched decode (matches vLLM behavior)
  - No need to maintain separate sampling logic
  - Future vLLM sampling features come for free

  Trade-off:
  - Uses mlx_to_torch conversion instead of native MLX sampling (mlx-lm's make_sampler)
  - Open to discussion: is this acceptable, or should we implement MLX-native sampling?

  Changes

  - vllm_metal/v1/model_runner.py: Updated _prefill_single, _batched_decode, _sequential_decode
  - tests/test_sampling.py: Added integration tests

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -316,12 +316,13 @@ class MetalModelRunner:
                     gen.manual_seed(sp.seed)
                 generators[i] = gen
 
-        # top_k: pass None if all values are -1 (no filtering)
+        # top_k: pass None if all values indicate no filtering
+        # -1 = vLLM default (no filtering), 0 = OpenAI API convention (no filtering)
         # vLLM's sampler expects None to skip top-k entirely
         top_k_values = [sp.top_k for sp in sampling_params_list]
         top_k = (
             None
-            if all(k == -1 for k in top_k_values)
+            if all(k <= 0 for k in top_k_values)
             else torch.tensor(top_k_values, dtype=torch.int32)
         )
 


vllm_metal/v1/model_runner.py
```diff
@@ -379,8 +492,16 @@ def _sequential_decode(
...
             mx.eval(logits)
 
-            next_token_logits = logits[:, -1, :]
-            next_token = int(mx.argmax(next_token_logits, axis=-1)[0].item())
+            # Sample using vLLM's Sampler with request's params
+            # Cast to float32 for numpy conversion (numpy doesn't support bfloat16)
+            logits_torch = mlx_to_torch(
+                logits[:, -1, :].astype(mx.float32), device="cpu"
+            )
+            metadata = self._make_sampling_metadata(
```
ericcurtin, line 501: Will calling _make_sampling_metadata inside a loop be overhead heavy?

ericcurtin: I'm curious about this causing a synchronisation point between the GPU (Metal) and CPU. Also will mlx_to_torch trigger unnecessary copies?

WindChimeRan: Ideally _make_sampling_metadata should be on demand: 

https://github.com/vllm-project/vllm/blob/7b926e890195f026b60f36506503a80afc583b33/vllm/v1/worker/gpu_model_runner.py#L754-L762

We'll also need a whole new class input batch (instead of the current plain tensor), like this one: https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/gpu_input_batch.py#L82

I think mlx_to_torch is ok for a temporary solution - the conversion is only for sampling final logits[:, -1, :], which is small.

WindChimeRan: It's also plausible to contribute to vllm upstream, by adding some mlx backend support, similar to the TPU: https://github.com/vllm-project/vllm/blob/main/vllm/v1/sample/tpu/metadata.py

not sure if this is feasible or an overkill.

Issue #26, opened 2025-12-22
[Feature] Benchmark the latency/throughput against llamacpp-metal


WindChimeRan:
see: https://github.com/ggml-org/llama.cpp?tab=readme-ov-file#llama-bench

This will help user choose vllm-metal or llamacpp. I believe vllm-metal would be uniquely effective in high concurrency scenarios and offline scenarios.

ericcurtin:
@WindChimeRan I think we have the scripts to collect the benchmarks, it's probably more about running them and tracking the data, rather than the scripts themselves

WindChimeRan:
`vllm bench throughput --model Qwen/Qwen3-0.6B --input-len 128 --output-len 128 --num-prompts 4`


```bash
(EngineCore_DP0 pid=8459)     raise self._exception
(EngineCore_DP0 pid=8459)   File "/Users/ran/workspace/vllm-metal/.venv-vllm-metal/lib/python3.12/site-packages/vllm/v1/executor/uniproc_executor.py", line 79, in collective_rpc
(EngineCore_DP0 pid=8459)     result = run_method(self.driver_worker, method, args, kwargs)
(EngineCore_DP0 pid=8459)              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=8459)   File "/Users/ran/workspace/vllm-metal/.venv-vllm-metal/lib/python3.12/site-packages/vllm/v1/serial_utils.py", line 461, in run_method
(EngineCore_DP0 pid=8459)     return func(*args, **kwargs)
(EngineCore_DP0 pid=8459)            ^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=8459)   File "/Users/ran/workspace/vllm-metal/.venv-vllm-metal/lib/python3.12/site-packages/vllm/v1/worker/worker_base.py", line 369, in execute_model
(EngineCore_DP0 pid=8459)     return self.worker.execute_model(scheduler_output, *args, **kwargs)
(EngineCore_DP0 pid=8459)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #27, opened 2025-12-23
Fix bfloat16 MLX→PyTorch tensor conversion

WindChimeRan:
## Problem

  np.array() fails on MLX bfloat16 arrays because numpy doesn't support bfloat16:
  RuntimeError: Item size 2 for PEP 3118 buffer format string B does not match the dtype B item size 1

   ##  Solution

  Use torch.frombuffer(memoryview(array), dtype=...) for zero-copy conversion, bypassing numpy entirely.

  Reference: https://github.com/ml-explore/mlx/issues/403

   ##  Changes

  - vllm_metal/pytorch_backend/tensor_bridge.py: Use memoryview path for all dtypes
  - tests/test_tensor_bridge.py: Add bfloat16 test coverage

## Impact

`vllm bench throughput --model Qwen/Qwen3-0.6B --input-len 128 --output-len 128 --num-prompts 4` now is runnable.

Before this patch,  had runtime error see https://github.com/vllm-project/vllm-metal/issues/26#issuecomment-3684958313

## Discussion

now the vllm bench is runnable. Should we include this into the README as one of the contributors guide (run benchmark before making PR to compare performance)?

### DIFF
--- a/tests/test_tensor_bridge.py
+++ b/tests/test_tensor_bridge.py
@@ -100,6 +100,21 @@ class TestTensorConversion:
         assert torch_tensor.dtype == torch.int32
         np.testing.assert_array_equal(torch_tensor.numpy(), np.array(mlx_array))
 
+    def test_mlx_to_torch_bfloat16(self) -> None:
+        """Test MLX to PyTorch conversion for bfloat16."""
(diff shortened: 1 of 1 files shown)


WindChimeRan:
updated the PR, to use `vllm bench throughput ` natively

PR #32, opened 2025-12-25
Fix: Support MoE models & Mamba Cache logic on Metal

wanghaohao994:
## Summary
This PR introduces support for **Mixture-of-Experts (MoE)** models (e.g., Qwen-MoE) on Apple Silicon by implementing missing logic for `MambaCache` handling in the Metal runner.

## Key Changes
1.  **Mamba Cache Support (`model_runner.py`)**:
    -   Refactored `ModelRunner` to correctly distinguish between Attention layers and Mamba layers.
    -   Previously, the runner treated all layers as standard KV-cache, which caused `AttributeError: 'MambaCache' object has no attribute ...` when running Hybrid/MoE models.
    -   Added logic to process Mamba state layers correctly.

2.  **Installation Fix (`install.sh`)**:
    -   Removed `set -u` to prevent installation failures on certain shell environments (where `BASH_SOURCE` might be empty).

## ⚠️ Upstream Compatibility (Action Required for Users)

To fully enable MoE models on Metal, **two upstream files in the `vllm` library** (not this repo) currently require patching due to device capability detection issues on Apple Silicon.

I am documenting them here so the community and maintainers are aware of the root cause:

### 1. Device Capability Issue (`vllm/platforms/interface.py`)
**The Issue:** On Metal devices, `current_capability` can return a `tuple` or `None`, but the upstream code expects an object with a `.to_int()` method.
**The Fix:** Modify the code to handle tuple/None types gracefully.

```python
# In vllm/platforms/interface.py

# Original (Crash):
# return current_capability.to_int()

# Fixed (Workaround):
if isinstance(current_capability, tuple):
    # Handle tuple (major, minor) -> major * 10 + minor
    return current_capability[0] * 10 + current_capability[1]
elif current_capability is None:
    return -1
else:
    return current_capability.to_int()

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -78,7 +78,7 @@ download_and_install_wheel() {
 }
 
 main() {
-  set -eu -o pipefail
+  set -e -o pipefail
 
   local repo_owner="vllm-project"
   local repo_name="vllm-metal"
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -1,11 +1,17 @@
+#!/usr/bin/env python3
 # SPDX-License-Identifier: Apache-2.0
-"""Metal Model Runner for vLLM v1 engine.
+"""Metal Model Runner for vLLM v1 engine (patched for hybrid caches).
 
 Optimized for performance with:
 - True batched decode using BatchKVCache for O(1) forward passes per batch
 - Async evaluation pipeline for pipelined computation
 - Pre-allocated input buffers to reduce allocation overhead
 - Rust-based token state management for efficient batch operations
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by WindChimeRan** [comment]
Why did you remove "Sampler with per-request params"?

File: install.sh [hunk 1]
```diff
@@ -78,7 +78,7 @@ download_and_install_wheel() {
 }
 
 main() {
-  set -eu -o pipefail
```
ericcurtin, line 81: Can we leave this script as is, the u is helpful.

ericcurtin: If we need to fix something, there should be a way of fixing without removing -u

PR #33, opened 2025-12-27
Fix/metal platform compatibility

catundercar:
# Fix Metal Platform Compatibility Issues

## Summary

This PR fixes several compatibility issues with the Metal platform implementation, addressing device mismatch errors and pin_memory problems.

### 1. Implement `is_pin_memory_available()` for Metal Platform
- **File**: `vllm_metal/platform.py`
- **Change**: Add `is_pin_memory_available()` method that returns `False`
- **Reason**: Prevents vLLM from using `pin_memory` on Metal/MLX, which causes issues because:
  - PyTorch's `pin_memory` is designed for CUDA, not CPU/MPS
  - Our architecture uses CPU tensors for MLX interop
  - Unified memory already provides fast CPU-GPU transfers without pinning
- **Benefit**: Replaces the need for monkey patching `penalties._convert_to_tensors`

### 2. Fix Device Mismatch in Sampling Metadata
- **File**: `vllm_metal/v1/model_runner.py`
- **Changes**:
  - Explicitly set `device=self.device` for all tensors (top_k, top_p, temperature, penalties)
  - Create `prompt_token_ids` tensor instead of `None` to satisfy vLLM's sampler validation
  - Update `mlx_to_torch()` calls to use `self.device` instead of hardcoded `"cpu"`
- **Fixes**: `RuntimeError: Expected all tensors to be on the same device`

### 3. Use Platform Device Selection in Worker
- **File**: `vllm_metal/v1/worker.py`
- **Change**: Replace hardcoded `torch.device("cpu")` with `MetalPlatform.get_torch_device(0)`
- **Benefit**: Properly supports MPS when available and ensures consistency with platform's device selection logic

## Testing

- [x] Tested on Apple Silicon with MPS available
- [x] Tested on Apple Silicon with CPU fallback

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -342,7 +342,7 @@ class MetalModelRunner:
         # Pass empty lists and set pin_memory=False to avoid device mismatch
         batch_size = len(sampling_params_list)
         vocab_size = self.model_args.get("vocab_size", 32000)
-        empty_prompt_lists = [[] for _ in range(batch_size)]
+        empty_prompt_lists: list[list[int]] = [[] for _ in range(batch_size)]
         prompt_token_ids = make_tensor_with_pad(
             empty_prompt_lists,
             pad=vocab_size,


vllm_metal/v1/worker.py
```diff
@@ -98,7 +99,9 @@ def init_device(self) -> None:
             logger.info(f"MLX device set to: {mx.default_device()}")
 
         # Set CPU as PyTorch device for interop
```
ericcurtin: If 101 and 102 are truly not needed we should remove the commented out code and in the comment write why MetalPlatform.get_torch_device(0) makes sense

catundercar: Thanks for the review, I've made the changes:

- Removed the commented-out `torch.device("cpu")` code
- Added a comment explaining that `MetalPlatform.get_torch_device(0)` is used to properly support MPS when available, ensuring consistency with the platform's device selection logic, while falling back to CPU if MPS is not available

This keeps the code clean and makes the device selection rationale clear.

PR #45, opened 2026-01-02
fix: set SDKROOT for macOS C++ builds

Leundai:
On macOS, CMake builds can fail with 'cstdlib file not found' errors when the SDK path isn't properly configured. This adds a setup_macos_sdk function that exports SDKROOT and CPLUS_INCLUDE_PATH before building vLLM.

The function is called automatically in install.sh to ensure C++ standard library headers are found during compilation.

---

Its possible that this might be a self-inflicted problem with my xcode paths being misconfigured but this might help with some issues I was having building this

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -116,6 +116,8 @@ main() {
     exit 1
   fi
 
+  setup_macos_sdk
+
   local venv="$HOME/.venv-vllm-metal"
   if [[ -n "$local_lib" && -f "$local_lib" ]]; then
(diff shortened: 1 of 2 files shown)


ericcurtin:
@Leundai the better fix is to ensure we get macOS vllm wheels published in the main vllm project for next release, that way we don't need to build at all.

Leundai:
closing and will look into getting that wheel published if it's not worked on yet

PR #63, opened 2026-01-07
Add VLM support with mlx-vlm integration

ericcurtin:
vLLM Metal now supports vision-language models like Qwen2-VL using
mlx-vlm. The MetalModelRunner class has been updated to detect VLM
models and load them with the appropriate processor and configuration.
The generate method now accepts an optional images parameter for
multimodal input. The server has been enhanced to handle multimodal
messages in the OpenAI-compatible API, extracting image URLs from
content parts. Documentation has been added to the README with usage
examples for VLMs.

Signed-off-by: Eric Curtin <eric.curtin@docker.com>

### DIFF
--- a/vllm_metal/model_runner.py
+++ b/vllm_metal/model_runner.py
@@ -4,11 +4,19 @@
 import logging
 from typing import TYPE_CHECKING, Any
 
+# Third-party imports
 import mlx.core as mx
 from mlx_lm import load as mlx_load
 from mlx_lm import stream_generate
 from mlx_lm.sample_utils import make_sampler
 
+# mlx-vlm for vision-language models
+from mlx_vlm import generate as mlx_vlm_generate
+from mlx_vlm import load as mlx_vlm_load
+from mlx_vlm.prompt_utils import apply_chat_template
+from mlx_vlm.utils import load_config as mlx_vlm_load_config
+
+# First-party imports
 from vllm_metal.config import get_config
 from vllm_metal.mlx_backend.cache import PagedKVCache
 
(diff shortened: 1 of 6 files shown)


vllm_metal/model_runner.py
```diff
@@ -4,11 +4,19 @@
...
 import mlx.core as mx
 from mlx_lm import load as mlx_load
 from mlx_lm import stream_generate
 from mlx_lm.sample_utils import make_sampler
 
+# mlx-vlm for vision-language models
+from mlx_vlm import generate as mlx_vlm_generate
+from mlx_vlm import load as mlx_vlm_load
+from mlx_vlm.prompt_utils import apply_chat_template
+from mlx_vlm.utils import load_config as mlx_vlm_load_config
```
LxYuan0420: Note: given we are also adding PEP508 gating for MLX deps on non‑macOS #64 : it becomes more important to avoid top-level mlx_vlm importsin vllm_metal/*. 

With gating, non‑macOS installs will succeed (will skip installing mlx, mlx-lm, mlx-vlm because the PEP508 markers evaluate false) but import vllm_metal.model_runner would still raise ImportError if mlx_vlm is imported at module import time. 

Lazy-importing mlx_vlm only inside the VLM code path keeps non‑macOS installs usable for tooling/CI while still failing clearly when VLM is actually requested.

```python
def _load_mlx_vlm():
    try:
        from mlx_vlm import generate as mlx_vlm_generate
        from mlx_vlm import load as mlx_vlm_load
        from mlx_vlm.prompt_utils import apply_chat_template
        from mlx_vlm.utils import load_config as mlx_vlm_load_config
        return mlx_vlm_generate, mlx_vlm_load, apply_chat_template, mlx_vlm_load_config
    except ImportError as e:
        raise RuntimeError(
            "VLM requested but mlx-vlm is not installed. Install vllm-metal on macOS arm64."
        ) from e

# and then call _load_mlx_vlm() only when is_vlm is true.
```

PR #66, opened 2026-01-08
Add Ray distributed inference with documentation

ericcurtin:
This commit adds comprehensive support for distributed inference using
Ray across multiple Apple Silicon Macs. A new documentation file
explains the Ray integration, including setup instructions, usage
examples, and troubleshooting guidance. The platform now automatically
detects Ray initialization and switches to the Ray executor backend. The
Ray resource key is set to CPU for proper scheduling on Apple Silicon
systems with unified memory. Multiprocessing is forced to use spawn
method to prevent Metal/MLX state corruption when forking processes. An
optional Ray dependency is added to pyproject.toml and the README is
updated with a reference to the new distributed inference documentation.

Signed-off-by: Eric Curtin <eric.curtin@docker.com>

### DIFF
--- a/docs/distributed-inference.md
+++ b/docs/distributed-inference.md
@@ -1,0 +1,249 @@
+# Distributed Inference with Ray
+
+vLLM-Metal can be used with [Ray](https://ray.io/) for scaling across multiple Apple Silicon Macs. This enables tensor parallelism and data parallelism for larger models and higher throughput.
+
+## Current Status
+
+> ⚠️ **HIGHLY EXPERIMENTAL**: Ray integration on macOS with Metal is experimental and has significant limitations. Manual configuration is required and success is not guaranteed. Known issues include:
+>
+> 1. **gRPC communication issues** - Upstream Ray bugs causing connection problems on macOS
+> 2. **IP address mismatch** - Ray and vLLM may use different node addresses, causing placement group scheduling to hang
+> 3. **Resource allocation conflicts** - Custom resource keys (like "METAL") don't appear in Ray's accelerator IDs, causing KeyError in vLLM's Ray executor
+> 4. **Upstream vLLM changes needed** - Proper Ray support for non-GPU platforms requires changes in vLLM's Ray executor to handle custom accelerator keys properly
+>
+> **Known Working Configurations**:
+> - Single-node Ray cluster with explicit IP address configuration (see troubleshooting)
+> - Simple model serving with minimal parallelism
+> - Ray configured with custom "METAL" resource: `--resources='{"METAL": 1}'`
+>
+> **Not Yet Supported**:
+> - Multi-node clusters across different machines
+> - Advanced tensor/data parallelism configurations
+> - Production deployments
+> - Automatic resource detection and allocation
+>
+> The integration code is provided as-is for testing purposes. See troubleshooting section for manual configuration steps.
+>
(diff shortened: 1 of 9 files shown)


docs/distributed-inference.md
```diff
@@ -0,0 +1,171 @@
...
+
+```bash
+# Start Ray cluster first
+ray start --head
+
+# Serve with Ray executor
+vllm serve mlx-community/Llama-3.2-1B-Instruct-4bit \
+    --distributed-executor-backend ray \
+    --max-model-len 4096
+```
```
LxYuan0420, line 116: The first issue I hit:  Ray/vLLM node IP mismatch and so placement group never schedules

some reproducible code:

```bash
ray stop -f
ray start --head --port=6379 --disable-usage-stats
vllm serve HuggingFaceTB/SmolLM2-135M --port 8000 --distributed-executor-backend ray

# Ray shows the node is 127.0.0.1:

python -c "import ray; ray.init(address='auto'); print([n['NodeManagerAddress'] for n in ray.nodes() if n.get('Alive')])"
# ['127.0.0.1']

# But vLLM requests a placement group pinned to node:192.168.1.10 (from logs):

Error: No available node types can fulfill resource request {'CPU': 1.0, 'node:192.168.1.10': 0.001}
```

Workaround is to standardise the IPs:
```bash
ray stop -f
ray start --head --node-ip-address=127.0.0.1 --port=6379 --disable-usage-stats
VLLM_HOST_IP=127.0.0.1 vllm serve HuggingFaceTB/SmolLM2-135M --port 8000 --distributed-executor-backend ray

```

vllm_metal/platform.py
```diff
@@ -53,6 +54,7 @@ class MetalPlatform(Platform):
     device_name: str = "cpu"  # PyTorch device name (use CPU for compatibility)
     device_type: str = "cpu"  # PyTorch device type (use CPU for compatibility)
     dispatch_key: str = "CPU"  # PyTorch dispatch key
+    ray_device_key: str = "CPU"  # Ray resource key for scheduling
```
LxYuan0420: Second issue I hit: ray_device_key="CPU" breaks Ray actor creation (resources vs num_cpus)


vLLM’s Ray executor uses this in resources={...} for non-GPU platforms:
```python
# vllm/v1/executor/ray_executor.py (vllm side)
ray.remote(num_cpus=0, num_gpus=0, resources={current_platform.ray_device_key: num_gpus})(...)
```

That leads to the runtime error:
```bash
ValueError: Use the 'num_cpus' and 'num_gpus' keyword instead of 'CPU' and 'GPU' in 'resources' keyword
```

I tried switching to a custom resource key (e.g. METAL) and advertising it in Ray:

```python
# vllm_metal/platform.py
ray_device_key: str = "METAL"
```

```bash
ray stop -f
ray start --head --node-ip-address=127.0.0.1 --port=6379 --disable-usage-stats --resources='{"METAL": 1}'
```

but then vLLM fails later with:

```bash
KeyError: 'METAL'
```

because it expects ray.get_runtime_context().get_accelerator_ids()[device_key], and custom resources don’t show up as accelerator IDs.

I only have a single Mac so I can’t validate the multi‑Mac flow, but based on the above it looks like making Ray work for Metal likely needs an upstream vLLM change (Ray executor shouldn’t assume accelerator IDs for non‑GPU platforms), not just setting ray_device_key.

ericcurtin:
I'm gonna let @IntuitIntelLLC look at this one. Feel free to take anything from here if you'd like @IntuitIntelLLC

PR #74, opened 2026-01-15
Fix api server should not require vllm dependencies

Alex-ai-future:
## Summary

Refactor vLLM-metal to use an independent `utils.py` module, enabling cleaner code organization and supporting independent operation without vLLM runtime dependencies.

## Background

vLLM-metal is designed as a hardware plugin for vLLM, unifying MLX and PyTorch under a single lowering path for Apple Silicon. During development, it was discovered that vLLM-metal can operate completely independently for deployment services without requiring vLLM as a runtime dependency. The package functions as a standalone MLX-based inference engine for model loading and inference, reducing deployment footprint and unnecessary dependencies.

## Changes

- Extract `set_wired_limit` function from `vllm_metal/platform.py` to new `vllm_metal/utils.py` module
- Update import in `vllm_metal/v1/worker.py` to use the new utils module
- Improve code organization for better maintainability and modularity

### DIFF
--- a/vllm_metal/v1/worker.py
+++ b/vllm_metal/v1/worker.py
@@ -22,14 +22,13 @@ from vllm.v1.kv_cache_interface import KVCacheConfig, KVCacheSpec
 from vllm.v1.outputs import ModelRunnerOutput
 from vllm.v1.worker.worker_base import WorkerBase
 
-from vllm_metal.platform import MetalPlatform
-from vllm_metal.utils import set_wired_limit
 from vllm_metal.config import (
     AUTO_MEMORY_MIN_BLOCKS_BUFFER_FACTOR,
     AUTO_MEMORY_OVERHEAD_FACTOR,
     get_config,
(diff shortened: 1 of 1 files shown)


vllm_metal/model_runner.py
```diff
@@ -11,7 +11,7 @@
 
 from vllm_metal.config import get_config
 from vllm_metal.mlx_backend.cache import PagedKVCache
-from vllm_metal.platform import set_wired_limit
+from vllm_metal.utils import set_wired_limit
```
LxYuan0420, line 14: Can we add a regression test (subprocess preferred) that import vllm_metal.server / import vllm_metal.model_runner does not import vllm as a side effect (e.g., 'vllm' not in sys.modules in a clean process), to prove the stated goal of this PR?

Alex-ai-future: Thanks for your advice.

PR #75, opened 2026-01-19
Update installer platform checks and uv minimum

LxYuan0420:
This PR is:
- To prevent installs from running on non-Apple Silicon machines
- To enforce a minimum uv version while avoiding automatic system upgrades

Context
This addresses the installer mis-gating issue where Linux ARM64 could pass the arch check and fail later in MLX/Metal setup, plus the lack of a uv version guard that caused confusing toolchain failures. The changes tighten the platform check to OS+arch, add an explicit fail-fast in install.sh, and require uv >= 0.9.18 with a clear manual upgrade message.

How to verify
- ./install.sh on non-Darwin/arm64 should exit immediately with the new error.
- bash -lc 'source scripts/lib.sh; ensure_uv' should fail when uv < 0.9.18, then pass after upgrading uv.

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -24,7 +24,10 @@ is_apple_silicon() {
 # Compare dotted versions (return 0 if $1 >= $2)
 version_ge() {
   local IFS=.
-  local -a v1=($1) v2=($2)
+  local -a v1=()
+  local -a v2=()
+  read -r -a v1 <<< "$1"
+  read -r -a v2 <<< "$2"
   local i
 
   for ((i=${#v1[@]}; i<${#v2[@]}; i++)); do


scripts/lib.sh
```diff
@@ -18,21 +18,64 @@ section() {
 
 # Check if running on Apple Silicon
 is_apple_silicon() {
-  [ "$(uname -m)" = "arm64" ]
+  [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "arm64" ]
```
ericcurtin, line 21: I don't think we need any code from this PR in general. Funnily enough:

```
[ "$(uname -m)" = "arm64" ]
```

basically guarantees we are on Apple Silicon, it's called aarch64 on Linux

PR #80, opened 2026-01-26
[Feature] add content hash prefix caching for KV cache reuse

ricky-chaoju:
## Summary
Implements content hash based prefix caching to reuse computed KV cache states across requests with identical prompts. This avoids redundant prefill computation for shared prefixes like system prompts.

### Changes

**vllm_metal/v1/model_runner.py:**
- Add `PrefixCacheManager` class for KV state caching with content hash lookup
- Cache full prompt KV states and restore on subsequent identical requests
- Add periodic stats logging for hit rate monitoring

### Test Results

Sequential requests with identical prompt:
```
Request 1: MISS (0 hits, 1 misses, rate=0.0%)   <- First request, cache stored
Request 2: HIT  (1 hits, 1 misses, rate=50.0%)  <- Cache reused
Request 3: HIT  (2 hits, 1 misses, rate=66.7%)
Request 4: HIT  (3 hits, 1 misses, rate=75.0%)
Request 5: HIT  (4 hits, 1 misses, rate=80.0%)
```

### DIFF
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -1,0 +1,92 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+import mlx.core as mx
+
+import vllm_metal.v1.model_runner as mr
+
+
+class StubMambaCache:
+    @property
+    def state(self):
+        return []
+
+
+class TestPrefixCacheHybridGuard:
+    def _make_runner(self) -> mr.MetalModelRunner:
+        runner = mr.MetalModelRunner.__new__(mr.MetalModelRunner)
+        runner.model = MagicMock()
+        runner._is_vlm = False
+        runner._prefix_cache = mr.PrefixCacheManager()
+        return runner
+
+    def test_hybrid_model_skips_prefix_cache(self, monkeypatch) -> None:
+        lookup_spy = MagicMock(return_value=None)
+        insert_spy = MagicMock()
+
+        def fake_make_prompt_cache(model):
+            kv = mr.KVCache()
+            kv.keys = mx.zeros((1, 8, 0, 64))
+            kv.values = mx.zeros((1, 8, 0, 64))
+            kv.offset = 0
+            return [kv, StubMambaCache(), kv]
+
+        monkeypatch.setattr(mr, "make_prompt_cache", fake_make_prompt_cache)
+
+        runner = self._make_runner()
+        monkeypatch.setattr(runner._prefix_cache, "lookup", lookup_spy)
+        monkeypatch.setattr(runner._prefix_cache, "insert", insert_spy)
+
+        fake_logits = mx.zeros((1, 5, 100))
+        runner.model.return_value = MagicMock(logits=fake_logits)
+
+        token_ids = [1, 2, 3, 4, 5]
+        sampling_params = MagicMock(
+            temperature=0.0,
+            top_p=1.0,
+            top_k=0,
+            frequency_penalty=0,
+            presence_penalty=0,
+            repetition_penalty=1.0,
+        )
+
+        runner._prefill_single("req-1", token_ids, sampling_params)
+
+        lookup_spy.assert_not_called()
+        insert_spy.assert_not_called()
+
+    def test_pure_kvcache_uses_prefix_cache(self, monkeypatch) -> None:
+        lookup_spy = MagicMock(return_value=None)
+        insert_spy = MagicMock()
+
+        def fake_make_prompt_cache(model):
+            kv = mr.KVCache()
+            kv.state = [mx.zeros((1, 4, 8, 64)), mx.zeros((1, 4, 8, 64))]
+            return [kv, kv]
+
+        monkeypatch.setattr(mr, "make_prompt_cache", fake_make_prompt_cache)
+
+        runner = self._make_runner()
+        monkeypatch.setattr(runner._prefix_cache, "lookup", lookup_spy)
+        monkeypatch.setattr(runner._prefix_cache, "insert", insert_spy)
+
+        fake_logits = mx.zeros((1, 1, 100))
+        runner.model.return_value = MagicMock(logits=fake_logits)
+
+        token_ids = [1, 2, 3, 4, 5]
+        sampling_params = MagicMock(
+            temperature=0.0,
+            top_p=1.0,
+            top_k=0,
+            frequency_penalty=0,
+            presence_penalty=0,
+            repetition_penalty=1.0,
+        )
+
+        runner._prefill_single("req-1", token_ids, sampling_params)
+
+        lookup_spy.assert_called_once_with([1, 2, 3, 4])
+        insert_spy.assert_called_once()


vllm_metal/v1/model_runner.py
```diff
@@ -610,18 +724,43 @@ def _prefill_single(
...
+                self.model.language_model
+                if self._is_vlm and hasattr(self.model, "language_model")
+                else self.model
+            )
+            cache = make_prompt_cache(cache_model)
+
+        # Prefill: process tokens (or remaining tokens if prefix was cached)
+        if cached_prefix_len > 0 and cached_prefix_len == len(token_ids):
+            input_ids = mx.array([token_ids[-1:]], dtype=mx.int32)
+            model_output = self.model(input_ids, cache=cache)
```
LxYuan0420: A full‑prefix hit should *not* reprocess the last token from a cache that already includes it.
Doing so makes the effective sequence A B C C.

Correct:
  cache used:   A B
  input token:  C
  cache after:  A B C
  logits for:   next token after C (D)

Incorrect:
  cache used:   A B C
  input token:  C
  cache after:  A B C C
  logits for:   next token after “A B C C”

Thats why vLLM caps cache‑hit length to prompt_len − 1, so we feed the last token once and keep the sequence correct.

And the pseudo:
```python
prompt = [A, B, C]

self.model(input=[C], cache=[A,B,C])  # wrong, duplicates C

#vLLM‑style:

self.model(input=[C], cache=[A,B])  # correct, predicts next token after C
```
This matches vLLM v1 cache logic.

vllm_metal/v1/model_runner.py
```diff
@@ -64,6 +66,113 @@
...
+                if c is min_ref:
+                    del self._cache[h]
+                    break
+
+        prefix_hash = _compute_prefix_hash(token_ids)
+        cache_state = []
+        for layer_cache in cache:
+            k = layer_cache.state[0]
+            v = layer_cache.state[1]
+            cache_state.append((mx.array(k), mx.array(v)))
```
LxYuan0420: Prefix caching assumes KVCache.state exists. Hybrid/Mamba caches use a different structure, so this will break or corrupt state. Please guard to KV cache types only or disable prefix cache for non‑KV caches.

vllm_metal/v1/model_runner.py
```diff
@@ -64,6 +66,113 @@
...
+# Prefix cache configuration
+_PREFIX_CACHE_ENABLED = True
+_PREFIX_CACHE_MAX_ENTRIES = 100
+
+
+def _compute_prefix_hash(token_ids: list[int]) -> bytes:
+    """Compute content hash for a token sequence."""
+    h = hashlib.sha256()
+    h.update(str(token_ids).encode())
+    return h.digest()
```
LxYuan0420, line 80: Hashing via str(token_ids) allocates a big string and is slow on long prompts. Consider a byte‑stable representation (e.g., array('I', token_ids).tobytes() or struct.pack) to reduce overhead.


im interested to see how others repo do this kind of prefix hash and do we have better way than tobytes()

vllm_metal/v1/model_runner.py
```diff
@@ -64,6 +67,130 @@
...
+        prefix_hash = _compute_prefix_hash(token_ids)
+        cache_state = []
+        for layer_cache in cache:
+            if isinstance(layer_cache, KVCache):
+                k = layer_cache.state[0]
+                v = layer_cache.state[1]
+                cache_state.append((mx.array(k), mx.array(v)))
+            else:
+                # MambaCache/ArraysCache - skip (not supported)
+                cache_state.append(None)
```
LxYuan0420, line 150: Prefix cache stores None for non‑KV layers, but cache hits also skip the prefix tokens. 

For Mamba/hybrid models that means SSM state is missing and logits diverge. Please disable prefix caching when any non‑KV cache types are present

we need to handle proper Mamba/Arrays snapshot/restore (including left_padding) in a separate PR.

vllm_metal/v1/model_runner.py
```diff
@@ -610,18 +742,41 @@ def _prefill_single(
...
             if self._is_vlm and hasattr(self.model, "language_model")
             else self.model
         )
-        cache = make_prompt_cache(cache_model)
 
-        # Prefill: process the entire prompt with cache
-        input_ids = mx.array([token_ids], dtype=mx.int32)
+        # Try to reuse cached prefix
+        if self._prefix_cache is not None and len(prefix) > 0:
+            cached = self._prefix_cache.lookup(prefix)
```
LxYuan0420, line 764: Minimal guard (in _prefill_single, near the prefix cache block):

```python
cache = make_prompt_cache(cache_model)
supports_prefix_cache = all(isinstance(c, KVCache) for c in cache)

if supports_prefix_cache and self._prefix_cache is not None and len(prefix) > 0:
    cached = self._prefix_cache.lookup(prefix)
    if cached is not None:
        cache = self._prefix_cache.restore_cache(cached, self.model, self._is_vlm)
        cached_prefix_len = len(cached.token_ids)
    else:
        prefix_ids = mx.array([prefix], dtype=mx.int32)
        _ = self.model(prefix_ids, cache=cache)
        self._prefix_cache.insert(prefix, cache)
        cached_prefix_len = len(prefix)
# else: fall through; cached_prefix_len stays 0, so we process full prompt
```

This disables prefix caching automatically for non‑KV caches while keeping normal behavior for KV‑only models.

ricky-chaoju:
> Prefix cache is safe only for pure KVCache models. For Mamba/hybrid, insert/restore drops non‑KV state, but cache hits skip the prefix tokens, so SSM state is missing and outputs diverge.
> 
> Please gate prefix caching to **KVCache‑only** models and handle proper Mamba/Arrays snapshot/restore (incl left_padding) in a separate PR.
> 
> minimal reproduce snippet:
> 
> ```python
> from mlx_lm import load
> from mlx_lm.models.cache import make_prompt_cache
> from vllm_metal.v1.model_runner import PrefixCacheManager
> 
> model, _ = load("mlx-community/mamba-130m-hf-f32")
> cache = make_prompt_cache(model)
> 
> mgr = PrefixCacheManager()
> mgr.insert([1, 2, 3], cache)
> cached = mgr.lookup([1, 2, 3])
> 
> print([c is None for c in cached.cache_state])  # all True
> ```

Thanks for catching this! Added the supports_prefix_cache guard as you suggested - prefix caching now only kicks in for pure KVCache models.

LxYuan0420 (changes requested):
Thanks for the fix. 

Before LGTM, please rebase on main so the non‑v1 fix from #81 is included, and add a minimal unit test that verifies prefix caching is disabled when make_prompt_cache yields any non‑KV cache type (e.g., stub MambaCache). That will keep the hybrid guard from regressing

tests/test_prefix_cache.py
```diff
@@ -0,0 +1,54 @@
...
+    pass
+
+
+class TestPrefixCacheHybridGuard:
+    """Tests for prefix caching guard with hybrid models."""
+
+    def test_pure_kvcache_enables_prefix_caching(self) -> None:
+        """Prefix caching should be enabled for pure KVCache models."""
+        # Simulate cache from a pure transformer model
+        cache = [KVCache(), KVCache(), KVCache()]
```
LxYuan0420: Could you update tests/test_prefix_cache.py to exercise the actual guard in _prefill_single instead of re‑implementing all(isinstance...)?

One option is to monkeypatch vllm_metal.v1.model_runner.make_prompt_cache to return a list containing a non‑KV cache, and patch PrefixCacheManager.lookup/insert to a spy so you can assert they are not called. This keeps the test on the production path without pulling in real MLX deps.

LxYuan0420 (changes requested):

tests/test_prefix_cache.py
ricky-chaoju: done

ricky-chaoju:
Thanks a lot, I learned a lot from this discussion.

PR #82, opened 2026-01-27
[Feature] memory-based eviction for prefix cache

ricky-chaoju:
## Summary

Changes prefix cache eviction from entry count to memory-based. Tracks actual memory usage via `array.nbytes` and evicts when approaching the limit.

Default limit is 5% of total memory, which is more predictable and conservative for unified memory systems like Apple Silicon.

### Changes

- Track cache memory usage with `array.nbytes`
- Evict based on memory limit instead of entry count
- Default to 5% of total memory (e.g., 9.6GB on 192GB machine)
- Add `VLLM_METAL_PREFIX_CACHE_FRACTION` env var for customization
- Add memory stats to periodic logging

### DIFF
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -144,41 +144,70 @@ class TestPrefixCacheEnableFlag:
 
     def test_enabled_when_env_set(self, monkeypatch) -> None:
         monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE", "1")
-        # Re-evaluate the module-level flag
-        assert "VLLM_METAL_PREFIX_CACHE" in mr.os.environ
+        assert mr._prefix_cache_enabled() is True
 
     def test_disabled_when_env_unset(self, monkeypatch) -> None:
         monkeypatch.delenv("VLLM_METAL_PREFIX_CACHE", raising=False)
-        assert "VLLM_METAL_PREFIX_CACHE" not in mr.os.environ
+        assert mr._prefix_cache_enabled() is False
+
+
+_TEN_GB = 10 * 1024**3
+
+
+def _mock_device_info():
+    return {"max_recommended_working_set_size": _TEN_GB}
 
 
 class TestPrefixCacheFractionParsing:
     def test_valid_fraction(self, monkeypatch) -> None:
         monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "0.1")
+        monkeypatch.setattr(mr.mx.metal, "device_info", _mock_device_info)
         result = mr._get_prefix_cache_max_bytes()
-        assert result > 0
+        assert result == int(_TEN_GB * 0.1)
 
-    def test_invalid_string(self, monkeypatch) -> None:
+    def test_default_fraction(self, monkeypatch) -> None:
+        monkeypatch.delenv("VLLM_METAL_PREFIX_CACHE_FRACTION", raising=False)
+        monkeypatch.setattr(mr.mx.metal, "device_info", _mock_device_info)
+        result = mr._get_prefix_cache_max_bytes()
+        assert result == int(_TEN_GB * mr._PREFIX_CACHE_DEFAULT_FRACTION)
+
+    def test_invalid_string_uses_default(self, monkeypatch) -> None:
         monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "abc")
+        monkeypatch.setattr(mr.mx.metal, "device_info", _mock_device_info)
         result = mr._get_prefix_cache_max_bytes()
-        assert result > 0
+        assert result == int(_TEN_GB * mr._PREFIX_CACHE_DEFAULT_FRACTION)
 
-    def test_out_of_range_zero(self, monkeypatch) -> None:
+    def test_out_of_range_zero_uses_default(self, monkeypatch) -> None:
         monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "0")
+        monkeypatch.setattr(mr.mx.metal, "device_info", _mock_device_info)
         result = mr._get_prefix_cache_max_bytes()
-        assert result > 0
+        assert result == int(_TEN_GB * mr._PREFIX_CACHE_DEFAULT_FRACTION)
 
-    def test_out_of_range_above_one(self, monkeypatch) -> None:
+    def test_out_of_range_above_one_uses_default(self, monkeypatch) -> None:
         monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "2")
+        monkeypatch.setattr(mr.mx.metal, "device_info", _mock_device_info)
         result = mr._get_prefix_cache_max_bytes()
-        assert result > 0
+        assert result == int(_TEN_GB * mr._PREFIX_CACHE_DEFAULT_FRACTION)
 
-    def test_nan(self, monkeypatch) -> None:
+    def test_nan_uses_default(self, monkeypatch) -> None:
         monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "nan")
+        monkeypatch.setattr(mr.mx.metal, "device_info", _mock_device_info)
         result = mr._get_prefix_cache_max_bytes()
-        assert result > 0
+        assert result == int(_TEN_GB * mr._PREFIX_CACHE_DEFAULT_FRACTION)
 
-    def test_inf(self, monkeypatch) -> None:
+    def test_inf_uses_default(self, monkeypatch) -> None:
         monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "inf")
+        monkeypatch.setattr(mr.mx.metal, "device_info", _mock_device_info)
         result = mr._get_prefix_cache_max_bytes()
-        assert result > 0
+        assert result == int(_TEN_GB * mr._PREFIX_CACHE_DEFAULT_FRACTION)
+
+    def test_device_info_fallback(self, monkeypatch) -> None:
+        monkeypatch.delenv("VLLM_METAL_PREFIX_CACHE_FRACTION", raising=False)
+        monkeypatch.setattr(
+            mr.mx.metal,
+            "device_info",
+            lambda: {},
+        )
+        result = mr._get_prefix_cache_max_bytes()
+        fallback = 8 * 1024**3
+        assert result == int(fallback * mr._PREFIX_CACHE_DEFAULT_FRACTION)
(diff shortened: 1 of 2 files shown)


vllm_metal/v1/model_runner.py
```diff
@@ -64,6 +66,171 @@
...
+    import os
+
+    import psutil
+
+    fraction_str = os.environ.get("VLLM_METAL_PREFIX_CACHE_FRACTION", "")
+    if fraction_str:
+        fraction = float(fraction_str)
+    else:
+        fraction = _PREFIX_CACHE_DEFAULT_FRACTION
+
```
LxYuan0420, line 109: VLLM_METAL_PREFIX_CACHE_FRACTION is parsed with float() and no validation. Invalid values will throw, and values <=0 or >1 create bad limits. 

Please add try/except + clamp to (0, 1] and fall back to default with a log.

vllm_metal/v1/model_runner.py
```diff
@@ -64,6 +66,171 @@
...
+        # Evict if needed
+        self._evict_until_fits(entry_bytes)
+
+        self._cache[prefix_hash] = CachedPrefix(
+            token_ids=list(token_ids),
+            cache_state=cache_state,
+            size_bytes=entry_bytes,
+            ref_count=1,
+        )
+        self._current_bytes += entry_bytes
```
LxYuan0420, line 245: ` _evict_until_fit`s can clear the cache but still insert an entry larger than _max_bytes, so the cap is violated. 

Suggest skipping caching when entry_bytes > _max_bytes (or adjust the cap with a warning)

vllm_metal/v1/model_runner.py
```diff
@@ -64,6 +66,171 @@
...
+    import psutil
+
+    fraction_str = os.environ.get("VLLM_METAL_PREFIX_CACHE_FRACTION", "")
+    if fraction_str:
+        fraction = float(fraction_str)
+    else:
+        fraction = _PREFIX_CACHE_DEFAULT_FRACTION
+
+    total = psutil.virtual_memory().total
+    max_bytes = int(total * fraction)
```
LxYuan0420: Using total system RAM can be too aggressive on low‑memory Macs. Consider basing the default on MLX `max_recommended_working_set_size` (or available memory), with env override

```python
import mlx.core as mx

info = mx.metal.device_info()
print("max_recommended_working_set_size:", info.get("max_recommended_working_set_size"))
print("device_info:", info)

```

LxYuan0420 (changes requested):
I reviewed and left a few pitfalls inline. Even after those are fixed, I still feel hesitant to land memory‑based eviction as the default because it changes behavior in a hard‑to‑predict way on low‑memory Macs and we do not yet have tests that validate eviction behavior under pressure. 

I would prefer making this opt‑in (via env) or keying the default off MLX’s recommended working set, then revisiting after we have coverage and some real‑world data like hit/eviction rates, memory pressure impact, and latency trade‑offs on 16–32GB machines

vllm_metal/v1/model_runner.py
```diff
@@ -65,6 +69,199 @@
 # Performance tuning
 _CACHE_CLEAR_INTERVAL = 50  # Clear cache every N finished requests
 
+# Prefix cache configuration
+_PREFIX_CACHE_ENABLED = True
```
LxYuan0420: This feature still has correctness + hybrid issues. Consider making it opt‑in via env/config until the prefix‑correctness and hybrid guard land and also please document `VLLM_METAL_PREFIX_CACHE_FRACTION` in README and add a small test for invalid/out‑of‑range values (e.g., “abc”, 0, 2) to ensure we fall back to the default

vllm_metal/v1/model_runner.py
```diff
@@ -65,6 +69,199 @@
...
+                fraction = _PREFIX_CACHE_DEFAULT_FRACTION
+        except ValueError:
+            logger.warning(
+                "Invalid VLLM_METAL_PREFIX_CACHE_FRACTION=%r, using default %.2f",
+                fraction_str,
+                _PREFIX_CACHE_DEFAULT_FRACTION,
+            )
+            fraction = _PREFIX_CACHE_DEFAULT_FRACTION
+    else:
+        fraction = _PREFIX_CACHE_DEFAULT_FRACTION
```
LxYuan0420, line 108: Fraction parsing still unsafe for nan/inf please add math.isfinite(fraction) and fallback to default on nan/inf

vllm_metal/v1/model_runner.py
```diff
@@ -611,18 +813,43 @@ def _prefill_single(
...
+        cached_prefix_len = 0
+
+        # Try to reuse cached prefix
+        if self._prefix_cache is not None:
+            cached = self._prefix_cache.lookup(token_ids)
+            if cached is not None:
+                cache = self._prefix_cache.restore_cache(
+                    cached, self.model, self._is_vlm
+                )
+                cached_prefix_len = len(cached.token_ids)
```
LxYuan0420, line 861: hmmm..I remember we fixed this issue in #80 : This caches by full prompt and on hit runs the last token again with a cache that already includes it

LxYuan0420 (changes requested):
This PR looks like it re-implements the prefix-cache behavior instead of layering on PR80, and it reintroduces two issues PR80 already solved (full-prompt hit re-feeds last token; missing KVCache-only guard). 

To keep this production-grade and avoid churn, can you please rebase this work on top of PR80 so this PR becomes eviction policy + config only? That should keep the diff minimal and preserve the correct prefix-caching semantics

Also need to add few more tests to ensure:
- eviction triggers under max_bytes
- fraction parsing handles nan/inf/out-of-range

tests/test_prefix_cache.py
```diff
@@ -90,3 +90,82 @@ def fake_make_prompt_cache(model):
...
+
+        mgr.insert([1, 2], [self._make_kv()])
+        assert mgr._current_bytes == bytes_after_first
+
+
+class TestPrefixCacheFractionParsing:
+    def test_valid_fraction(self, monkeypatch) -> None:
+        monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "0.1")
+        result = mr._get_prefix_cache_max_bytes()
+        assert result > 0
```
LxYuan0420: These tests cover fraction parsing only; they do not validate the enable flag behavior. The `VLLM_METAL_PREFIX_CACHE` bool parsing bug is elsewhere

vllm_metal/v1/model_runner.py
```diff
@@ -69,8 +71,48 @@
 _CACHE_CLEAR_INTERVAL = 50  # Clear cache every N finished requests
 
 # Prefix cache configuration
-_PREFIX_CACHE_ENABLED = True
-_PREFIX_CACHE_MAX_ENTRIES = 100
+_PREFIX_CACHE_ENABLED = bool(os.environ.get("VLLM_METAL_PREFIX_CACHE", ""))
```
LxYuan0420: This treats any non‑empty value (0, false, etc.) as True.

If we want a simple, non‑parsing approach, use presence only: "VLLM_METAL_PREFIX_CACHE" in os.environ, and please document it in README

vllm_metal/v1/model_runner.py
```diff
@@ -69,8 +71,48 @@
...
+            logger.warning(
+                "Invalid VLLM_METAL_PREFIX_CACHE_FRACTION=%r, using default %.2f",
+                fraction_str,
+                _PREFIX_CACHE_DEFAULT_FRACTION,
+            )
+            fraction = _PREFIX_CACHE_DEFAULT_FRACTION
+    else:
+        fraction = _PREFIX_CACHE_DEFAULT_FRACTION
+
+    device_info = mx.metal.device_info()
```
LxYuan0420: This is right but mx.metal.device_info() can throw on Non-metal?

Please wrap in try/except and fallback to (I think a reasonable number would be) 8GB so this optimal feature doesn't crash.

LxYuan0420 (changes requested):
Thanks for the update. 

I still see a few minor polish items and I personally feel there is a bit of code smell, but once those are addressed it looks fine to land. We can revisit the design if we see issues in real usage

tests/test_prefix_cache.py
```diff
@@ -90,3 +90,95 @@ def fake_make_prompt_cache(model):
...
+        assert mgr._current_bytes == bytes_after_first
+
+
+class TestPrefixCacheEnableFlag:
+    """Verify VLLM_METAL_PREFIX_CACHE presence-based enabling."""
+
+    def test_enabled_when_env_set(self, monkeypatch) -> None:
+        monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE", "1")
+        # Re-evaluate the module-level flag
+        assert "VLLM_METAL_PREFIX_CACHE" in mr.os.environ
```
LxYuan0420: These tests only assert env presence, not the actual enable logic (the flag is computed at import time). To make this test real without module reloads, i suggest a tiny helper and use it in the test:

```python
# in vllm_metal/v1/model_runner.py
def _prefix_cache_enabled() -> bool:
    return "VLLM_METAL_PREFIX_CACHE" in os.environ

_PREFIX_CACHE_ENABLED = _prefix_cache_enabled()
```


Then in tests:
```python
def test_enabled_when_env_set(monkeypatch):
    # Arrange
    monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE", "1")
    # Act
    enabled = mr._prefix_cache_enabled()
    # Assert
    assert enabled is True
```

this way we can keep tests deterministic and avoids reloading the module

tests/test_prefix_cache.py
```diff
@@ -90,3 +90,95 @@ def fake_make_prompt_cache(model):
...
+
+    def test_disabled_when_env_unset(self, monkeypatch) -> None:
+        monkeypatch.delenv("VLLM_METAL_PREFIX_CACHE", raising=False)
+        assert "VLLM_METAL_PREFIX_CACHE" not in mr.os.environ
+
+
+class TestPrefixCacheFractionParsing:
```
LxYuan0420, line 161: These tests currently only `assert > 0` and still call real `mx.metal.device_info()`, so they dont validate parsing or fallback behavior and are hardware‑dependent. 

we need to ensure they are deterministic tests by monkeypatching device_info and asserting the exact byte result:

```python
def test_valid_fraction(monkeypatch):
    # Arrange
    monkeypatch.setenv("VLLM_METAL_PREFIX_CACHE_FRACTION", "0.1")
    monkeypatch.setattr(
        mr.mx.metal,
        "device_info",
        lambda: {"max_recommended_working_set_size": 10 * 1024**3},
    )
    # Act
    result = mr._get_prefix_cache_max_bytes()
    # Assert
    assert result == int(10 * 1024**3 * 0.1)
```

PR #86, opened 2026-01-29
native mlx sampler

solarpunkin:
perf: replaces pytorch sampler with mlx implementation

* Added `vllm_metal/mlx_sampler.py`: native sampling in MLX.

Routes some requests to mlx as slow path is really slow (especially on lower-end Macs), keeping feature completeness with vllm.


solarpunkin:
I will post the tps gains.

PR #90, opened 2026-02-04
Refactor token id extraction in Metal runner

LxYuan0420:
This PR is:
- To refactor token id extraction to handle None and non-iterables consistently
- To add focused tests that cover those edge cases

Context:
- Simple refactor only; behavior is unchanged except for clearer error handling.

Tests:
- ruff check .
- pytest

### DIFF
--- a/tests/test_model_runner_extract_tokens.py
+++ b/tests/test_model_runner_extract_tokens.py
@@ -1,0 +1,26 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+import vllm_metal.model_runner as mr  # noqa: E402
+
+
+class DummySeqData:
+    def __init__(self, token_ids) -> None:
+        self._token_ids = token_ids
+
+    def get_token_ids(self):
+        return self._token_ids
+
+
+def test_extract_tokens_none_returns_empty() -> None:
+    assert mr.MetalModelRunner._extract_tokens(DummySeqData(None)) == []
+
+
+def test_extract_tokens_non_iterable_raises() -> None:
+    with pytest.raises(ValueError, match="Sequence data lacks token ids"):
+        mr.MetalModelRunner._extract_tokens(DummySeqData(123))


ericcurtin:
Hey @LxYuan0420 want to join the slack channel? We got a surprise for you, we want to make you a co-maintainer

WindChimeRan:
@LxYuan0420  is this ‎vllm_metal/model_runner.py‎ dead code? Should we delete this file?

LxYuan0420:
> @LxYuan0420 is this ‎vllm_metal/model_runner.py‎ dead code? Should we delete this file?

Not dead code today

`vllm_metal/model_runner.py` is still used by the `vllm-metal` standalone server (vllm_metal/server.py) and the legacy `vllm_metal/worker.py`, and we also have tests that import it. The main “production” path is the v1 runner (`vllm_metal/v1/model_runner.py`) via `vllm serve`, but we still keep the non-v1 runner as a lightweight smoke- test / demo entrypoint.

Example commands:

```bash
# Lightweight smoke test (standalone server; uses vllm_metal/model_runner.py)
vllm-metal --model Qwen/Qwen2.5-1.5B-Instruct

# Production / upstream parity (vLLM engine v1; uses vllm_metal/v1/model_runner.py)
vllm serve Qwen/Qwen2.5-1.5B-Instruct
```

WindChimeRan:
@LxYuan0420 I don't think so.

looking at the code:
  - CI already smoke-tests via vllm serve, not vllm-metal
  - The non-v1 model_runner allocates a PagedKVCache but never uses it during inference
  - It only supports greedy sampling, so it's not representative of the real system
  - The CLI isn't documented anywhere
  - Having two parallel systems with separate memory allocation logic adds maintenance burden and naming confusion (e.g. __init__.py
  exported the non-v1 MetalWorker as the top-level name)

see PR #95

LxYuan0420:
@WindChimeRan 

Thanks for calling out the issues here

agree the legacy runner has gaps (e.g. unused PagedKVCache, limited feature). CI only smoke-tests vllm serve, but that doesnt make the non‑v1 path dead: we still can use vllm-metal for quick local smoke/demo runs and it’s referenced in code/examples.

So im not in favor of deleting it in PR #95 right now because it’s a user-facing breaking change. If we want less
confusion, I prefer we:

1. clearly label vllm-metal as a legacy smoke/demo entrypoint (not v1 / not production),
2. stop exposing legacy MetalWorker / MetalModelRunner as top-level imports from vllm_metal/__init__.py (this won’t break the vllm-metal CLI since it imports the legacy runner directly),
3. clean up obvious issues in the legacy runner (e.g. remove the unused PagedKVCache allocation), then revisit removal later with a proper deprecation + migration plan

WindChimeRan:
@LxYuan0420  I think keeping the legacy path creates a concrete problem right now.

  In PR #92 I'm redesigning the paged KV cache, it touches mlx_backend/cache.py, and other memory allocation logic. It will break the compatibility of non-v1. 

  On the "user-facing breaking change" concern, the vllm-metal CLI isn't documented in the README or any guide, so there are no   known users to break. The README only documents vllm serve with the Metal plugin.

  I'd prefer to remove the legacy code now in a small focused PR, rather than carry it through the paged attention redesign and deal with it later.

LxYuan0420:
@WindChimeRan 

I think PR92 can move forward without deleting non‑v1? we can treat the legacy path as best‑effort/unsupported while the paged‑KV redesign lands. If we decide to retire it, I’d rather do a proper deprecation: keep vllm-metal as a thin wrapper that prints a deprecation notice and points people to vllm serve, plus update the README/migration notes

Issue #84, opened 2026-01-27
[Feature Request] Add TTS and STT support


miesgre:
I would like to request the addition of Speech-to-Text (STT) and Text-to-Speech (TTS) capabilities to the project.

I suggest integrating [mlx-audio](https://github.com/Blaizzy/mlx-audio), which provides a simple interface for running audio models natively on Apple Silicon.

Thanks!

ricky-chaoju:
STT support has been merged. See the documentation for usage: docs/stt.md
https://github.com/vllm-project/vllm-metal/blob/main/docs/stt.md

PR #91, opened 2026-02-04
[Feature] Add OpenAI-compatible STT support for Whisper models (#84)

ricky-chaoju:
## Summary
Adds `/v1/audio/transcriptions` and `/v1/audio/translations` endpoints so vllm-metal can serve Whisper models (e.g. `openai/whisper-small`, MLX-converted variants) as an OpenAI-compatible STT server.

Related #84

## included

- Whisper inference on Metal — encoder-decoder forward pass via MLX, with mel spectrogram preprocessing using the same pipeline as upstream
- All five response formats — `json`, `verbose_json`, `text`, `srt`, `vtt`
- Timestamp extraction — segment-level timestamps via Whisper's timestamp tokens
- Audio chunking — long audio is split at low-energy points to avoid cutting words mid-sentence
- Prompt support — `prompt` parameter for guiding proper noun spelling (injected via `<|startofprev|>` tokens, same as upstream)
- Three-tier language validation — matching upstream vLLM behaviour: officially supported codes pass silently, known-but-unofficial codes pass with a warning, unknown codes return an error
- Auto-detection — `--model` automatically detects LLM vs STT from `config.json`, no separate flag needed

## test

```bash
vllm serve --model /Volumes/MacPro/llm-dev/models/whisper-small-mlx 
curl -X POST http://localhost:8000/v1/audio/transcriptions \
  -F "file=@test.wav" -F "response_format=verbose_json"
```

### DIFF
--- a/vllm_metal/stt/transcribe.py
+++ b/vllm_metal/stt/transcribe.py
@@ -132,7 +132,11 @@ class WhisperTranscriber:
         for chunk_audio, chunk_start in chunks:
             features = _encode_chunk(self.model, chunk_audio)
             output_tokens = _greedy_decode(
-                self.model, features, language, task, prompt,
+                self.model,
+                features,
+                language,
+                task,
+                prompt,
                 with_timestamps=with_timestamps,
             )
 
@@ -144,7 +148,9 @@ class WhisperTranscriber:
                 seg_id_offset += len(segments)
                 # Fallback if no segments extracted
                 if not segments:
-                    text = self.tokenizer.decode(output_tokens, skip_special_tokens=True)
+                    text = self.tokenizer.decode(
+                        output_tokens, skip_special_tokens=True
+                    )
                     if text.strip():
                         all_text_parts.append(text.strip())
             else:


pyproject.toml
```diff
@@ -38,6 +38,9 @@ dependencies = [
     # Core utilities
     "numpy>=1.24.0",
     "psutil>=5.9.0",
+    # Audio processing — required by vLLM's speech_to_text pipeline
+    "librosa>=0.10.2",
+    "numba>=0.59.0",
```
LxYuan0420: These are heavy deps. Please move to an extra (ex: [project.optional-dependencies] stt = ["vllm[audio]>=0.13.0"] or stt =["librosa>=...", "soundfile>=..."]) and update docs/CI accordingly

vllm_metal/__init__.py
```diff
@@ -88,6 +88,39 @@ def __getattr__(name):
...
+    import vllm.transformers_utils.config as _tc
+
+    _original = _tc.try_get_safetensors_metadata
+
+    def _patched(model: str, *, revision: str | None = None):  # type: ignore[override]
+        if Path(model).exists():
+            return None
+        return _original(model, revision=revision)
+
+    _tc.try_get_safetensors_metadata = _patched
```
LxYuan0420: This is a code smell and very risky: it mutates vLLM internals globally for every users. Please gate behind an env flag or maybe an upstream fix. Prefer a separate PR so STT related review stay focused.

vllm_metal/stt/config.py
```diff
@@ -0,0 +1,158 @@
...
+logger = logging.getLogger(__name__)
+
+# STT model types that can be auto-detected from config.json
+_STT_MODEL_TYPES = {"whisper"}
+
+# Full Whisper language map (100 languages recognised by the tokenizer).
+# fmt: off
+WHISPER_LANGUAGES: dict[str, str] = {
+    "en": "english", "zh": "chinese", "de": "german", "es": "spanish",
+    "ru": "russian", "ko": "korean", "fr": "french", "ja": "japanese",
```
LxYuan0420: Please derive these from a canonical source (transformers Whisper tokeniser tables or vLLM) instead of hardcoding large maps into the repo

vllm_metal/stt/config.py
```diff
@@ -0,0 +1,158 @@
...
+    "ha": "hausa", "ba": "bashkir", "jw": "javanese", "su": "sundanese",
+    "yue": "cantonese",
+}
+
+# Officially supported languages (from OpenAI docs) — a subset of the above.
+# Languages in WHISPER_LANGUAGES but not here are accepted with a warning.
+# Reference: https://platform.openai.com/docs/guides/speech-to-text/supported-languages
+ISO639_1_SUPPORTED_LANGS: dict[str, str] = {
+    "af": "afrikaans", "ar": "arabic", "hy": "armenian", "az": "azerbaijani",
+    "be": "belarusian", "bs": "bosnian", "bg": "bulgarian", "ca": "catalan",
```
LxYuan0420: Please derive these from a canonical source (transformers Whisper tokeniser tables or vLLM) instead of hardcoding large maps into the repo

vllm_metal/stt/config.py
```diff
@@ -0,0 +1,158 @@
...
+    matching upstream Whisper behaviour).
+    """
+    if code is None:
+        if default is not None:
+            logger.warning(
+                "Defaulting to language=%r. Pass the `language` field "
+                "to transcribe audio in a different language.",
+                default,
+            )
+        return default
```
LxYuan0420, line 112: This basically will warn on every request, might be too noisy

vllm_metal/platform.py
```diff
@@ -239,6 +248,59 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+
+            for fname in missing:
+                try:
+                    src = hf_hub_download(
+                        repo_id=cls._STT_CANONICAL_REPO, filename=fname
+                    )
+                    shutil.copy2(src, str(dest / fname))
+                    logger.info(f"Copied {fname} into {model_path}")
+                except Exception:
+                    logger.debug(f"Could not fetch {fname} from HF, skipping")
```
LxYuan0420: Few issues here:

this writes files into the user's model directory at startup. Breaks on read-only mounts, fails silently. Load the tokenizer from the canonical repo at runtime instead of copying files around

LxYuan0420: with only logger.debug, it swallows all errors silently (permission denied, disk full, etc)

vllm_metal/stt/audio.py
```diff
@@ -0,0 +1,208 @@
...
+N_SAMPLES = CHUNK_LENGTH * SAMPLE_RATE
+N_FRAMES = N_SAMPLES // HOP_LENGTH
+N_MELS = 80
+
+
+def load_audio(file_path: str, sample_rate: int = SAMPLE_RATE) -> mx.array:
+    """Load audio file using ffmpeg."""
+    cmd = [
+        "ffmpeg",
+        "-nostdin",
```
LxYuan0420, line 31: ffmpeg is used here but it is an undocumented dependency?

vllm_metal/stt/transcribe.py
```diff
@@ -0,0 +1,362 @@
...
+    # Handle trailing tokens without a closing timestamp
+    if seg_start is not None and seg_tokens:
+        seg_text = tokenizer.decode(seg_tokens, skip_special_tokens=True)
+        if seg_text.strip():
+            segments.append(
+                TranscriptionSegment(
+                    id=seg_id,
+                    seek=int(seg_start * 100),
+                    start=round(seg_start + time_offset, 2),
+                    end=round(seg_start + time_offset + 30.0, 2),
```
LxYuan0420: 100 and 30.0 are magic numbers?

vllm_metal/stt/audio.py
```diff
@@ -0,0 +1,208 @@
...
+import numpy as np
+
+# Whisper audio constants
+SAMPLE_RATE = 16000
+N_FFT = 400
+HOP_LENGTH = 160
+CHUNK_LENGTH = 30
+N_SAMPLES = CHUNK_LENGTH * SAMPLE_RATE
+N_FRAMES = N_SAMPLES // HOP_LENGTH
+N_MELS = 80
```
LxYuan0420: ideally we shouldn't hardcoded these and should use the loaded model specific config

LxYuan0420: this applies to functions that use these constants like log_mel_spectrogram and others.

vllm_metal/stt/transcribe.py
```diff
@@ -0,0 +1,362 @@
...
+
+    for _ in range(max_tokens):
+        logits, kv_cache = model.decode(tokens, audio_features, kv_cache)
+        next_token = int(mx.argmax(logits[:, -1, :], axis=-1).item())
+        if next_token == eot_token:
+            break
+        output_tokens.append(next_token)
+        tokens = mx.array([[next_token]], dtype=mx.int32)
+
+    return output_tokens
```
LxYuan0420, line 277: These 2 functions are almost identical with only difference is one append:  prefix.append(_get_token_id("<|notimestamps|>"))

vllm_metal/v1/model_runner.py
```diff
@@ -1129,6 +1050,173 @@ def execute_model(
...
+            return audio_features
+
+        return None
+
+    def _greedy_decode_stt(
+        self,
+        audio_features: mx.array,
+        prompt_token_ids: list[int],
+        eot_token: int,
+    ) -> list[int]:
```
LxYuan0420, line 1449: Seems like now we have 3 copies of the same greedy decode loop?

vllm_metal/v1/model_runner.py
```diff
@@ -514,12 +428,38 @@ def _extract_logits(self, model_output: Any) -> mx.array:
...
+        """Get supported tasks for this model.
+
+        Returns:
+            Tuple of supported task names. STT models support
+            ``("transcription",)``; LLM/VLM models support ``("generate",)``.
+        """
+        if self._is_stt:
+            return ("transcription",)
+        return ("generate",)
+
```
LxYuan0420: Avoid stringly-typed tasks and align types with vLLM.MetalWorker.get_supported_tasks returns tuple[SupportedTask, ...], but this returns tuple[str, ...]

Please import SupportedTask (or GenerationTask) from vllm.tasks and type this method accordingly to prevent typos and make the contract explicit

vllm_metal/v1/model_runner.py
```diff
@@ -514,12 +428,38 @@ def _extract_logits(self, model_output: Any) -> mx.array:
...
+        if self._is_stt:
+            block_size = self.metal_config.block_size
+            return {
+                "layers.0.self_attn": FullAttentionSpec(
+                    block_size=block_size,
+                    num_kv_heads=1,
+                    head_size=64,
+                    dtype=torch.float16,
+                ),
+            }
```
LxYuan0420, line 677: this is a code smell: why this is safe and doesnt allocate unnecessary cache?

tests/test_stt.py
```diff
@@ -0,0 +1,412 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for Speech-to-Text functionality."""
+
```
LxYuan0420, line 3: Once the code comments are addressed, the tests will need updating too. 

Specifically: add tests for the core audio pipeline (log_mel_spectrogram, _stft), add tests for sanitize weight mapping, and add error-path tests for malformed mm_features instead of relying on broad except Exception catches

ricky-chaoju:
Thanks for the thorough review! I’ve addressed all the feedback:

- Removed the monkeypatch (will handle it in a separate PR)
- Moved audio dependencies to an optional `[stt]` extra
- Now deriving supported languages directly from transformers
- Removed noisy warnings and `_ensure_stt_hf_files`
- Added a `WhisperTranscriber` class
- Fixed `n_mels` to be read from the model config (adds large-v3 support)
- Fixed audio truncation in non-timestamp mode
- Unsupported sampling parameters now raise a `ValueError`
- Added tests covering the audio pipeline, weight sanitization, and mm_features error paths

Ready for re-review whenever you have time.

docs/stt.md
LxYuan0420, line 1: this `docs/stt.md` currently claims subtitle output and response_format supports srt/vtt, but vLLM OpenAI STT only accepts json/text/verbose_json (afaik it errors on srt/vtt). Since vllm-metal is a plugin, please align docs/examples to vLLM core behavior, or move srt/vtt support to an upstream vLLM PR first (and then document it here)

vllm_metal/platform.py
```diff
@@ -212,6 +214,11 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
         if config.debug:
             logger.info(f"Metal config: {config}")
 
+        # STT models: set tokenizer path for runtime loading
+        if model_config is not None and is_stt_model(model_config.model):
+            model_config.tokenizer = model_config.model
+            logger.info("STT model detected")
```
LxYuan0420, line 220: this unconditionally sets model_config.tokeniser = model_config.model

it can override user supplied --tokenizer and caused some unwanted side effects from a plugin hook.

please only set a fallback when tokeniser is unset or avoid mutating model_config  and instead use model as a fallback in the STT runner

vllm_metal/v1/model_runner.py
```diff
@@ -689,6 +754,17 @@ def warm_up(self) -> None:
 
         logger.info("Warming up model...")
 
+        if self._is_stt:
+            # STT warm-up: run a short silent mel through the encoder
+            try:
+                dummy_mel = mx.zeros((1, 80, 3000), dtype=mx.float16).transpose(0, 2, 1)
```
LxYuan0420, line 760: avoid hardcoded config/value

this part hardcodes dummy_mel shape as (80, 3000) but it breaks large-v3 which use 128 Mel bins

so please derive warm-up tensor shape from the loaded model config, for example:
  - n_mels = self.model.config.n_mels
  - n_audio_ctx = self.model.config.n_audio_ctx
  - and set a time dimension consistent with the model (or reuse the real mm_features shape if available)

vllm_metal/v1/model_runner.py
LxYuan0420: this part still confuses me: is this necessary?

and the docstring says empty dict but it returns dummy spec?

tests/test_stt.py
```diff
@@ -0,0 +1,596 @@
...
+    @pytest.fixture(autouse=True)
+    def _import_extract(self):
+        """Import _extract_segments; skip if tokenizer unavailable."""
+        try:
+            from vllm_metal.stt.transcribe import _extract_segments, _get_token_id
+
+            self._extract = _extract_segments
+            self._tid = _get_token_id
+        except Exception:
+            pytest.skip("Whisper tokenizer not available")
```
LxYuan0420, line 303: This can hide real regressions by turning failures into skips. Please use pytest.importorskip("transformers") or only catch ImportError

tests/test_stt.py
```diff
@@ -0,0 +1,596 @@
...
+
+    @pytest.fixture(autouse=True)
+    def _import_encode(self):
+        try:
+            from vllm_metal.stt.transcribe import _encode_prompt, _get_token_id
+
+            self._encode = _encode_prompt
+            self._tid = _get_token_id
+        except Exception:
+            pytest.skip("Whisper tokenizer not available")
```
LxYuan0420, line 357: This can hide real regressions by turning failures into skips. Please use pytest.importorskip("transformers") or only catch ImportError.

docs/stt.md
```diff
@@ -0,0 +1,113 @@
+# Speech-to-Text (STT)
+
+vllm-metal includes OpenAI-compatible Speech-to-Text support built on Whisper. It supports transcription, translation, timestamps, and subtitle output — all running natively on Apple Silicon via MLX.
+
+## Requirements
+
+```bash
+pip install vllm-metal[stt]  # Install audio processing dependencies
+brew install ffmpeg          # Required for audio file loading
```
LxYuan0420, line 9: we generally want to be careful introducing heavy dependency: the ffmpeg note is too strong: librosa.load(...) works for WAV without ffmpeg, but formats like mp3/m4a often require ffmpeg/audioread

Please clarify when ffmpeg is required vs recommended?

ricky-chaoju:
Thanks for taking the time to review this so thoroughly  I learned a lot from your feedback.
I’ve split the changes into smaller PRs as suggested. PR #96 is the first one (packaging).

ricky-chaoju:
Thanks @LxYuan0420 ! All review items have been addressed across #96, #98, #126, #133, #137, and #143. Close this pr.

PR #71, opened 2026-01-11
Add paged KV cache adapter and integrate

ericcurtin:
Tests and implementation for paged KV cache adapters that wrap
PagedKVCache to match mlx_lm's KVCache interface. Replaces the
traditional KV cache with block-based storage for memory-efficient
inference. Updates model runner to use paged cache with pre-allocation
and batched decode operations.

Signed-off-by: Eric Curtin <eric.curtin@docker.com>

PR #92, opened 2026-02-07
[Paged KV] Paged KV cache by HF Metal kernel. 

WindChimeRan:
Usage 

Page KV Cache On
```
VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.3 vllm serve Qwen/Qwen3-0.6B --max-model-len 2048
```

```
vllm bench serve --backend vllm --model Qwen/Qwen3-0.6B \
    --endpoint /v1/completions \
    --dataset-name sharegpt \
    --dataset-path ShareGPT_V3_unfiltered_cleaned_split.json \
    --num-prompts 100 \
    --request-rate 10 \
    --max-concurrency 32
```

baseline: default, Page KV Cache Off, mlx_lm
```
vllm serve Qwen/Qwen3-0.6B --max-model-len 2048
```

### Benchmark

Apple M2 Max 36GB, ShareGPT, num-prompts 100, request-rate 10, max-concurrency 32

bench_comparison

* TTFT: better
* output troughput: better
* Mean ITL: worse

output equivelence (paged kv cache vs mlx_lm), both:
* Total input tokens:23260
* Total generated tokens:22061

Memory allocation

* mlx_lm use `auto` to only use just enough memory
* paged kv cache use VLLM_METAL_MEMORY_FRACTION, allocate as much memory as possible.

Paged KV cache trades higher memory usage for better concurrency, making it systemwide faster than mlx_lm. Whether it's also faster at the kernel level is unclear, but advanced features like continuous batching and chunked prefilling are infeasible to support with mlx_lm alone.

### PR Summary

<details> 
<summary> Patch the mlx models with paged attention kernel. </summary>

  - mlx_lm requires contiguous kv cache, and this PR use paged kv cache (not contiguous). 
  - Paged kv cache is a prerequisite of future continuous batching and real chunked prefilling. 
  - Integrates the https://huggingface.co/kernels-community/paged-attention Metal shader for paged KV cache on Apple Silicon (This can be replaced by mlx native page attention or other better kernels in the future)
  - Patches existing mlx_lm model attention layers at runtime with a wrapper that routes to the external Metal kernel for cache read/write, while keeping MLX for projections and other layers. 
  - Prefill: standard MLX causal SDPA, then writes K/V to MPS paged cache via reshape_and_cache
  - Decode: zero-copy attention via paged_attention_v1 — reads K/V directly from block tables on GPU, eliminating the O(seq_len) gather/copy per layer per step
  - Falls back to original mlx_lm attention when the env var is not set

</details>

<details> 
<summary> Implement the model runner <--> vllm scheduler contract, so they are aligned.  </summary>

- for chunked prefilling 0:n-1: sample-then-drop the last token
- for chunked prefilling n: sample-and-keep the last token
- for decoding: generate 1 new token

</details>

### Known Limitation & Planned Future PRs:

* **[High Priority]** when setting too small `VLLM_METAL_MEMORY_FRACTION=0.1`. hit `RuntimeError: Not enough free blocks: need 21, have 0` . This is because: All kv blocks have been consumed, while the prefilling/decoding have not been finished, then deadlock. Need to implement vllm-metal's paged kv cache preemption to align with vllm scheduler contract. 
* **[High Priority]** torch_to_mlx in the tensor bridge may not be true zero copy. 
* [Bug] not working with HuggingFaceTB/SmolVLM-Instruct, https://github.com/vllm-project/vllm-metal/pull/114 might be the fix
* [Feature] re-enable prefix caching under paged kv cache. Prior version: https://github.com/vllm-project/vllm-metal/pull/80
* [Medium Priority] real chunked prefilling. This PR's implementation is wasteful. Expected: chunk n prefill read the 0:n-1 kv cache, and only prefill n. Actual: prefill all 0:n kv cache each time. The time complexity is quadratic in terms of the number of chunks. Why? just to satisfy vllm scheduler. 
* [Medium Priority] real continuous batching to align with upstream vllm. This requires var len prefilling & decoding operating on `[total_num_token, *]` instead of the current `[batch, seq, *]`. 
* [Refactor] https://github.com/vllm-project/vllm-metal/issues/97
* [Refactor] five separate forward paths (_prefill_single, _prefill_single_request_paged .etc) that share the same pre/post processing. Maybe we can merged duplicate codes, but that's for the sake of aesthetics. 
* [Doc] Readme architecture figure is no longer accurate. 

* ~~[Testing] Need to test on macos 14/15, metal 3.2. It is expected to work.~~ It works. 

### FAQ

<details>
<summary>Why hack the paged KV cache as a global variable?</summary>

The model's `__call__` signature is `(input_ids, cache=...)` — and `mlx_lm`'s call requires contiguous tensors with no additional parameters. There's no way to pass `slot_mapping`, `block_tables`, or any other per-forward metadata down to the attention layers. This design is inspired by [nano-vllm](https://github.com/GeeeekExplorer/nano-vllm).

</details>

<details>
<summary>Why use this attention kernel?</summary>

This kernel supports variable-length prefilling and decoding, so attention can be computed over `[total_tokens, *]` instead of `[batch, seq, *]`. This is essential for supporting real continuous batching in the future.

</details>

<details>
<summary>Each call to the attention kernel triggers an MLX ↔ Torch round trip?</summary>

Yes, but it's okay as long as the MLX-to-Torch round trip is implemented in zero-copy mode. Besides, this kernel can be replaced by better ones in the future if they become available. The most important thing is to get the whole system working end-to-end (chunked prefill, continuous batching) first; then we can swap in better modules later.

</details>

<details>
<summary>What would a future paged attention kernel look like?</summary>

It would need to support variable sequence lengths, `slot_map`, etc. — similar to `flash_attn_varlen_func` and `flash_attn_with_kvcache` from FlashAttention. The difficulties are:

1. HuggingFace kernel libraries only expose PyTorch bindings, which require type conversion from our MLX tensors.
2. As far as I understand, a proper FlashAttention-style implementation would need to be written directly in Metal, not in MLX.

</details>

### Acknowledgement: 
Early prototype https://github.com/vllm-project/vllm-metal/pull/71

### DIFF
--- a/tests/test_kernel_loader.py
+++ b/tests/test_kernel_loader.py
@@ -16,8 +16,10 @@ from unittest import mock
 
 import pytest
 
+pytest.importorskip("kernels")
+
 # ---------------------------------------------------------------------------
-# Unit tests (no network, no GPU — run everywhere)
+# Unit tests (no network, no GPU)
 # ---------------------------------------------------------------------------
 
 


WindChimeRan:
known issue: 

hit `RuntimeError: Not enough free blocks: need 21, have 0` when setting too small `VLLM_METAL_MEMORY_FRACTION=0.1`. 

All kv blocks have been consumed, while the prefilling/decoding have not been finished, then deadlock. Need to check with the upstream vllm's eviction strategy.

LxYuan0420:
Thanks Ran. Appreciate the thorough PR write-up 

One integration point to address up front: 

since paged attention is experimental and off-by-default, can we make the kernels dependency an optional extra (e.g. vllm-metal[paged]) instead of a hard dependency in pyproject.toml (around L:39)? 

That keeps the default install lean and reduces risk for users who never enable the flag


Will review this in detail later

vllm_metal/v1/worker.py
```diff
@@ -147,10 +150,161 @@ def load_model(self) -> None:
...
+        num_attention_heads = runner.model_args.get("num_attention_heads") or 32
+        num_kv_heads = (
+            runner.model_args.get("num_key_value_heads")
+            or runner.model_args.get("n_kv_heads")
+            or num_attention_heads
+        )
+        hidden_size = runner.model_args.get("hidden_size") or 4096
+        head_dim = runner.model_args.get("head_dim") or (
+            hidden_size // num_attention_heads
+        )
```
LxYuan0420: this fall back to defaults for num_layers/num_heads/hidden_size/head_dim when keys are missing from runner.model_args (e.g. or 32, or 4096)

I think we should avoid guessing shapes here. Either reliably derive dims from the model/config, or raise a clear error. Silent mis-sizing of the paged KV cache would be a hard-to-debug correctness issue

vllm_metal/v1/model_runner.py
```diff
@@ -1082,6 +1183,215 @@ def _sequential_decode(
...
+        # - reads get_context() for slot_mapping
+        # - computes attention via MLX SDPA
+        # - writes K/V to MPS paged cache via reshape_and_cache
+        input_ids = mx.array([token_ids], dtype=mx.int32)
+        model_output = self.model(input_ids, cache=offset_caches)
+        logits = self._extract_logits(model_output)
+        last_logits = logits[:, -1, :]
+
+        clear_context()
+
```
LxYuan0420, line 1239: prepare_prefill/prepare_decode sets a thread-local paged-attention context, but clear_context() is only reached on the happy path

If the forward/kernel path throws any error (e.g. metallib incompat or OOM), the stale context can leak into later forwards with mismatched slot/block metadata

Can we wrap the "after prepare_*" region in try/finally so we always clear_context()?

tests/test_kernel_loader.py
```diff
@@ -0,0 +1,135 @@
...
+        from vllm_metal.metal_kernel_backend.kernel_loader import _needs_compat_revision
+
+        with mock.patch("platform.mac_ver", return_value=("26.3", ("", "", ""), "")):
+            assert _needs_compat_revision() is False
+
+    def test_empty_version_no_compat(self):
+        from vllm_metal.metal_kernel_backend.kernel_loader import _needs_compat_revision
+
+        with mock.patch("platform.mac_ver", return_value=("", ("", "", ""), "")):
+            assert _needs_compat_revision() is False
```
LxYuan0420: Nit: can we avoid repeated local imports in tests? _needs_compat_revision can be imported once at module scope (top of file)

tests/test_paged_attention.py
```diff
@@ -0,0 +1,55 @@
...
+        from vllm_metal.paged_attention_common import get_context
+
+        ctx = get_context()
+        assert ctx is not None
+        assert not ctx.is_prefill
+        # new_pos=7, block_ids[7//4]=block_ids[1]=6, slot=6*4+(7%4)=24+3=27
+        assert ctx.slot_mapping == [27]
+        assert ctx.context_lens == [8]
+        assert ctx.offsets == [7]
+        clear_context()
```
LxYuan0420: Nit: get_context can be imported alongside OffsetCache/prepare_* at the top of the file. Local imports here don’t seem necessary and reduce readability

Try to stick to AAA style for python tests

vllm_metal/v1/model_runner.py
WindChimeRan: done

vllm_metal/v1/worker.py
WindChimeRan: done

WindChimeRan:
merged @LxYuan0420

vllm_metal/v1/worker.py
```diff
@@ -262,10 +262,9 @@ def _setup_paged_attention(self) -> None:
         )
 
         # --- Extract model dimensions ---
-        num_layers = (
-            runner.model_args.get("num_hidden_layers")
-            or runner.model_args.get("n_layers")
-        )
+        num_layers = runner.model_args.get(
+            "num_hidden_layers"
+        ) or runner.model_args.get("n_layers")
```
LxYuan0420: Can we validate + fail fast on required dims here? Otherwise we can pass None into MPSPagedKVCache and crash later with a vague error

vllm_metal/metal_kernel_backend/cache.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # The key cache uses a 5D layout for vectorized memory access:
+        #   [num_blocks, num_kv_heads, head_dim // x, block_size, x]
+        # where x = 16 // element_size ensures each innermost vector is
+        # exactly 16 bytes, matching the Metal kernel's load granularity.
+        # This layout is required by the HF paged-attention Metal kernel
+        # (ported from vLLM CUDA / mistral.rs):
+        #   https://github.com/huggingface/kernels-community/blob/main/paged-attention/paged-attention-metal/paged_attention.mm
+        element_size = torch.tensor([], dtype=dtype).element_size()
+        self.x = 16 // element_size  # 8 for float16, 4 for float32
```
LxYuan0420, line 50: Small but important: please add if head_dim % x != 0: raise ValueError(...) before head_dim // x (x = 16 // element_size). Otherwise we silently truncate the key-cache shape

vllm_metal/paged_attention_common.py
```diff
@@ -0,0 +1,198 @@
...
+def _find_layers_and_attr(model: Any) -> tuple[list[Any], str]:
+    """Find transformer layers and the attention attribute name.
+
+    Returns (layer_list, attn_attr_name) where each layer has
+    getattr(layer, attn_attr_name) pointing to the attention module.
+
+    Supports mlx_lm model structures like:
+        model.model.layers[i].self_attn
+        model.layers[i].self_attn
+    """
```
LxYuan0420, line 116: How do we want to handle VLM wrappers (model.language_model)? _find_layers_and_attr() wont find layers there today

LxYuan0420 (changes requested):
I think we need a small set of “fail fast” hardening fixes (all should be quick)

Once those are in, I'm comfortable approving + merging, and I’m looking forward to the follow-up PRs

vllm_metal/v1/worker.py
WindChimeRan: refactored.

vllm_metal/metal_kernel_backend/cache.py
WindChimeRan: done

vllm_metal/paged_attention_common.py
WindChimeRan: fixed

tests/test_kernel_loader.py
WindChimeRan: done

tests/test_paged_attention.py
WindChimeRan: done

LxYuan0420:
> @LxYuan0420
> 
> when trying vlm, hit the same issue `RuntimeError: sample_tokens called without pending output from execute_model` #114
> 
> vlm: HuggingFaceTB/SmolVLM-Instruct.
> 
> I haven't check if your PR will fix this.

Quick clarification: PR #114 is narrowly about the v1 async sample_tokens() pending-output contract (so we don’t mask the real execute_model() failure).

If you try HuggingFaceTB/SmolVLM-Instruct and run into errors, that likely falls under a separate VLM support gap (wiring vLLM multimodal inputs into mlx_vlm’s pixel_values), rather than the sample_tokens fix itself

If you do test it, please paste the full EngineCore traceback (the part above any sample_tokens line) + the exact vllm serve … command, and I will help triage whether it’s #111/#114 or a VLM-specific issue

tests/test_kernel_loader.py
```diff
@@ -0,0 +1,126 @@
...
+                revision=_MACOS15_COMPAT_REVISION,
+            )
+        self._reset_kernel_cache()
+
+    def test_macos_26_uses_latest(self):
+        self._reset_kernel_cache()
+        with (
+            mock.patch("platform.mac_ver", return_value=("26.3", ("", "", ""), "")),
+            mock.patch("kernels.get_kernel", return_value=mock.MagicMock()) as mk,
+        ):
```
LxYuan0420, line 76: These "unit" tests patch kernels.get_kernel via a string path, which forces importing the optional kernels module. If vllm-metal[paged] isn’t installed, this will fail at import/collection time. Can we gate with pytest.importorskip("kernels")

WindChimeRan: Done. guard the whole test file by pytest.importorskip("kernels")

PR #94, opened 2026-02-08
Update Mamba cache to support ArraysCache 

solarpunkin:
Addresses changes introduced in **ml-explore/mlx-lm#842**, which deprecated `MambaCache` for `ArraysCache` in inference loop. Removed `BatchMambaCache` class as `ArraysCache` now supports `merge()` and `extract()`. Older mlx-lm versions still support MambaCache.

Tests

- **Unit Tests:** All prefix cache guard tests passed after updating sampling_params mock to include missing attributes (logprobs, seed, logits_bias, logits_processors).
- **Linter/Type Check:** `ruff` and `mypy` checks passed.

### DIFF
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -5,8 +5,8 @@ import platform
 
 import pytest
 import torch
-from vllm.attention.backends.registry import AttentionBackendEnum
-from vllm.attention.selector import AttentionSelectorConfig
+from vllm.v1.attention.backends.registry import AttentionBackendEnum
+from vllm.v1.attention.selector import AttentionSelectorConfig
 
 from vllm_metal.platform import MetalPlatform
 
--- a/vllm_metal/v1/worker.py
+++ b/vllm_metal/v1/worker.py
@@ -15,7 +15,7 @@ from vllm.distributed import (
 )
 from vllm.logger import init_logger
 from vllm.lora.request import LoRARequest
-from vllm.model_executor import set_random_seed
+from vllm.utils.torch_utils import set_random_seed
 from vllm.tasks import SupportedTask
 from vllm.v1.core.sched.output import SchedulerOutput
 from vllm.v1.kv_cache_interface import KVCacheConfig, KVCacheSpec
--- a/vllm_metal/platform.py
+++ b/vllm_metal/platform.py
@@ -7,13 +7,13 @@ from typing import TYPE_CHECKING
 
 import psutil
 import torch
-from vllm.attention.backends.registry import AttentionBackendEnum
+from vllm.v1.attention.backends.registry import AttentionBackendEnum
 from vllm.platforms.interface import DeviceCapability, Platform, PlatformEnum
 
 from vllm_metal.config import get_config
 
 if TYPE_CHECKING:
-    from vllm.attention.selector import AttentionSelectorConfig
+    from vllm.v1.attention.selector import AttentionSelectorConfig
     from vllm.config import VllmConfig
 
 logger = logging.getLogger(__name__)


ericcurtin:
This seems reasonably fine. But we should change to a model where we strictly define what mlx versions we use vllm-metal side. So we are deliberate in what version of mlx we use at any point in time in the commit history.

vllm_metal/v1/model_runner.py
```diff
@@ -284,59 +283,12 @@ def get_stats(self) -> dict:
...
-
-        Returns:
-            MambaCache for the individual request
-        """
-        cache = MambaCache()
-        for i in range(self._cache_size):
-            if self.cache[i] is not None:
-                cache.cache[i] = self.cache[i][idx : idx + 1]
-        return cache
+AnyCache: TypeAlias = KVCache | ArraysCache | Any
```
LxYuan0420, line 286: this make the alias effectively type safety lost?

vllm_metal/v1/model_runner.py
```diff
@@ -284,59 +283,12 @@ def get_stats(self) -> dict:
...
 
 
 def _is_mamba_cache(cache: AnyCache) -> bool:
-    """Check if a cache is a Mamba-style cache (ArraysCache or MambaCache)."""
-    return isinstance(cache, (MambaCache, ArraysCache))
+    """Check if a cache is a Mamba-style cache (has .cache attribute)."""
+    return hasattr(cache, "cache") and not isinstance(cache, (KVCache, BatchKVCache))
```
LxYuan0420, line 291: Prefer isinstance(cache, ArraysCache) since MambaCache is a subclass of ArraysCache in mlx-lm; this is clearer and avoids accidental matches

```bash
╰─➤  python
Python 3.12.7 (main, Oct 16 2024, 07:12:08) [Clang 18.1.8 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from mlx_lm.models.cache import ArraysCache, MambaCache
>>> print('issubclass:', issubclass(MambaCache, ArraysCache))
issubclass: True
>>> print('isinstance:', isinstance(MambaCache(), ArraysCache))
isinstance: True
```

vllm_metal/v1/worker.py
```diff
@@ -15,7 +15,7 @@
 )
 from vllm.logger import init_logger
 from vllm.lora.request import LoRARequest
-from vllm.model_executor import set_random_seed
+from vllm.utils.torch_utils import set_random_seed
```
LxYuan0420, line 18: importing set_random_seed from vllm.utils.torch_utils breaks on vLLM 0.13.0 (our declared minimum); either keep compatibility with the version we declared or bumped the min version to v0.14.0

tests/test_platform.py
```diff
@@ -5,8 +5,8 @@
 
 import pytest
 import torch
-from vllm.attention.backends.registry import AttentionBackendEnum
-from vllm.attention.selector import AttentionSelectorConfig
+from vllm.v1.attention.backends.registry import AttentionBackendEnum
+from vllm.v1.attention.selector import AttentionSelectorConfig
```
LxYuan0420, line 9: vllm.v1.attention.* imports in vllm_metal/platform.py and tests/test_platform.py do not exist in vLLM 0.13.0; it only exists in 0.14.0?

vllm_metal/v1/model_runner.py
```diff
@@ -432,7 +384,18 @@ def _extract_kv_cache(
...
-    return [cache.extract(idx) for cache in batch_caches]
+    extracted: list[AnyCache] = []
+    for cache in batch_caches:
+        if hasattr(cache, "extract"):
+            extracted.append(cache.extract(idx))
+        elif hasattr(cache, "cache") and not isinstance(cache, BatchKVCache):
+            # Fallback for older ArraysCache/MambaCache versions where .extract is missing
+            new_cache = type(cache)(len(cache.cache))
+            new_cache.cache = [c[idx : idx + 1] for c in cache.cache]
+            extracted.append(new_cache)
```
LxYuan0420, line 395: hmm... can you confirm whether cache can include None entries in this path (hybrid/Mamba/ArraysCache)? The fallback list comprehension slices every entry, which would crash on None. If None is possible, we should preserve those entries in the fallback (or add a test showing it cant happen)

solarpunkin:
I see this getting addressed in #110 so it would be better to close this one. 

[PS: I took a lil break from coding due to fatigue]

PR #95, opened 2026-02-09
clean non-v1 code

WindChimeRan:
clean up dead code of non-v1 engine, because
* it messed up with the memory allocation, while I'm trying to re-design it to fit with new paged kv cache
* it does not support advanced vllm features, e.g., Sampling, vlm, continuous batching
* the usage is not documented anywhere.

### DIFF
--- a/vllm_metal/utils.py
+++ b/vllm_metal/utils.py
@@ -12,6 +12,9 @@ def get_model_download_path(model_repo_name: str) -> str:
     """
     Get the path to the model, downloading from ModelScope if configured, otherwise will pass the model_repo_name.
 
+    When VLLM_USE_MODELSCOPE=True, downloads the model from ModelScope (modelscope.cn)
+    instead of HuggingFace. Useful in regions where HuggingFace is slow or blocked.
+
     Args:
         model_repo_name: Model repo name from HuggingFace or ModelScope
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -38,15 +38,8 @@ dependencies = [
...
     "psutil>=5.9.0",
-    # Server dependencies
-    "fastapi>=0.100.0",
-    "uvicorn>=0.23.0",
-    "pydantic>=2.0.0",
 ]
 
-[project.scripts]
-vllm-metal = "vllm_metal.server:main"
-
```
LxYuan0420, line 49: As discussed, this removes the vllm-metal cli command, which still can be used for quick local smoke tests.

LxYuan0420 (commented):
I’m not comfortable merging PR95 as-is

it hard-deletes the non‑v1 path and breaks downstream imports (e.g. from vllm_metal import MetalModelRunner) without a deprecation/migration plan.

Also, even if we do want to remove vllm-metal, there are still in-repo references that would become stale. For example:

```bash
rg -n "vllm-metal" vllm_metal/utils.py
```

shows docs/examples that would need updating (perhaps you should update it)

Finally, I dont think PR92 is blocked on deleting legacy code. We can treat the non‑v1 entrypoint as legacy/unsupported (best-effort) while the paged-KV redesign lands, and do any removal later as an explicit deprecation with clear migration notes.

@ericcurtin  would you mind giving a second opinion on whether we want to retire the vllm-metal entrypoint now vs keep it (legacy) for local smoke/demo for a bit longer?

ericcurtin:
All usages of vllm-metal should be via vllm core executable. We are a plugin, not an independent project. Whether we keep unsupported code around or not, I don't mind either way.

pyproject.toml
WindChimeRan: @LxYuan0420 The vllm-metal CLI isn't a meaningful smoke test. Let's remove these.

PR #96, opened 2026-02-11
[Docs] Add optional [stt] extra for Whisper audio processing

ricky-chaoju:
## Summary

Adds optional `[stt]` extra for Speech-to-Text support with Whisper models.

Part 1 of STT PR series (split from #91 per reviewer request).

## Changes

- Add `[stt]` optional dependency group in `pyproject.toml`
  - `librosa>=0.10.2`
  - `numba>=0.59.0` (required by librosa on Python 3.12+)
- Add `docs/stt.md` with installation instructions
  - Clarifies `ffmpeg` is only needed for non-WAV formats (mp3, m4a, etc.)

## Installation

```bash
pip install vllm-metal[stt]
```

I will continue working on the Core STT module after this PR is merged.

### DIFF
--- a/docs/stt.md
+++ b/docs/stt.md
@@ -1,0 +1,83 @@
+# Speech-to-Text (STT)
+
+vllm-metal supports OpenAI-compatible Speech-to-Text using Whisper models, running natively on Apple Silicon via MLX.
+
+## Installation
+
+STT requires optional audio processing dependencies:
+
+```bash
+pip install vllm-metal[stt]
+```
+
+### ffmpeg (Optional)
+
+ffmpeg is only needed for non-WAV audio formats (mp3, m4a, flac, etc.):
+
+```bash
+# macOS
+brew install ffmpeg
+
+# Not required for WAV files - librosa handles those directly
+```
+
+## Quick Start
+
+```bash
+# Start server with a Whisper model
+vllm serve openai/whisper-small --port 8000
+
+# Transcribe audio
+curl -X POST http://localhost:8000/v1/audio/transcriptions \
+  -F "file=@recording.wav"
+```
+
+## Supported Models
+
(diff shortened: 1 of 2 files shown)


docs/stt.md
```diff
@@ -0,0 +1,83 @@
...
+
+vllm-metal supports OpenAI-compatible Speech-to-Text using Whisper models, running natively on Apple Silicon via MLX.
+
+## Installation
+
+STT requires optional audio processing dependencies:
+
+```bash
+pip install vllm-metal[stt]
+```
```
LxYuan0420, line 11: Nit:  this currently shows pip install vllm-metal[stt] but the quick start uses vllm serve ...; could you add a one-liner that the recommended path is install.sh (per README) and then install [stt] inside that venv, since [stt] alone doesn’t provide the vllm CLI?

pyproject.toml
```diff
@@ -42,6 +42,11 @@ dependencies = [
 
 [project.optional-dependencies]
 vllm = ["vllm>=0.13.0"]
+stt = [
+    # Speech-to-text audio processing (Whisper models)
+    "librosa>=0.10.2",
+    "numba>=0.59.0",  # Required by librosa on Python 3.12+
+]
```
LxYuan0420, line 49: Nit: now that [stt] exists in pyproject.toml, consider updating all = [...] to include stt (vllm,stt,dev) so the all extra stays true to its name (fine to defer if you want to keep this PR minimal)

LxYuan0420 (approved):
Thanks for breaking this down into a smaller PR

 Much easier to review now

For the follow-up implementation PR (actual Whisper/STT support), would be really helpful to include a brief “Verification / Repro” section in the PR description: exact setup + server start commands, 1–2 curl examples (with expected status/response), and optionally a short before/after note or screenshot/log snippet. That will make validation and review much smoother.

PR #98, opened 2026-02-12
Add STT data types, config, and formatting

ricky-chaoju:
## Summary
Add the foundational STT types, configuration, and subtitle formatting utilities.

## Included
- `vllm_metal/stt/protocol.py` — `TranscriptionSegment` data type
- `vllm_metal/stt/config.py` — language validation, `SpeechToTextConfig`, model detection
- `vllm_metal/stt/formatting.py` — SRT / WebVTT output
- `tests/test_stt.py` — 18 unit tests

## Reviewer Feedback Addressed
Related #91

- Languages derived from `transformers.models.whisper.tokenization_whisper.LANGUAGES`
- `except Exception` → specific exception types
- Docstrings follow project Google-style convention

### DIFF
--- a/docs/stt.md
+++ b/docs/stt.md
@@ -4,9 +4,16 @@ vllm-metal supports OpenAI-compatible Speech-to-Text using Whisper models, runni
 
 ## Installation
 
-STT requires optional audio processing dependencies:
+First, install vllm-metal using the install script (see [README](../README.md)):
 
 ```bash
+./install.sh
+```
+
+Then install the optional STT dependencies inside the virtual environment:
+
+```bash
+source .venv/bin/activate
 pip install vllm-metal[stt]
 ```
 
--- a/vllm_metal/stt/formatting.py
+++ b/vllm_metal/stt/formatting.py
@@ -1,0 +1,51 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Subtitle formatting utilities (SRT / WebVTT)."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from vllm_metal.stt.protocol import TranscriptionSegment
+
+
+def _format_timestamp(seconds: float, ms_sep: str = ".") -> str:
+    """Format *seconds* as ``HH:MM:SS<sep>mmm``.
+
+    Args:
+        seconds: Time in seconds.
+        ms_sep: Millisecond separator — ``,`` for SRT, ``.`` for WebVTT.
+
+    Returns:
+        Formatted timestamp string, e.g. ``"01:02:03.456"``.
+    """
+    hrs = int(seconds // 3600)
+    mins = int((seconds % 3600) // 60)
+    secs = int(seconds % 60)
+    millis = int(round((seconds - int(seconds)) * 1000))
+    return f"{hrs:02d}:{mins:02d}:{secs:02d}{ms_sep}{millis:03d}"
+
+
+def format_as_srt(segments: list[TranscriptionSegment]) -> str:
+    """Render segments as an SRT subtitle string."""
+    lines: list[str] = []
+    for seg in segments:
+        lines.append(str(seg.id + 1))
+        start = _format_timestamp(seg.start, ms_sep=",")
+        end = _format_timestamp(seg.end, ms_sep=",")
+        lines.append(f"{start} --> {end}")
+        lines.append(seg.text.strip())
+        lines.append("")
+    return "\n".join(lines)
+
+
+def format_as_vtt(segments: list[TranscriptionSegment]) -> str:
+    """Render segments as a WebVTT subtitle string."""
+    lines: list[str] = ["WEBVTT", ""]
+    for seg in segments:
+        start = _format_timestamp(seg.start, ms_sep=".")
+        end = _format_timestamp(seg.end, ms_sep=".")
+        lines.append(f"{start} --> {end}")
+        lines.append(seg.text.strip())
+        lines.append("")
+    return "\n".join(lines)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ dev = [
     "mypy>=1.19.1",
 ]
 all = [
-    "vllm-metal[vllm,dev]",
+    "vllm-metal[vllm,stt,dev]",
 ]
 
 [project.urls]
--- a/tests/test_stt.py
+++ b/tests/test_stt.py
@@ -1,0 +1,179 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for STT data types, config, and formatting."""
+
+from __future__ import annotations
+
+import pytest
+
+from vllm_metal.stt.config import (
+    SpeechToTextConfig,
+    get_supported_languages,
+    get_whisper_languages,
+    validate_language,
+)
+from vllm_metal.stt.formatting import format_as_srt, format_as_vtt
+from vllm_metal.stt.protocol import TranscriptionSegment
+
+# ===========================================================================
+# SpeechToTextConfig
+# ===========================================================================
+
+
+class TestSpeechToTextConfig:
+    """Tests for SpeechToTextConfig dataclass."""
+
+    def test_default_values(self) -> None:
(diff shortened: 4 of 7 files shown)


vllm_metal/stt/audio.py
```diff
@@ -0,0 +1,213 @@
...
+        audio = mx.array(audio, mx.float32)
+
+    window = _hanning(N_FFT)
+    freqs = _stft(audio, window, N_FFT, HOP_LENGTH)
+    magnitudes = mx.abs(freqs).square()
+
+    filters = _mel_filters(SAMPLE_RATE, N_FFT, n_mels)
+    mel_spec = filters @ magnitudes
+
+    log_spec = mx.maximum(mel_spec, 1e-10).log10()
```
LxYuan0420: The audio constants at module top are fine, but the inline magic numbers need documentation or named constants, for instance:


1e-10 in log_mel_spectrogram can be written as ` _LOG_FLOOR = 1e-10` ; basically we try to make it clear why `1e-10` in this case

vllm_metal/stt/audio.py
```diff
@@ -0,0 +1,213 @@
...
+    search_radius: int | None = None,
+) -> int:
+    """Find a low-energy split point near *center*.
+
+    Searches within *search_radius* samples (default: ``window_size * 4``)
+    around *center* and returns the sample index at the start of the
+    quietest window.
+    """
+    if search_radius is None:
+        search_radius = window_size * 4
```
LxYuan0420: Why 4x?

vllm_metal/stt/audio.py
```diff
@@ -0,0 +1,213 @@
...
+        for j in range(left, center):
+            if center != left:
+                filters[i, j] = (j - left) / (center - left)
+        for j in range(center, right):
+            if right != center:
+                filters[i, j] = (right - j) / (right - center)
+
+    # Slaney normalization
+    enorm = 2.0 / (hz_points[2 : n_mels + 2] - hz_points[:n_mels])
+    return filters * enorm[:, None]
```
LxYuan0420: This part is O(n2) using nested for loop. Ok for a one-time computation but `log_mel_spectrogram` call it every invocation.

And it seems like it produces the same output for the same (sample_rate, n_fft, n_mels) tuple since they are constants? 

Consider to cache it with @lru_cache or compute it once at module level or vectorizing with mx/np operations

vllm_metal/stt/audio.py
```diff
@@ -0,0 +1,213 @@
...
+    window = _hanning(N_FFT)
+    freqs = _stft(audio, window, N_FFT, HOP_LENGTH)
+    magnitudes = mx.abs(freqs).square()
+
+    filters = _mel_filters(SAMPLE_RATE, N_FFT, n_mels)
+    mel_spec = filters @ magnitudes
+
+    log_spec = mx.maximum(mel_spec, 1e-10).log10()
+    log_spec = mx.maximum(log_spec, log_spec.max() - 8.0)
+    return (log_spec + 4.0) / 4.0
```
LxYuan0420: 8.0 and 4.0 are magic number again: would be better to park these somewhere else and make it clear the meaning

vllm_metal/stt/audio.py
```diff
@@ -0,0 +1,213 @@
...
+def _stft(audio: mx.array, window: mx.array, n_fft: int, hop_length: int) -> mx.array:
+    pad_amount = n_fft // 2
+    audio = mx.pad(audio, [(pad_amount, pad_amount)])
+    num_frames = (len(audio) - n_fft) // hop_length + 1
+    frames = mx.stack(
+        [
+            audio[i * hop_length : i * hop_length + n_fft] * window
+            for i in range(num_frames)
+        ]
+    )
```
LxYuan0420: This creates num_frames separate mx arrays in Python. For 30s audio at 16kHz, that is about ~3000 iterations of Python-level slicing

We can use mx.as_strided or equivalent vectorized windowing. The Python loop over 3000 frames is a performance bottleneck

vllm_metal/stt/formatting.py
```diff
@@ -0,0 +1,52 @@
...
+    return f"{hrs:02d}:{mins:02d}:{secs:02d},{millis:03d}"
+
+
+def _format_timestamp_vtt(seconds: float) -> str:
+    """Format *seconds* as ``HH:MM:SS.mmm`` for WebVTT."""
+    hrs = int(seconds // 3600)
+    mins = int((seconds % 3600) // 60)
+    secs = int(seconds % 60)
+    millis = int(round((seconds - int(seconds)) * 1000))
+    return f"{hrs:02d}:{mins:02d}:{secs:02d}.{millis:03d}"
```
LxYuan0420: these 2 functions differ only in the millisecond separator

vllm_metal/stt/transcribe.py
```diff
@@ -0,0 +1,374 @@
...
+        except Exception as e:
+            logger.debug("Local tokenizer load failed for %s: %s", model_path, e)
+
+    # Fall back to openai/whisper-small (online or cached)
+    try:
+        return WhisperTokenizer.from_pretrained("openai/whisper-small")
+    except OSError:
+        return WhisperTokenizer.from_pretrained(
+            "openai/whisper-small", local_files_only=True
+        )
```
LxYuan0420: _get_tokenizer() is a module-level @lru_cache singleton used by both the class methods AND the free functions. This means _extract_segments and _encode_prompt silently depend on global state. If you ever need two models with different tokenizers loaded, this breaks

vllm_metal/stt/transcribe.py
```diff
@@ -0,0 +1,374 @@
...
+
+    config = WhisperConfig.from_dict(config_dict)
+    model = WhisperModel(config, dtype)
+
+    # Load weights
+    weight_files = glob.glob(str(model_path / "*.safetensors"))
+    if not weight_files:
+        weight_files = glob.glob(str(model_path / "*.npz"))
+    if not weight_files:
+        raise FileNotFoundError(f"No weight files in {model_path}")
```
LxYuan0420: since you are already using Path everywhere else. list(model_path.glob("*.safetensors")) is more consistent. Also, what happens if there are BOTH safetensors and npz files?

vllm_metal/stt/transcribe.py
LxYuan0420, line 1: The WhisperTranscriber class is a thin shell that delegates everything to module-level free functions `(_greedy_decode, _encode_prompt, _extract_segments, _encode_chunk)`. The class doesn't encapsulate anything meaningful so it just calls the free functions and passes self.model to them. 

This is the worst of both worlds: you have a class that provides no encapsulation, and free functions that rely on a global cached tokenizer instead of instance state

LxYuan0420 (changes requested):
This is a lot of good foundational work, but I cant merge a 1,660-line PR as it's too much to review thoroughly and too risky to merge as one unit.

 Let's break it up:
1. Start with the data types and config. These are leaf modules with no internal dependencies
2. Then the audio pipeline. We can benchmark it and iterate on the vectorization issues before it gets entangled with everything else
3. Then the model (whisper.py).
4. Finally the transcription orchestration (transcribe.py) that wires it all together.

Each PR should be independently testable including the relevant subset of test_stt.py with each one. This way we can iterate faster and you won't end up resubmitting the same large diff

ricky-chaoju:
Restructured this PR to only include data types, config, and formatting (~400 lines, 18 tests). The audio pipeline, model, and transcriber will follow in separate PRs with all feedback addressed.

docs/stt.md
```diff
@@ -4,9 +4,16 @@ vllm-metal supports OpenAI-compatible Speech-to-Text using Whisper models, runni
...
+First, install vllm-metal using the install script (see [README](../README.md)):
 
 ```bash
+./install.sh
+```
+
+Then install the optional STT dependencies inside the virtual environment:
+
+```bash
+source .venv/bin/activate
```
LxYuan0420, line 16: Nit: activation path should match our install flow (it’s .venv-vllm-metal, not .venv)

You can fix this in future PR

LxYuan0420 (approved):
Approving so we can merge and unblock the rest of the STT work

Issue #83, opened 2026-01-27
ImportError: cannot import name 'ALLOWED_LAYER_TYPES'


KennethEnevoldsen:
Hi, I would love to use this, but I can't seem to get it to work. I suspect that it might be a transformers versioning issue, but I can't see fix it.

```bash
$ python                            
Python 3.12.0 (main, Oct  2 2023, 20:56:14) [Clang 16.0.3 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from vllm import LLM
INFO 01-27 16:58:16 [__init__.py:43] Available plugins for group vllm.platform_plugins:
INFO 01-27 16:58:16 [__init__.py:45] - metal -> vllm_metal:register
INFO 01-27 16:58:16 [__init__.py:48] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 01-27 16:58:16 [__init__.py:217] Platform plugin metal is activated
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/au561649/.venv-vllm-metal/lib/python3.12/site-packages/vllm/__init__.py", line 74, in __getattr__
    module = import_module(module_name, __package__)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/au561649/.local/share/uv/python/cpython-3.12.0-macos-aarch64-none/lib/python3.12/importlib/__init__.py", line 95, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/au561649/.venv-vllm-metal/lib/python3.12/site-packages/vllm/entrypoints/llm.py", line 20, in <module>
    from vllm.config import (
  File "/Users/au561649/.venv-vllm-metal/lib/python3.12/site-packages/vllm/config/__init__.py", line 18, in <module>
    from vllm.config.model import (
  File "/Users/au561649/.venv-vllm-metal/lib/python3.12/site-packages/vllm/config/model.py", line 14, in <module>
    from transformers.configuration_utils import ALLOWED_LAYER_TYPES
[...]

PR #99, opened 2026-02-13
vLLM 0.14.x compatibility: async sample_tokens fix + dependency cleanup

krystophny:
## Summary

1. Upgrade vLLM target version
- `install.sh`: `vllm_v` from `0.13.0` to `0.14.1`
- `pyproject.toml` optional extra: `vllm>=0.13.0` to `vllm>=0.14.0`

2. Align imports with vLLM 0.14 API moves
- `vllm_metal/platform.py`: `vllm.attention.*` -> `vllm.v1.attention.*`
- `vllm_metal/v1/worker.py`: `set_random_seed` import to `vllm.utils.torch_utils`
- `tests/test_platform.py`: update imports to `vllm.v1.attention.*`

3. Fix async scheduling runtime path
- `vllm_metal/v1/worker.py`: implement `sample_tokens(...)`
- `vllm_metal/v1/model_runner.py`:
  - store pending output in `execute_model(...)`
  - return it in `sample_tokens(...)`

## Why

vLLM 0.14 changes API paths and the execution flow in async scheduling.
Without these updates, plugin import/runtime compatibility regresses on v0.14.x.

## Validation

Ran locally on macOS/Apple Silicon:

- `./scripts/lint.sh`
  - `ruff check`: pass
  - `ruff format --check`: pass
  - `mypy`: pass
- `./scripts/test.sh`
  - smoke test chat completion: passed
  - pytest suite: `110 passed, 1 deselected`
- `pytest -m "not slow" tests/ -q --tb=short`
  - `110 passed, 1 deselected`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-vllm = ["vllm>=0.13.0"]
+vllm = ["vllm>=0.14.0"]
 stt = [
     # Speech-to-text audio processing (Whisper models)
     "librosa>=0.10.2",
--- a/install.sh
+++ b/install.sh
@@ -123,7 +123,7 @@ main() {
 
   ensure_venv "$venv"
 
-  local vllm_v="0.13.0"
+  local vllm_v="0.14.1"
   local url_base="https://github.com/vllm-project/vllm/releases/download"
   local filename="vllm-$vllm_v.tar.gz"
   curl -OL $url_base/v$vllm_v/$filename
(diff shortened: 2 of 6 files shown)


Faylixe:
Any update on this ? Current version doesn't work due to https://github.com/vllm-project/vllm-metal/issues/83 when installing with the `install.sh` script, and I believe this would fix it :)

I would gladly help if needed but I feel like the job is already done :D.

pyproject.toml
```diff
@@ -41,7 +41,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-vllm = ["vllm>=0.13.0"]
```
ericcurtin, line 44: The better long-term fix I think is it put an == here, so we don't suddenly break from time to time

PR #105, opened 2026-02-20
Ensure install venv always provides pip entrypoints

tburleyinfo:
When using the venv provided by vllm-metal you can't install pip dependencies in the environment for projects that might incorporate vllm-metal because pip seems to be missing. This fix is to bootstrap pip with ensurepip when missing and create pip/pip3 wrappers in scripts/lib.sh ensure_venv.

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -39,38 +39,11 @@ ensure_uv() {
 ensure_venv() {
   if [ ! -d "$1" ]; then
     section "Creating virtual environment"
-    uv venv "$1" --clear --python 3.12
+    uv venv "$1" --clear --python 3.12 --seed
   fi
 
   # shellcheck source=/dev/null
   source "$1/bin/activate"
-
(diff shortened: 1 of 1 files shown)


scripts/lib.sh
```diff
@@ -44,6 +44,33 @@ ensure_venv() {
...
+    chmod +x "$1/bin/pip"
+  fi
+
+  if [ ! -x "$1/bin/pip3" ]; then
+    cat > "$1/bin/pip3" <<'EOF'
+#!/usr/bin/env sh
+exec "$(dirname "$0")/python" -m pip "$@"
+EOF
+    chmod +x "$1/bin/pip3"
+  fi
```
LxYuan0420: I think we can simplify this by using uv built-in seeding instead of bootstrapping via ensurepip + writing wrapper scripts. Something like:

```diff
- uv venv "$1" --clear --python 3.12
+ uv venv "$1" --clear --python 3.12 --seed
```

LxYuan0420 (approved):

PR #116, opened 2026-02-24
chore: Update dependencies

reneleonhardt:
### Chores
- Update cargo dependencies and GitHub Actions (use pinning to improve supply chain security)
- Let Dependabot update `cargo`, `github-actions` and `uv` (use cooldown to improve supply chain security)
- Update install.sh to latest stable `vllm 0.16.0`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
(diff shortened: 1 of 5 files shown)


ericcurtin:
> ### Notes
> * Dependabot `uv` is a "no-op" for now (no uv.lock, no pinned dependencies)
> * Could install.sh be hardened (check hashes and `cosign verify`)?
> * Instead of building from source in install.sh, could a vllm wheel be installed?

We are missing a macOS wheel from vllm core project to make this happen, please poke the maintainers there

PR #117, opened 2026-02-25
[Bugfix] RotatingKVCache merge broadcast shape error in batched decode

ricky-chaoju:
## Summary
- Fix `BatchRotatingKVCache.merge` crash under concurrent multi-turn load with models using `RotatingKVCache` (e.g. gpt-oss-20b)
- Handle missing pending output in `sample_tokens` for async scheduling (credit: @LxYuan0420)

## Cause
`BatchRotatingKVCache.merge` in mlx-lm 0.29.1 uses `c.offset` (total tokens ever seen) as the slice width. After the cache rotates past `max_size`, `c.offset` can exceed the actual array size, causing:

```ValueError: [broadcast_shapes] Shapes (1,8,311,64) and (1,8,128,64) cannot be broadcast```

This was fixed upstream in ml-explore/mlx-lm#738, but upgrading to mlx-lm 0.30.8 currently introduces a separate crash in `RotatingKVCache._update_in_place`, so upgrading is not viable for now.

## Changes
- Add `_merge_rotating_kv_caches` workaround that pre-computes temporally ordered keys/values and uses their actual shape for the copy width (same pattern as `_merge_arrays_caches`)
- Add regression test covering the `offset > max_size` merge scenario
- Make `sample_tokens` return `None` under async scheduling when `execute_model` fails, allowing vLLM to surface the original exception

Related #111 #114

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -359,11 +359,10 @@ def _merge_rotating_kv_caches(
     if not caches:
         raise ValueError("caches must be non-empty")
 
-    for c in caches:
-        if c.keys is None or c.values is None:
-            raise ValueError(
-                "Cannot merge unpopulated RotatingKVCache (keys/values is None)"
-            )
+    if any(c.keys is None or c.values is None for c in caches):
+        raise ValueError(
+            "Cannot merge unpopulated RotatingKVCache (keys/values is None)"
+        )
 
     if not all(c.max_size == caches[0].max_size for c in caches):
         raise ValueError(
@@ -373,14 +372,21 @@ def _merge_rotating_kv_caches(
     # Pre-compute temporal-ordered keys/values and trim to the effective
     # sliding-window length.  ``_temporal_order`` may return an array larger
     # than ``len(cache)`` when the internal buffer has not been trimmed yet
-    # (e.g. after a large prefill), so we slice to ``len(cache)`` which is
-    # ``min(offset, max_size)``.
+    # (e.g. after a large prefill), so we trim via ``_trim`` to preserve
+    # the ``keep`` prefix semantics used by RotatingKVCache internally.
     ordered: list[tuple[mx.array, mx.array]] = []
     for c in caches:
-        n = len(c)  # effective length: min(offset, max_size)
-        k = c._temporal_order(c.keys)[..., :n, :]
-        v = c._temporal_order(c.values)[..., :n, :]
-        ordered.append((k, v))
+        effective_len = len(c)  # min(offset, max_size)
+        ordered_keys = c._temporal_order(c.keys)
+        ordered_values = c._temporal_order(c.values)
+        if ordered_keys.shape[2] > effective_len:
+            trim_size = ordered_keys.shape[2] - effective_len
+            ordered_keys = c._trim(trim_size, ordered_keys)
+            ordered_values = c._trim(trim_size, ordered_values)
+        else:
+            ordered_keys = ordered_keys[..., :effective_len, :]
+            ordered_values = ordered_values[..., :effective_len, :]
+        ordered.append((ordered_keys, ordered_values))
 
     lengths = [k.shape[2] for k, _ in ordered]
     max_length = max(lengths)

### CODE REVIEW

File: vllm_metal/v1/model_runner.py [hunk 1]
```diff
@@ -335,6 +341,52 @@ def _extract_arrays_cache(batch_cache: ArraysCache, idx: int) -> ArraysCache:
...
+) -> BatchRotatingKVCache:
+    """Merge per-request RotatingKVCache objects into a single BatchRotatingKVCache.
+
+    This mirrors ``BatchRotatingKVCache.merge`` but pre-computes the temporal-ordered
+    keys/values and uses their actual shape for the copy width.  The upstream
+    implementation uses ``c.offset`` which can exceed the underlying array size
+    after the cache has rotated, causing a broadcast shape error.
+    """
+    if not caches:
+        raise ValueError("caches must be non-empty")
```
LxYuan0420, line 360: Can we add a fail-fast guard for empty caches? _temporal_order(c.keys) will crash with a less actionable exception if any RotatingKVCache hasn’t been populated yet (keys/values is None). A clear ValueError would be much easier to debug

vllm_metal/v1/model_runner.py
```diff
@@ -335,6 +341,52 @@ def _extract_arrays_cache(batch_cache: ArraysCache, idx: int) -> ArraysCache:
...
+    # Pre-compute temporal-ordered keys/values so we know the real size of each.
+    ordered = [(c._temporal_order(c.keys), c._temporal_order(c.values)) for c in caches]
+    lengths = [k.shape[2] for k, _ in ordered]
+    max_length = max(lengths)
+    padding = [max_length - length for length in lengths]
+    batch_size = len(caches)
+    n_heads = max(k.shape[1] for k, _ in ordered)
+    k_dim = max(k.shape[3] for k, _ in ordered)
+    v_dim = max(v.shape[3] for _, v in ordered)
+    dtype = next(iter(k.dtype for k, _ in ordered))
```
LxYuan0420: I think _merge_rotating_kv_caches still needs to batch by the effective sliding-window length (len(cache)), not k.shape[2]. In real prefill paths RotatingKVCache(max_size=70) can temporarily hold keys.shape[2]==128 while len(cache)==70, which is the broadcast_shapes (…128…) vs (…70…) crash. Suggest trimming the temporal-ordered buffers down to len(cache) (ideally via _trim to preserve keep) before copying.

Reference: mlx-lm RotatingKVCache.__len__ is min(offset, max_size); see mlx_lm/models/cache.py in the RotatingKVCache class : https://github.com/ml-explore/mlx-lm/blob/v0.29.0/mlx_lm/models/cache.py#L486

Take note of the package version when we read the source code

vllm_metal/v1/model_runner.py
```diff
@@ -335,6 +341,52 @@ def _extract_arrays_cache(batch_cache: ArraysCache, idx: int) -> ArraysCache:
...
+def _merge_rotating_kv_caches(
+    caches: list[RotatingKVCache],
+) -> BatchRotatingKVCache:
+    """Merge per-request RotatingKVCache objects into a single BatchRotatingKVCache.
+
+    This mirrors ``BatchRotatingKVCache.merge`` but pre-computes the temporal-ordered
+    keys/values and uses their actual shape for the copy width.  The upstream
+    implementation uses ``c.offset`` which can exceed the underlying array size
+    after the cache has rotated, causing a broadcast shape error.
+    """
```
LxYuan0420, line 358: Could we also add a short note here about when we can delete this workaround (e.g. once we can bump mlx-lm to a version that includes the upstream fix and we have verified gpt-oss works)? This will help keep the workaround maintainable

vllm_metal/v1/model_runner.py
```diff
@@ -335,6 +341,52 @@ def _extract_arrays_cache(batch_cache: ArraysCache, idx: int) -> ArraysCache:
...
+        n = k.shape[2]
+        keys[i : i + 1, :, pad : pad + n] = k
+        values[i : i + 1, :, pad : pad + n] = v
+
+    cache = BatchRotatingKVCache(caches[0].max_size, padding)
+    cache.keys = keys
+    cache.values = values
+    cache.offset = mx.array([c.offset for c in caches])
+    cache._idx = keys.shape[2]
+    cache._offset = keys.shape[2]
```
LxYuan0420, line 412: Since _merge_rotating_kv_caches manually sets BatchRotatingKVCache internals (_idx/_offset), can we add a regression test that does merge -> update_and_fetch(S=1) -> extract on rotated inputs? That would prove the merged cache is correct beyond just merge/extract.

tests/test_prefix_cache.py
```diff
@@ -214,6 +232,50 @@ def test_mixed_kv_and_arrays_cache_merge_extract_roundtrip(self) -> None:
...
+        extracted_req0 = mr._extract_kv_cache(merged, 0)[0]
+        extracted_req1 = mr._extract_kv_cache(merged, 1)[0]
+
+        assert isinstance(merged[0], mr.BatchRotatingKVCache)
+        assert isinstance(extracted_req0, mr.RotatingKVCache)
+        assert isinstance(extracted_req1, mr.RotatingKVCache)
+        assert extracted_req0.offset == cache_req0.offset
+        assert extracted_req1.offset == cache_req1.offset
+
+    def test_rotating_kvcache_merge_handles_offset_exceeding_max_size(self) -> None:
```
LxYuan0420, line 251: Nice regression for offset > max_size after rotation and I like the AAA style.

Could we also add a test for prefill_len > max_size (e.g. first update with S=128, max_size=70) and ensure merge/extract works? That’s the other common trigger for the broadcast-shape mismatch

vllm_metal/v1/model_runner.py
```diff
@@ -347,20 +347,41 @@ def _merge_rotating_kv_caches(
...
+    to work with gpt-oss models end-to-end.
     """
     if not caches:
         raise ValueError("caches must be non-empty")
 
+    for c in caches:
+        if c.keys is None or c.values is None:
+            raise ValueError(
+                "Cannot merge unpopulated RotatingKVCache (keys/values is None)"
+            )
```
LxYuan0420: Can we do it like:

```python
if any(c.keys is None or c.values is None for c in caches):
    raise ValueError("Cannot merge unpopulated RotatingKVCache (keys/values is None)")
```

vllm_metal/v1/model_runner.py
```diff
@@ -347,20 +347,41 @@ def _merge_rotating_kv_caches(
...
+    # than ``len(cache)`` when the internal buffer has not been trimmed yet
+    # (e.g. after a large prefill), so we slice to ``len(cache)`` which is
+    # ``min(offset, max_size)``.
+    ordered: list[tuple[mx.array, mx.array]] = []
+    for c in caches:
+        n = len(c)  # effective length: min(offset, max_size)
+        k = c._temporal_order(c.keys)[..., :n, :]
+        v = c._temporal_order(c.values)[..., :n, :]
+        ordered.append((k, v))
+
```
LxYuan0420, line 390: Can we trim via RotatingKVCache._trim(...) instead of [..., :n, :]? 

That preserves mlx-lm keep semantics and matches how rotating caches are trimmed internally

Something like

```python

n = len(c)
k = c._temporal_order(c.keys)
v = c._temporal_order(c.values)
if k.shape[2] > n:
    trim = k.shape[2] - n
    k = c._trim(trim, k)
    v = c._trim(trim, v)
else:
    k = k[..., :n, :]
    v = v[..., :n, :]
```

Note: please change to use descriptive variable name in this case

ricky-chaoju:
update both in 7dfae52:                                                                                                                                          
  1. Guard changed to any() one-liner                                                                                             
  2. Trim now uses c._trim() to preserve keep prefix semantics instead of plain slice

LxYuan0420 (approved):

PR #121, opened 2026-02-27
Fix batched decode crash for hybrid cache models (Qwen3.5)

laudney:
## Summary

Interim fix for the **mlx-native (non-paged) path** so hybrid models like Qwen3.5 can serve today.

- Hybrid models (e.g. Qwen3.5-35B-A3B) use mixed cache types (`ArraysCache` for linear/SSM layers + `KVCache` for attention layers). `BatchKVCache.offset` returns `mx.array` but hybrid attention code uses `cache.offset` as a Python `int` for mask slicing, causing `ValueError: Slice indices must be integers or None.`
- Detects hybrid caches at model load time via `make_prompt_cache()` and falls back to sequential decode for incompatible models.
- Core detection logic in `vllm_metal/v1/cache_utils.py` (standalone pure function), keeping `model_runner.py` changes minimal per #122.

**NOTE:** This is a band-aid until per-layer attention dispatching (#201) and a paged linear attention kernel (roadmap #148) land, at which point hybrid models will be handled properly at the dispatch level.

## Test plan
- [x] Verified fix resolves `ValueError` when serving Qwen3.5 via vllm-metal
- [x] Verified standard (non-hybrid) models still use batched decode path (Qwen3-0.6B: all `KVCache` → `True`)
- [x] Verified Qwen3.5-35B-A3B hybrid model falls back gracefully (`{ArraysCache, KVCache}` → `False`)

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -61,6 +61,7 @@ from vllm_metal.stt.config import (
 from vllm_metal.stt.runtime import STTRuntimeAdapter
 from vllm_metal.stt.serve import VLLMSTTRequestAdapter
 from vllm_metal.utils import get_model_download_path
+from vllm_metal.v1.cache_utils import supports_batched_decode
 
 logger = init_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


vllm_metal/v1/model_runner.py
```diff
@@ -664,6 +664,13 @@ def is_stt(self) -> bool:
         """Whether the loaded model is a Speech-to-Text model."""
         return self._is_stt
 
+        # Whether the model supports batched decode via BatchKVCache.
+        # Hybrid models (e.g. Qwen3.5 with mixed attention/linear layers)
+        # use ArraysCache for some layers, and their attention code assumes
+        # scalar cache.offset which is incompatible with BatchKVCache's
+        # per-element mx.array offset.  Determined in load_model().
+        self._supports_batched_decode: bool = True
+
```
WindChimeRan, line 704: unreachable code?

laudney: Fixed — the init was placed after a `return` in the `is_stt` property. Now correctly in `__init__`.

PR #123, opened 2026-02-28
Add Qwen3.5 model support

otarkhan:
## Summary

- **Upgrade vLLM to main branch** which includes the Qwen3.5 model registry (`Qwen3_5MoeForConditionalGeneration`, `Qwen3_5ForConditionalGeneration`)
- **Upgrade mlx-lm (>=0.30.7) and mlx-vlm (>=0.3.12)** which include Qwen3.5 and Qwen3.5-MoE model implementations
- **Upgrade transformers (>=5.2.0)** as required by newer mlx-lm/mlx-vlm
- **Patch rope validation compatibility** between vLLM and transformers 5.x — vLLM's model configs pass `ignore_keys_at_rope_validation` as a list, but transformers 5.x expects a set (uses the `|` operator). The patch coerces list→set in `convert_rope_params_to_dict` and is idempotent (guarded against double-application).

## Test plan

- [x] Verified `vllm serve` with `Qwen3.5-397B-A17B-4bit` (MoE, multimodal) loads and serves successfully
- [x] Verified `vllm serve` with `Qwen3-Coder-Next` loads successfully

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -123,16 +123,22 @@ main() {
 
   ensure_venv "$venv"
 
-  local vllm_v="0.14.1"
-  local url_base="https://github.com/vllm-project/vllm/releases/download"
-  local filename="vllm-$vllm_v.tar.gz"
-  curl -OL $url_base/v$vllm_v/$filename
-  tar xf $filename
-  cd vllm-$vllm_v
+  local vllm_repo="https://github.com/vllm-project/vllm"
+  local vllm_ref="main"
+  local vllm_dir="vllm-${vllm_ref}"
+  if ! curl -fsSL "$vllm_repo/archive/refs/heads/${vllm_ref}.tar.gz" -o "${vllm_dir}.tar.gz"; then
+    error "Failed to download vLLM source."
+    exit 1
+  fi
+  tar xf "${vllm_dir}.tar.gz"
+  cd "$vllm_dir"
   uv pip install -r requirements/cpu.txt --index-strategy unsafe-best-match
-  uv pip install .
-  cd -
-  rm -rf vllm-$vllm_v*
+  SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0.dev0 uv pip install .
+  cd - > /dev/null
+  rm -rf "${vllm_dir}" "${vllm_dir}.tar.gz"
+
+  # Upgrade dependencies for newer model support (e.g., Qwen3.5)
+  uv pip install 'mlx-lm>=0.30.7' 'mlx-vlm>=0.3.12' 'transformers>=5.2.0'
 
   if [[ -n "$local_lib" && -f "$local_lib" ]]; then
     uv pip install .
--- a/vllm_metal/__init__.py
+++ b/vllm_metal/__init__.py
@@ -99,7 +99,35 @@ def _register_ops() -> None:
     """Register Metal operations with vLLM.
 
     This is the entry point for vLLM's general plugin system.
-    Currently a no-op as operations are handled internally.
+    Also applies compatibility patches between vLLM and transformers.
     """
-    # Operations are registered implicitly through the MLX backend
-    pass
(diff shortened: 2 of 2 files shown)


install.sh
```diff
@@ -123,16 +123,22 @@ main() {
...
   ensure_venv "$venv"
 
-  local vllm_v="0.14.1"
-  local url_base="https://github.com/vllm-project/vllm/releases/download"
-  local filename="vllm-$vllm_v.tar.gz"
-  curl -OL $url_base/v$vllm_v/$filename
-  tar xf $filename
-  cd vllm-$vllm_v
+  local vllm_repo="https://github.com/vllm-project/vllm"
+  local vllm_ref="main"
```
LxYuan0420, line 127: This switches installer from a pinned vLLM release to building from vllm@main. That’s not reproducible and is too risky for the default install path. 

Can we keep install.sh pinned to a known-good vLLM tag/commit, and move “Qwen3.5 requires vLLM main + transformers 5.x” into an opt-in path (separate extra)?

install.sh
```diff
@@ -123,16 +123,22 @@ main() {
...
   uv pip install -r requirements/cpu.txt --index-strategy unsafe-best-match
-  uv pip install .
-  cd -
-  rm -rf vllm-$vllm_v*
+  SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0.dev0 uv pip install .
+  cd - > /dev/null
+  rm -rf "${vllm_dir}" "${vllm_dir}.tar.gz"
+
+  # Upgrade dependencies for newer model support (e.g., Qwen3.5)
+  uv pip install 'mlx-lm>=0.30.7' 'mlx-vlm>=0.3.12' 'transformers>=5.2.0'
```
LxYuan0420, line 141: Please dont force-upgrade these globally in the default installer; it will drift users’ environments and can break unrelated models. 

Either make the Qwen3.5 dependency bumps opt-in (extra/flag like .[qwen35]), or add enough testing to show the new default stack is stable.

vllm_metal/__init__.py
```diff
@@ -99,7 +99,35 @@ def _register_ops() -> None:
...
+        if isinstance(ignore_keys_at_rope_validation, list):
+            ignore_keys_at_rope_validation = set(ignore_keys_at_rope_validation)
+        return orig(
+            self,
+            ignore_keys_at_rope_validation=ignore_keys_at_rope_validation,
+            **kwargs,
+        )
+
+    _patched._metal_patched = True
+    RotaryEmbeddingConfigMixin.convert_rope_params_to_dict = _patched
```
LxYuan0420, line 133: I’m wary of a global monkeypatch in __init__.py. If we keep this, can we (1) isolate it in a compat module, (2) guard by transformers version / feature detection, and (3) add a focused unit test proving the failure + fix? Otherwise this is a hidden global behavior change.

LxYuan0420 (changes requested):
Nice work but few comments.

Could you split this into smaller pieces?

- Keep install.sh pinned to a known-good vLLM release/commit (no main), and make the Qwen3.5 dependency bumps opt-in (extra/flag/separate script).
- If we need the rope compatibility shim, please isolate it (compat module), add tight guards, and include a unit test demonstrating the list then set mismatch and confirming the fix.


One extra concern: this switches the installer to vllm@main + newer mlx-lm/mlx-vlm/transformers, which can drift the torch stack (we’ve previously hit torch/torchvision mismatches on macOS when vLLM bumps pulled torch 2.10). 

Also, mlx-lm upgrades can include breaking API changes: we already saw CI failures from the RotatingKVCache API change (len(cache) → cache.size() in mlx-lm v0.30.7). That’s exactly why we need to be extra careful with “upgrade the stack” PRs.

https://github.com/ml-explore/mlx-lm/blob/v0.30.7/mlx_lm/models/cache.py#L494

PR #124, opened 2026-02-28
Enable custom logits processors via build_logitsprocs()

ereid7:
## Summary

`MetalModelRunner` currently creates an empty `LogitsProcessors()` instance, bypassing vLLM's plugin discovery system. This means custom logits processors registered via the `vllm.logits_processors` entry-point group are never loaded or applied on Metal.

This PR mirrors `GPUModelRunner`'s pattern:
- Call `build_logitsprocs()` in `__init__` to discover and instantiate both built-in and entry-point logits processors
- Pass the populated instance to `SamplingMetadata` instead of an empty one

## Changes

**`vllm_metal/v1/model_runner.py`** — 9 insertions, 2 deletions:
1. Import `build_logitsprocs` alongside `LogitsProcessors`
2. In `__init__`, call `build_logitsprocs(vllm_config, device, ...)` and store as `self._logitsprocs`
3. In `_make_sampling_metadata`, use `getattr(self, '_logitsprocs', LogitsProcessors())` for backward compatibility with tests that construct via `__new__`

## Why this matters

vLLM V1 supports [custom logits processors](https://docs.vllm.ai/en/stable/features/custom_logitsprocs/) via entry-points. Without this fix, any package that registers under `vllm.logits_processors` (e.g., guided decoding plugins, custom sampling strategies) silently does nothing on Metal — the processors are discovered by `importlib.metadata.entry_points()` but never instantiated or passed to the sampler.

## Testing

- All 105 existing tests pass (23 sampling tests + 82 others)
- Verified end-to-end with a custom logits processor plugin registered via entry-points — plugin loads, initializes with correct vocab size, and processes every token during inference on Apple Silicon (M-series)

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -1029,7 +1029,7 @@ class MetalModelRunner:
             no_penalties=no_penalties,
             allowed_token_ids_mask=None,
             bad_words_token_ids={},
-            logitsprocs=getattr(self, '_logitsprocs', LogitsProcessors()),
+            logitsprocs=getattr(self, "_logitsprocs", None) or LogitsProcessors(),
         )
 
     def _prefill_single(


vllm_metal/v1/model_runner.py
```diff
@@ -587,6 +587,13 @@ def __init__(
         # vLLM Sampler for token sampling with temperature, top_k, top_p support
         self._sampler = Sampler()
 
+        # Build logits processors (includes custom plugins from entry-points)
+        custom_lp = vllm_config.model_config.logits_processors
+        custom_logitsprocs = tuple(custom_lp) if custom_lp is not None else ()
+        self._logitsprocs = build_logitsprocs(
+            vllm_config, device, False, False, custom_logitsprocs,
```
LxYuan0420: Can we avoid hardcoding the is_pin_memory / is_pooling_model flags passed to build_logitsprocs? 

Upstream derives these (pin memory via vllm.utils.platform_utils.is_pin_memory_available(), pooling via model_config.runner_type == "pooling")

Even if Metal ends up false for pinning, computing it keeps us aligned with vLLM semantics and avoids surprises for pooling runners.

vllm_metal/v1/model_runner.py
```diff
@@ -1015,7 +1022,7 @@ def _make_sampling_metadata(
             no_penalties=no_penalties,
             allowed_token_ids_mask=None,
             bad_words_token_ids={},
-            logitsprocs=LogitsProcessors(),
+            logitsprocs=getattr(self, '_logitsprocs', LogitsProcessors()),
```
LxYuan0420: Can we add a small unit test that asserts _make_sampling_metadata() uses self._logitsprocs when present (and falls back to an empty LogitsProcessors() when __init__ is bypassed via __new__)? 

This locks in the behavior without needing a real entry-point package installed in tests.

LxYuan0420 (changes requested):
Nice work! A couple minor tweaks and we are good to go

vllm_metal/v1/model_runner.py
ereid7: Addressed in 15ed369.

`MetalModelRunner.__init__` now derives the flags the same way upstream does:
- `pin_memory = is_pin_memory_available()`
- `is_pooling_model = getattr(self.model_config, "runner_type", None) == "pooling"`

These are now passed into `build_logitsprocs(...)` instead of hardcoded `False, False`.

vllm_metal/v1/model_runner.py
ereid7: Added in 15ed369.

I added `TestV1SamplingMetadataLogitsProcessors` in `tests/test_sampling.py` with both requested checks:
- `test_make_sampling_metadata_uses_runner_logitsprocs`
- `test_make_sampling_metadata_falls_back_without_init`

This verifies `_make_sampling_metadata()` uses `self._logitsprocs` when present and falls back to an empty `LogitsProcessors()` when `__init__` is bypassed via `__new__`.

I also ran `pytest tests/test_sampling.py -q` locally (`25 passed`).

ereid7:
@LxYuan0420 Addressed feedback. Let me know if you have any more comments.

vllm_metal/v1/model_runner.py
```diff
@@ -1015,7 +1029,7 @@ def _make_sampling_metadata(
...
```
LxYuan0420: Minor minor performance nit: `LogitsProcessors()` gets created every call even when _logitsprocs exists, because Python builds the default value before checking if it's needed. Use `getattr(self, '_logitsprocs', None) or LogitsProcessors()` so the fallback is only created when the attribute is missing.

ereid7: Thanks for fixing!

Issue #119, opened 2026-02-27
Metal paged-attention parity mismatch vs standard path


LxYuan0420:
Two slow tests fail with token mismatches between the standard MLX cache path and the Metal paged-attention kernel path.

Repro (on main):

  source .venv-vllm-metal/bin/activate
  pytest -m slow -q tests/test_metal_kernel_paged.py

Failures:
- TestMetalKernelPagedVsStandard.test_greedy_output_matches
- TestMetalKernelPagedVsStandard.test_batched_decode_matches

Example mismatch (from local run):
Standard tokens: [12095, 13, 576, 6722, 315, 9625, ...]
Metal kernel:   [12095, 13, 576, 6722, 315, 15344, ...]

Likely kernel/cache parity issue. Tracking here; will temporarily xfail these tests in the PR.

WindChimeRan:
Ideally, the outputs of the two paths should be identical. However, due to
implementation differences under the hood of paged attention, e.g., floating-point
non-associativity, softmax numerical ordering. There can be token-level divergences.

These differences may or may not indicate an actual correctness issue in the
algorithm.

One quantitative way to assess correctness (beyond exact token matching) is to run
both paths on a benchmark dataset and compare aggregate metrics like response length
distribution and perplexity. If the paged path produces statistically similar
outputs, that's a strong signal the implementation is sound even when individual
tokens diverge.

I'll follow up with a more detailed investigation into the root cause of the specific
 mismatches reported here.

PR #125, opened 2026-03-01
Fix paged-attention KV cache dtype + size accounting (issue #119)

LxYuan0420:
This PR is:
- To align the Metal paged-attention KV cache dtype with the model's dtype (fixes batched decode parity for #119).
- To compute KV cache byte sizes via `torch.dtype.itemsize` instead of allocating temporary tensors.

Notes:
- `tests/test_metal_kernel_paged.py::test_batched_decode_matches` now passes.
- `tests/test_metal_kernel_paged.py::test_greedy_output_matches` remains xfailed (tracked in #119). This is a remaining single-request greedy parity mismatch between the paged-kernel path and the standard path; fixing it likely requires deeper kernel/offset semantics work, so I'm keeping it out of this PR to keep scope tight.

Quick manual smoke test:

Terminal 1:
```bash
vllm serve Qwen/Qwen3-0.6B --host 127.0.0.1 --port 8000 --max-model-len 2048
```

Terminal 2 (single request):
```bash
curl -fsS http://127.0.0.1:8000/v1/chat/completions \
  -d '{"model":"Qwen/Qwen3-0.6B","messages":[{"role":"user","content":"Write a 2-sentence apple story."}],"max_tokens":512,"temperature":0.8}' \
| jq -r '.choices[0].message.content'
```

Terminal 2 (concurrent 4 requests):
```bash
for i in 1 2 3 4; do
  (
    echo "===== req $i ====="
    curl -fsS http://127.0.0.1:8000/v1/chat/completions \
      -d "{\"model\":\"Qwen/Qwen3-0.6B\",\"messages\":[{\"role\":\"user\",\"content\":\"Write a 2-sentence apple story (${i}).\"}],\"max_tokens\":256,\"temperature\":0.8}" \
    | jq -r '.choices[0].message.content'
    echo
  ) &
done
wait
```

Related: #119

### DIFF
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -302,6 +302,7 @@ class TestHybridCacheMergeExtract:
 
         assert isinstance(merged[0], mr.BatchRotatingKVCache)
         assert isinstance(extracted_req0, mr.RotatingKVCache)
+        assert isinstance(extracted_req1, mr.RotatingKVCache)
         assert extracted_req0.offset == cache_req0.offset
         assert extracted_req1.offset == cache_req1.offset
 


LxYuan0420:
I think the current result is good enough to show stability under pressure. This PR is a bug fix, not a performance improvement: it fixes the block-exhaustion failure (RuntimeError: Not enough free blocks) by aligning allocation/preemption behavior with scheduler-driven recompute

@ericcurtin  PTAL

PR #127, opened 2026-03-01
[Paged KV] Add ShareGPT avg generation length tool

WindChimeRan:
## Summary
 Add a diagnostic tool for paged KV cache development. Runs offline vLLM inference on ShareGPT prompts and reports response length statistics (mean/std tokens). 

The intended workflow is to compare the non-paged (standard MLX cache) and paged (Metal kernel) paths: if a KV cache bugfix improves alignment between the two distributions, that's a strong signal the fix is correct. 

Also useful for comparing across batch sizes (--max-num-seqs 1 vs 8) to verify batched decode consistency. Related: #119

I learnt this from https://arxiv.org/pdf/2601.11580 table 1. 

relevant post from thinking machine lab: https://thinkingmachines.ai/blog/defeating-nondeterminism-in-llm-inference/

## Usage
```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.7 \
    python tools/avg_gen_length.py --max-num-seqs 1 8
```

### DIFF
--- a/tools/avg_gen_length.py
+++ b/tools/avg_gen_length.py
@@ -4,7 +4,7 @@
 
 Usage:
     # Download ShareGPT dataset (~642 MB):
-    huggingface-cli download anon8231489123/ShareGPT_Vicuna_unfiltered \
+    hf download anon8231489123/ShareGPT_Vicuna_unfiltered \
         --repo-type dataset --local-dir . ShareGPT_V3_unfiltered_cleaned_split.json
 
     # Batch size 1 (sequential):
@@ -55,6 +55,7 @@ def run_offline(
     max_tokens: int,
     max_model_len: int,
     max_num_seqs: int,
+    seed: int,
 ) -> list[int]:
     """Run offline inference, return list of completion token counts."""
     llm = LLM(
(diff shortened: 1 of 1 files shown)


tools/avg_gen_length.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""Measure average response length (tokens) via offline vLLM inference on ShareGPT.
+
+Usage:
+    # Download ShareGPT dataset (~642 MB):
+    huggingface-cli download anon8231489123/ShareGPT_Vicuna_unfiltered \
+        --repo-type dataset --local-dir . ShareGPT_V3_unfiltered_cleaned_split.json
```
LxYuan0420: Can we update the example command to use hf download ... because hugging face-cli download is deprecated

tools/avg_gen_length.py
```diff
@@ -0,0 +1,130 @@
...
+    max_model_len: int,
+    max_num_seqs: int,
+) -> list[int]:
+    """Run offline inference, return list of completion token counts."""
+    llm = LLM(
+        model=model,
+        max_model_len=max_model_len,
+        max_num_seqs=max_num_seqs,
+    )
+    sampling_params = SamplingParams(temperature=0, seed=42, max_tokens=max_tokens)
```
LxYuan0420: the 42 is hardcoded value ; should be wired to use --seed

tools/avg_gen_length.py
WindChimeRan: done

tools/avg_gen_length.py
WindChimeRan: done

PR #128, opened 2026-03-02
[Paged KV] Page request preemption when block full, remove rust block allocator

WindChimeRan:
### Summary

  - Remove the Rust `BlockAllocator` from the paged attention path.  Use the upstream vLLM scheduler as the sole source of truth for  KV cache block allocation
  - Wire scheduler-assigned `block_ids` through to the model runner (new requests, cached requests, and batched decode)
  - Handle **preemption-with-recompute**: when the scheduler preempts a request, reset model runner state and re-prefill the full   sequence (prompt + previously generated tokens) before resuming generation. This aligns with the upstream vllm strategy. 

###  Motivation

  Previously, the model runner ran its own independent Rust block allocator alongside the scheduler's KVCacheManager. When blocks   ran out, the Rust allocator crashed with RuntimeError("Not enough free blocks") instead of triggering the scheduler's preemption  logic. Now block management is unified and preemption works as designed upstream.

### Related discussion

https://discuss.vllm.ai/t/request-preemption-option/1672

### Possible limitation

from the [doc](https://docs.vllm.ai/en/stable/configuration/optimization/#preemption), I would expect warnning message like below. But it's not shown up in vllm-metal. 
```
WARNING 05-09 00:49:33 scheduler.py:1057 Sequence group 0 is preempted by PreemptionMode.RECOMPUTE mode because there is not enough KV cache space. This can affect the end-to-end performance. Increase gpu_memory_utilization or tensor_parallel_size to provide more KV cache memory. total_cumulative_preemption_cnt=1
```

I indeed observed the preemption behavior from the log

<details>
<summary>log</summary>

```
INFO 03-02 21:39:41 [v1/metrics/loggers.py:257] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 355.9 tokens/s, Running: 160 reqs, Waiting: 0 reqs, GPU KV cache usage: 98.3%, Prefix cache hit rate: 0.0%
INFO 03-02 21:39:52 [v1/metrics/loggers.py:257] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 337.5 tokens/s, Running: 142 reqs, Waiting: 18 reqs, GPU KV cache usage: 99.7%, Prefix cache hit rate: 0.0%
INFO 03-02 21:40:02 [v1/metrics/loggers.py:257] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 298.3 tokens/s, Running: 113 reqs, Waiting: 47 reqs, GPU KV cache usage: 99.2%, Prefix cache hit rate: 0.0%
```

</details>

### Test 1

run 
```bash
python tools/repro_block_exhaustion.py
```
(fail fast by assigning very small memory to paged kv, to exhaust block fast)

**before this fix**:  when setting too small `VLLM_METAL_MEMORY_FRACTION=0.1`. hit `RuntimeError: Not enough free blocks: need 21, have 0`

**after this fix**: Now the preepmtion works as expected

### Test 2
```
VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.3 vllm serve Qwen/Qwen3-0.6B --max-model-len 2048
```

```
vllm bench serve --backend vllm --model Qwen/Qwen3-0.6B \
    --endpoint /v1/completions \
    --dataset-name sharegpt \
    --dataset-path ShareGPT_V3_unfiltered_cleaned_split.json \
    --num-prompts 100 \
    --request-rate 10 
```
(remove the --max-concurrency 32 limitation)

<details>
<summary>results:</summary>

```
============ Serving Benchmark Result ============
Successful requests:                     100
Failed requests:                         0
Request rate configured (RPS):           10.00
Benchmark duration (s):                  276.46
Total input tokens:                      23260
Total generated tokens:                  22061
Request throughput (req/s):              0.36
Output token throughput (tok/s):         79.80
Peak output token throughput (tok/s):    275.00
Peak concurrent requests:                100.00
Total token throughput (tok/s):          163.93
---------------Time to First Token----------------
Mean TTFT (ms):                          13550.27
Median TTFT (ms):                        13442.19
P99 TTFT (ms):                           24532.78
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          682.65
Median TPOT (ms):                        427.85
P99 TPOT (ms):                           2767.03
---------------Inter-token Latency----------------
Mean ITL (ms):                           395.56
Median ITL (ms):                         338.87
P99 ITL (ms):                            2835.78
==================================================
```

</details>

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -1646,7 +1646,6 @@ class MetalModelRunner:
 
                         if target_len < len(state.token_ids):
                             # Intermediate chunk: sample then drop
-                            prev_seq_len = self._paged_request_seq_lens.get(req_id, 0)
                             _discarded = self._prefill_single_request_paged(
                                 req_id,
                                 state.token_ids[:target_len],
@@ -1654,16 +1653,11 @@ class MetalModelRunner:
                                 block_ids=state.block_ids,
                                 generator=state.generator,
                             )
-                            if self._rust_state_manager is not None:
-                                for tid in state.token_ids[prev_seq_len:target_len]:
-                                    self._rust_state_manager.append_token(req_id, tid)
                             req_ids.append(req_id)
                             req_id_to_index[req_id] = len(req_ids) - 1
                             sampled_tokens.append([])
                         else:
                             # Last chunk: sample and keep (drains async placeholder)
-                            prev_seq_len = self._paged_request_seq_lens.get(req_id, 0)
-                            seq_len_before = len(state.token_ids)
                             next_token = self._prefill_single_request_paged(
                                 req_id,
                                 state.token_ids,
@@ -1677,8 +1671,6 @@ class MetalModelRunner:
                                 len(state.token_ids) - state.prompt_len
                             )
                             if self._rust_state_manager is not None:
-                                for tid in state.token_ids[prev_seq_len:seq_len_before]:
-                                    self._rust_state_manager.append_token(req_id, tid)
                                 self._rust_state_manager.append_token(
                                     req_id, next_token
                                 )


tools/repro_block_exhaustion.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""Reproduce KV cache block exhaustion with vLLM offline inference.
+
+Usage:
+    VLLM_METAL_MEMORY_FRACTION=0.1 \
+    VLLM_METAL_USE_PAGED_ATTENTION=1 \
+    python scripts/repro_block_exhaustion.py
```
LxYuan0420: Typo here: change scripts to tools

tests/test_metal_kernel_paged.py
```diff
@@ -48,8 +46,6 @@
         allow_module_level=True,
```
LxYuan0420, line 45: Skip message still references "Rust extension" and uv pip install -e ., but this PR removes the Rust allocator from this path. please update to match.

tools/repro_block_exhaustion.py
ericcurtin: typo, it's in tools/

WindChimeRan: done!

vllm_metal/v1/model_runner.py
```diff
@@ -1606,6 +1588,30 @@ def execute_model(
...
+                        assert new_block_ids is not None
+                        state.block_ids = list(new_block_ids[0])
+                        state.generated_tokens = 0
+                        state.prompt_len = len(state.token_ids)
+                        self._paged_request_seq_lens.pop(req_id, None)
+                        if self._rust_state_manager is not None:
+                            self._rust_state_manager.remove_request(req_id)
+                            self._rust_state_manager.add_request(
+                                req_id, list(state.token_ids[: state.prompt_len])
+                            )
```
LxYuan0420: I want to double-check the invariants around prompt_len during preemption-with-recompute

Today on resume we do state.prompt_len = len(state.token_ids) (L1607). That makes sense as a recompute target, but prompt_len is also the boundary used later when building sampling metadata:

- prompt tokens = state.token_ids[:state.prompt_len]
- output/history tokens = state.token_ids[state.prompt_len:]

After the “last chunk” (L1652-1666) we set generated_tokens = 1, but I don’t see prompt_len being restored to the original prompt boundary. So after a preemption, the runner will treat the entire previously generated prefix as “prompt” forever, and only tokens generated after resume count as "output/history". That can change repetition/frequency/presence penalty behavior and any logits processors that depend on output history (effectively “resets” history on preemption).

Is that reset intended? If yes, can we add a short note in code/PR description explaining it as a deliberate behavioral choice. If not, I think we should keep prompt_len immutable (original prompt length) and introduce a separate variable for "recompute_upto_len" / "prefill_upto_len" to drive the chunked prefill loop.

tests/test_metal_kernel_paged.py
WindChimeRan: done. But now all rust dependecies are dead code. Will clean it up in another PR.

vllm_metal/v1/model_runner.py
WindChimeRan: good catch! 

now: Keep `prompt_len` immutable as the original prompt boundary and use `len(state.token_ids)` to drive the recompute scope

vllm_metal/v1/model_runner.py
```diff
@@ -1645,20 +1657,20 @@ def execute_model(
                         else:
                             # Last chunk: sample and keep (drains async placeholder)
                             prev_seq_len = self._paged_request_seq_lens.get(req_id, 0)
+                            seq_len_before = len(state.token_ids)
                             next_token = self._prefill_single_request_paged(
```
LxYuan0420, line 1661: This is a correctness issue in the resumed path.

This call samples through `_prefill_single_request_paged`, and that helper currently builds metadata as `prompt=[token_ids]` and `output=[[]]` (L1347-L1348). vLLM applies presence/frequency penalties from `output_token_ids`, so this step can ignore prior generated history after resume.

I think we should pass the original prompt boundary into `_prefill_single_request_paged` and split explicitly:
prompt = token_ids[:prompt_len]
output = token_ids[prompt_len:]

vllm_metal/v1/model_runner.py
```diff
@@ -1606,6 +1588,35 @@ def execute_model(
...
+                        # includes previously generated output tokens —
+                        # to determine the recompute scope, matching
+                        # upstream vLLM's use of request.num_tokens.
+                        assert new_block_ids is not None
+                        state.block_ids = list(new_block_ids[0])
+                        state.generated_tokens = 0
+                        self._paged_request_seq_lens.pop(req_id, None)
+                        if self._rust_state_manager is not None:
+                            self._rust_state_manager.remove_request(req_id)
+                            self._rust_state_manager.add_request(
```
LxYuan0420, line 1622: This can duplicate historical tokens in Rust state after resume.

We re-seed full history here with `add_request(req_id, state.token_ids)`, then recompute branches append historical tokens again (L1652-L1653 and L1673-L1674). Since seq-len tracking is reset at L1613, those loops replay from 0.

I think we should keep a single strategy here: re-seed full history on resume, remove historical re-append loops, and only append newly sampled tokens.

tests/test_metal_kernel_paged.py
```diff
@@ -235,11 +230,13 @@ def test_batched_decode_matches(self, qwen3_model):
         all_seq_lens = []
```
LxYuan0420, line 230: This PR changes resume/preemption behavior in `v1/model_runner.py`, but current tests here are parity/routing and don’t cover those invariants.

Can we add focused model-runner tests for preemption-with-recompute:
- penalty history stays consistent after resume,
- Rust token state is not duplicated,
- block_ids handling on resume (replace vs append) is correct.

LxYuan0420 (changes requested):

vllm_metal/v1/model_runner.py
WindChimeRan: done

vllm_metal/v1/model_runner.py
WindChimeRan: I fixed it as you suggested for now. 

But I think eventually we'll remove all rust dependencies, because they are dead code. We don't actually need rust for anything.

tests/test_metal_kernel_paged.py
WindChimeRan: I prefer to defer these tests.

* the model runner is about to be rewritten for v1 chunked prefill and continuous batching, which will
  restructure the preemption path. 
* The Rust state manager is also dead code slated for cleanup. Will add focused preemption tests   against the new design.
* need to make metal kernel inline before writing strict tests.

PR #129, opened 2026-03-02
Fix Qwen3.5 MoE load path in vLLM Metal

manmal:
## Summary\n- bump minimum \ to \ (includes qwen3_5/qwen3_5_moe model modules)\n- add a guarded mlx_lm alias shim in \ so older mlx_lm builds can still map \ to \ as a fallback\n- avoid overriding native modules when mlx_lm already provides \n\n## Motivation\nQwen3.5 MoE snapshots currently fail to load on some vLLM Metal installations with:\n\.\n\nThis patch ensures native support is used when present and provides a fallback alias path for older environments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 dependencies = [
     # MLX - Required for Apple Silicon GPU acceleration
     "mlx>=0.29.2; platform_system == 'Darwin' and platform_machine == 'arm64'",
-    "mlx-lm>=0.28.4; platform_system == 'Darwin' and platform_machine == 'arm64'",
+    "mlx-lm>=0.30.8; platform_system == 'Darwin' and platform_machine == 'arm64'",
     "mlx-vlm>=0.3.0; platform_system == 'Darwin' and platform_machine == 'arm64'",  # Vision-language model support
     # Model loading and weights
     "transformers>=4.40.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers = [
 dependencies = [
     # MLX - Required for Apple Silicon GPU acceleration
     "mlx>=0.29.2; platform_system == 'Darwin' and platform_machine == 'arm64'",
-    "mlx-lm>=0.28.4; platform_system == 'Darwin' and platform_machine == 'arm64'",
+    "mlx-lm>=0.30.8; platform_system == 'Darwin' and platform_machine == 'arm64'",
```
LxYuan0420, line 32: This change mixes two conflicting strategies: it raises the minimum `mlx-lm` version to one that should already include native `qwen3_5*` support, and also adds a runtime alias shim for older versions.

vllm_metal/v1/model_runner.py
```diff
@@ -61,6 +63,47 @@
...
+        # If mlx_lm already ships this architecture, do not override it.
+        try:
+            importlib.import_module(alias_module)
+            continue
+        except ModuleNotFoundError:
+            pass
+
+        try:
+            resolved = importlib.import_module(target_module)
+        except ModuleNotFoundError:
```
LxYuan0420, line 96: `except ModuleNotFoundError` is too broad here and can hide real import problems inside the imported module. Please only suppress the error when `e.name` matches the exact module you attempted to import; otherwise re-raise.

PR #133, opened 2026-03-03
[Feature] Add Whisper model definition and weight sanitization

ricky-chaoju:
## Summary
- Add `vllm_metal/stt/whisper.py` — Whisper encoder/decoder model (MLX)
- WhisperConfig supports both HuggingFace and MLX config formats
- Weight sanitization: HF key renaming, Conv1d transpose, dtype casting
- Add 5 unit tests (TestWeightSanitize)

Related #91

### Run tests

```bash
source ~/.venv-vllm-metal/bin/activate
pytest tests/test_whisper.py -v
```

All 9 tests should pass, including the 4 new decoder cache-path tests.

### Minimal repro (before fix)

The mask bug triggers when `q_len > 1` during cached decode:

```python
import mlx.core as mx
from vllm_metal.stt.whisper import WhisperConfig, WhisperModel

config = WhisperConfig(
    n_mels=80, n_audio_ctx=10, n_audio_state=64,
    n_audio_head=2, n_audio_layer=1, n_vocab=100,
    n_text_ctx=32, n_text_state=64, n_text_head=2, n_text_layer=1,
)
model = WhisperModel(config, dtype=mx.float32)

mel = mx.random.normal((1, 20, 80))
audio_features = model.encode(mel)

# Prefill 3 tokens
_, kv_cache = model.decode(mx.array([[1, 2, 3]]), audio_features)

# Cached decode 2 tokens — before fix this raised:
# ValueError: [broadcast_shapes] Shapes (1,2,2,5) and (2,2) cannot be broadcast
logits, _ = model.decode(mx.array([[4, 5]]), audio_features, kv_cache)
mx.eval(logits)
print(f"OK — logits shape: {logits.shape}")  # (1, 2, 100)
```

### DIFF
--- a/tests/test_stt.py
+++ b/tests/test_stt.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Tests for STT data types, config, formatting, audio pipeline, and model."""
+"""Tests for STT data types, config, formatting, and audio pipeline."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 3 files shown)


vllm_metal/stt/whisper.py
```diff
@@ -0,0 +1,450 @@
...
+        n_batch, n_ctx, n_state = q.shape
+        scale = (n_state // self.n_head) ** -0.25
+
+        q = q.reshape(*q.shape[:2], self.n_head, -1).transpose(0, 2, 1, 3) * scale
+        k = k.reshape(*k.shape[:2], self.n_head, -1).transpose(0, 2, 3, 1) * scale
+        v = v.reshape(*v.shape[:2], self.n_head, -1).transpose(0, 2, 1, 3)
+
+        qk = q @ k
+        if mask is not None:
+            qk = qk + mask[:n_ctx, :n_ctx]
```
LxYuan0420: This is a correctness bug in cached decode.

When `kv_cache` is present and we decode more than 1 token in a step, attention scores are `(B, H, q_len, k_len)` with `k_len = offset + q_len`. This line adds `mask[:q_len, :q_len]`, which only matches `(q_len, q_len)` and breaks when `k_len != q_len`.

I can reproduce with prefill 3 tokens, then cached decode 2 tokens:
`ValueError: [broadcast_shapes] Shapes (1,2,2,5) and (2,2) cannot be broadcast.`

Please make mask slicing offset-aware for cached decode.

tests/test_stt.py
```diff
@@ -288,3 +288,73 @@ def test_split_with_energy_based_split_point(self) -> None:
...
+# ===========================================================================
+
+
+class TestWeightSanitize:
+    """Tests for WhisperModel.sanitize() weight mapping."""
+
+    @pytest.fixture()
+    def model(self):
+        """Create a minimal WhisperModel for testing sanitize()."""
+        from vllm_metal.stt.whisper import WhisperConfig, WhisperModel
```
LxYuan0420: Can we avoid lazy import in this fixture and use top-level imports for `WhisperConfig` / `WhisperModel`?

Also, we don't need to duplicate all default config values here. Using `WhisperConfig()` (and overriding only when needed) keeps the test aligned with source defaults and avoids drift.

tests/test_stt.py
```diff
@@ -288,3 +288,73 @@ def test_split_with_energy_based_split_point(self) -> None:
...
         _, second_start = chunks[1]
         assert 26.0 <= second_start <= 31.0
+
+
+# ===========================================================================
+# Weight sanitization
+# ===========================================================================
+
+
+class TestWeightSanitize:
```
LxYuan0420: `whisper.py` adds model internals (attention/cache/encoder/decoder), but tests in this PR only cover `sanitize()`.

Can we split Whisper tests into a dedicated file (e.g. `tests/test_whisper.py`) and add a focused decoder cache-path test:
- prefill without cache,
- decode with cache and `q_len > 1`,
- compare with non-cached full decode tail (or at least assert no broadcast/shape failure)?

That keeps STT tests scoped and protects the highest-risk path in this module.

LxYuan0420 (changes requested):
Main blocker is the decoder cached-attention mask is incorrect for `q_len > 1` and currently reproducible as a broadcast-shape failure (see inline comment with repro code).

Also please add a short "How to verify" section with exact commands and the minimal repro snippet used for this PR.

LxYuan0420 (approved):

PR #134, opened 2026-03-03
Bump vllm to 0.16.0 and install from wheel

ericcurtin:
Install vllm 0.16.0 from the macOS wheel on PyPI instead of building from source, and update the optional dependency lower bound to match.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -123,16 +123,8 @@ main() {
 
   ensure_venv "$venv"
 
-  local vllm_v="0.14.1"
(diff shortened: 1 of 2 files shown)


ericcurtin:
Thanks @mgoin

LxYuan0420:
+1 on the direction.

Since this bump also brings Torch 2.10 with vLLM 0.16.0, could we add a quick compatibility pass before merge (install path + first/second completion smoke + a focused torch/mlx conversion check)? 

Just want to keep the transition smooth

PR #136, opened 2026-03-05
[Paged KV] Inline metal kernel. deprecate hf pytorch kernel

WindChimeRan:
bench_comparison

##  Overview

  This PR eliminates the external `kernels-community/paged-attention` HuggingFace dependency by vendoring the Metal shader files  directly and replacing the PyTorch MPS dispatch layer with a nanobind C++ extension that dispatches through MLX's own Metal command  encoder. 

This removes the MLX↔PyTorch MPS fence synchronization bubble.

## Key architectural change: 
* original: mlx.array → PyTorch MPS bridge → HF kernel → PyTorch MPS bridge  → mlx.array
* this PR: mlx.array → native Metal dispatch →  mlx.array 
 
fully zero-copy. single Metal command queue. 

## FAQ: 

* Q: Why `mlx.core.fast.metal_kernel` Can't Be Used Here? This is much more simpler.
* A: `mlx.core.fast.metal_kernel` always allocates fresh output arrays and provides inputs as read-only. Paged KV
   cache requires **in-place, scatter-write mutation** of a persistent buffer (via `reshape_and_cache`) and **zero-copy reads** from that same   buffer (via `paged_attention_v1`). If we copied the entire cache in and out on every token, it would negate the whole purpose of   paging. The C++ nanobind bridge is the minimum complexity needed to get direct Metal buffer access through MLX's internal command   encoder.

* Q: What's the difference between `kernel` and `kernel_v1`?
* A: `kernel` is a drop-in replacement for the HuggingFace kernel, originally vendored from an older version of mistral.rs. This PR's end-to-end deterministic tests confirm no regression from the replacement. `kernel_v1` is the latest Metal kernel from the mistral.rs repo — more mature, with preliminary scaffolding for variable-length kernels and gpt-oss sink attention support. **That said, neither will persist beyond this PR**. Both are slated for deprecation once we introduce first-class variable-length kernel support, which is a prerequisite for continuous batching, chunked prefill, and MQA Scorer speculative decoding.

## Test
```
python -m pytest tests/test_paged_deterministic.py -v -s
```

5 PASSED. The model output is identical to the hf kernel version before. 

## Benchmark (M1 Pro 32G RAM, macos 26.1)

Usage 

Page KV Cache On
```
VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.3 vllm serve Qwen/Qwen3-0.6B --max-model-len 2048
```

```
[...]

PR #138, opened 2026-03-05
[Paged KV] Add paged attention deterministic smoke test

WindChimeRan:
## Summary
- Add `test_paged_deterministic.py`: 5-prompt smoke test using vLLM offline inference (temp=0, greedy) against hardcoded golden token IDs from Qwen3-0.6B
- Golden values generated on `main` from both MLX inline cache and HF paged KV cache paths
- Add `tools/gen_golden.py` helper to regenerate golden values

## Motivation
Prerequisite for the native Metal kernel PR (#136). After inlining the vendored Metal shaders, paged attention output must remain identical to the current HF kernel baseline. This test anchors that.

## Test
- `python -m pytest tests/test_paged_deterministic.py -v -s` (paged path by default)
- Passes on `main` with HF kernel: 5/5

## Relevant Issue & PR

* Issue #119 
* PR #136 : This inline metal kernel need to either pass this test, or explain the possible non-deterministics from the kernel. 

upstream batch invariant feature
* blog: https://thinkingmachines.ai/blog/defeating-nondeterminism-in-llm-inference/
* main feature: https://github.com/vllm-project/vllm/issues/27433
* vllm upstream batch invariant feature is only compatible with H / B series NVIDIA GPU. A100 not working. See my exp results https://github.com/WindChimeRan/spec_deterministic
* community work: https://github.com/vllm-project/vllm/pull/30018

Batch invariant is hardware & kernel dependent. Supporting this feature is non-trivial on metal. 

output example: 
image

### DIFF
--- a/tests/test_paged_deterministic.py
+++ b/tests/test_paged_deterministic.py
@@ -36,6 +36,9 @@ from vllm import LLM, SamplingParams
 
 MODEL_NAME = "Qwen/Qwen3-0.6B"
 MAX_TOKENS = 10
+DEFAULT_USE_PAGED_ATTENTION = "1"
+DEFAULT_PAGED_MEMORY_FRACTION = "0.2"
+DEFAULT_MLX_MEMORY_FRACTION = "auto"
 
 PROMPTS = [
     "The capital of France is",
@@ -69,6 +72,15 @@ GOLDEN_PAGED = {
 # fmt: on
 
 
+def _setenv_default(mp: pytest.MonkeyPatch, key: str, default: str) -> str:
+    """Set an env var only when absent and return the effective value."""
+    value = os.environ.get(key)
+    if value is None:
+        mp.setenv(key, default)
+        return default
+    return value
+
+
 @pytest.fixture(autouse=True, scope="module")
 def _set_env():
     """Set default env vars for this test.
@@ -83,19 +95,20 @@ def _set_env():
     with pytest.MonkeyPatch.context() as mp:
         mp.setenv("VLLM_ENABLE_V1_MULTIPROCESSING", "0")
 
-        # Default to paged attention, but allow the caller to override.
-        use_paged = os.environ.get("VLLM_METAL_USE_PAGED_ATTENTION")
-        if use_paged is None:
-            mp.setenv("VLLM_METAL_USE_PAGED_ATTENTION", "1")
-            use_paged = "1"
+        # Default to paged attention, but allow explicit caller override.
+        use_paged = _setenv_default(
+            mp,
+            "VLLM_METAL_USE_PAGED_ATTENTION",
+            DEFAULT_USE_PAGED_ATTENTION,
+        )
 
-        # Set a sensible default memory setting for the selected path, unless
-        # the caller has already specified one.
-        if os.environ.get("VLLM_METAL_MEMORY_FRACTION") is None:
-            if use_paged == "1":
-                mp.setenv("VLLM_METAL_MEMORY_FRACTION", "0.2")
-            else:
-                mp.setenv("VLLM_METAL_MEMORY_FRACTION", "auto")
+        # Choose a path-specific memory default, while preserving caller override.
+        memory_default = (
+            DEFAULT_PAGED_MEMORY_FRACTION
+            if use_paged == "1"
+            else DEFAULT_MLX_MEMORY_FRACTION
+        )
+        _setenv_default(mp, "VLLM_METAL_MEMORY_FRACTION", memory_default)
         yield
 
 


tests/test_paged_deterministic.py
```diff
@@ -0,0 +1,112 @@
...
+
+from __future__ import annotations
+
+import os
+
+# Default: test the paged KV cache path through vLLM offline inference.
+# Change to "0" to test the MLX inline cache path.
+os.environ.setdefault("VLLM_METAL_USE_PAGED_ATTENTION", "1")
+os.environ.setdefault("VLLM_METAL_MEMORY_FRACTION", "0.2")
+os.environ.setdefault("VLLM_ENABLE_V1_MULTIPROCESSING", "0")
```
LxYuan0420: These os.environ.setdefault(...) run at module import (pytest collection) and leak global state into the process. 

Please move env setup into the fixture and use monkeypatch.setenv so it only affects this test

LxYuan0420: Quick python code to confirm that side effects:

```bash
import os
os.environ.pop("VLLM_METAL_USE_PAGED_ATTENTION", None)
os.environ.pop("VLLM_METAL_MEMORY_FRACTION", None)
print("before:", os.environ.get("VLLM_METAL_USE_PAGED_ATTENTION"), os.environ.get("VLLM_METAL_MEMORY_FRACTION"))

import tests.test_paged_deterministic

print("after: ", os.environ.get("VLLM_METAL_USE_PAGED_ATTENTION"), os.environ.get("VLLM_METAL_MEMORY_FRACTION"))

>>
before: None None
after:  1 0.2
```

tests/test_paged_deterministic.py
```diff
@@ -0,0 +1,112 @@
...
+  Likely caused by floating-point non-determinism in the attention kernel
+  where top-2 logits are very close.
+
+The assert accepts EITHER golden set (mlx-cache or paged-cache) and prints
+which path matched.
+
+Run (paged KV path, the default):
+    python -m pytest tests/test_paged_deterministic.py -v -s
+
+To test the MLX inline cache path instead, change the env vars below.
```
LxYuan0420: I think we should be clear on the doc and env vars setting.

MLX requires VLLM_METAL_MEMORY_FRACTION=auto (numeric fractions are only valid for paged attention)

```bash
ValueError: VLLM_METAL_MEMORY_FRACTION=0.2 requires VLLM_METAL_USE_PAGED_ATTENTION=1. The MLX path must use VLLM_METAL_MEMORY_FRACTION=auto.
```

tests/test_paged_deterministic.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+@pytest.fixture(scope="module")
+def vllm_outputs():
+    """Run vLLM offline inference once for all prompts."""
+    llm = LLM(model=MODEL_NAME, max_model_len=512)
+    sp = SamplingParams(temperature=0, max_tokens=MAX_TOKENS)
+    outputs = llm.generate(PROMPTS, sp)
+    return {o.prompt: o for o in outputs}
+
```
LxYuan0420, line 125: Key comment for this PR: I ran this locally and hit a flake with the current batched llm.generate(PROMPTS, ...) run. On macOS 15 (HF paged-attention compat kernel 8968951), pytest failed with:

```bash
AssertionError: Output for 'One plus one equals' matched neither golden set.
Got:            [825, 13, 3776, 5519, 825, 16819, 1378, 13, 3776, 5519]
Expected (mlx): [825, 11, 825, 5519, 825, 16819, 1378, 13, 2055, 11]
Expected (pgd): [825, 11, 825, 5519, 825, 16819, 1378, 13, 2055, 11]
```

(I will share a  screenshot later)


Fix that made it stable for me: force single-seq execution by setting `LLM(..., max_num_seqs=1)` (or generate prompts one-by-one).

So: if the intention is a deterministic golden smoke test to gate #136 (inline kernel must match the HF baseline), we should explicitly disable batching (max_num_seqs=1) to keep CI stable on macOS 15. If the intention is to test batch-invariant behavior, exact token-ID goldens will be flaky on Metal and the assertion should be changed.

LxYuan0420:
```bash
# Repro the flake (batched run)
python -m pytest tests/test_paged_deterministic.py -v -s

# After the fix (set max_num_seqs=1 or run prompts one-by-one), this becomes stable:
python -m pytest tests/test_paged_deterministic.py -v -s

```

tools/gen_golden.py
```diff
@@ -0,0 +1,48 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""Generate golden token IDs for the e2e smoke test via vLLM offline inference.
+
+Usage:
+    # MLX inline cache (default):
+    VLLM_ENABLE_V1_MULTIPROCESSING=0 python tools/gen_golden.py
+
+    # Paged KV cache:
+    VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_ENABLE_V1_MULTIPROCESSING=0 \
+        python tools/gen_golden.py
```
LxYuan0420: Nit: This is a useful script but prefer to have better and more descriptive script name; I expect us to continue adding more scripts into tools/ and want make it clear to people which script to be used in what scenario.

LxYuan0420 (changes requested):
Nice direction and the tools/ helper is a good addition.

To recap the main issue I ran into locally: on macOS 15 (HF compat kernel 8968951), this test can flake under the current batched llm.generate(PROMPTS, ...) setup. Specifically, greedy output for `One plus one equals` sometimes diverges from both golden sets.

For CI stability, I think we should make the execution mode explicit. Would you be open to forcing single-seq execution (LLM(..., max_num_seqs=1)) for this test?

If your intent is to also cover batched behavior, maybe we can add a separate batched smoke test with a weaker assertion (since exact token IDs aren’t batch-invariant on Metal today)

LxYuan0420:
On my machine, I need to set max_seq_nums=1 to pass the test

Screenshot 2026-03-06 at 11 17 08 AM

tests/test_paged_deterministic.py
WindChimeRan: done. For the batch invariant feature & test, let's delay it until we support real continuous batching.

tests/test_paged_deterministic.py
WindChimeRan: done

tools/gen_golden.py
WindChimeRan: new name: tools/gen_golden_token_ids_for_deterministics.py

tests/test_paged_deterministic.py
WindChimeRan: Will address it in the next PR, by updating README.md

WindChimeRan:
> On my machine, I need to set max_seq_nums=1 to pass the test
> 
> Screenshot 2026-03-06 at 11 17 08 AM
> EDIT Typo: it is max_num_seqs=1 `(vLLM LLM(..., max_num_seqs=1))`, not max_seq_nums

This is really intriguing... Maybe it's due to Metal 3 vs. Metal 4, or the 0.2 × RAM auto-calculated max sequence length, or differences in our chips.

tests/test_paged_deterministic.py
WindChimeRan: see #140

LxYuan0420 (approved):
Thanks for iterating on this. Forcing single-seq execution (LLM(..., max_num_seqs=1)) resolves the macOS 15 flake I hit with batched scheduling.

I also pushed a small maintainer tweak on top to (1) fix the generator script name in comments/usage and (2) make the autouse env fixture respect any user-provided env overrides so the MLX run instructions match actual behavior.

LGTM.

Re batch-invariant determinism: I’m not expecting that in this PR (I only wanted a CI-stable single-seq golden smoke test). Let’s defer any batch-invariant feature/test until we support real continuous batching on Metal.

PR #143, opened 2026-03-07
[Feature] Add v1 STT integration for Whisper models

ricky-chaoju:
## Summary

Wire the STT pipeline (PRs #96, #98, #126, #133, #137) into vLLM's v1 engine so that `vllm serve openai/whisper-small` serves `/v1/audio/transcriptions` and `/v1/audio/translations` endpoints.

- **model_runner**: STT model loading with caching, dummy KV cache spec (Whisper self-manages KV), warm-up from model config, greedy decode loop, audio feature extraction (handles `MultiModalKwargsItem`/UserDict, bfloat16 torch tensors, shape transpose)
- **platform**: STT auto-detection, tokenizer fallback (only when unset), disable `async_scheduling`
- **worker**: skip paged attention for STT, nominal memory for scheduler, `get_supported_tasks` returns `("transcription",)`
- **docs**: add `whisper-large-v3-turbo` to model table

## Test

截圖 2026-03-07 中午12 54 23
截圖 2026-03-07 中午12 55 16

截圖 2026-03-07 下午1 06 37

### Unit tests
```
pytest tests/test_v1_stt_integration.py tests/test_stt.py tests/test_whisper.py tests/test_transcribe.py -v -m "not slow"
```

### End-to-end
```
pytest tests/test_v1_stt_integration.py -v -m slow
```

### Server smoke test (requires local model)
```
vllm serve /path/to/whisper-small-mlx --port 8000
```

```
curl http://localhost:8000/v1/audio/transcriptions \
  -F file=@test.wav \
  -F model=whisper
```

### DIFF
--- a/tests/test_v1_stt_integration.py
+++ b/tests/test_v1_stt_integration.py
@@ -273,7 +273,7 @@ class TestSamplingParamsValidation:
 
 
 class TestExecuteSTTProtocol:
-    """Tests for _execute_stt_inner output protocol and request lifecycle."""
+    """Tests for _execute_stt output protocol and request lifecycle."""
 
     def _run_stt(self, runner, sched):
         """Run _execute_stt (STTExecutor is pre-cached in _make_runner)."""


vllm_metal/v1/model_runner.py
```diff
@@ -821,6 +865,19 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
         Returns:
             Dictionary mapping attention layer names to KV cache specs
         """
+        if self._is_stt:
```
LxYuan0420, line 1000: STT scheduler placeholder values are currently scattered across multiple locations without one shared contract. Could we centralize these as named STT scheduler constants and reuse them consistently?

Anchoring here, but same contract points also exist at `model_runner.py:L915` (`return 1`) and `worker.py:L325-L330` (`return 1 << 30`).

Suggested constants: `STT_SCHED_BLOCK_BYTES`, `STT_SCHED_NUM_BLOCKS`, `STT_SCHED_AVAILABLE_BYTES`.

vllm_metal/v1/worker.py
```diff
@@ -134,8 +134,10 @@ def load_model(self) -> None:
         """Load the model onto the Metal device."""
         self.model_runner.load_model()
 
-        # Patch model for paged attention if enabled
-        if self.metal_config.use_paged_attention:
+        # Patch model for paged attention if enabled (skip for STT)
+        if self.metal_config.use_paged_attention and not getattr(
```
LxYuan0420: `MetalWorker` currently reaches into `MetalModelRunner` private state via `getattr(self.model_runner, "_is_stt", False)`. Could we expose a public runner contract (`is_stt` or capability method) and consume that here instead?

Anchoring this line, but same private-state coupling appears at `worker.py:L328` and `worker.py:L476`.

This will keep class boundaries explicit and reduce hidden coupling during future refactors.

vllm_metal/v1/model_runner.py
```diff
@@ -1916,6 +1990,199 @@ def sample_tokens(
         self._pending_output = None
         return output
 
+    # ------------------------------------------------------------------
+    # STT (Speech-to-Text) helpers
+    # ------------------------------------------------------------------
+
+    def _execute_stt(
```
LxYuan0420, line 2129: This STT block is handling validation, feature extraction, decode, and output assembly in one place. Could we extract an `STTExecutor` and have `MetalModelRunner` delegate? It will keep responsibilities clearer and easier to refactor.

vllm_metal/v1/model_runner.py
```diff
@@ -1916,6 +1990,199 @@ def sample_tokens(
...
+            List of decoded token IDs ending with eot_token.
+        """
+        if not prompt_token_ids:
+            return [eot_token]
+
+        tokens = mx.array([prompt_token_ids], dtype=mx.int32)
+        kv_cache = None
+        output_tokens: list[int] = []
+
+        for _ in range(_WHISPER_MAX_DECODE_TOKENS):
```
LxYuan0420: This decode loop is near-duplicate of `WhisperTranscriber._greedy_decode` in `vllm_metal/stt/transcribe.py`. Could we keep a single decode owner (preferably transcriber) to avoid drift?

vllm_metal/v1/model_runner.py
```diff
@@ -63,6 +65,9 @@
 _model_cache: dict[str, tuple[Any, Any]] = {}  # model_name -> (model, tokenizer)
 _model_cache_lock = Lock()
 
+# Maximum decode tokens for Whisper STT models
+_WHISPER_MAX_DECODE_TOKENS = 448
```
LxYuan0420: Could we move this `448` into a shared STT constant with a short source note? Keeping a single source of truth (DRY) will reduce drift risk as STT paths evolve.

vllm_metal/v1/model_runner.py
```diff
@@ -1916,6 +1990,199 @@ def sample_tokens(
...
+
+        Args:
+            audio_features: Encoded audio from the Whisper encoder.
+            prompt_token_ids: Prefix tokens (language, task, etc.).
+            eot_token: End-of-text token ID.
+
+        Returns:
+            List of decoded token IDs ending with eot_token.
+        """
+        if not prompt_token_ids:
```
LxYuan0420: Returning EOT for empty `prompt_token_ids` can mask upstream contract issues. Could we at least warning-log with request context (or fail-fast once the caller contract is finalized)? This helps keep failure modes explicit.

tests/test_v1_stt_integration.py
```diff
@@ -0,0 +1,511 @@
...
+    mock_tokenizer = MagicMock()
+    mock_tokenizer.convert_tokens_to_ids.return_value = 50257
+    mock_transcriber = MagicMock()
+    mock_transcriber.tokenizer = mock_tokenizer
+    runner._stt_transcriber = mock_transcriber
+
+    # Import the real methods and bind them
+    from vllm_metal.v1.model_runner import MetalModelRunner
+
+    runner._execute_stt = MetalModelRunner._execute_stt.__get__(runner)
```
LxYuan0420: These tests are helpful for wiring, but `MagicMock + bound real methods` can bypass class invariants. Could we add at least one `@pytest.mark.slow` real Whisper end-to-end smoke test through the actual runner path?

vllm_metal/v1/model_runner.py
LxYuan0420: I noticed a runtime edge case here: this loop always allows up to 448 decode steps, but Whisper `n_text_ctx` includes prompt tokens too. I was able to reproduce an overflow case where non-empty prompt + no early EOT pushes decode past context (`offset >= n_text_ctx`). Can we cap decode steps by remaining context, e.g. `n_text_ctx - len(prompt_token_ids)`?

vllm_metal/platform.py
```diff
@@ -231,6 +231,21 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
         if model_config is not None:
             model_config.disable_cascade_attn = True
 
+        # STT model detection — set tokenizer fallback if not already configured
+        from vllm_metal.stt.config import is_stt_model
+
+        if model_config is not None and is_stt_model(model_config.model):
```
LxYuan0420: One more thing I noticed: STT detection here uses raw `model_config.model`, while runner checks STT after `get_model_download_path(...)`. If raw-ID detection misses, this block can skip STT config updates (especially disabling async scheduling) even though runner still enters STT mode. Can we detect STT from the resolved model path in this config phase too?

tests/test_v1_stt_integration.py
```diff
@@ -0,0 +1,567 @@
...
+    runner._pending_output = None
+
+    # Create a real STTExecutor with pre-cached transcriber
+    runner._stt_executor = _make_executor()
+    runner._stt_executor.model = runner.model
+
+    # Import the real methods and bind them
+    from vllm_metal.v1.model_runner import MetalModelRunner
+
+    runner._execute_stt = MetalModelRunner._execute_stt.__get__(runner)
```
LxYuan0420: Could we avoid method rebinding onto `MagicMock` here? This bypasses runner invariants and can let tests pass while real integration breaks. Prefer a small concrete test double or lightweight real object.

vllm_metal/stt/transcribe.py
```diff
@@ -180,6 +180,56 @@ def transcribe(
...
+            max_tokens = WHISPER_MAX_DECODE_TOKENS
+
+        if not prompt_token_ids:
+            logger.warning("Empty prompt_token_ids; returning no tokens")
+            return []
+
+        # Cap by context window to prevent overflow
+        n_text_ctx = getattr(self.model.config, "n_text_ctx", None)
+        if n_text_ctx is not None:
+            max_tokens = min(max_tokens, n_text_ctx - len(prompt_token_ids))
```
LxYuan0420: `n_text_ctx - len(prompt_token_ids)` can be negative, which currently results in a silent zero-iteration decode. Please clamp explicitly (`max(0, ...)`) and consider a debug/warning log when prompt already consumes context.

vllm_metal/v1/model_runner.py
```diff
@@ -570,6 +576,129 @@ def _extract_kv_cache(
...
+        elif isinstance(first, dict):
+            input_features = first.get("input_features")
+
+        if input_features is None:
+            return None
+
+        # Convert to MLX array — handle numpy, torch tensors, and lists
+        if isinstance(input_features, np.ndarray):
+            mel = mx.array(input_features, dtype=mx.float16)
+        elif hasattr(input_features, "numpy"):
```
LxYuan0420: `hasattr(input_features, "numpy")` is too broad here; objects can match this but still not support `.cpu().float().numpy()`. Could we use explicit tensor-type checks (or guarded conversion) to avoid this fragile path?

vllm_metal/v1/model_runner.py
```diff
@@ -570,6 +576,129 @@ def _extract_kv_cache(
...
+            # torch tensor — .cpu() for device safety, .float() because
+            # bfloat16 has no numpy dtype support.
+            mel = mx.array(input_features.cpu().float().numpy(), dtype=mx.float16)
+        else:
+            mel = mx.array(np.array(input_features), dtype=mx.float16)
+
+        # HF WhisperFeatureExtractor output shape: (n_mels, time)
```
LxYuan0420, line 660: Please add explicit rank validation for mel inputs. Right now non-2D/3D tensors fall through to `model.encode` without checks, which can fail later with less actionable errors.

vllm_metal/v1/model_runner.py
```diff
@@ -1916,6 +2118,92 @@ def sample_tokens(
...
+    # STT (Speech-to-Text) helpers
+    # ------------------------------------------------------------------
+
+    def _execute_stt(
+        self, scheduler_output: SchedulerOutput
+    ) -> ModelRunnerOutput | None:
+        """Execute STT inference for all new requests in the batch."""
+        try:
+            return self._execute_stt_inner(scheduler_output)
+        except Exception:
```
LxYuan0420: Could we narrow this broad `except Exception` (or add a short comment for why broad catch is required)? Re-raising after a generic catch reduces error contract clarity.

tests/test_v1_stt_integration.py
```diff
@@ -0,0 +1,594 @@
...
+        runner._execute_stt(sched)
+
+
+# ===========================================================================
+# TestExecuteSTTProtocol
+# ===========================================================================
+
+
+class TestExecuteSTTProtocol:
+    """Tests for _execute_stt_inner output protocol and request lifecycle."""
```
LxYuan0420: Small nit: this docstring still references `_execute_stt_inner`, but the current path runs through `_execute_stt`.

ricky-chaoju: Wait a moment, I'll handle it now.

ricky-chaoju:
Thanks for catching this! Fixed .

PR #147, opened 2026-03-08
[varlen Kernel] Add test spec for unified varlen paged attention [2/n]

WindChimeRan:
## Goal

This is the spec to smoothly migrate v1 kernel (only paged) to v2 kernel (paged + varlen), without break anything on main, but enable us to support continuous batching. 

## Summary

Adds `test_metal_unified_attention.py`, adapted from vLLM upstream's `test_triton_unified_attention.py`.

Three test functions form a triangle validation:

```
        v1 kernel (production)
           /        \
  PASS    /          \   xfail
 (runs   /            \  (v2 not built)
  now)  /              \
       v                v
  ref_paged_attn ----> v2 kernel (unified varlen)
     (naive MLX)  xfail
```

- `test_v1_kernel_vs_reference` - v1 == ref (12 cases, runs now, validates the reference)
- `test_metal_unified_attn_decode_only` - v2 == v1 (24 cases, xfail, proves v2 is a drop-in replacement for decode)
- `test_metal_unified_attn` - v2 == ref (192 cases, xfail, full varlen with mixed prefill+decode, GQA, sliding window, soft cap)

## Migration plan

```
Step 1: Build v2 kernel, test decode-only (q_len=1)
        Scaffolding test goes green: v2 == v1
                    |
Step 2: Extend v2 to handle varlen (q_len > 1)
        Full test goes green: v2 == ref
                    |
Step 3: Delete scaffolding test, replace v1 with v2 in production
```

### DIFF
--- a/tests/test_metal_unified_attention.py
+++ b/tests/test_metal_unified_attention.py
@@ -15,6 +15,15 @@ import pytest
 # TODO: import from vllm_metal.metal once the kernel is implemented
 # from vllm_metal.metal import metal_unified_attention
 
+# Original upstream parameters (vLLM Triton/CUDA test_triton_unified_attention.py):
+#   HEAD_SIZES = [128, 256]
+#   NUM_BLOCKS = [32768, 2048]
+#   sliding_window = [None, 64, 128, 256]
+#   DTYPES = [torch.bfloat16]
+#   Also tested: FP8 output quantization (QDTYPES), 3D decode kernel
+#     (SEQ_THRESHOLD_3D) — both CUDA-specific, omitted here.
+# Current sizes are reduced for Apple Silicon unified memory.
+# TODO: try head_size=256 and larger num_blocks as @pytest.mark.slow variants.
 NUM_HEADS = [(4, 4), (8, 2), (5, 1)]
 HEAD_SIZES = [128]
 BLOCK_SIZES = [16]


tests/test_metal_unified_attention.py
```diff
@@ -0,0 +1,461 @@
...
+#
+# Adapted from vLLM's test_triton_unified_attention.py for Metal/MLX.
+#
+# Compares metal_unified_attention (the Metal kernel under development)
+# against ref_paged_attn (a naive pure-MLX loop implementation that is
+# trivially correct).  Both receive the same paged KV cache and query
+# inputs; the test asserts their outputs match within FP tolerance.
+
+import mlx.core as mx
+import numpy as np
```
LxYuan0420, line 12: Please add module-level guards here (e.g., pytest.importorskip("mlx.core") plus Metal availability check, similar to test_metal_kernel_paged.py). Without this, the file can fail on non-Apple Silicon or CI environments without MLX/Metal.

LxYuan0420: Clarification: this is not a current CI blocker because tests run on macos-15 only. The guard request is for portability on non-Darwin local runs and to keep us safe if/when we expand CI platforms.

tests/test_metal_unified_attention.py
```diff
@@ -0,0 +1,461 @@
...
+# Triangle edge: v1 == ref (decode-only)
+#
+# Validates that the v1 kernel and the pure-MLX reference produce the same
+# results for decode-only inputs.  This test runs TODAY (no v2 needed) and
+# also validates ref_paged_attn itself, so we can trust it as ground truth.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "seq_lens",
```
LxYuan0420, line 162: This test triggers Metal kernel JIT build and real kernel execution. Please gate it from the default fast suite (either @pytest.mark.slow or an env flag like VLLM_METAL_RUN_KERNEL_TESTS=1).

tests/test_metal_unified_attention.py
```diff
@@ -0,0 +1,461 @@
...
+
+import mlx.core as mx
+import numpy as np
+import pytest
+
+# TODO: import from vllm_metal.metal once the kernel is implemented
+# from vllm_metal.metal import metal_unified_attention
+
+NUM_HEADS = [(4, 4), (8, 2), (5, 1)]
+HEAD_SIZES = [128]
```
LxYuan0420, line 28: Non-blocking suggestion: current params are intentionally small for speed (head_size=128, num_blocks=256). In a follow-up, we can add one larger stress combo (head_size=256 and bigger num_blocks) behind @pytest.mark.slow or an env flag for broader coverage.

tests/test_metal_unified_attention.py
WindChimeRan: mlx and mlx-lm are all hard dependencies in pyproject.toml, and the repo is macOS-only. If MLX isn't   importable, the environment is broken and we want a loud failure, not a silent skip. The guard in test_metal_kernel_paged.py was likely from claude code's overdefensive coding style and we shouldn't propagate it.

tests/test_metal_unified_attention.py
WindChimeRan: Kernel JIT tests are intentionally in the fast suite. They're cheap (small tensors, no model loading, around 12 seconds) and validate the critical  path. `@pytest.mark.slow` is reserved for full model load + generation. 

An env var gate goes against the direction upstream vLLM is   heading (see vllm-project/vllm#25700,  reducing env var proliferation). If the kernel can't JIT, we want CI to tell us  immediately.

tests/test_metal_unified_attention.py
WindChimeRan: Agreed. 

Added the original upstream Triton parameters back as comments so they're tracked. 

These were designed for NVIDIA HBM and may not be realistic for unified memory on Apple Silicon,  so we'll need to try them out once we have a v2 kernel prototype running. At that point we can pick the ones that make sense.

tests/test_metal_unified_attention.py
LxYuan0420: Agree. Keeping this in the fast suite is fine as long as runtime stays stable in CI.

Issue #148, opened 2026-03-09
[RoadMap] [Paged KV] Continuous Batching + Chunked Prefilling + paged varlen flash att


WindChimeRan:
## Roadmap: Paged KV → Continuous Batching

**Goal:** Enable real continuous batching on Apple Silicon by replacing mlx_lm's   KVCache with block-managed paged attention, closing the gap with upstream vLLM's scheduling model.

**Problem:** The mlx_lm KVCache path cannot participate in vLLM's block-level scheduling — it reports fake memory capacity to avoid crashing, effectively disabling real continuous batching & chunked prefilling (#145).

### Stage 1: Paged KV Prototype

**Core challenge: Intercept the mlx_lm attention layer, replace it with paged attention, and make sure it works.**

- [x] Integrate paged KV cache kernel (from HuggingFace), prototype with torch MPS KV cache, let the scheduler operate on it #92
- [x] nanobind + vendored Metal kernel #136
- [x] Listen to the scheduler to preempt sequences and release blocks when paged KV is full #128
- [x] Separate env vars for paged KV path and mlx_lm path #115
- [x] Deterministic test: paged path vs original mlx_lm path #138
- [x] Benchmark 

<img width="2661" height="926" alt="Image" src="https://github.com/user-attachments/assets/8e386fa0-6ee7-4aec-a1a9-09a00abb38b1" />

* 83x TTFT
* 3.6x throughput

### Stage 2: Varlen Paged KV Kernel

**why varlen matters**: Varlen attention lets us pack multiple sequences of different lengths into a single kernel launch, which is the prerequisite for mixed prefill+decode batches.

- [x] Simplify KV layout from `[num_blocks, kv_heads, head_dim//x, block_size, x]` to `[num_blocks, block_size, num_kv_heads, head_size]` #146
- [x] Unit test for paged KV varlen kernel #147
- [x] fork kernel_v1 to kernel_v2, then implement one-pass online-softmax (v1 is using textbook two-pass softmax), then direcly replace v1 kernel (by freeze some unified attention parameters). #156
- [x] extend kernel_v2 to varlen, pass all triangle tests  #166
- [x] wire kernel_v2 to Stage 3 packed prefilling (mx.SDPA version) #170
- [x] Log End2End benchmark results. Performance regreesions are expected. #170
- [x] revisit the numeric stability issue #167 cleanup stale tests. 

Non-blocking:
[...]

PR #151, opened 2026-03-09
[Continuous Batching] Packed prefill with cu_seq_lens for multiple requests

ricky-chaoju:
## Summary
截圖 2026-03-10 上午9 15 51

- Pack multiple complete prefill requests into a single forward pass using cumulative sequence lengths (`cu_seq_lens`) as separators
- Build block-diagonal causal mask so packed requests don't cross-attend
- Apply per-request RoPE position reset within packed sequences
- Fall back to single-request path when only 1 prefill is scheduled

This is the first step of Stage 3 (Chunked Prefilling & Continuous Batching) in the roadmap (#148).

## Changes

- `paged_attention_common.py`: Add `cu_seq_lens` field to `PagedAttentionContext`; add `prepare_prefill_packed()`
- `paged_attention.py`: Add `_build_packed_causal_mask()`; update `_metal_kernel_prefill_attention()` for packed mode
- `model_runner.py`: Add `_prefill_packed_paged()`; restructure Phase 1 to collect complete prefills and batch them
- `test_paged_attention.py`: Add tests for packed slot_mapping, cu_seq_lens, and block-diagonal causal mask isolation

## Test

- [x] Unit tests: 9/9 pass (`pytest tests/test_paged_attention.py`)
- [x] E2E: Qwen3-0.6B with `VLLM_METAL_USE_PAGED_ATTENTION=1`, 3 concurrent requests all return correct responses
- [x] Verified packed path triggered via debug log: scheduler batches 2 complete requests into packed prefill

### DIFF
--- a/tests/test_paged_attention.py
+++ b/tests/test_paged_attention.py
@@ -170,3 +170,53 @@ class TestPackedRoPE:
         assert q_out.shape == (1, 1, 5, 2)
         assert mx.allclose(q_out, mx.zeros_like(q_out)).item()
         assert mx.allclose(k_out, mx.zeros_like(k_out)).item()
+
+
+class TestBatchSplitting:
+    """Tests for the packed-prefill batch splitting logic."""
+
+    @staticmethod
+    def _split_batches(
+        entries: list[tuple[int, int]],
+        max_tokens: int,
+    ) -> list[list[tuple[int, int]]]:
+        """Reproduce the batch splitting algorithm from _run_packed_prefill.
+
+        entries: list of (index, num_tokens) for simplicity.
+        """
+        batches: list[list[tuple[int, int]]] = [[]]
+        batch_tokens = 0
+        for entry in entries:
+            entry_tokens = entry[1]
+            if batch_tokens + entry_tokens > max_tokens and batches[-1]:
+                batches.append([])
+                batch_tokens = 0
+            batches[-1].append(entry)
+            batch_tokens += entry_tokens
+        return batches
+
+    def test_all_fit_single_batch(self):
+        entries = [(0, 100), (1, 200), (2, 300)]
+        batches = self._split_batches(entries, max_tokens=4096)
+        assert len(batches) == 1
+        assert batches[0] == entries
+
+    def test_split_into_two_batches(self):
+        entries = [(0, 3000), (1, 2000)]
+        batches = self._split_batches(entries, max_tokens=4096)
+        assert len(batches) == 2
+        assert batches[0] == [(0, 3000)]
+        assert batches[1] == [(1, 2000)]
+
+    def test_single_large_request_not_dropped(self):
+        # A request exceeding the cap should still go into its own batch
+        entries = [(0, 5000)]
+        batches = self._split_batches(entries, max_tokens=4096)
+        assert len(batches) == 1
+        assert batches[0] == [(0, 5000)]
+
+    def test_preserves_all_entries(self):
+        entries = [(i, 1000) for i in range(10)]
+        batches = self._split_batches(entries, max_tokens=4096)
+        flat = [e for batch in batches for e in batch]
+        assert flat == entries
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -704,6 +704,12 @@ class STTExecutor:
         return tokens
 
 
+# SCAFFOLDING: remove when varlen kernel is ready.
+# Cap total packed-prefill tokens to bound the O(N²) dense causal mask.
+# Batches exceeding this limit are split into multiple forward passes.
+MAX_PACKED_PREFILL_TOKENS = 4096
+
+
 class MetalModelRunner:
     """Model runner for MLX-based inference on Metal.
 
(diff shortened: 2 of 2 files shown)


vllm_metal/v1/model_runner.py
```diff
@@ -1757,90 +1856,141 @@ def execute_model(
...
+                    cache=[],
+                    sampling_params=sp,
+                    generator=gen,
+                    generated_tokens=1,
+                    block_ids=bids,
+                )
+                if self._rust_state_manager is not None:
+                    self._rust_state_manager.add_request(rid, list(tids) + [nt])
+        elif len(paged_complete) == 1:
+            # Single request: no packing overhead
```
WindChimeRan: Could we always go through the packed path and drop this elif branch? don't want to overcomplicate the codebase. In the future, once we have a varlen kernel, one request vs multiple request will be the same.

WindChimeRan:
@ricky-chaoju failed on the preemption test

```bash
python tools/repro_block_exhaustion.py
```

could you please take a look? 
```
[rank0]:   File "/Users/ran/workspace/vllm-metal/vllm_metal/metal_kernel_backend/paged_attention.py", line 338, in __call__
[rank0]:     return _metal_kernel_prefill_attention(
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/Users/ran/workspace/vllm-metal/vllm_metal/metal_kernel_backend/paged_attention.py", line 153, in _metal_kernel_prefill_attention
[rank0]:     output = mx.fast.scaled_dot_product_attention(
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: ValueError: [scaled_dot_product_attention] Mask type must promote to output type bfloat16.
```

ricky-chaoju:
Fixed in latest commit (59aed34). The packed causal mask was float32 but the model runs bfloat16 — added .astype(queries.dtype).
                                                                                                                                                                                                                                 
repro_block_exhaustion.py passes locally but KV cache usage only reached 31% (my machine has 192GB RAM), so preemption likely didn't trigger. What fraction triggers exhaustion on your setup? I can rerun with that value.
<img width="1935" height="1021" alt="截圖 2026-03-10 下午1 46 48" src="https://github.com/user-attachments/assets/3446be73-dfe9-49b5-8ecd-95f20be0440a" />

WindChimeRan (approved):
LGTM, do you want to take another look? @LxYuan0420

vllm_metal/metal_kernel_backend/packed_prefill_compat.py
```diff
@@ -0,0 +1,55 @@
...
+    Each request only attends to its own tokens (causally).  Returns an
+    additive mask of shape ``(1, 1, total_len, total_len)`` with 0 for
+    allowed positions and ``-inf`` for blocked positions, suitable for
+    ``mx.fast.scaled_dot_product_attention``.
+
+    SCAFFOLDING: remove when varlen kernel is ready.
+    """
+    # Start with all-blocked, then open causal windows per request
+    mask = mx.full((total_len, total_len), -mx.inf)
+    for i in range(len(cu_seqlens) - 1):
```
LxYuan0420, line 34: I understand this is temporary scaffolding. Since it is active in production path now, please add a packed-length safety cap (or split batches) to avoid O(N²) mask blowups until varlen kernel replaces it. (Packing currently happens unconditionally in model_runner complete-prefill path.)

ricky-chaoju: done

vllm_metal/metal_kernel_backend/paged_attention.py
```diff
@@ -89,25 +93,36 @@ def _metal_kernel_prefill_attention(
...
+        queries, keys = apply_packed_rope(attn_module, queries, keys, ctx.cu_seqlens)
+    else:
+        offset = offset_cache.offset if offset_cache is not None else 0
+        queries = attn_module.rope(queries, offset=offset)
+        keys = attn_module.rope(keys, offset=offset)
+
+    # Causal SDPA
+    # SCAFFOLDING: dense mask — remove when varlen kernel is ready.
+    if ctx.cu_seqlens is not None and len(ctx.cu_seqlens) > 2:
+        attn_mask = build_packed_causal_mask(ctx.cu_seqlens, L).astype(queries.dtype)
```
LxYuan0420: Small follow-up suggestion: could we pass dtype into build_packed_causal_mask and construct the mask directly in that dtype? Right now we allocate default float32 and then cast to queries.dtype, which can increase transient peak memory in the packed path.

tests/test_paged_attention.py
```diff
@@ -61,3 +86,46 @@ def test_prepare_decode(self):
         assert ctx.slot_mapping == [27]
         assert ctx.context_lens == [8]
         assert ctx.offsets == [7]
+
+
+class TestPackedCausalMask:
```
LxYuan0420, line 91: Could we add one focused unit test for apply_packed_rope as well? We already validate packed mask isolation here; a small RoPE-reset test would lock in the other key packed-prefill contract.

vllm_metal/v1/model_runner.py
```diff
@@ -1950,26 +1950,52 @@ def execute_model(
...
+                        token_ids=list(tids) + [nt],
+                        prompt_len=len(tids),
+                        cache=[],
+                        sampling_params=sp,
+                        generator=gen,
+                        generated_tokens=1,
+                        block_ids=bids,
+                    )
+                    if self._rust_state_manager is not None:
+                        self._rust_state_manager.add_request(rid, list(tids) + [nt])
```
LxYuan0420: This commit makes this path harder to reason about.
Could you share your reasoning/trade-offs for this long inlined block so we can align before more changes?

ricky-chaoju: The intent is to split packed requests into batches capped at 4096 tokens, bounding the O(N²) dense mask allocation until the varlen kernel replaces it. I'll extract the batching + state-update loop into a `_run_packed_prefill` helper to keep the main path readable.

LxYuan0420: Good, that is the right intent

Please keep the follow-up scoped to:

- extract only batching + dispatch + state writeback into _run_packed_prefill (no behavior change)
- make 4096 a named constant with a short rationale comment
- add one focused test for split-batch mapping correctness
- keep this as scaffolding-only and avoid adding more policy logic here.

ricky-chaoju: Done in 074c865
1. Extracted `_run_packed_prefill` helper — batching + dispatch + state writeback
2. `MAX_PACKED_PREFILL_TOKENS = 4096` as module-level named constant with rationale
3. Added `TestBatchSplitting` with 4 focused tests (single batch, split, oversized request, entry preservation)
4. No behavior change, scaffolding only

PR #150, opened 2026-03-09
[Feature] Add Qwen3-ASR MLX model and transcriber (1/2)

ricky-chaoju:
## Summary

Add Qwen3-ASR-0.6B as a second STT model alongside Whisper — model implementation and offline transcriber only. No vLLM server changes.

- MLX model: Conv2d audio encoder → Qwen3 causal LM (GQA, QK-norm, RoPE, tied embeddings)
- `Qwen3ASRTranscriber`: prompt construction, greedy decode, post-processing
- Weight sanitization: HF `thinker.*` prefix mapping, Conv2d NCHW→NHWC transpose
- 38 unit tests

Part 1 of 2. Part 2 (`stt/qwen3-asr-integration`) wires this into `vllm serve`.

## Test

```bash
# Unit tests (38, <1s)
pytest tests/test_qwen3_asr.py -v -m "not slow"

# Regression (existing tests unaffected)
pytest tests/ -m "not slow" -q

# Slow tests (requires local model)
pytest tests/test_qwen3_asr.py -v -m slow

```
截圖 2026-03-09 晚上8 59 16
截圖 2026-03-09 晚上9 01 57

截圖 2026-03-09 晚上9 02 19

PR #152, opened 2026-03-11
[Feature] Wire Qwen3-ASR into vLLM serve pipeline (2/2)

ricky-chaoju:
## Summary

- Register `qwen3_asr` model type with `transformers.AutoConfig`, vLLM `_CONFIG_REGISTRY`, and `ModelRegistry`
- Add `Qwen3ASRStub` implementing `SupportsTranscription` + `SupportsMultiModal` so `vllm serve` accepts Qwen3-ASR
- Wire STTExecutor dispatch: transcriber, eot_token, extract_audio_features, decode all branch by model_type
- Token-level post-processing (`_extract_asr_text_tokens`) extracts content between `<asr_text>` and `<|im_end|>`

Depends on #150.

## Design decisions

- **`supports_transcription_only = True`**: Qwen3-ASR is pure ASR, does not support text generation
- **`supports_segment_timestamp = False`**: blocks `verbose_json` at vLLM validation level
- **`get_num_audio_tokens`**: uses `_get_feat_extract_output_lengths()` (Conv2d + chunking), not Whisper hop-length
- **Prompt rebuild in decode**: Qwen3-ASR uses a fixed prompt format — language/task/prompt controls are not supported by this model
- **`_model_type` cached in STTExecutor**: avoids repeated `getattr` dispatch (6 call sites)
- **Token ID caching**: `_asr_text_token_id` / `_im_end_token_id` resolved once, not per-decode
- **`_get_languages`**: reuses `config.py:get_whisper_languages()` (DRY)
- **Audio hash**: full bytes, no truncation

## Test
截圖 2026-03-11 上午11 46 48
截圖 2026-03-11 上午11 47 27

### DIFF
--- a/tests/test_v1_stt_integration.py
+++ b/tests/test_v1_stt_integration.py
@@ -565,8 +565,7 @@ class TestSTTExecutorQwen3ASRDispatch:
         executor.model.encode = capture_encode
 
         mel = np.zeros((128, 500), dtype=np.float32)
-        req = _make_new_req(mm_features=[{"input_features": mel}])
-        result = executor.extract_audio_features(req)
+        result = executor.extract_audio_features(mel)
         assert result is not None
         assert result.shape == (50, 1024)
 
@@ -582,8 +581,7 @@ class TestSTTExecutorQwen3ASRDispatch:
         executor.model.encode = capture_encode
 
         mel = np.zeros((1, 128, 500), dtype=np.float32)
-        req = _make_new_req(mm_features=[{"input_features": mel}])
-        result = executor.extract_audio_features(req)
+        result = executor.extract_audio_features(mel)
         assert result is not None
 
     def test_extract_audio_features_1d_raises(self) -> None:
@@ -591,9 +589,8 @@ class TestSTTExecutorQwen3ASRDispatch:
         executor = _make_qwen3_executor()
 
         mel = np.zeros((500,), dtype=np.float32)
-        req = _make_new_req(mm_features=[{"input_features": mel}])
         with pytest.raises(ValueError, match="rank"):
-            executor.extract_audio_features(req)
+            executor.extract_audio_features(mel)
 
     def test_decode_rebuilds_prompt_with_audio_frames(self) -> None:
         """Qwen3-ASR decode should rebuild prompt using build_prompt_tokens."""


vllm_metal/stt/hf_config.py
```diff
@@ -0,0 +1,481 @@
...
+        """Minimal stub so vLLM's API server can initialise for Qwen3-ASR.
+
+        Never instantiated for inference — vllm-metal's MLX path handles that.
+
+        Implements ``SupportsTranscription`` (for STT API) and
+        ``SupportsMultiModal`` (so audio data flows through the pipeline).
+        Transcription-only — text generation is not supported.
+        """
+
+        supported_languages: Mapping[str, str] = _get_languages()
```
LxYuan0420: Nice progress here. I noticed `supported_languages` is still sourced from Whisper (`_get_languages()`). Could we derive this from Qwen3-ASR config (`support_languages`) so the API does not advertise unsupported languages?

LxYuan0420: Repro code:

```python
from transformers import AutoConfig
from vllm_metal.stt.hf_config import _get_languages, register_qwen3_asr_config
register_qwen3_asr_config()
qwen = AutoConfig.from_pretrained("Qwen/Qwen3-ASR-0.6B")
print(len(_get_languages()) == len(qwen.support_languages))
```

LxYuan0420: The len and the content are different. 

While reproducing this locally, I noticed the naming indirection itself made the issue harder to spot: get_whisper_languages is wrapped/aliased as _get_languages and then used in Qwen stub code. That confusion is a useful signal in itself. In future refactors, it may help to keep model-specific providers explicit so these cross-model couplings are easier to catch early.

vllm_metal/v1/model_runner.py
```diff
@@ -700,11 +721,59 @@ def decode(
             logger.warning("STT: empty prompt_token_ids, returning EOT")
             return [eot]
 
+        if self._model_type == "qwen3_asr":
+            # Qwen3-ASR uses a fixed prompt format — language, task, and
+            # user prompt controls are not supported by this model.
+            # Rebuild prompt with the correct number of audio_pad tokens
+            # matching the audio encoder output length.
+            n_audio_frames = audio_features.shape[0]
+            prompt_token_ids = self.transcriber.build_prompt_tokens(n_audio_frames)
```
LxYuan0420, line 696: I noticed this path rebuilds prompt tokens from audio frame count and overwrites caller-provided `prompt_token_ids`. Could we align this with the public request contract: either honor request-level language/prompt controls end-to-end or reject unsupported controls explicitly?

vllm_metal/v1/model_runner.py
```diff
@@ -660,23 +673,32 @@ def extract_audio_features(self, new_req: Any) -> "mx.array | None":
...
-        elif mel.ndim == 3:
-            mel = mel.transpose(
-                0, 2, 1
-            )  # (batch, n_mels, time) → (batch, time, n_mels)
+        if self._model_type == "qwen3_asr":
+            # Qwen3-ASR encoder expects: (n_mels, time) or (batch, n_mels, time)
+            # HF WhisperFeatureExtractor output shape is already (n_mels, time)
+            if mel.ndim == 3:
+                mel = mel[0]  # drop batch dim → (n_mels, time)
+            features = self.model.encode(mel)
```
LxYuan0420, line 650: Small robustness ask: please add explicit mel-rank validation on this Qwen path (`2D/3D` only). Right now malformed 1D input can pass through and fail deeper in `encode` with a less actionable error.

vllm_metal/stt/hf_config.py
```diff
@@ -0,0 +1,481 @@
...
+    except ValueError:
+        pass  # Already registered
+
+    # 2. Register with vLLM's internal _CONFIG_REGISTRY if available
+    try:
+        from vllm.transformers_utils.config import _CONFIG_REGISTRY
+
+        if "qwen3_asr" not in _CONFIG_REGISTRY:
+            _CONFIG_REGISTRY["qwen3_asr"] = Qwen3ASRHFConfig
+    except (ImportError, AttributeError):
```
LxYuan0420: I understand why we keep compatibility guards here. One request: please avoid fully silent failure paths for registration and log at least debug/warning with exception detail, so real integration breakages are diagnosable.

vllm_metal/stt/hf_config.py
```diff
@@ -0,0 +1,481 @@
...
+                mm_kwargs=mm_kwargs,
+                mm_hashes=mm_hashes,
+                mm_placeholders=mm_placeholders,
+            )
+
+    # ------------------------------------------------------------------
+    # Stub model class
+    # ------------------------------------------------------------------
+
+    class Qwen3ASRStub(nn.Module, SupportsTranscription, SupportsMultiModal):
```
LxYuan0420, line 335: Not blocking this PR. I noticed model-specific behavior is currently split across the stub and runner conditionals; in a follow-up, it would be good to move more of this behind transcriber/executor interfaces so modality boundaries stay clearer as STT support grows.

docs/stt.md
```diff
@@ -54,6 +56,12 @@ Any OpenAI Whisper checkpoint (HuggingFace or MLX format):
 
 MLX-format weights (e.g. from `mlx-community`) are also supported.
 
+### Qwen3-ASR
```
LxYuan0420, line 59: Could we add one model-specific capability note in this Qwen3-ASR section (for example: transcription-only, translation unsupported) so docs and runtime behavior stay aligned?

vllm_metal/stt/hf_config.py
ricky-chaoju: Fixed. Now derived from HF config support_languages via _build_supported_languages() in get_speech_to_text_config (classmethod, runs during server init before validate_language).

vllm_metal/v1/model_runner.py
ricky-chaoju: Fixed. get_generation_prompt now rejects translate with ValueError, and warns on language/prompt parameters that Qwen3-ASR ignores.

docs/stt.md
ricky-chaoju: Added. Qwen3-ASR section now notes transcription-only and auto-detect behavior.

vllm_metal/v1/model_runner.py
ricky-chaoju: Fixed. Added explicit rank check , 1D mel now raises ValueError with actionable message.

vllm_metal/stt/hf_config.py
ricky-chaoju: Fixed. All three except blocks now log at debug level with exception detail.

vllm_metal/v1/model_runner.py
```diff
@@ -700,11 +723,59 @@ def decode(
             logger.warning("STT: empty prompt_token_ids, returning EOT")
```
LxYuan0420: For Qwen, prompt tokens are rebuilt from audio frames, but the empty-prompt return happens first. Please move the empty-prompt check after Qwen prompt rebuild (or keep that check only for Whisper), so empty upstream prompt ids do not return EOT too early.

vllm_metal/stt/hf_config.py
```diff
@@ -0,0 +1,531 @@
...
+        def get_num_audio_tokens(
+            cls,
+            audio_duration_s: float,
+            stt_config: SpeechToTextConfig,
+            model_config: ModelConfig,
+        ) -> int | None:
+            from vllm_metal.stt.qwen3_asr import _get_feat_extract_output_lengths
+
+            # mel frames from audio duration (hop_length=160, 10ms at 16kHz)
+            hop_length = 160
```
LxYuan0420: `get_num_audio_tokens()` hardcodes `hop_length=160` and depends on default `n_window=50`. Please derive both from model/feature config so this stays correct for future Qwen variants.

vllm_metal/v1/model_runner.py
```diff
@@ -596,21 +596,34 @@ def __init__(self, model: Any, model_path: str) -> None:
...
 
     @property
     def eot_token(self) -> int:
-        """End-of-text token ID resolved from the tokenizer."""
+        """End-of-text token ID resolved from the tokenizer or config."""
+        if self._model_type == "qwen3_asr":
+            return self.model.config.eos_token_id
         return self.transcriber.tokenizer.convert_tokens_to_ids("<|endoftext|>")
 
     def extract_audio_features(self, new_req: Any) -> "mx.array | None":
```
LxYuan0420: `main` recently merged the STT serve-boundary adapter (`VLLMSTTRequestAdapter` in `vllm_metal/stt/serve.py`) that unwraps multimodal inputs once at the boundary.

After rebasing onto latest `main`, please keep `mm_features` unwrapping there and change `STTExecutor.extract_audio_features` to accept raw `input_features` (not the full `new_req`). This removes the runtime `getattr/hasattr` + dict fallback and keeps the runner thin/model-agnostic.

This also means `_execute_stt` should normalize via `VLLMSTTRequestAdapter.from_vllm_request(new_req)` and pass `stt_req.input_features` into the executor (and drop the “missing audio => return EOT” path).

Suggested shape:
```python
def extract_audio_features(self, input_features: Any) -> "mx.array":
    ...
```

ricky-chaoju: done

PR #156, opened 2026-03-11
[varlen Kernel] Online softmax v2 kernel (decode-only) [3/n]

WindChimeRan:
## Summary

- Fork `kernels_v1/` → `kernels_v2/` as a clean starting point 
- Replace v1's two-pass softmax with single-pass **online softmax** in the v2 Metal kernel (commit 3)

### How to review

Commit 3 is the main change. +174 LOC

- Commit 1 : merge main — skip this
- Commit 2 : Pure copy of `kernels_v1/` → `kernels_v2/`. No logic changes — skip this.
- **Commit 3** (`5ca89ed`): All actual changes. Review this diff to see the online softmax implementation, C++ bridge additions, Python entry point, and tests.
- Commit 4: swap in kernels_v2, so that we can run end-to-end testing and benchmarks. — one line change

### Why do we need online softmax 
The v1 kernel stores ALL QK logits in threadgroup memory, finds the global max across all KV tokens, then computes softmax — this requires O(seq_len) shared memory per warp. The v2 kernel uses **online softmax**: each warp maintains running `(max, sum, output_accumulator)` state and processes KV blocks one at a time, updating with a correction factor. This eliminates the large shared memory requirement and unblocks variable-length query support (prefill+decode in a single kernel launch).

### Triangle Test Status

```
        ref (pure-MLX naive)
       /         \
  edge 1        edge 3
     /             \
   v1  ── edge 2 ── v2
```

**Edge 1: v1 == ref** (`test_v1_kernel_vs_reference`)
- Already passing before this PR, unchanged.
- Establishes that both v1 and the reference are trustworthy.

**Edge 2: v2 == v1** (`test_metal_unified_attn_decode_only`)
- Before: all 18 cases xfail (kernel was a stub raising `NotImplementedError`)
- After: **all 18 pass**
- This proves the online softmax v2 kernel is a safe drop-in replacement for v1 on decode.

**Edge 3: v2 == ref** (`test_metal_unified_attn`)
- Before: all 24 cases xfail (same stub)
- After: **3 pass** (decode-only, no sliding window, no soft cap), **21 xfail** (varlen / sliding window / soft cap — gated by features not yet in the kernel)
- The 3 passing cases close the triangle for decode: v1==ref ✓, v2==v1 ✓, v2==ref ✓.
- The 21 xfails are the roadmap for the varlen PR.

**Summary**: 0 → 21 passing, 42 → 21 xfail. The full triangle is closed for decode. The remaining xfails are varlen features, not regressions.

## Benchmark

<details>
<summary>baseline: v1 kernel</summary>

```
============ Serving Benchmark Result ============
Successful requests:                     100
Failed requests:                         0
Maximum request concurrency:             32
Request rate configured (RPS):           10.00
Benchmark duration (s):                  101.40
Total input tokens:                      23260
Total generated tokens:                  22061
Request throughput (req/s):              0.99
Output token throughput (tok/s):         217.57
Peak output token throughput (tok/s):    352.00
Peak concurrent requests:                35.00
Total token throughput (tok/s):          446.96
---------------Time to First Token----------------
Mean TTFT (ms):                          966.17
Median TTFT (ms):                        488.17
P99 TTFT (ms):                           3113.86
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          130.50
Median TPOT (ms):                        118.08
P99 TPOT (ms):                           531.20
---------------Inter-token Latency----------------
Mean ITL (ms):                           110.12
Median ITL (ms):                         96.84
P99 ITL (ms):                            469.06
==================================================
```

</details>

<details>
<summary>This PR: v2 kernel online softmax</summary>

```
============ Serving Benchmark Result ============
Successful requests:                     100
Failed requests:                         0
Maximum request concurrency:             32
Request rate configured (RPS):           10.00
Benchmark duration (s):                  102.98
Total input tokens:                      23260
Total generated tokens:                  22061
Request throughput (req/s):              0.97
Output token throughput (tok/s):         214.22
Peak output token throughput (tok/s):    350.00
Peak concurrent requests:                35.00
Total token throughput (tok/s):          440.09
---------------Time to First Token----------------
Mean TTFT (ms):                          584.46
Median TTFT (ms):                        374.08
P99 TTFT (ms):                           2228.14
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          127.54
Median TPOT (ms):                        121.12
P99 TPOT (ms):                           339.99
---------------Inter-token Latency----------------
Mean ITL (ms):                           113.12
Median ITL (ms):                         100.30
P99 ITL (ms):                            491.78
==================================================
```

</details>

## Benchmark Interpretation

* From me: I would expect no significant difference between v1 and v2. Because the prefilling is still using mlx sdpa, and our v1 and v2 kernel only take effect on the decoding side (no prefilling side). 
* From claude code opus 4.6: The worst-case decode latency dropped significantly. This makes sense: v1's threadgroup memory scales with `O(seq_len)`, so the   longest sequences hit higher memory pressure. v2's constant-size shared memory avoids that cliff.

### DIFF
--- a/vllm_metal/metal_kernel_backend/paged_attention.py
+++ b/vllm_metal/metal_kernel_backend/paged_attention.py
@@ -226,8 +226,8 @@ def _metal_kernel_decode_attention(
     max_seq_len = max(ctx.context_lens)
     scale = attn_module.scale
 
-    # Zero-copy paged attention
-    ops.paged_attention_v1(
+    # Zero-copy paged attention (v2, online softmax)
+    ops.paged_attention_v2_online(
         out,
         q_3d,
         cache.key_caches[layer_idx],
@@ -240,7 +240,7 @@ def _metal_kernel_decode_attention(
         max_seq_len,
     )
 
-    # Synchronize GPU: paged_attention_v1 wrote to out's buffer via a raw
+    # Synchronize GPU: paged_attention_v2_online wrote to out's buffer via a raw
     # Metal dispatch that MLX's lazy graph doesn't track.  mx.eval(out) would
     # be a no-op here (out was already evaluated as zeros), so we must use
     # mx.synchronize() to flush the command encoder and wait for the kernel.


WindChimeRan:
@LxYuan0420 Request for review. 

This is a targeted PR on replacing v1's multi-pass softmax to flashatt's one-pass online softmax, which will pave the way for varlen kernel. 

The actual content of this PR is only +174 LOC (in commit 3). And it's only about online softmax.

vllm_metal/metal/paged_ops.cpp
```diff
@@ -245,6 +254,114 @@ void paged_attention_v1_impl(
...
+  int head_size  = static_cast<int>(query.shape(2));
+  int max_blocks = static_cast<int>(block_tables.shape(1));
+
+  // Same kernel name format as v1 — the template instantiation is identical.
+  auto dt = dtype_to_metal(query.dtype());
+  std::string kname =
+      "paged_attention_" + dt + "_cache_" + dt +
+      "_hs" + std::to_string(head_size) +
+      "_bs" + std::to_string(block_size) +
+      "_nt256_nsl32_ps0";
```
ericcurtin, line 294: Should this be ps0 or ps512?

Also I'd prefer we used Rust + Python, if we could rewrite this in Rust that would be preferable. Maybe for the next PR.

WindChimeRan: I use cpp because metal kernel and mlx are all natively C++. 

I only found some unofficial rust bindings, e.g., https://github.com/oxiglade/mlx-rs

LxYuan0420: `ps0` seems intentional for this v2 decode-only path (non-partitioned). Probably a good idea to add a short comment saying why we dont use `ps512` here, and which buffer slots are intentionally not bound when `use_partitioning/use_alibi/use_fp8/use_sinks` are false?

PR #157, opened 2026-03-12
[STT] Refactor request boundary and enforce fail-fast runner behavior

LxYuan0420:
This PR is:
- To move STT request parsing and validation out of the runtime path.
- To make the v1 STT runner use one normalized internal request shape.
- To fail fast on bad STT requests instead of silently returning fallback output.

### Refactor (why this is needed)
The STT runtime path was doing two jobs:
- parsing raw vLLM request shapes
- running STT encode/decode

That made the code harder to maintain and could hide invalid input.

This refactor keeps responsibilities clear:
- serve boundary: parse + validate
- runtime path: execute only

### Note
- Added STT request adapter and normalized DTO in `vllm_metal/stt/serve.py`.
- Added fail-fast validation for missing/invalid `req_id`, `prompt_token_ids`, `mm_features`, and `input_features`.
- Updated `_execute_stt` to normalize request first, then run extract/decode.
- Removed silent malformed-input fallback behavior.

### DIFF
--- a/vllm_metal/stt/serve.py
+++ b/vllm_metal/stt/serve.py
@@ -3,12 +3,8 @@
 
 from __future__ import annotations
 
-from collections.abc import Mapping
 from dataclasses import dataclass
-from numbers import Integral
 from typing import Any
-
-import numpy as np
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


vllm_metal/stt/serve.py
```diff
@@ -0,0 +1,136 @@
...
+        if input_features is None:
+            raise ValueError(f"STT request {req_id!r} must include input_features.")
+
+        if isinstance(input_features, np.ndarray):
+            return input_features
+
+        try:
+            unwrapped_input_features = input_features.data
+        except AttributeError:
+            return input_features
```
WindChimeRan: these try-except block with uncondictional return is a bit confusing. Can we make it cleaner?

LxYuan0420: This path is intentionally handling both payload shapes at the serve boundary: wrapped values (`.data`) and plain values. The current logic is for compatibility and fail-fast behavior, not a functional bug. I agree readability can be improved, but I prefer to keep this PR focused on boundary correctness and do the cleanup in the upcoming STT structural refactor so we do it consistently in one pass.

PR #161, opened 2026-03-15
[Refactor] Move TranscriptionResult to shared STT base module

ricky-chaoju:
## Summary

This PR is a small preparatory STT refactor with no intended behavior change.

It moves `TranscriptionResult` out of `vllm_metal/stt/transcribe.py` into a shared module:

- `vllm_metal/stt/base.py`

To preserve compatibility, `TranscriptionResult` is still re-exported from:

- `vllm_metal.stt.transcribe`
- `vllm_metal.stt`

## Why

This is a minimal first step toward cleaning up the STT package structure before any larger Whisper/Qwen3 split.

I intentionally kept this PR narrow to make review easier and avoid mixing structural cleanup with runtime changes.

## Next Steps
- move model-specific STT code into dedicated subpackages such as `stt/whisper/` and `stt/qwen3/`
- keep `transcribe.py` focused on orchestration and shared entrypoints

### DIFF
--- a/tests/test_transcribe.py
+++ b/tests/test_transcribe.py
@@ -10,6 +10,7 @@ from types import SimpleNamespace
 import mlx.core as mx
 import pytest
 
+from vllm_metal.stt.base import TranscriptionResult as SharedTranscriptionResult
 from vllm_metal.stt.transcribe import (
     _MAX_PROMPT_TOKENS,
     TranscriptionResult,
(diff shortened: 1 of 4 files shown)


LxYuan0420 (commented):
Thanks for the PR, Ricky. This overlaps with the STT refactor currently in progress (we’re already relocating TranscriptionResult as part of that work), and I’d like to avoid introducing a new stt/base.py module that we may undo shortly.

Let’s close this PR for now to reduce churn/merge conflicts. Once the refactor lands, I’ll point you to a small follow-up slice you can help with.

PR #168, opened 2026-03-17
Refactor STT model loading

LxYuan0420:
This PR is:
- To move STT model loading out of `vllm_metal/stt/transcribe.py` into `vllm_metal/stt/loader.py`.
- To centralize `model_type -> constructor` mapping in `vllm_metal/stt/registry.py`.
- To keep `vllm_metal/stt/transcribe.py` orchestration-only while preserving the public `load_model()` API.

Next:
- Extract STT runtime glue out of `vllm_metal/v1/model_runner.py` into `vllm_metal/stt/runtime.py`.
- Move model-specific request/mm payload interpretation into model-owned adapters under `stt/<model>/`.

### DIFF
--- a/vllm_metal/stt/registry.py
+++ b/vllm_metal/stt/registry.py
@@ -16,6 +16,18 @@ STTModel = WhisperModel | Qwen3ASRModel
 STTModelConstructor = Callable[[dict, mx.Dtype], STTModel]
 
 
+def get_stt_model_constructor(model_type: str) -> STTModelConstructor:
+    """Return the model constructor for an STT ``model_type``."""
+    model_type = model_type.lower()
+    try:
+        return _STT_MODEL_CONSTRUCTORS[model_type]
+    except KeyError:
+        raise ValueError(
+            f"Unsupported STT model_type: {model_type!r}. "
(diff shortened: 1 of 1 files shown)


vllm_metal/stt/registry.py
WindChimeRan, line 1: The original dispatch in transcribe.py was:
```
  if model_type == "qwen3_asr":
      return _load_qwen3_asr_model(model_path, config_dict, dtype)
  if model_type in ("", "whisper"):
      return _load_whisper_model(model_path, config_dict, dtype)
  raise ValueError(...)
```

The new registry.py introduces many new abstractions and it seems to be overcomplicating a trivial model loading.

LxYuan0420: This split is intentional. We want transcribe.py/load_model to stay orchestration-only and avoid growing a model-specific if/elif chain (which becomes a "god module" once we add the 3rd/4th STT model and more run time glue).

registry.py is the single source of truth mapping model_type -> constructor, so adding a model becomes "add constructor + register it" without touching loader/transcribe. It’s a straightforward application of separation of concerns, single responsibility, and Open/Closed.

Issue #171, opened 2026-03-17
When running "vllm bench serve --num-prompts=3" `assert request.num_output_placeholders >= 0`


starpit:
When running the vllm-project/vllm's `bench serve --num-prompts=3 --model mlx-community/Llama-3.2-3B-Instruct-4bit`, vllm-metal errors out with an assertion failure. For some reason, this worked for me last week. I'm not quite sure what changed. Using `--num-prompts=1` or `=2` works fine. Any value >3 seems to fail. with `assert request.num_output_placeholders >= 0`

It fails in the same way with model `unsloth/Llama-3.2-3B-Instruct`

full output:
```
INFO 03-17 12:25:52 [__init__.py:43] Available plugins for group vllm.platform_plugins:
INFO 03-17 12:25:52 [__init__.py:45] - metal -> vllm_metal:register
INFO 03-17 12:25:52 [__init__.py:48] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 03-17 12:25:52 [__init__.py:217] Platform plugin metal is activated
INFO 03-17 12:25:53 [importing.py:68] Triton not installed or not compatible; certain GPU-related functions will not be available.
(APIServer pid=42148) INFO 03-17 12:25:56 [api_server.py:1272] vLLM API server version 0.14.1
(APIServer pid=42148) INFO 03-17 12:25:56 [utils.py:263] non-default args: {'model_tag': 'mlx-community/Llama-3.2-3B-Instruct-4bit', 'model': 'mlx-community/Llama-3.2-3B-Instruct-4bit'}
(APIServer pid=42148) INFO 03-17 12:25:57 [model.py:530] Resolved architecture: LlamaForCausalLM
(APIServer pid=42148) INFO 03-17 12:25:57 [model.py:1545] Using max model len 131072
(APIServer pid=42148) INFO 03-17 12:25:57 [scheduler.py:229] Chunked prefill is enabled with max_num_batched_tokens=2048.
(APIServer pid=42148) INFO 03-17 12:25:57 [vllm.py:630] Asynchronous scheduling is enabled.
(APIServer pid=42148) INFO 03-17 12:25:57 [vllm.py:637] Disabling NCCL for DP synchronization when using async scheduling.
(APIServer pid=42148) [2026-03-17 12:25:57] INFO platform.py:258: Metal memory: 68.7GB total, 44.1GB available
[...]

PR #179, opened 2026-03-19
Disable chunked prefill on Metal

LxYuan0420:
This PR is:
- To disable chunked prefill on Metal until the default non-paged MLX path handles chunked-prefill scheduler boundaries correctly.
- To fix the long-prompt async serving crash reported in issue #171 on the default Metal text-generation path.
- To add a regression test that keeps `enable_chunked_prefill` disabled for Metal.

## Note

This issue looked like an async-scheduling problem at first, but the actual root cause is narrower: the default non-paged MLX path does not fully support chunked prefill yet.

In simple terms, vLLM can split long prompt prefill into chunks, but the current non-paged MLX path still behaves like prefill always finishes in one shot. That mismatch breaks scheduler expectations and can crash the server.

This change fixes the issue by disabling chunked prefill for Metal so the scheduler stays on the full-prefill behavior that the current runtime actually supports.

The tradeoff is that long-prompt serving on the default non-paged MLX path may be slower for now. The long-term fix is to make the non-paged MLX path truly chunk-aware, and then re-enable chunked prefill safely instead of broad-brushing the platform by disabling async scheduling.

### DIFF
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -2,6 +2,7 @@
 """Tests for Metal platform."""
 
 import platform
+from types import SimpleNamespace
 
 import pytest
 import torch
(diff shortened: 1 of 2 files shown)


WindChimeRan:
Thanks for the fix @LxYuan0420 . 

I would suggest keeping the mlx_lm path simple and just for correctness reference and not making it more complicated in the future.

PR #183, opened 2026-03-20
[Benchmark] Add prefix caching benchmark script 

ricky-chaoju:
## Summary

- Add `tools/bench_prefix_caching.py` — a wrapper around `vllm bench serve --dataset-name prefix_repetition` to measure prefix caching TTFT  
- Automatically starts a vllm server (or connects to an existing one), runs a shared-prefix workload, and reports TTFT / TPOT / E2EL  
- Establishes a TTFT baseline before implementing prefix caching  

## Usage

```bash
# Auto-starts server:
VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.5 \
    python tools/bench_prefix_caching.py

# Against running server:
python tools/bench_prefix_caching.py --server-url http://localhost:8000

# Custom config:
python tools/bench_prefix_caching.py \
    --prefix-len 512 --suffix-len 128 --num-prefixes 5

### CODE REVIEW

**Review by LxYuan0420 — COMMENTED** [review]
@ricky-chaoju This looks like a thin subprocess.call wrapper around two existing CLIs (vllm serve + vllm bench serve) — no custom logic, no result parsing. I think it's better served as a documented recipe in tools/README.md rather than a standalone script.

Something like:

**Start the server:**
```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.7 \
  vllm serve Qwen/Qwen3-0.6B \
    --port 8000 --max-model-len 2048 --max-num-seqs 8
```

**Run the benchmark:**
```bash
vllm bench serve \
  --backend openai \
  --base-url http://localhost:8000 \
  --model Qwen/Qwen3-0.6B \
  --dataset-name prefix_repetition \
  --num-prompts 100 \
  --prefix-repetition-prefix-len 256 \
  --prefix-repetition-suffix-len 256 \
  --prefix-repetition-num-prefixes 10 \
  --prefix-repetition-output-len 128 \
  --request-rate inf \
  --percentile-metrics ttft,tpot,e2el \
  --metric-percentiles 50,99 \
  --save-result --label baseline
```

Also please park it after the attention benchmark section and so keep tools/README.md ordered easy > complex.

If the script grows to add real value later (e.g., before/after comparison, sweep across prefix lengths), then a script makes sense. For now, docs are enough.

PR #193, opened 2026-03-21
Remove dead BlockAllocator and InputPreparer code from Rust extension

sshkhr:
Removed the `BlockAllocator`, `InputPreparer` and the compute functions from the Rust library since these are now dead code after #128 was merged. Left the `RequestStateManager` (used in model_runner.py) and Rust toolchain unchanged based on discussion regarding the Roadmap in #148


sshkhr:
> LGTM. Please sign the DCO, then it's good to merge.

Done.

> see [#148 (comment)](https://github.com/vllm-project/vllm-metal/issues/148#issuecomment-4104493400) for the followup discussion and the PR.

I went through the discussion. And I agree that the RequestStateManager cleanup can be a separate PR. I will get started on it.

WindChimeRan (approved):

Issue #182, opened 2026-03-20
[RoadMap] [Paged KV] Prefix Caching on Paged KV Cache Path


ricky-chaoju:
## Goal

Enable prefix caching on the paged KV path — requests sharing the same token prefix skip redundant prefill and improve TTFT.

## What upstream already handles

Block hash, cache hit detection (`get_computed_blocks` → `find_longest_cache_hit`), `NewRequestData.num_computed_tokens`, block allocation (cached + new blocks), LRU eviction, ref counting. `enable_prefix_caching` defaults to `True`. No hash/LRU code needed on our side.

## Current state

After #172, prefill and decode share a unified forward path using `paged_attention_v2_online` (varlen kernel). `prepare_unified()` already accepts per-request `start_pos` for slot mapping offset, RoPE offset, and `context_lens = start_pos + num_tokens`. Token IDs are already sliced to `token_ids[computed_tokens:cur_len]` and `computed_tokens` is passed as `start_pos`.

The remaining gap: an explicit assert in `execute_model()` (model_runner.py:1878) blocks `start_pos > 0` for new complete prefills — prefix caching not yet wired.

## What needs to change

1. **Remove the assert** at model_runner.py:1878 — allow `start_pos > 0` for complete prefills with cache hits
2. **Fix `RequestState` creation** — when `start_pos > 0`, the `token_ids` stored in state and passed to `_make_sampling_metadata` must be the full prompt, not just the suffix slice
3. **Fix `_paged_request_seq_lens`** — model_runner.py:1610 sets `len(token_ids)` which is the suffix length; needs to be `start_pos + len(token_ids)` (full sequence length) for correct decode RoPE/slot in subsequent steps
4. **Verify intermediate chunk path** — cached request path (model_runner.py:~1800) also passes `computed` as `start_pos`; verify it works correctly when `computed` reflects a prefix cache hit on a previously chunked request

## Implementation plan

- [x] Remove assert, fix `RequestState` + `_paged_request_seq_lens` tracking #195
- [ ] Test: two requests with shared prefix #283                                                              
- [x] Test: intermediate chunk with num_computed_tokens > 0 #211                                              
- [x] Benchmark: instructions in tools/README.md (#183)

WindChimeRan:
[...]

PR #195, opened 2026-03-22
[Paged KV] Wire prefix cache hits in unified prefill path

ricky-chaoju:
When the upstream scheduler reports `num_computed_tokens > 0` (prefix cache hit), the unified prefill path now handles it correctly instead of asserting.

Changes:
- Remove the `start_pos == 0` assert that blocked prefix cache hits
- Fix `RequestState` to store full prompt `token_ids` and correct `prompt_len`
- Fix `_paged_request_seq_lens` to track `start_pos + suffix_len` for correct decode RoPE/slot mapping

Part of #182.


tests/test_paged_prefix_caching.py
```diff
@@ -0,0 +1,289 @@
...
+                return_value=mx.array(0),
+            ),
+            patch(
+                "vllm_metal.paged_attention_common.prepare_unified",
+            ),
+            patch(
+                "vllm_metal.paged_attention_common.clear_context",
+            ),
+        ):
+            new_req = _make_new_req("req-1", prompt, num_computed_tokens=num_computed)
```
LxYuan0420, line 200: All three tests use `temperature=0.0` with no penalties so the `prompt_for_meta` branch in `_unified_prefill_decode_paged` is never reached. Can we add at least one test with `repetition_penalty != 1.0` to actually exercise the path this PR is fixing?

vllm_metal/v1/model_runner.py
```diff
@@ -1598,16 +1604,23 @@ def _unified_prefill_decode_paged(
...
+                prompt_for_meta = (
+                    full_prompt_token_ids
+                    if full_prompt_token_ids is not None
+                    else token_ids
+                )
                 metadata = self._make_sampling_metadata(
                     [sampling_params],
-                    [token_ids[:prompt_len]],
-                    [token_ids[prompt_len:]],
+                    [prompt_for_meta[:prompt_len]],
```
LxYuan0420, line 1640: Removing the assert opens up a new case: intermediate chunk with `start_pos > 0` and penalties. Here `prompt_len` was set to `len(token_ids)` (suffix length) earlier, so `prompt_for_meta[:prompt_len]` slices the wrong range of the full prompt. The token gets discarded so no output impact today, but worth fixing: `prompt_len = len(full_prompt_token_ids) if full_prompt_token_ids is not None else len(token_ids)`

vllm_metal/v1/model_runner.py
```diff
@@ -1840,18 +1854,41 @@ def execute_model(
...
+                # Recover full prompt for sampling metadata when only
+                # a suffix was forwarded (start_pos > 0).
+                if start_pos > 0:
+                    if is_new:
+                        full_prompt = list(new_reqs_by_id[rid].prompt_token_ids)
+                    else:
+                        state = self._request_states[rid]
+                        full_prompt = state.token_ids[: state.prompt_len]
+                else:
+                    full_prompt = None
```
LxYuan0420: Not quite sure if I get this part?
Feel like it has too many branching logic here and `full_prompt` can be refactored to one-liner and so better readability?

vllm_metal/v1/model_runner.py
```diff
@@ -1471,7 +1476,7 @@ def _unified_prefill_decode_paged(
         all_token_ids.extend(last_tokens)
 
         # Prefill: tokens per request
-        for _, token_ids, _, _, _, _, _ in prefill_reqs:
+        for _, token_ids, _, _, _, _, _, _ in prefill_reqs:
```
LxYuan0420: prefill_reqs is now an 8-element positional tuple. Worth considering a NamedTuple or data class before this grows furthe.

LxYuan0420 (changes requested):

vllm_metal/v1/model_runner.py
ricky-chaoju: Added PrefillRequest(NamedTuple) with 8 named fields.

tests/test_paged_prefix_caching.py
ricky-chaoju: Added TestSamplingMetadataWithPenalties with temperature=0.8, repetition_penalty=1.2 to exercise the advanced sampling path. Asserts _make_sampling_metadata receives the full prompt.

vllm_metal/v1/model_runner.py
ricky-chaoju: Fixed. prompt_len now prioritizes len(full_prompt_token_ids) , before falling back to pr.prompt_len or len(pr.token_ids).

vllm_metal/v1/model_runner.py
ricky-chaoju: Simplified to a chained ternary using self._request_states.get(rid) , state exists for cached and new intermediate requests, only new complete prefills fall back to new_reqs_by_id. Removed the is_new branch.

ericcurtin (approved):

LxYuan0420 (approved):

PR #203, opened 2026-03-23
Refactor to introduce PagedAttentionBackend protocol to decouple worker from Metal kernel

LxYuan0420:
This PR is:
- To decouple model_runner and worker from the concrete MetalPagedKVCache and Metal kernel, replacing the direct dependency with a PagedAttentionBackend protocol
- To lay the groundwork for PR2 (GLM/MLA paged attention support) without scattering if-MLA branches across the codebase
- To correctly size the KV cache for MLA models via a virtual-head override in _resolve_model_dims

Context: 
vllm-metal hardwires model_runner and worker to MetalPagedKVCache and the vendored Metal kernel. GLM uses Multi-head Latent Attention (MLA), which has no q_proj/k_proj/v_proj and cannot go through that kernel. Without this abstraction layer, adding MLA support would require if-is_mla branches scattered across both files. This PR introduces the seam before that happens

Note: 
PR2 adds MLAPagedAttentionBackend in paged_attention_backend/, using MLX-native ops to store the joint latent vector. The worker dispatch becomes a single`if runner.is_mla` check in _setup_paged_attention. Nothing in model_runner changes

### DIFF
--- a/tests/test_paged_deterministic.py
+++ b/tests/test_paged_deterministic.py
@@ -128,7 +128,9 @@ def vllm_outputs():
     # Verify paged KV path is active when requested
     if os.environ.get("VLLM_METAL_USE_PAGED_ATTENTION", "0") == "1":
         runner = llm.llm_engine.model_executor.driver_worker.model_runner
-        assert runner._paged_attention_backend is not None, "Paged attention backend not initialised"
+        assert runner._paged_attention_backend is not None, (
+            "Paged attention backend not initialised"
+        )
         from vllm_metal.metal_kernel_backend.paged_attention import (
             MetalKernelPagedAttentionWrapper,
         )


vllm_metal/v1/model_runner.py
```diff
@@ -82,6 +83,10 @@
 _MIN_BATCH_SIZE_FOR_BATCHING = 2  # Minimum requests to use BatchKVCache
 _MAX_BATCH_SIZE = 64  # Maximum batch size for decode
 
+# MLA default rope head dim (GLM/DeepSeek lineage; used when qk_rope_head_dim
+# is absent from model config).
+_MLA_DEFAULT_QK_ROPE_HEAD_DIM = 64
```
WindChimeRan, line 88: MLA doesn't seem to be common code. Maybe we should move it out of the model_runner.

vllm_metal/v1/model_runner.py
```diff
@@ -669,6 +674,16 @@ def is_stt(self) -> bool:
         """Whether the loaded model is a Speech-to-Text model."""
         return self._is_stt
 
+    @property
+    def is_mla(self) -> bool:
```
WindChimeRan, line 678: None-blocking notes: 

`is_mla` is conceptually similar to `is_sdpa` in #201. maybe we need to reconcile them in future PRs. (Currently they are in different places)

vllm_metal/paged_attention_backend/__init__.py
WindChimeRan, line 1: Why do you create a new folder instead of reusing the old one `metal_kernel_backends`?

WindChimeRan:
@LxYuan0420 

Agree. We can land both. I re-opened #201

vllm_metal/v1/model_runner.py
LxYuan0420: yes this will be moved into mla.py in my next PR

vllm_metal/v1/model_runner.py
LxYuan0420: That's right. will be easier to follow up to reconcile after both PRs merge

vllm_metal/paged_attention_backend/__init__.py
LxYuan0420: that is intentional because the two packages sit at different abstraction levels:

```tree
vllm_metal/
├── metal_kernel_backend/        # vendored C++/Metal kernel code
│   ├── cache.py                 # MetalPagedKVCache - Metal buffer alloc
│   ├── paged_attention.py       # Metal kernel wrappers
│   └── packed_prefill_compat.py
│
├── paged_attention_backend/     # protocol + backends (this PR)
│   ├── protocol.py              # PagedAttentionBackend - backend-agnostic
│   ├── mha.py                   # MHAPagedAttentionBackend (wraps metal_kernel_backend)
│   └── mla.py                   # MLAPagedAttentionBackend - will add in my next PR
│
└── v1/
    ├── worker.py                # selects and wires the backend; no kernel details
    └── model_runner.py
```

So my next PR will add mla.py under paged_attention_backend/. It uses pure MLX ops and has no dependency on metal_kernel_backend/ at all. Putting the Protocol (or MLA) inside a metal_kernel_* namespace would be misleading since MLA never touches the Metal kernel.

Note: worker.py depends on the abstraction, not the implementation, follow classic dependency inversion principles(DIP).

Issue #194, opened 2026-03-22
[Qwen3.5] Tracking issue & roadmap on Qwen3.5 collaboration


WindChimeRan:
Qwen3.5 support is non-trivial. It requires a new linear attention kernel, not just wiring up an existing model. Since multiple folks want to work on this, let’s coordinate before diving in so we’re not duplicating effort or stepping on each other’s toes.

Existing PRs: 

- #201 
- #174
- #121
- #186 
- #189

Challenges: 
* The current attention interception & dispatching system re-route every different attention to the same unified attention backend. We actually nee d a slightly more complex system to dispatch it, e.g., to linear attention for Qwen3.5
* We need a linear attention kernel that can work with our page block and also support varlen. 
* Let's add more TODOs together! 

PR Authors @ericcurtin @ricky-chaoju @laudney 

---

Proposal on the Qwen3.5 roadmap (Please feel free to give any comments, break stage to smaller, or claim any of them!):

**Stage A: Attention dispatch refactor** → #201 
- Per-type dispatch in the paged attention wrapper: SDPA vs linear attention, with per-layer detection
- Stub for `linear_attention_forward` that raises `NotImplementedError`
- Better abstraction, e.g., import from upstream vllm, e.g., refer to `vllm/v1/kv_cache_interface.py`. But if we only have two attention backend, complicated abstraction would be an overkill. 

**Stage B: Hybrid cache allocation** (parallel with A)
- Extend (or create alongside) `MetalPagedKVCache` to allocate per-layer buffers based on layer type: `[num_blocks, block_size, num_kv_heads, head_dim]` for SDPA layers, `[num_blocks, Hv, Dv, Dk]` for linear attention layers
- The hybrid cache must participate in the same vLLM scheduler block management as the existing SDPA cache
- Generate golden tokens from mlx_lm inline cache path on Qwen3.5-0.8B for future kernel validation

**Stage C: Linear attention Metal kernel** (blocked by A + B)
- C++ nanobind kernel that reads/writes recurrent state in-place via block tables
- Implement `linear_attention_forward` to call this kernel
- Deterministic golden token test: paged path should match mlx_lm ground truth, or read as normal English, not gibberish. 
- Kernel benchmark: linear attention kernel vs mlx_lm's `gated_delta_kernel`; end-to-end Qwen3.5 serving throughput

WindChimeRan:
[...]

PR #210, opened 2026-03-24
[Qwen3.5] Hybrid cache allocation for SDPA + linear attention (Stage B)

ricky-chaoju:
## Summary

Allocate per-layer-type cache buffers for hybrid models (Qwen3.5) where SDPA and GDN linear attention layers coexist. This is Stage B of the Qwen3.5 roadmap (#194), builds on the dispatch refactor (Stage A, #201).

- Unwrap `text_config` in `_extract_model_args` so Qwen3.5 dimensions are accessible
- Add `is_hybrid` detection and GDN dimensions to `_resolve_model_dims`
- Emit `FullAttentionSpec` for SDPA layers and `MambaSpec` for GDN layers in `get_kv_cache_spec`
- Fix `get_cache_block_size_bytes` to count only SDPA layers
- Add `LinearAttentionCache` with layout `[num_blocks, Hv, Dv, Dk]` per linear layer
- Add `HybridPagedAttentionBackend` that allocates both `MetalPagedKVCache` (SDPA) and `LinearAttentionCache` (GDN)
- Fail fast with `RuntimeError` when hybrid model enables paged attention (gated until Stage C)
- Only SDPA layers patched; linear layers keep original mlx_lm forward

Ref: #194 (Stage B: Hybrid cache allocation)

## Cache layout

| Layer type | Cache class | Shape per layer |
|---|---|---|
| SDPA | `MetalPagedKVCache` | `[num_blocks, block_size, num_kv_heads, head_dim]` |
| Linear (GDN) | `LinearAttentionCache` | `[num_blocks, Hv, Dv, Dk]` |

Both caches use the same `num_blocks` from the scheduler's memory budget. `get_kv_cache_spec` emits `MambaSpec` for GDN layers so the scheduler groups them separately.

This PR delivers allocation infrastructure to unblock Stage C kernel work.

### DIFF
--- a/vllm_metal/v1/worker.py
+++ b/vllm_metal/v1/worker.py
@@ -328,19 +328,26 @@ class MetalWorker(WorkerBase):
         return 0
 
     def _one_sequence_kv_bytes(self) -> int:
-        """Bytes for one max-length sequence of KV cache (K + V)."""
+        """Bytes for one max-length sequence of cache state."""
         runner = self.model_runner
-        dtype_size = (
-            runner.kv_cache_dtype.size if runner.kv_cache_dtype is not None else 2
+        if runner.kv_cache_dtype is None:
+            raise RuntimeError("KV cache dtype not initialized; runner.load_model()")
+        dtype_size = runner.kv_cache_dtype.size
+
+        num_kv_layers = (
+            runner.num_sdpa_layers if runner.is_hybrid else runner.num_layers
         )
-        return (
-            2  # K and V
-            * runner.num_layers
+        sdpa_kv_bytes = (
+            2
+            * num_kv_layers
             * self.model_config.max_model_len
             * runner.num_kv_heads
             * runner.head_dim
             * dtype_size
         )
+        if runner.is_hybrid:
+            return sdpa_kv_bytes + runner.linear_cache_bytes_per_slot()
+        return sdpa_kv_bytes
 
     def determine_available_memory(self) -> int:
         """Determine available memory for KV cache.
--- a/tests/test_v1_worker.py
+++ b/tests/test_v1_worker.py
@@ -110,3 +110,50 @@ class TestWorkerRunnerBoundaryDelegation:
 
         assert tasks == ("transcription",)
         model_runner.supported_worker_tasks.assert_called_once_with()
+
+
+class TestOneSequenceKvBytes:
+    """_one_sequence_kv_bytes must account for hybrid linear state."""
+
+    def test_non_hybrid_counts_all_layers(self) -> None:
+        # Arrange
+        import mlx.core as mx
+
+        model_runner = SimpleNamespace(
+            is_hybrid=False,
+            num_layers=16,
+            num_kv_heads=8,
+            head_dim=64,
+            kv_cache_dtype=mx.float16,
+        )
+        worker = _make_worker(model_runner, use_paged_attention=False)
+        worker.model_config = SimpleNamespace(max_model_len=2048)
+
+        # Act
+        result = MetalWorker._one_sequence_kv_bytes(worker)
+
+        # Assert — 2 * 16 * 2048 * 8 * 64 * 2
+        assert result == 2 * 16 * 2048 * 8 * 64 * 2
+
+    def test_hybrid_adds_linear_state(self) -> None:
+        # Arrange
+        import mlx.core as mx
+
+        linear_bytes = 1_000_000
+        model_runner = SimpleNamespace(
+            is_hybrid=True,
+            num_sdpa_layers=8,
+            num_kv_heads=4,
+            head_dim=256,
+            kv_cache_dtype=mx.float16,
+            linear_cache_bytes_per_slot=MagicMock(return_value=linear_bytes),
+        )
+        worker = _make_worker(model_runner, use_paged_attention=False)
+        worker.model_config = SimpleNamespace(max_model_len=2048)
+
+        # Act
+        result = MetalWorker._one_sequence_kv_bytes(worker)
+
+        # Assert — SDPA bytes + linear state
+        sdpa_bytes = 2 * 8 * 2048 * 4 * 256 * 2
+        assert result == sdpa_bytes + linear_bytes


tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,346 @@
...
+    text_config flattening at the end.
+    """
+
+    @staticmethod
+    def _unwrap(model_args: dict) -> dict:
+        """Replicate the text_config unwrapping from _extract_model_args."""
+        if (
+            "text_config" in model_args
+            and isinstance(model_args["text_config"], dict)
+            and "num_hidden_layers" not in model_args
```
LxYuan0420: `_unwrap` is a private reimplementation of the logic under test, not a call to `_extract_model_args`. If the production method changes, these three tests pass and no one notices. Test through the real method or an integration path. See `test_mla_paged_backend.py` for the pattern we use.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,346 @@
...
+        # 2 * 8 sdpa_layers * 16 block_size * 4 kv_heads * 256 head_dim * 2 bytes
+        expected = 2 * 8 * 16 * 4 * 256 * 2
+        assert result == expected
+
+    def test_hybrid_linear_cache_bytes(self) -> None:
+        from vllm_metal.v1.model_runner import MetalModelRunner
+
+        runner = self._make_runner(QWEN35_4B_ARGS)
+        result = MetalModelRunner.linear_cache_bytes_per_slot(runner)
+        # 24 linear layers * ((3 * 8192) + (32 * 128 * 128)) * 2 bytes
```
LxYuan0420: I notice `not hasattr(runner, "num_sdpa_layers")` is asserted on a `MagicMock`. Mock objects do not grow arbitrary attributes automatically, so this assertion is trivially true regardless of what `_resolve_model_dims` does.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,346 @@
...
+class TestKVCacheSpec:
+    @staticmethod
+    def _make_runner(args: dict) -> MagicMock:
+        import mlx.core as mx
+
+        from vllm_metal.v1.model_runner import MetalModelRunner
+
+        runner = MagicMock(spec=MetalModelRunner)
+        runner.model_args = dict(args)
+        runner.is_mla = MetalModelRunner.is_mla.fget(runner)
```
LxYuan0420: The expected value is the formula re-derived inline. If the production formula changes and someone updates both sides to the same wrong number, the test stays green. Assert against a structural invariant instead or call a helper that the production code also calls.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,346 @@
...
+        # linear fixed cost.
+        total_budget = 1_000_000_000  # 1 GB
+        expected_sdpa_budget = total_budget - linear_fixed
+        expected_blocks = expected_sdpa_budget // sdpa_per_block
+
+        # Verify the math is correct
+        assert expected_blocks > 0
+        assert (
+            expected_blocks < total_budget // sdpa_per_block
+        )  # fewer than without linear
```
LxYuan0420: I think this is not a test. It does arithmetic and asserts `expected_blocks > 0` without calling any production code. Drop it or replace with a call into the actual worker budget computation.

tools/gen_qwen35_golden_tokens.py
```diff
@@ -0,0 +1,50 @@
...
+        next_token = mx.argmax(out[:, -1, :], axis=-1)
+        mx.eval(next_token)
+        tid = int(next_token.item())
+        tokens.append(tid)
+        out = model(mx.array([[tid]], dtype=mx.int32), cache=cache)
+        mx.eval(out)
+
+    assert tokens == GOLDEN, (
+        f"Golden token mismatch.\nExpected: {GOLDEN}\nGot:      {tokens}"
+    )
```
LxYuan0420: This test downloads `Qwen/Qwen3.5-0.8B` from the network, has no `@pytest.mark.slow` marker, and tests the `mlx_lm` inline cache path, not the paged path. Drop it. If a golden baseline is ever needed, `tools/` is a better home than the test suite.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,193 @@
...
+
+        self._kv_cache: Any = None  # MetalPagedKVCache
+        self._linear_cache: Any = None  # LinearAttentionCache
+        self._num_blocks: int = 0
+
+    def initialize(self, num_blocks: int) -> None:
+        from vllm_metal.metal_kernel_backend.cache import MetalPagedKVCache
+        from vllm_metal.metal_kernel_backend.cache_linear import (
+            LinearAttentionCache,
+        )
```
LxYuan0420: I notice all three methods import inside the function body with no performance justification. These are init/setup paths, not hot paths. Move `MetalPagedKVCache`, `LinearAttentionCache`, `patch_model_attention_metal_kernel`, and `get_ops` to the top of the file.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,193 @@
...
+        self._sdpa_indices: list[int] = []
+        self._linear_indices: list[int] = []
+        for i in range(num_layers):
+            if (i + 1) % full_attention_interval == 0:
+                self._sdpa_indices.append(i)
+            else:
+                self._linear_indices.append(i)
+
+        self._kv_cache: Any = None  # MetalPagedKVCache
+        self._linear_cache: Any = None  # LinearAttentionCache
```
LxYuan0420: Typed as `Any`. Should be `MetalPagedKVCache | None` and `LinearAttentionCache | None`. The `Any` annotation defeats the type checker for all downstream accesses.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,193 @@
...
+            logger.info("Hybrid paged attention Metal kernel warm-up complete")
+        except RuntimeError as e:
+            if _METAL_LANGUAGE_VERSION_ERROR in str(e):
+                raise RuntimeError(
+                    f"Metal kernel incompatible with macOS {macos_version}: {e}"
+                ) from e
+            raise
+
+    def num_blocks(self) -> int:
+        return self._num_blocks
```
LxYuan0420: I think this should raise instead of returning `0` when called before `initialize()`. `MLAPagedAttentionBackend` and `MHAPagedAttentionBackend` both raise `RuntimeError` here. Use the same `_require_initialized` guard for consistency.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,193 @@
...
+                dummy_slot,
+            )
+            mx.eval(cache.key_caches[0])
+            logger.info("Hybrid paged attention Metal kernel warm-up complete")
+        except RuntimeError as e:
+            if _METAL_LANGUAGE_VERSION_ERROR in str(e):
+                raise RuntimeError(
+                    f"Metal kernel incompatible with macOS {macos_version}: {e}"
+                ) from e
+            raise
```
LxYuan0420: This is a copy-paste of `MHAPagedAttentionBackend.warm_up()`: same dummy tensor creation, same `ops.reshape_and_cache` call, same error handling. Extract a shared `_warm_up_sdpa_cache(kv_cache)` helper and call it from both.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,193 @@
...
+            dtype=self._dtype,
+        )
+
+        # Linear cache: fixed-size recurrent state, one slot per concurrent
+        # request.  Uses max_num_seqs (not num_blocks) because linear state
+        # is O(1) per request, unlike SDPA KV which grows with seq_len.
+        conv_dim = (
+            self._linear_num_k_heads * self._linear_key_head_dim * 2
+            + self._linear_num_v_heads * self._linear_value_head_dim
+        )
```
LxYuan0420: I notice the same `conv_dim` formula appears in three places: here, `model_runner.get_kv_cache_spec`, and `model_runner.linear_cache_bytes_per_slot`. Extract it once so a formula change does not require three edits.

vllm_metal/metal_kernel_backend/cache_linear.py
```diff
@@ -0,0 +1,64 @@
...
+        *,
+        num_layers: int,
+        num_blocks: int,
+        conv_kernel_dim: int,
+        conv_dim: int,
+        num_v_heads: int,
+        value_head_dim: int,
+        key_head_dim: int,
+        dtype: mx.Dtype = mx.float16,
+    ) -> None:
```
LxYuan0420: I notice there's no dtype validation here. `MLAPagedLatentCache` guards against unsupported dtypes with `if dtype not in (mx.float16, mx.bfloat16, mx.float32): raise ValueError(...)`. This class should do the same.

vllm_metal/v1/model_runner.py
```diff
@@ -916,6 +937,21 @@ def _resolve_model_dims(self) -> None:
                 args.get("qk_rope_head_dim", _MLA_DEFAULT_QK_ROPE_HEAD_DIM)
```
LxYuan0420: `_MLA_DEFAULT_QK_ROPE_HEAD_DIM` is renamed to `MLA_DEFAULT_QK_ROPE_HEAD_DIM` (made public, moved to `mla.py`) in PR #213. This will conflict at merge time.

vllm_metal/v1/model_runner.py
```diff
@@ -962,16 +998,39 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
...
-            layer_name = f"layers.{layer_idx}.self_attn"
-            specs[layer_name] = FullAttentionSpec(
-                block_size=block_size,
-                num_kv_heads=self.num_kv_heads,
-                head_size=self.head_dim,
-                dtype=torch_dtype,
-            )
+            if self.is_hybrid and (layer_idx + 1) % self.full_attention_interval != 0:
+                # GDN linear attention layer — fixed-size recurrent state
+                from vllm.v1.kv_cache_interface import MambaSpec
```
LxYuan0420: Lazy import inside the for-loop body, executed once per GDN layer iteration. Move `from vllm.v1.kv_cache_interface import MambaSpec` to the top of `get_kv_cache_spec`.

vllm_metal/v1/model_runner.py
```diff
@@ -962,16 +998,39 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
...
+
+                conv_dim = (
+                    self.linear_num_k_heads * self.linear_key_head_dim * 2
+                    + self.linear_num_v_heads * self.linear_value_head_dim
```
LxYuan0420: I think the `conv_dim` formula (`linear_num_k_heads * linear_key_head_dim * 2 + linear_num_v_heads * linear_value_head_dim`) is duplicated three times: here, again in `linear_cache_bytes_per_slot` (line ~1083), and in `hybrid.py:initialize`. Extract a shared helper or a runner property.

vllm_metal/v1/worker.py
```diff
@@ -260,23 +271,44 @@ def _setup_paged_attention(self) -> None:
...
-            num_layers=runner.num_layers,
-            num_kv_heads=runner.num_kv_heads,
-            head_dim=runner.head_dim,
-            block_size=block_size,
-            dtype=runner.kv_cache_dtype,
-        )
+        if runner.is_hybrid:
+            from vllm_metal.paged_attention_backend.hybrid import (
+                HybridPagedAttentionBackend,
+            )
```
LxYuan0420: Lazy import inside the `if runner.is_hybrid` branch. Move `from vllm_metal.paged_attention_backend.hybrid import HybridPagedAttentionBackend` to the top of `_setup_paged_attention` alongside the existing `MHAPagedAttentionBackend` import.

vllm_metal/v1/worker.py
```diff
@@ -260,23 +271,44 @@ def _setup_paged_attention(self) -> None:
...
+                head_dim=runner.head_dim,
+                linear_num_k_heads=runner.linear_num_k_heads,
+                linear_num_v_heads=runner.linear_num_v_heads,
+                linear_key_head_dim=runner.linear_key_head_dim,
+                linear_value_head_dim=runner.linear_value_head_dim,
+                linear_conv_kernel_dim=runner.linear_conv_kernel_dim,
+                block_size=block_size,
+                dtype=runner.kv_cache_dtype,
+            )
+        else:
```
LxYuan0420: I notice the `if runner.is_hybrid / else` dispatch here will grow into an elif chain as MLA paged attention lands. Consider a `_make_backend()` factory so the branching stays in one place.

LxYuan0420 (changes requested):
Thanks for picking this up. A few design notes for future PRs:

- **DRY at the concept level.** `conv_dim` is duplicated three times because it is not modeled as a named property. Derived values should live in one place, not reconstructed inline by every caller.

- **Open/closed on backend dispatch.** The `if is_hybrid / else` branch in the worker will need editing every time a new model type arrives. A `_make_backend()` factory fixes this. I will handle this separately.

- **Tests as specification, not mirrors.** A test that reimplements the same formula proves nothing. Tests should verify real behavior against the actual implementation.


- **Avoid `Any` fields.** It loses type safety locally and poisons every downstream call site. Use the real type or a Protocol.

tools/gen_qwen35_golden_tokens.py
LxYuan0420, line 1: Let's delete it. We already have sample files that work for the same purpose. One .py per model in tools/ doesn't scale.

tests/test_hybrid_model_dims.py
LxYuan0420, line 1: Few things: 
 
- QWEN35_4B_ARGS / LLAMA_ARGS must come from a config source, not be defined inline in the test
- Byte-count assertions (524_288, 26_345_472, 8192) are magic numbers so compute them from the input constants
- HybridPagedAttentionBackend(...) constructor duplicated verbatim in two tests, extract a helper
- test_non_hybrid_no_linear_attrs catches AttributeError from MagicMock(spec=...), not from real runner behaviour
- test is passing for the wrong reason

tests/conftest.py
```diff
@@ -41,3 +41,37 @@ def _seed_random_generators() -> None:
     if mlx_seed is None:
         return
     mlx_seed(seed)
+
+
+# === Model config fixtures ===
```
LxYuan0420: Blocking: Remove model config fixtures from `conftest.py`. `conftest` is shared test infrastructure only. Move these fixtures into `tests/test_hybrid_model_dims.py` and keep only the keys required by each test case.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,229 @@
...
+            + args["linear_num_value_heads"] * args["linear_value_head_dim"]
+        ),
+        "block_size": 16,
+        "dtype": mx.float16,
+    }
+    defaults.update(overrides)
+    return HybridPagedAttentionBackend(**defaults)
+
+
+# === _extract_model_args ===
```
LxYuan0420: Remove banner section comments (`# === ... ===`). They add noise and are inconsistent with this repo’s test style.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,229 @@
...
+    runner.is_mla = MetalModelRunner.is_mla.fget(runner)
+    runner.is_hybrid = MetalModelRunner.is_hybrid.fget(runner)
+    runner._is_stt = False
+    runner.kv_cache_dtype = mx.float16
+    runner.metal_config = MagicMock(block_size=16)
+    MetalModelRunner._resolve_model_dims(runner)
+    return runner
+
+
+def _make_hybrid_backend(args: dict, **overrides):
```
LxYuan0420: No lazy import here. Import `HybridPagedAttentionBackend` at module scope with the rest of imports. Lazy imports need a documented perf or circular-dependency reason.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,229 @@
...
+    @staticmethod
+    def _call_extract(model_args_dict: dict) -> dict:
+        runner = MagicMock(spec=MetalModelRunner)
+        runner.metal_config = MagicMock(debug=False)
+        runner.model = SimpleNamespace(args=SimpleNamespace(**model_args_dict))
+        runner._is_vlm = False
+        MetalModelRunner._extract_model_args(runner)
+        return runner.model_args
+
+    def test_qwen35_text_config_unwrapped(self, qwen35_4b_args: dict) -> None:
```
LxYuan0420: This should test behavior, not config constants. Assert merge semantics only: keys from `text_config` are populated into `model_args`, and top-level keys still win when both exist. Avoid asserting copied model default values.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,229 @@
...
+        from vllm.v1.kv_cache_interface import FullAttentionSpec
+
+        specs = MetalModelRunner.get_kv_cache_spec(_make_runner(llama_args))
+        assert all(isinstance(s, FullAttentionSpec) for s in specs.values())
+
+
+# === HybridPagedAttentionBackend ===
+
+
+class TestHybridBackend:
```
LxYuan0420: Rewrite in strict AAA structure. Keep Arrange, Act, Assert visually separated so intent is obvious and maintenance is easier.

vllm_metal/v1/worker.py
```diff
@@ -260,28 +269,57 @@ def _setup_paged_attention(self) -> None:
...
+            runner.is_hybrid,
         )
 
         runner._paged_attention_backend = backend
         runner._paged_block_size = block_size
 
+    @staticmethod
+    def _make_backend(runner: MetalModelRunner, block_size: int) -> Any:
+        """Create the right paged attention backend for the model type."""
+        from vllm_metal.paged_attention_backend.hybrid import (
```
LxYuan0420: No lazy imports in backend factory. Move these imports to module scope unless there is a measured startup or cycle issue documented in this PR.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,145 @@
...
+            model, cache, self._block_size, cache_idx_map=sdpa_cache_idx
+        )
+
+    def warm_up(self) -> None:
+        warm_up_paged_cache(self._require_initialized("warm_up"))
+
+    def num_blocks(self) -> int:
+        return self._require_initialized("num_blocks").num_blocks
+
+    @property
```
LxYuan0420: Blocking: kv_cache and linear_cache properties must return concrete types, not optional. Call _require_initialized() in both properties and fail fast before initialization.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,229 @@
...
+        assert args["num_hidden_layers"] == qwen35_4b_args["num_hidden_layers"]
+        assert (
+            args["full_attention_interval"] == qwen35_4b_args["full_attention_interval"]
+        )
+
+    def test_flat_args_unchanged(self, llama_args: dict) -> None:
+        args = self._call_extract(llama_args)
+        assert args["num_hidden_layers"] == llama_args["num_hidden_layers"]
+
+    def test_top_level_takes_precedence(self) -> None:
```
LxYuan0420: Test name and behavior are currently mismatched. Today, if top-level num_hidden_layers exists, text_config is not merged at all. Either fix _extract_model_args or rename this test and explicitly assert current behavior.

vllm_metal/v1/model_runner.py
```diff
@@ -962,16 +1008,34 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
 
         torch_dtype = MLX_TO_TORCH_DTYPE[self.kv_cache_dtype]
 
-        # Create a spec for each layer
+        # Create a spec for each layer.  Hybrid models (Qwen3.5) emit
```
LxYuan0420: Keep ModelRunner thin. Move hybrid MambaSpec shape construction to vllm_metal/paged_attention_backend/hybrid.py (add a top-level builder there) and call it from here. Runner should orchestrate, not own model-family-specific spec assembly.

vllm_metal/v1/model_runner.py
```diff
@@ -916,6 +942,26 @@ def _resolve_model_dims(self) -> None:
                 args.get("qk_rope_head_dim", _MLA_DEFAULT_QK_ROPE_HEAD_DIM)
             )
 
+        # Hybrid (Qwen3.5): mix of SDPA and GDN linear attention layers.
+        # Store per-type layer counts and GDN dimensions for cache allocation.
+        if self.is_hybrid:
```
LxYuan0420, line 950: Stop re-deriving layer classification in multiple places. In _resolve_model_dims, store self._sdpa_layer_indices: frozenset[int]; then use that set in get_kv_cache_spec instead of repeating (layer_idx + 1) % full_attention_interval logic.

vllm_metal/v1/model_runner.py
```diff
@@ -997,20 +1061,41 @@ def get_cache_block_size_bytes(self) -> int:
...
             2
-            * self.num_layers
+            * num_kv_layers
             * block_size
             * self.num_kv_heads
             * self.head_dim
             * dtype_size
         )
 
+    def linear_cache_bytes_per_slot(self) -> int:
```
LxYuan0420, line 1076: Add explicit guard: if not self.is_hybrid: raise RuntimeError(...). This method currently crashes with opaque AttributeError on non-hybrid runners.

vllm_metal/v1/model_runner.py
```diff
@@ -857,6 +872,17 @@ def _extract_model_args(self) -> None:
                 "Cannot extract model config: model has neither .args nor "
                 ".config attribute."
             )
+        # Qwen3.5 wrapper: model.args has {model_type, text_config: {…}}
```
LxYuan0420: Blocking: Flattening logic is too narrow. Remove the and "num_hidden_layers" not in self.model_args gate. Always merge text_config with setdefault so top-level values keep precedence while missing hybrid keys are still populated.

vllm_metal/metal_kernel_backend/paged_attention.py
```diff
@@ -119,8 +133,13 @@ def patch_model_attention_metal_kernel(
...
+        cache_idx = (
+            cache_idx_map[layer_idx]
+            if cache_idx_map is not None and layer_idx in cache_idx_map
+            else layer_idx
+        )
         wrapper = MetalKernelPagedAttentionWrapper(
-            attn, layer_idx, kv_cache, block_size
+            attn, layer_idx, kv_cache, block_size, cache_idx=cache_idx
         )
         setattr(layer, attn_attr, wrapper)
```
WindChimeRan, line 152: does it wrap EVERY layer that has attention? Are you expecting a skip from the cache_idx?

vllm_metal/metal_kernel_backend/paged_attention.py
```diff
@@ -71,12 +78,11 @@ def __call__(self, x: mx.array, mask: Any = None, cache: Any = None) -> mx.array
...
         # Dispatch to the right attention backend
+        cache_idx = self._mk_cache_idx
         if is_sdpa(inner):
-            return sdpa_forward(inner, x, ctx, self._mk_kv_cache, self._mk_layer_idx)
+            return sdpa_forward(inner, x, ctx, self._mk_kv_cache, cache_idx)
         elif is_linear_attention(inner):
-            return linear_attention_forward(
-                inner, x, ctx, self._mk_kv_cache, self._mk_layer_idx
-            )
+            return linear_attention_forward(inner, x, ctx, self._mk_kv_cache, cache_idx)
```
LxYuan0420, line 85: Blocking: hybrid linear layers are still dispatched to linear_attention_forward(...), which is a known stub that raises. We should fail fast during backend setup with a clear RuntimeError, not on first decode token.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,167 @@
...
+from vllm_metal.metal_kernel_backend.cache_linear import LinearAttentionCache
+from vllm_metal.metal_kernel_backend.paged_attention import (
+    patch_model_attention_metal_kernel,
+)
+from vllm_metal.paged_attention_backend.mha import warm_up_paged_cache
+
+logger = init_logger(__name__)
+
+
+def build_linear_layer_spec(runner: Any, torch_dtype: Any) -> Any:
```
LxYuan0420: build_linear_layer_spec(runner: Any, torch_dtype: Any) -> Any is too loose and runner-coupled. Use explicit typed args and return MambaSpec directly.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,167 @@
...
+    def patch_model(self, model: Any) -> int:
+        cache = self._require_initialized("patch_model")
+
+        # Map SDPA layer indices to compact cache indices so the wrapper
+        # indexes into the compact MetalPagedKVCache correctly.
+        sdpa_cache_idx = {
+            layer_idx: cache_idx
+            for cache_idx, layer_idx in enumerate(self._sdpa_indices)
+        }
+        return patch_model_attention_metal_kernel(
```
LxYuan0420: Blocking: this path patches hybrid layers even though linear execution is not implemented yet. Please gate hybrid paged-attention setup with a clear RuntimeError until Stage C wiring exists.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,167 @@
...
+        self,
+        *,
+        num_layers: int,
+        full_attention_interval: int,
+        max_num_seqs: int,
+        # SDPA dims
+        num_kv_heads: int,
+        head_dim: int,
+        # GDN dims
+        linear_num_k_heads: int,
```
LxYuan0420: linear_num_k_heads is passed through but not consumed by backend logic. Remove it from constructor/factory/tests unless it is actually used.

vllm_metal/v1/worker.py
```diff
@@ -260,45 +272,55 @@ def _setup_paged_attention(self) -> None:
...
+    def _make_backend(runner: MetalModelRunner, block_size: int) -> Any:
+        """Create the right paged attention backend for the model type."""
+        if runner.is_hybrid:
+            return HybridPagedAttentionBackend(
                 num_layers=runner.num_layers,
+                full_attention_interval=runner.full_attention_interval,
+                max_num_seqs=runner.scheduler_config.max_num_seqs,
                 num_kv_heads=runner.num_kv_heads,
                 head_dim=runner.head_dim,
+                linear_num_k_heads=runner.linear_num_k_heads,
```
LxYuan0420: linear_num_k_heads is passed through but not consumed by backend logic. Remove this pass-through unless it is actually used.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,277 @@
...
+        assert args["shared_key"] == "top"
+
+    def test_no_text_config_unchanged(self) -> None:
+        """Models without text_config pass through unchanged."""
+        args = self._call_extract({"num_hidden_layers": 24})
+
+        assert args["num_hidden_layers"] == 24
+
+
+class TestIsHybrid:
```
LxYuan0420: This is tautological for a one-line property (fai > 0). Drop this block or fold into a higher-value behavior test.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,277 @@
...
+        assert _make_runner(LLAMA_ARGS).is_hybrid is False
+
+
+class TestResolveModelDims:
+    def test_hybrid_layer_counts(self) -> None:
+        # Arrange
+        runner = _make_runner(QWEN35_4B_ARGS)
+        n = QWEN35_4B_ARGS["num_hidden_layers"]
+        fai = QWEN35_4B_ARGS["full_attention_interval"]
+        expected_sdpa = sum(1 for i in range(n) if (i + 1) % fai == 0)
```
LxYuan0420: This expected value mirrors production logic. Assert fixed ground truth for this fixture (e.g. SDPA=8, linear=24) instead of recomputing the same formula.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,277 @@
...
+        assert runner.num_layers == n
+        assert runner.num_sdpa_layers == expected_sdpa
+        assert runner.num_linear_layers == n - expected_sdpa
+
+    def test_hybrid_conv_dim(self) -> None:
+        # Arrange
+        runner = _make_runner(QWEN35_4B_ARGS)
+        a = QWEN35_4B_ARGS
+        expected = (
+            a["linear_num_key_heads"] * a["linear_key_head_dim"] * 2
```
LxYuan0420: This expected value mirrors the implementation formula. Use fixed ground truth here (conv_dim=8192 for this fixture).

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,277 @@
...
+        # Arrange
+        runner = _make_runner(QWEN35_4B_ARGS)
+
+        # Act
+        specs = MetalModelRunner.get_kv_cache_spec(runner)
+
+        # Assert
+        n = QWEN35_4B_ARGS["num_hidden_layers"]
+        assert len(specs) == n
+        assert (
```
LxYuan0420: This assertion is circular because counts come from the same runner state. Assert concrete layer keys/types instead (e.g. layers.3.self_attn -> FullAttentionSpec, layers.0.linear_attn -> MambaSpec).

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,277 @@
...
+        assert result == expected
+
+    def test_linear_cache_bytes_positive(self) -> None:
+        # Act
+        result = MetalModelRunner.linear_cache_bytes_per_slot(
+            _make_runner(QWEN35_4B_ARGS)
+        )
+
+        # Assert
+        assert result > 0
```
LxYuan0420: assert result > 0 is too weak. Use an exact expected byte value or remove this test.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,277 @@
...
+        # Arrange
+        budget = MetalWorker._kv_budget_bytes(
+            metal_limit=16_000_000_000, model_memory=4_000_000_000, fraction=0.5
+        )
+        runner = _make_runner(QWEN35_4B_ARGS)
+        linear_fixed = MetalModelRunner.linear_cache_bytes_per_slot(runner) * 8
+        sdpa_per_block = MetalModelRunner.get_cache_block_size_bytes(runner)
+
+        # Act
+        blocks_without = budget // sdpa_per_block
```
LxYuan0420: This only checks arithmetic direction. Please test actual _setup_paged_attention behavior and assert resulting num_blocks after linear-cache subtraction.

vllm_metal/v1/worker.py
```diff
@@ -174,6 +175,14 @@ def _setup_paged_attention(self) -> None:
         runner = self.model_runner
         block_size = self.metal_config.block_size
 
+        if runner.is_hybrid:
```
LxYuan0420, line 177: Blocking: this PR now hard-raises for hybrid here, so the hybrid-only `linear_cache_bytes` block below and the hybrid branch in `_make_backend()` are unreachable in the same patch. Pick one direction: either keep the fail-fast and delete the dead Stage C scaffolding, or make Stage B runnable.

vllm_metal/paged_attention_backend/hybrid.py
```diff
@@ -0,0 +1,173 @@
...
+from vllm_metal.metal_kernel_backend.cache_linear import LinearAttentionCache
+from vllm_metal.metal_kernel_backend.paged_attention import (
+    patch_model_attention_metal_kernel,
+)
+from vllm_metal.paged_attention_backend.mha import warm_up_paged_cache
+
+logger = init_logger(__name__)
+
+
+def build_linear_layer_spec(
```
LxYuan0420: This should not be a public module function. There is one internal caller and no API boundary here. If we keep the helper in this file, make it `_build_linear_layer_spec`.

vllm_metal/v1/model_runner.py
```diff
@@ -973,16 +1011,31 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
 
         torch_dtype = MLX_TO_TORCH_DTYPE[self.kv_cache_dtype]
 
-        # Create a spec for each layer
+        if self.is_hybrid:
+            from vllm_metal.paged_attention_backend.hybrid import (
```
LxYuan0420: No lazy import here. If this helper stays, import it at module scope. I do not see a concrete import cycle requiring a function-scope import.

vllm_metal/paged_attention_backend/mha.py
```diff
@@ -18,6 +18,41 @@
 _METAL_LANGUAGE_VERSION_ERROR = "language version"
 
 
+def warm_up_paged_cache(cache: MetalPagedKVCache) -> None:
+    """Trigger Metal shader compilation with a dummy reshape_and_cache call.
+
+    Shared by MHA and Hybrid backends to avoid duplicating warm-up logic.
+    """
+    from vllm_metal.metal import get_ops
```
LxYuan0420: No lazy import here unless there is a real circular dependency. I do not see one through `vllm_metal.metal`, so please move `get_ops` to module scope or document the cycle.

tests/test_hybrid_model_dims.py
```diff
@@ -0,0 +1,211 @@
+# SPDX-License-Identifier: Apache-2.0
```
LxYuan0420: Please delete this file for now. It is built around hand-copied model config dicts and mostly re-asserts copied constants or local setup state. I do not want to add fake config scaffolding here.

ricky-chaoju:
Kept fail-fast, deleted dead scaffolding (HybridPagedAttentionBackend, LinearAttentionCache, worker hybrid branch, budget subtraction, test file). Fixed lazy imports in model_runner and mha.py. Renamed to _build_linear_layer_spec.

vllm_metal/v1/worker.py
LxYuan0420, line 330: Blocking: `_one_sequence_kv_bytes()` still uses the old “all layers have token-growing KV” formula, but this PR changes the hybrid cache contract to SDPA KV + fixed linear state.

That matters here because hybrid paged attention is now hard-disabled, so Qwen3.5 still goes through `determine_available_memory()` -> `_one_sequence_kv_bytes()` on the live MLX path. I do not want to defer this to follow-up; it should be fixed in the same PR that changes hybrid cache sizing semantics.

Suggested to have something like:
```py
def _one_sequence_kv_bytes(self) -> int:
    """Bytes for one max-length sequence of cache state."""
    runner = self.model_runner
    if runner.kv_cache_dtype is None:
        raise RuntimeError("KV cache dtype not initialized; runner.load_model()")
    dtype_size = runner.kv_cache_dtype.size

    num_kv_layers = runner.num_sdpa_layers if runner.is_hybrid else runner.num_layers
    sdpa_kv_bytes = (
        2
        * num_kv_layers
        * self.model_config.max_model_len
        * runner.num_kv_heads
        * runner.head_dim
        * dtype_size
    )
    if runner.is_hybrid:
        return sdpa_kv_bytes + runner.linear_cache_bytes_per_slot()
    return sdpa_kv_bytes
```

Please also add one worker-path unit test so this does not regress and once this is fixed, I am okay to approve from my side

PR #213, opened 2026-03-25
Add paged attention backend for MLA models

LxYuan0420:
This PR is:
- To add MLAPagedAttentionBackend and MLAPagedAttentionWrapper implementing the PagedAttentionBackend protocol from #203, using MLX-native scatter/gather in place of the vendored Metal kernel
- To add MLAPagedLatentCache storing a joint [kv_norm || k_pe_roped] latent per token per layer
- To wire MLA backend dispatch in the worker and add mla_latent_dim to avoid depending on the _resolve_model_dims head_dim override
- To drop the NotImplementedError("MLA is not supported") gate, letting MLA models load on Metal/MLX
- To move MLA_DEFAULT_QK_ROPE_HEAD_DIM into mla.py as committed in #203 review

Context:

GLM/DeepSeek models use Multi-head Latent Attention (MLA). Unlike standard transformers that cache separate K and V tensors per head, MLA compresses all KV information into a single low-rank latent vector per token. The existing Metal paged attention kernel expects per-head K/V arrays and cannot be used. This backend stores the compressed latent in an MLX-native paged cache and computes attention by splitting the latent back into its positional (RoPE) and content (no-positional-encoding, nope) components at query time.

Smoke test:

```bash
VLLM_HOST_IP=127.0.0.1 \
vllm serve mlx-community/GLM-4.7-Flash-4bit --max-model-len 512
```

```bash
curl -s http://localhost:8000/v1/chat/completions \
-d '{"model":"mlx-community/GLM-4.7-Flash-4bit","messages":[{"role":"user","content":"What is MLA?"}],"max_tokens":64}' | jq -r '.choices[0].message.content'
```

Confirm in server logs: `MLA paged attention enabled`

Note:

`get_cache_block_size_bytes()` currently reports ~2x conservative capacity for MLA models due to the virtual-head override (`num_kv_heads=1`, `head_dim=latent_dim`) going through the standard 2 * K+V sizing path. This is safe but wastes capacity, tracked for follow-up.

### DIFF
--- a/tests/test_mla_paged_backend.py
+++ b/tests/test_mla_paged_backend.py
@@ -1,0 +1,426 @@
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import math
+from collections.abc import Generator
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import mlx.core as mx
+import mlx.nn as nn
+import pytest
+
+import vllm_metal.paged_attention_common as pac
+
+from vllm_metal.mlx_backend.mla_cache import MLAPagedLatentCache
+from vllm_metal.paged_attention_backend.mla import (
+    MLAPagedAttentionBackend,
+    MLAPagedAttentionWrapper,
+    patch_model_attention_mla,
+)
+from vllm_metal.paged_attention_backend.protocol import PagedAttentionBackend
+
+# Fixture dimensions matching GLM/DeepSeek-V2 defaults
+_KV_LORA_RANK = 512
+_QK_ROPE_HEAD_DIM = 64
+_LATENT_DIM = _KV_LORA_RANK + _QK_ROPE_HEAD_DIM
+
+
+class TestMLAPagedLatentCache:
+    def test_latent_dim_is_sum_of_lora_rank_and_rope_head_dim(self) -> None:
+        cache = MLAPagedLatentCache(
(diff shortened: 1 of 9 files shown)


WindChimeRan (changes requested):
Nice scaffolding PR. This would unblock MLA models on vllm-metal and the development of MLA kernel. Two requests before merging:

1. **Add a deterministic gold-token test**

 The unit tests validate plumbing (shapes, slot writes, causal mask isolation) but not end-to-end correctness. Since this PR hand-rolls chunked prefilling and decoding in a per-request loop, the correctness risk is real and shape-level tests can't catch it.

Please add a test (can live in `tools/`) that loads a small MLA model (e.g., `mlx-community/GLM-4.7-Flash-4bit`), runs greedy decode via `mlx_lm.generate` for gold tokens, then runs the same prompt through the paged MLA path and asserts token-for-token match.
 
2. Add a brief README to `mlx_backend/` (or just move everything to `metal/` so we can directly deprecate it when we have a kernel.)

The directory name is ambiguous. it holds shared MLX utilities and now gains MLA-specific cache code. A short README clarifying this is a scaffolding module (pure-MLX paged attention, to be superseded by varlen paged MLA Metal kernels) would help future contributors understand the intended lifecycle.

WindChimeRan (approved):

PR #224, opened 2026-04-02
Refactor execute_model into typed model runner helpers

LxYuan0420:
This PR is:
- To split `MetalModelRunner.execute_model()` into focused private helpers without changing execution order or runtime behavior.
- To replace tuple-heavy local bookkeeping with typed `_ExecutionBatch` and `_PendingPrefillEntry` state.
- To add regression coverage for empty-batch execution, cached-request block updates, and paged cached-request continuation paths.

Verification:
- Ran `tools/gen_golden_token_ids_for_deterministics.py` on `origin/main` and this branch.
- Compared both the default MLX path and the paged path with `Qwen/Qwen3-0.6B`.
- `GOLDEN_MLX` and `GOLDEN_PAGED` matched exactly, so this targeted check showed no behavior drift.

Notes:
- This keeps the current runtime shape intact while making the execution steps easier to reason about and change safely.
- The next planned work is to keep separating step-local batching flow from persistent runner state so later model-runner cleanup can happen in smaller, behavior-preserving pieces.

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -1678,7 +1678,8 @@ class MetalModelRunner:
             if state is None:
                 logger.warning(
                     "Paged cached request %s has no RequestState; "
-                    "emitting placeholder token. This is a state tracking bug.",
+                    "emitting placeholder token. This indicates scheduler/runner "
+                    "state desync.",
                     req_id,
                 )
                 batch.add_output(req_id, [0])
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -1767,9 +1767,6 @@ class MetalModelRunner:
         prefill_pack: list[PrefillRequest],
     ) -> None:
         """Run paged prefill/decode and write results back into ``batch``."""
-        if not batch.has_paged_work():
-            return
-
         prefill_tokens, decode_tokens = self._unified_prefill_decode_paged(
             prefill_pack, batch.paged_decode_reqs
         )
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -1575,7 +1575,7 @@ class MetalModelRunner:
 
             if self._paged_attention_backend is not None:
                 sched_block_ids = list(new_req.block_ids[0])
-                scheduled_tokens = scheduler_output.num_scheduled_tokens.get(req_id, 0)
+                scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]
                 computed_tokens = new_req.num_computed_tokens
                 prompt_len = len(token_ids)
                 cur_len = computed_tokens + scheduled_tokens
@@ -1852,9 +1852,7 @@ class MetalModelRunner:
                 if new_req is not None:
                     prompt_len = len(new_req.prompt_token_ids or [])
                     computed_tokens = new_req.num_computed_tokens
-                    scheduled_tokens = scheduler_output.num_scheduled_tokens.get(
-                        req_id, 0
-                    )
+                    scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]
                     is_intermediate_ctx = (
                         computed_tokens + scheduled_tokens < prompt_len
                     )
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -38,7 +38,12 @@ from vllm.logger import init_logger
 from vllm.sampling_params import SamplingParams
 from vllm.tasks import SupportedTask
 from vllm.utils.platform_utils import is_pin_memory_available
-from vllm.v1.core.sched.output import GrammarOutput, SchedulerOutput
+from vllm.v1.core.sched.output import (
+    CachedRequestData,
+    GrammarOutput,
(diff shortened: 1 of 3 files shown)


vllm_metal/v1/model_runner.py
```diff
@@ -1615,306 +1611,331 @@ def execute_model(
...
-                    cached_full_prompt = prefill_pack[i].full_prompt_token_ids
-                    full_prompt = (
-                        cached_full_prompt if cached_full_prompt is not None else tids
+    def _run_paged_batch(
+        self,
+        batch: _ExecutionBatch,
+        prefill_pack: list[PrefillRequest],
+    ) -> None:
+        """Run paged prefill/decode and write results back into ``batch``."""
+        if not batch.has_paged_work():
```
WindChimeRan: dead code. double gated already in `execute_model`

vllm_metal/v1/model_runner.py
```diff
@@ -1925,15 +1946,44 @@ def execute_model(
...
-            pooler_output=[None] * len(req_ids),
+            pooler_output=[None] * len(batch.req_ids),
         )
         return None
 
+    def execute_model(
+        self, scheduler_output: SchedulerOutput
+    ) -> ModelRunnerOutput | None:
+        """Execute model inference with true batched decode."""
+        if self.model is None:
```
WindChimeRan, line 1950: We already have early catch in `load_model`

vllm_metal/v1/model_runner.py
```diff
@@ -1615,306 +1611,331 @@ def execute_model(
...
+                if state is not None:
+                    batch.valid_decode_reqs.append((req_id, state))
+            return
+
+        req_id_to_cached_idx = {
+            req_id: idx for idx, req_id in enumerate(cached_reqs.req_ids)
+        }
+        for req_id in decode_req_ids:
+            state = self._request_states.get(req_id)
+            if state is None:
```
WindChimeRan, line 1678: Can you elaborate this bug? If this is indeed a bug, should we open a new tracking issue for this?

vllm_metal/v1/model_runner.py
```diff
@@ -1615,306 +1611,331 @@ def execute_model(
...
+
+        if self._paged_attention_backend is None:
+            batch.scheduled_cached_req_ids.extend(decode_req_ids)
+            for req_id in decode_req_ids:
+                state = self._request_states.get(req_id)
+                if state is not None:
+                    batch.valid_decode_reqs.append((req_id, state))
+            return
+
+        req_id_to_cached_idx = {
```
WindChimeRan: this function (especially this dict build) is overcomplicated. It can be simplified and faster.

vllm_metal/v1/model_runner.py
```diff
@@ -1615,306 +1611,331 @@ def execute_model(
...
-                    block_ids=[],
                 )
+                batch.add_output(req_id, [0])
+                continue
 
-        # --- Cached requests ---
-        decode_req_ids = list(cached_reqs.req_ids)
+            if state.generated_tokens == 0:
+                idx = req_id_to_cached_idx.get(req_id)
+                if idx is not None and idx < len(cached_reqs.num_computed_tokens):
```
WindChimeRan: idx always not Note?

vllm_metal/v1/model_runner.py
```diff
@@ -1615,306 +1611,331 @@ def execute_model(
...
+            self._paged_request_seq_lens.pop(req_id, None)
+
+    def _collect_cached_requests(
+        self,
+        batch: _ExecutionBatch,
+        cached_reqs: CachedRequestData,
+        scheduler_output: SchedulerOutput,
+    ) -> None:
+        """Classify cached requests into prefill continuation or decode work."""
+        decode_req_ids = list(cached_reqs.req_ids)
```
WindChimeRan: why do we need to materialize it?

vllm_metal/v1/model_runner.py
```diff
@@ -1615,306 +1611,331 @@ def execute_model(
...
 
-        # --- Cached requests ---
-        decode_req_ids = list(cached_reqs.req_ids)
+            if state.generated_tokens == 0:
+                idx = req_id_to_cached_idx.get(req_id)
+                if idx is not None and idx < len(cached_reqs.num_computed_tokens):
+                    computed_tokens = cached_reqs.num_computed_tokens[idx]
+                else:
+                    computed_tokens = self._paged_request_seq_lens.get(req_id, 0)
+                scheduled_tokens = scheduler_output.num_scheduled_tokens.get(req_id, 0)
```
WindChimeRan: should be `scheduler_output.num_scheduled_tokens[req_id]`?

vllm_metal/v1/model_runner.py
LxYuan0420: This is meant as a warning for unexpected scheduler/runner state desync, not a known tracked bug. I reworded the log to make the intent clearer

vllm_metal/v1/model_runner.py
LxYuan0420: I think it is still useful as a fail-fast invariant check in a core runtime path

WindChimeRan (approved):

PR #226, opened 2026-04-03
[Qwen3.5] C++ Metal kernel for hybrid SDPA + GDN paged inference(Stage C)

ricky-chaoju:
## Summary
- Add C++ nanobind Metal kernel for GDN linear attention with in-place paged state management
- Enable end-to-end `vllm serve` for Qwen3.5 hybrid models (75% GDN + 25% SDPA)
- Decompose mlx_lm GDN forward: conv1d (per-request) + recurrent update (batched C++ kernel)
- Add rope validation compat patch for transformers 5.4+
- Add M-RoPE support for mlx_vlm Qwen3.5 SDPA layers
- Golden token test: 4/4 MATCH on Qwen3.5-0.8B (paged vs inline cache)

## Benchmark (Qwen3.5-0.8B, 10 reqs, input=128, output=64, 3-run avg)

qwen35_benchmark

|  | Prompt tok/s | Gen tok/s |
|---|---|---|
| MLX-native (per-request loop) | 257 | 28.6 |
| C++ Metal kernel (batched) | 998 | 110.9 |
| **Speedup** | **3.9x** | **3.9x** |

## design
- GDN recurrent state uses float32 pool to avoid float16 overflow during accumulation
- Conv1d remains per-request (stateful), recurrent update is batched via C++ kernel
- Auto-enables paged attention for hybrid models in worker.load_model()
- `support_hybrid_kv_cache() = True` enables vLLM's hybrid KV cache manager

Ref: #194 (Stage C)


ericcurtin:
@ricky-chaoju @WindChimeRan I'm ok with this, but are we sure we want to do C++ ? Maybe we'd have to remove the mlx dependancy, but it might be worth it:

https://github.com/ericcurtin/inferrs

ericcurtin (approved):

LxYuan0420:
My view is:
- there is nothing fundamentally wrong with the current repo, and it is still moving in a good direction
- Rust/Candle is exciting and promising, and I would be happy to explore it
- personally I would prefer to explore that path in a separate repo first

so if we do want to seriously explore that direction, it might be good to first agree on what the starting repo should be, so the effort stays focused?

CC: @ericcurtin

ricky-chaoju:
> @ricky-chaoju This PR modified Qwen3's hot path (e.g, `vllm_metal/metal_kernel_backend/attention_sdpa.py`)
> 
> I'm not sure if this is the right way to patch it. Maybe we'll have other model to hack and then attention_sdpa will be ballooned. At least we need some benchmark to show if Qwen3 has performance regression.

The improvement comes from #226's unified varlen kernel, not from #232. Direct A/B of #232 alone is not possible because main without #232 crashes on Qwen3 (head_dim AttributeError). Micro-benchmark confirms the per-call overhead of the fix is 0.25 µs on a ~ms-scale forward pass, no regression.
benchmark_232

PR #240, opened 2026-04-07
Add Qwen3-Next GDN support and fix multi-request GPU hang

gwlee-dev:
## Summary

Add Qwen3-Next GatedDeltaNet support and fix multi-request GPU hang on Apple Silicon.

### Problem

1. **Qwen3-Next models fail to load**: `AttributeError: 'Qwen3NextGatedDeltaNet' object has no attribute 'in_proj_qkv'` — Qwen3-Next uses combined projections (`in_proj_qkvz` + `in_proj_ba`) unlike Qwen3.5's separate ones.

2. **GPU hang on second request**: After the first successful inference, subsequent requests cause `kIOGPUCommandBufferCallbackErrorHang`. Root cause: `_gdn_free_slot` does not materialize the previous request's lazy computation graph before zeroing state, so the next request inherits a deep dependency chain that stalls the Metal GPU.

3. **MLX/MPS sync**: Missing synchronization between MLX Metal commands and PyTorch MPS tensor transfers causes `torch.AcceleratorError`.

### Changes

- **`attention_linear.py`**: Detect Qwen3-Next projection style (`in_proj_qkvz`) at runtime, handling both Qwen3-Next and Qwen3.5 variants.
- **`model_runner.py`**: Add `mx.eval` before and after slot zeroing in `_gdn_free_slot` to materialize existing arrays (breaking the lazy graph) then materialize the zeroed state. Only the freed slot is zeroed, preserving other active requests' state.
- **`tensor_bridge.py`**: Add `sync_mlx()` before MLX-to-PyTorch MPS tensor transfer.

### Test

Tested on M2 Ultra 192GB with `mlx-community/Qwen3-Next-80B-A3B-Instruct-8bit` (max-model-len=131072, max-num-seqs=15):

**Multi-request stability:**
| | Before | After |
|---|---|---|
| Request 1 | 1.5s | 3.1s (warm-up) |
| Request 2 | GPU hang | 0.5s |

**Concurrent throughput (2048 token output, realistic prompt):**
| Concurrent | Total throughput | Wall time | Failed |
|---|---|---|---|
| 1 | 26.6 tok/s | 44.4s | 0 |
| 6 | 89.8 tok/s | 126.2s | 0 |
| 15 | 131.6 tok/s | 195.4s | 0 |

Qwen3.5-0.8B and Qwen3-0.6B regression verified — no impact.

## Test plan

- [x] `pytest tests/test_qwen3_next_gdn.py -k "not slow"` — 6 unit tests pass
- [x] `pytest tests/ -m "not slow"` — 395 tests pass
- [x] `ruff check` — no lint errors
- [x] Golden token test: `mlx_lm` greedy reference IDs for 6 prompts (`@pytest.mark.slow`)
- [x] Qwen3-Next-80B-A3B multi-request inference (1, 6, 15 concurrent)
- [x] Qwen3.5-0.8B regression (existing GDN model)
- [x] Qwen3-0.6B regression (non-GDN model)

### DIFF
--- a/tests/test_qwen3_next_gdn.py
+++ b/tests/test_qwen3_next_gdn.py
@@ -23,7 +23,6 @@ from __future__ import annotations
 from unittest.mock import MagicMock, patch
 
 import mlx.core as mx
-import pytest
 
 from vllm_metal.mlx_backend.gdn_cache import GDNPagedStateCache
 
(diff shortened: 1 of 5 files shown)


vllm_metal/v1/model_runner.py
```diff
@@ -1042,13 +1042,14 @@ def _gdn_free_slot(self, req_id: str) -> None:
...
+                # Recreate fresh zero arrays to fully detach from
+                # previous request's computation graph.
                 for layer_idx in range(sc.num_layers):
-                    conv = sc.conv_states[layer_idx]
-                    conv[slot] = 0
-                    sc.conv_states[layer_idx] = conv
-                    rec = sc.recurrent_states[layer_idx]
-                    rec[slot] = 0
-                    sc.recurrent_states[layer_idx] = rec
+                    sc.conv_states[layer_idx] = mx.zeros_like(
```
WindChimeRan: could you please try with `max_num_seqs` more than 1? This seems to be a bug ( _gdn_free_slot zeros ALL slots, not just the freed one).

WindChimeRan (commented):
Could you please use `mlx_lm` to generate 5-6 groundtruth labels, then compare the outputs with this paged path model implementation, under greedy decoding? That's our common practice to add new models, to ensure the correctness. (reference `tests/test_paged_deterministic.py`)

gwlee-dev (commented):
Added golden token deterministic test in `tests/test_qwen3_next_gdn.py` — greedy-decoded reference IDs generated with `mlx_lm` on `mlx-community/Qwen3-Next-80B-A3B-Instruct-8bit`. Marked `@pytest.mark.slow` since it requires the 80B model download.

Also addressed the `_gdn_free_slot` issue — now zeros only the freed slot (with `mx.eval` before/after to detach the computation graph), preserving other active requests' state. Removed the per-layer `mx.eval(x)` and `mx.eval(conv_states)` calls that were unnecessary with this fix.

tests/test_qwen3_next_gdn.py
```diff
@@ -0,0 +1,243 @@
...
+    """Run vLLM offline inference once for all prompts."""
+    from vllm import LLM, SamplingParams
+
+    llm = LLM(model=MODEL_NAME, max_model_len=512, max_num_seqs=1)
+    sp = SamplingParams(temperature=0, max_tokens=MAX_TOKENS)
+    outputs = llm.generate(PROMPTS, sp)
+    return {o.prompt: o for o in outputs}
+
+
+class TestQwen3NextGolden:
```
WindChimeRan: even if you mark the gold test as slow, if someone just want to run all tests (including the slow ones), then this test will explode their machine by downloading & running a giant model. 

Maybe we should move it to `tools/`?

WindChimeRan (changes requested):

ricky-chaoju (commented):
Code changes LGTM. Verified locally:

- `fix_query_key_value_ordering` exists in mlx_lm 0.31.2 and returns `(q, k, v, z, b, a)` with correct shapes matching downstream `compute_g`/kernel dispatch
- The `mx.eval` before/after slot zeroing in `_gdn_free_slot` correctly breaks the lazy graph to prevent GPU hang
- `sync_mlx()` placement before MPS transfer is correct
- 6 unit tests pass, 295 existing tests pass with no regression

Only issue: agree with @WindChimeRan — please move `TestQwen3NextGolden` to `tools/` (where `test_qwen35_golden.py` already lives). I hit a PermissionError just running the slow tests because it tried to download the 80B model.

gwlee-dev:
Rebased to fix commit email — accidentally used work email. No code changes.

WindChimeRan (approved):

PR #246, opened 2026-04-08
Add vllm-project doc using mkdocs

WindChimeRan:
## Summary
- Set up mkdocs doc system same with upstream vllm
- Migrate existing content (README sections, docs/stt.md) into the doc site — no new content written

## Next steps
- Register project on ReadTheDocs
- Request sub-project linking under `docs.vllm.ai/projects/metal/`


docs/Makefile
mgoin, line 1: Instead of sphinx could you use mkdocs like upstream vllm?

WindChimeRan: image

done. Also made a quick style match. (content is a bit outdated. Will rework in future PRs)

ericcurtin:
Could we remove the .PNG and the other icon files. Binary data in git repos is a bad idea generally. Slows down git clone time.

ericcurtin:
And could we as closely follow the main vLLM project as much as possible (lowering the amount of tool dependencies etc.)

WindChimeRan:
@ericcurtin the tools dependency & logos are directly vendored from the upstream vllm. (we are using a very small subset of the upstream one)

ericcurtin:
Content duplication between README and docs, loose version constraints in requirements-docs.txt, site_url may be empty locally

WindChimeRan:
@ericcurtin 

> Content duplication between README and docs

The scope of this PR is just to setup things. Not focusing on the content. Currently both doc and the README are outdated. Will make following PRs on updating them. 

> loose version constraints in requirements-docs.txt, site_url may be empty locally

These are all aligned with the upstream vllm.

docs/Makefile
ericcurtin: This looks super cool!

ericcurtin:
I still think committing the .ico and .png is a mistake from upstream vLLM, no binary files whatsoever keeps clones super fast. Can we just curl these down at build time? Keeping the repo binary free. They are not very diffable and big (might say it's just a few kb now but that's how it starts)

WindChimeRan:
> I still think committing the .ico and .png is a mistake from upstream vLLM, no binary files whatsoever keeps clones super fast. Can we just curl these down at build time? Keeping the repo binary free. They are not very diffable and big (might say it's just a few kb now but that's how it starts)

@ericcurtin I'm not clear on how to achieve this (I tried some options and they are all fragile). Could you please push changes to this PR directly? If it works well, we can make a PR on the upstream vllm as well.

ericcurtin:
Added

ericcurtin (approved):

WindChimeRan:
@ericcurtin The fix pinned url to vllm upstream, so we cannot apply the fix to upstream recursively ... 

How about we put those logo into GitHub Release asset on vllm-metal itself, and curl it during pre-build? 

In the future we'll also need to make our own logo (similar to vllm-ascend)

PR #248, opened 2026-04-09
Fix/logger lower level

Alex-ai-future:
# Fix vllm_metal Logger INFO Level Issue

## Problem

The vllm_metal plugin's INFO level logs were being filtered out. Only WARNING and above were visible.

### Root Cause

Python's root logger defaults to `WARNING` (level 30). vLLM's logging configuration only configures the `"vllm"` logger, not `"vllm_metal"`. This caused all `vllm_metal` loggers to inherit the root logger's WARNING level, filtering out INFO messages.

**Logger hierarchy before fix:**
```
root (WARNING/30)
├── vllm (INFO/20) ← configured by vLLM
│   └── vllm.* (effective level: INFO) ✓
│
└── vllm_metal (NOTSET/0) ← not configured!
    └── vllm_metal.* (effective level: WARNING) ✗
```

## Solution

This PR includes three commits:

### 1. Standardize logger initialization (78ae769)

Changed all logger initialization from `logging.getLogger(__name__)` to `init_logger(__name__)`:

**Files modified (11):**
- `vllm_metal/__init__.py`
- `vllm_metal/platform.py`
- `vllm_metal/utils.py`
- `vllm_metal/compat.py`
- `vllm_metal/metal/__init__.py`
- `vllm_metal/metal/build.py`
- `vllm_metal/pytorch_backend/tensor_bridge.py`
- `vllm_metal/stt/detection.py`
- `vllm_metal/stt/loader.py`
- `vllm_metal/stt/whisper/adapter.py`
- `vllm_metal/stt/whisper/transcriber.py`

**Benefits:**
- Consistent with vLLM's logging pattern
- Access to vLLM's extended methods (`info_once()`, `warning_once()`)
- Better integration with vLLM's logging system

### 2. Set vllm_metal logger level to INFO (34a9a42)

Added explicit level setting in `vllm_metal/__init__.py`:

```python
logging.getLogger("vllm_metal").setLevel(logging.INFO)
```

This ensures all `vllm_metal` loggers have INFO level, regardless of root logger settings.

**Logger hierarchy after fix:**
```
root (WARNING/30)
├── vllm (INFO/20)
│   └── vllm.* (effective level: INFO) ✓
│
└── vllm_metal (INFO/20) ← explicitly set!
    └── vllm_metal.* (effective level: INFO) ✓
```

### 3. Add comprehensive logger tests (5045d29)

Created `tests/test_logger.py` with 15 test cases:

**Test classes:**
[...]

PR #252, opened 2026-04-09
[CI] Fix smoke test race condition between sequential tests

ricky-chaoju:
## Summary
- Add graceful shutdown with SIGKILL fallback after `kill` in smoke tests
- Wait for port 8000 to be released before starting the next test
- Fixes the intermittent Qwen3.5 smoke test failure discussed in #248

## Root cause
The second smoke test's health check could hit the still-alive first server (same port 8000), then the completion request would fail when that server shut down mid-request.

## Changes
- `wait` + 10s grace period + SIGKILL fallback to ensure process termination
- `lsof` port check loop before starting next test

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -65,6 +65,17 @@ run_smoke_test() {
   echo "Smoke test passed! Output matches golden."
 
   kill $vllm_pid
+  wait $vllm_pid 2>/dev/null || true
+  # Wait for port 8000 to be released before next test
+  local i=0
+  while lsof -i :8000 -sTCP:LISTEN >/dev/null 2>&1; do
+    sleep 1
+    i=$((i + 1))
+    if [ $i -ge 30 ]; then
+      echo "Warning: port 8000 still in use after 30s"
+      break
+    fi
+  done
 }
 
 smoke_tests() {
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -64,15 +64,25 @@ run_smoke_test() {
 
   echo "Smoke test passed! Output matches golden."
 
-  kill $vllm_pid
-  wait $vllm_pid 2>/dev/null || true
-  # Wait for port 8000 to be released before next test
+  # Graceful shutdown with SIGKILL fallback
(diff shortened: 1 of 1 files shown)


scripts/test.sh
```diff
@@ -64,7 +64,28 @@ run_smoke_test() {
...
+  # Wait for port to be released before next test
+  i=0
+  while lsof -i :8000 -sTCP:LISTEN >/dev/null 2>&1; do
+    sleep 1
+    i=$((i + 1))
+    if [ $i -ge 15 ]; then
+      echo "Warning: port 8000 still in use after 15s"
+      break
+    fi
+  done
```
LxYuan0420, line 91: Magic numbers and confusing counter placement here. `local i=0` is declared before a loop that never uses it, and `10`/`15` are duplicated in both logic and messages. Also `$(seq ...)` spawns a subshell.

Suggested refactor:

```bash
local shutdown_timeout=10
local port_timeout=15

kill $vllm_pid 2>/dev/null
for ((shutdown_wait=0; shutdown_wait<shutdown_timeout; shutdown_wait++)); do
  if ! kill -0 $vllm_pid 2>/dev/null; then break; fi
  sleep 1
done
if kill -0 $vllm_pid 2>/dev/null; then
  echo "Warning: vLLM did not terminate after ${shutdown_timeout}s, sending SIGKILL"
  kill -9 $vllm_pid 2>/dev/null
fi
wait $vllm_pid 2>/dev/null || true

local port_wait=0
while lsof -i :8000 -sTCP:LISTEN >/dev/null 2>&1; do
  sleep 1
  port_wait=$((port_wait + 1))
  if [ $port_wait -ge $port_timeout ]; then
    echo "Warning: port 8000 still in use after ${port_timeout}s"
    break
  fi
done
```

so this keeps timeouts in one place, removes dead state, and avoids `seq`.

LxYuan0420 (changes requested):

scripts/test.sh
ricky-chaoju: done!

LxYuan0420 (approved):

PR #253, opened 2026-04-09
Fix vllm_metal logging integration with vLLM

LxYuan0420:
This PR is:
- To ensure `vllm_metal` logs respect vLLM’s logging level and handlers at registration time.
- To prevent INFO logs from being silently dropped due to the root logger default.
- To stabilize log capture in tests under the new logger wiring.

Problem
`vllm_metal` loggers were inheriting the root logger (default `WARNING`) instead of vLLM’s configured logger, so INFO logs were filtered out even when vLLM was logging at INFO.

Before
```
root (WARNING/30)
├── vllm (INFO/20)        <- configured by vLLM
└── vllm_metal (NOTSET)   <- inherits root => WARNING
```

After
```
root (WARNING/30)
├── vllm (INFO/20)        <- configured by vLLM
└── vllm_metal (INFO/20)  <- explicitly synced + handlers copied
```

Before/After Capture
```bash
# main
effective_level 30 handlers 0
WARNING should appear on main

# fix/logger-init-minimal
effective_level 20 handlers 1
INFO 04-09 17:54:24 [<stdin>:7] INFO should appear after sync
WARNING 04-09 17:54:24 [<stdin>:8] WARNING should appear after sync
```

### DIFF
--- a/tests/test_macos_defaults.py
+++ b/tests/test_macos_defaults.py
@@ -37,7 +37,14 @@ def test_apply_macos_defaults_logs_when_setting(monkeypatch, caplog) -> None:
     monkeypatch.delenv("VLLM_WORKER_MULTIPROC_METHOD", raising=False)
     monkeypatch.setattr(sys, "platform", "darwin")
 
-    with caplog.at_level(logging.DEBUG):
+    metal_logger = logging.getLogger("vllm_metal")
+    original_level = metal_logger.level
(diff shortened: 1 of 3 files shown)


WindChimeRan (commented):
None blocking: test case can be simplified by `caplog.at_level(logging.DEBUG, logger="vllm_metal") `

WindChimeRan (approved):

PR #256, opened 2026-04-09
refactor env register in a vllm way

WindChimeRan:
similar fix to my upstream PR: https://github.com/vllm-project/vllm/pull/35007

* register all env vars in a vllm way
* dedicated `envs.py`
* update relevant unit tests

### The problem

**before**: unregistered envs warning
image

**after**: the warning disappeared.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -21,8 +21,18 @@ class TestMetalConfig:
         yield
         reset_config()
 
-    def test_default_config(self) -> None:
+    def test_default_config(self, monkeypatch) -> None:
         """Test default configuration values."""
+        for var in [
+            "VLLM_METAL_MEMORY_FRACTION",
+            "VLLM_METAL_USE_MLX",
+            "VLLM_MLX_DEVICE",
+            "VLLM_METAL_BLOCK_SIZE",
+            "VLLM_METAL_DEBUG",
+            "VLLM_METAL_USE_PAGED_ATTENTION",
+        ]:
+            monkeypatch.delenv(var, raising=False)
+
         config = MetalConfig.from_env()
 
         assert config.memory_fraction == AUTO_MEMORY_FRACTION
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -3,6 +3,7 @@
 
 import pytest
 
+import vllm_metal.envs as envs
 from vllm_metal.config import (
     AUTO_MEMORY_FRACTION,
     MetalConfig,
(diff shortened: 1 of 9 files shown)


ricky-chaoju (commented):
Want to mirror the if TYPE_CHECKING: type stub block from vllm/envs.py (lines 14–259) here too?

vllm_metal/envs.py
```diff
@@ -0,0 +1,83 @@
...
+        return environment_variables[name]()
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+def __dir__() -> list[str]:
+    # Mirrors vllm/envs.py; enables tab-completion and introspection.
+    return list(environment_variables.keys())
+
+
+def is_set(name: str) -> bool:
```
LxYuan0420: Non-blocking: Prefer removing is_set and keeping the module minimal. It is unused and the docstring says "not yet used," which is speculative API. Add it only when a caller exists.

tests/test_config.py
```diff
@@ -16,32 +14,12 @@
 class TestMetalConfig:
     """Tests for MetalConfig class."""
 
-    def setup_method(self) -> None:
-        """Reset config before each test."""
```
LxYuan0420, line 20: Blocking: The autouse fixture no longer clears VLLM_METAL_* / VLLM_MLX_* env vars. Tests like test_default_config now depend on external env state, which breaks determinism. Please restore env cleanup in the fixture before reset_config().

vllm_metal/envs.py
```diff
@@ -0,0 +1,83 @@
...
+    # Returns the raw string; config.py handles "auto" → sentinel conversion.
+    "VLLM_METAL_MEMORY_FRACTION": lambda: os.getenv(
+        "VLLM_METAL_MEMORY_FRACTION", "auto"
+    ),
+    # Whether to use MLX as the compute backend (default True).
+    "VLLM_METAL_USE_MLX": lambda: os.getenv("VLLM_METAL_USE_MLX", "1") == "1",
+    # MLX device type: "gpu" (default) or "cpu".
+    "VLLM_MLX_DEVICE": lambda: os.getenv("VLLM_MLX_DEVICE", "gpu"),
+    # Tokens per KV-cache block (default 16).
+    "VLLM_METAL_BLOCK_SIZE": lambda: int(os.getenv("VLLM_METAL_BLOCK_SIZE", "16")),
```
LxYuan0420: Blocking: Move VLLM_METAL_BLOCK_SIZE parsing into MetalConfig.from_env(). envs.py should return raw values, and parsing and validation should live in config so errors include the variable name and stay consistent with other validations.

LxYuan0420 (changes requested):
Good fix. Minor changes needed; then ready to merge.

ericcurtin (approved):

ericcurtin:
Minor: remaining direct os.environ in tool scripts (not core lib), no test for env var registration in _register()

Conflicts to fix also

@LxYuan0420 has the requested changes flag on which will block

LxYuan0420 (approved):
Approve to keep things moving. Currently occupied by other stuff.

PR #259, opened 2026-04-10
Fix hybrid model prefill hang and GDN slot overflow

gwlee-dev:
## Summary

Fix two critical bugs in the hybrid model (SDPA + GDN linear attention) paged attention path that prevented reliable serving on Apple Silicon:

1. **Long prefill hang**: Prompts over ~1145 tokens silently failed to schedule
2. **GDN slot overflow**: Concurrent request turnover caused slot exhaustion and crash

Also regenerates incorrect Qwen3-Next golden token values and adds unit tests.

### 1. Prefill hang (block_size mismatch)

`_build_linear_layer_spec` hardcoded `MambaSpec(block_size=1)`, causing the scheduler to require one block per token for GDN layers. With the unified block pool divided into groups (~4580 blocks / 4 groups = ~1145 per group), any prompt exceeding ~1145 tokens could not be scheduled. The scheduler silently refused the request with no error.

### 2. Slot overflow (async scheduling race)

In the async paged path, `execute_model` allocates new request slots before `sample_tokens` frees finished request slots. With `max_num_seqs=1`, a finished request's slot was unavailable for the next request in the same scheduling step, causing an out-of-bounds slot access and crash.

### `vllm_metal/paged_attention_backend/hybrid.py`
- Pass SDPA `block_size` through to `_build_linear_layer_spec` instead of hardcoding 1

### `vllm_metal/v1/model_runner.py`
- **Early slot release**: Pop finished request slot mappings in `execute_model` before `_start_paged_forward`, so slots can be reused within the same scheduling step
- **Alloc-time zeroing**: Zero conv/recurrent state in `_gdn_alloc_slot` using `mx.zeros_like` when reusing a slot from the free list, avoiding `mx.eval` synchronization issues that caused GPU hangs on large models
- **Simplify `_gdn_free_slot`**: Remove redundant zeroing, retain `mx.eval` to materialize state arrays and prevent lazy graph accumulation

### `tools/test_qwen3_next_golden.py`
- Regenerate golden token IDs via `mlx_lm generate_step` + argmax sampler — previous values had incorrect leading-space tokenization (e.g. token 59604 `'Paris'` instead of 12095 `' Paris'`)
- Replace `"The weather today is not"` with `"The speed of light is approximately"` — the original prompt produces a bitwise-equal logit tie (IEEE 754 `bff00000`, -1.875) at position 7, making argmax non-deterministic
- Set `VLLM_METAL_MEMORY_FRACTION=0.55` default for 80B model test

### `tests/test_qwen3_next_gdn.py`
- Add 5 unit tests: alloc-time zeroing, slot isolation, fresh allocation, double-free safety, early release reuse

## Test

**Environment**: M2 Ultra 192GB, mlx-community/Qwen3-Next-80B-A3B-Instruct-8bit

| Test | Result |
|------|--------|
| `pytest -m "not slow" tests/` | 416 passed |
| Qwen3-0.6B golden (6/6) | passed |
| Qwen3-Next 80B golden (6/6) | passed |
| Qwen3-Next 80B serve 1K×15 batch | 93.3 TPS |
| Qwen3-Next 80B serve 4K×15 batch | 31.2 TPS |
| `ruff check` + `ruff format` | clean |

## Test plan

- [x] `pytest -m "not slow" tests/` — 416 passed
- [x] Qwen3-0.6B paged deterministic golden — 6/6 exact match
- [x] Qwen3-Next 80B golden token test — 6/6 exact match
- [x] Lint + format clean

### DIFF
--- a/tests/test_qwen3_next_gdn.py
+++ b/tests/test_qwen3_next_gdn.py
@@ -143,6 +143,201 @@ class TestGDNFreeSlotZeroing:
             assert sc.recurrent_states[layer_idx].dtype == mx.float32
 
 
+class TestGDNAllocSlotZeroing:
+    """Verify that _gdn_alloc_slot zeros state for reused slots."""
+
+    def _make_runner_stub(self, max_seqs: int = 2):
+        """Build a minimal stub with GDN slot management wired up."""
+        from vllm_metal.mlx_backend.gdn_cache import GDNPagedStateCache
+
+        sc = GDNPagedStateCache(
+            num_layers=2,
+            max_seqs=max_seqs,
+            conv_kernel_dim=4,
+            conv_dim=64,
+            num_v_heads=4,
+            value_head_dim=16,
+            key_head_dim=16,
+            dtype=mx.float16,
+        )
+        backend = MagicMock()
+        backend._state_cache = sc
+
+        runner = MagicMock()
+        runner._gdn_req_to_slot = {}
+        runner._gdn_free_slots = []
+        runner._paged_attention_backend = backend
+        return runner, sc
+
+    def test_reused_slot_is_zeroed(self):
+        """A slot returned to the free list and re-allocated must have
+        zeroed conv and recurrent state."""
+        from vllm_metal.v1.model_runner import MetalModelRunner
+
+        runner, sc = self._make_runner_stub()
+
+        # Allocate slot 0 for req-A
+        slot = MetalModelRunner._gdn_alloc_slot(runner, "req-A")
+        assert slot == 0
+
+        # Write non-zero data to slot 0
+        for layer_idx in range(sc.num_layers):
+            sc.conv_states[layer_idx] = mx.ones_like(sc.conv_states[layer_idx])
+            sc.recurrent_states[layer_idx] = mx.ones_like(
+                sc.recurrent_states[layer_idx]
+            )
+        mx.eval(*sc.conv_states, *sc.recurrent_states)
+
+        # Free slot 0 (bare pop+append, no zeroing — matches early release path)
+        runner._gdn_req_to_slot.pop("req-A")
+        runner._gdn_free_slots.append(slot)
+
+        # Re-allocate — should trigger alloc-time zeroing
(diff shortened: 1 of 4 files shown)


ericcurtin:
minor test coverage gaps (no mock test for execute_model early-release path, test_slot_reuse doesn't verify zeroing values)

WindChimeRan (approved):
LGTM. critical bug fixed.

ericcurtin:
Early-free skips mx.eval (implicit dependency), block_size default=1 could be required, no integration test for early-free path

gwlee-dev:
Addressed in `005df37`:

**Test coverage:**
- `test_slot_reuse_after_early_release` now verifies zeroed conv/recurrent state values after reuse (not just slot number)
- Added `test_early_release_then_alloc_full_cycle`: 2-slot scenario — finish req-A, start req-C while req-B active. Verifies req-C gets zeroed state AND req-B's state is untouched

**Early-free skips mx.eval:**
Intentional. The original `_gdn_free_slot` zeroes state via `conv[slot] = 0` + `mx.eval()`, which caused GPU hangs on 80B models (the `mx.eval` forces a full GPU sync mid-forward). We moved zeroing to alloc time using `mx.zeros_like` (lazy, no sync barrier). The `_gdn_free_slot` retains `mx.eval` only for materializing state arrays to prevent lazy graph accumulation — but the actual zeroing happens at `_gdn_alloc_slot` when a slot is reused.

**block_size default=1:**
The `default=1` preserves the original hardcoded value for backward compatibility. The only call site (`model_runner.py:1140`) always passes `block_size=self.cache_config.block_size` explicitly, so the default is never reached in practice.

**Integration test coverage:**
The current CI smoke tests (`scripts/test.sh`) send a single request per model, so the early-release slot reuse path is not exercised in CI. It is covered by the unit tests above (`test_early_release_then_alloc_full_cycle` simulates the full execute_model early-release → alloc → zeroing cycle). Could add a multi-request smoke test in a follow-up PR if that would be useful.

WindChimeRan:
`block_size default=1 could be required` fixed

PR #265, opened 2026-04-13
Fix GDN slot release and enforce hybrid backend state cache

LxYuan0420:
This PR is:
- To unify GDN slot freeing so early release uses the same cleanup path.
- To materialize GDN state once per scheduling step to avoid repeated syncs.
- To enforce hybrid backend expectations and keep GDN tests aligned with runtime types.

Problem:
Early release bypassed `_gdn_free_slot`, skipping state materialization and creating two free paths. This could leave lazy MLX graphs attached to freed slots and made future cleanup changes inconsistent.

Fix:
- Centralized state materialization in `_gdn_materialize_state_cache`.
- Early release now materializes once per step and calls `_gdn_free_slot(..., materialize_state=False)`.
- Fail fast if the hybrid backend or state cache is missing.
- Tests now use a lightweight hybrid backend stub to match runtime expectations.

Tests:
- `python -m pytest tests/test_qwen3_next_gdn.py -v`

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -33,6 +33,7 @@ def make_stub_runner(
         "_paged_attention_backend": None,
         "_gdn_req_to_slot": {},
         "_gdn_free_slots": [],
+        "_gdn_needs_materialize": False,
         "_request_states": {},
         "_paged_request_seq_lens": {},
         "_finished_request_count": 0,
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -31,6 +31,8 @@ def make_stub_runner(
         "_is_stt": False,
         "_is_vlm": False,
         "_paged_attention_backend": None,
+        "_gdn_req_to_slot": {},
+        "_gdn_free_slots": [],
         "_request_states": {},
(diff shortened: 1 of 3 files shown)


WindChimeRan (commented):
The latter half is a bit overcomplicated: 

Both places that free slots iterate over `finished_req_ids`. That's a batch, not a per-slot operation. A single `_gdn_release_slots(req_ids)` that pops each, materializes once if anything was freed, then extends `_gdn_free_slots` would let you delete:

- `_gdn_free_slot` and the `materialize_state` flag
- The `materialized` latch in `execute_model`
- The `if req_id not in ...: continue` + internal `pop` (one `pop` covers both)

Each call site becomes one line, and the "second call is a no-op" property is structural (dict is already empty) rather than relying on early-return plumbing.

ericcurtin:
Critical regression in _cleanup_finished_requests (unprotected _gdn_free_slot crashes for non-hybrid paged models), missing test for materialize_state=False branch, stub_runner missing GDN defaults, tests unverifiable due to environment issue

WindChimeRan (approved):

PR #267, opened 2026-04-13
Refactor model logic into adapter boundary

LxYuan0420:
This PR is:
- To isolate model logic behind a single adapter boundary.
- To keep `MetalModelRunner` orchestration-only.
- To remove ad-hoc helper modules.

Note:
- Next refactor will extract model lifecycle responsibilities from the runner.
- Keep the adapter boundary as the only place for model rules.
- Avoid new helper modules unless they are true ownership boundaries.


WindChimeRan (commented):
Discussion: `_model_adapter`: Private attribute leak across module boundary. Can we use a cleaner way? 

Other parts LGTM

ericcurtin:
I'm pretty happy with this one:

Minor test organization nit (text_model tests in separate file from other adapter tests)

ericcurtin (approved):

PR #251, opened 2026-04-09
[Gemma4] Add text generation support with paged attention

ricky-chaoju:
## Summary
- Add Gemma4 text generation with paged KV cache on Metal
- **hs512 Metal kernel**: instantiate paged attention for head_dim=512 (v1 + v2)
- **Variable head_dim**: pad Q/K/V to cache's max head_dim for models with mixed sliding (256) / full (512) attention layers
- **YOCO KV sharing**: SDPA forward accepts `shared_kv` so later layers reuse K/V from earlier same-type layers
- **v_norm**: support per-head RMSNorm on values (Gemma4 uses q_norm + k_norm + v_norm)
- **global_head_dim**: use `max(head_dim, global_head_dim)` for KV cache block sizing
- **mlx_lm loading workaround**: Gemma4 loaded via mlx_lm (text backbone) because mlx_vlm produces garbled output; remove when mlx_vlm fixes this
- **K-eq-V**: handle Gemma4 variants without v_proj (26B/31B use values = keys)                                           
- **Variable KV head guard**: reject `num_global_key_value_heads != num_key_value_heads` with clear error (26B/31B need per-layer KV allocation)
## What's not included
- Multimodal support (vision/audio) — blocked by mlx_vlm Gemma4 forward pass bug, will track separately
- Sliding window enforcement in kernel (currently disabled)
- YOCO memory optimization (shared layers currently duplicate cache)
- Per-layer KV head allocation (needed for 26B/31B where global_kv_heads != kv_heads) 

## Test
- Gemma4-E2B-it verified with `vllm serve` (paged attention, thinking mode, max-model-len 4096)
- Qwen3-0.6B / Qwen3.5-0.8B unaffected — unit tests 411/411 passed
- hs512 kernel instantiation does not affect existing hs≤256 models
```
VLLM_METAL_MEMORY_FRACTION=0.5 \
  vllm serve /Users/rickychen/llm/models/gemma-4-E2B-it \
  --port 8000 --max-model-len 4096 --max-num-seqs 4 \
  --default-chat-template-kwargs '{"enable_thinking": true}' \
  --reasoning-parser gemma4   
```

PR #269, opened 2026-04-14
Reduce KV cache allocation for YOCO shared layers

### LINKED ISSUE #264: [Gemma4] kv cache only reuse by computation, not by memory
#251 achieved Gemma4 kv cache reuse, only by computation, but not by memory (still duplicate memory write)

ricky-chaoju:
Fixes #264

Gemma4 YOCO shared layers (last 20 of 35) now reuse the reference layer's paged cache instead of allocating and writing their own.

- `model_adapter.build_yoco_cache_mapping()` computes layer→cache_idx mapping
- `MHAPagedAttentionBackend` allocates only unique cache layers (15 vs 35)
- `sdpa_forward` skips scatter write when `shared_kv` is provided
- Gemma4-E2B-it deterministic test: 5/5 token-exact match with mlx_lm

Related #251

### DIFF
--- a/tests/test_model_adapter.py
+++ b/tests/test_model_adapter.py
@@ -98,10 +98,12 @@ class TestYocoCacheIntegration:
     _NUM_UNIQUE = _NUM_HIDDEN - _NUM_SHARED
     _NUM_KV_HEADS = 2
     _HEAD_DIM = 4
+    _BLOCK_SIZE = 16
+    _VOCAB_SIZE = 100
 
     def _gemma4_args(self) -> dict:
         return {
-            "vocab_size": 100,
+            "vocab_size": self._VOCAB_SIZE,
             "num_hidden_layers": self._NUM_HIDDEN,
             "num_kv_shared_layers": self._NUM_SHARED,
             "layer_types": list(self._LAYER_TYPES),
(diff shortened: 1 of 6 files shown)


WindChimeRan (commented):
This PR LGTM. But I want to ask for @LxYuan0420 's approval, since it touches `model_adapter.py`

tests/test_model_adapter.py
```diff
@@ -75,6 +75,66 @@ def test_returns_none_when_head_dim_none(self) -> None:
         assert result is None
 
 
+class TestBuildYocoCacheMapping:
```
LxYuan0420: These tests feel low-value to me. I want to see:
- reduced `num_kv_cache_layers` / cache-byte accounting for YOCO models
- `_make_backend()` wiring the compact layer count and `cache_idx_map`
- shared layers actually reusing the remapped cache slot instead of allocating/writing their own KV cache

LxYuan0420 (approved):

tests/test_model_adapter.py
ricky-chaoju: Replaced with integration tests covering cache-byte accounting, _make_backend wiring, and shared-layer cache slot reuse.

Issue #274, opened 2026-04-15
guidellm crashes vllm-metal due to out of memory


jrr-ghcp:
Create gemma4 chat template:
```
cat > ~/Documents/ChatTemplates/gemma_chat_template.jinja << 'EOF'                                                                                                                        
{{ bos_token }}{% if messages[0]['role'] == 'system' %}{% set system_message = messages[0]['content'] %}{% set messages = messages[1:] %}{% else %}{% set system_message = '' %}{% endif %}{% for message in messages %}{% if loop.first and system_message %}{% if message['role'] == 'user' %}{{ '<start_of_turn>user\n' + system_message + '\n\n' + message['content'] | trim + '\n<end_of_turn>\n' }}{% endif %}{% elif message['role'] == 'user' %}{{ '<start_of_turn>user\n' + message['content'] | trim + '\n<end_of_turn>\n' }}{% elif message['role'] == 'assistant' %}{{ '<start_of_turn>model\n' + message['content'] | trim + '\n<end_of_turn>\n' }}{% endif %}{% endfor %}{% if add_generation_prompt %}{{'<start_of_turn>model\n'}}{% endif %}
EOF
```

Run vllm-metal:

```
vllm serve ~/.lmstudio/models/mlx-community/gemma-4-31b-8bit --hf-overrides '{"architectures": ["Gemma4ForCausalLM"]}' --chat-template ~/Documents/ChatTemplates/gemma_chat_template.jinja
```
---

Install [guidellm](https://github.com/vllm-project/guidellm):

```
python3 -m venv guidellm-venv
source ./guidellm-venv/bin/activate
pip install guidellm
Collecting guidellm
  Downloading guidellm-0.6.0-py3-none-any.whl.metadata (24 kB)
<cut>
```

Run guidellm:
```
guidellm benchmark \
  --target "http://localhost:8000" \
  --profile sweep \
  --max-seconds 30 \
  --data "prompt_tokens=512,output_tokens=128"
```

---
Observe crash due to `Command buffer execution failed: Insufficient Memory`:

```
(APIServer pid=26468) INFO:     127.0.0.1:51494 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(APIServer pid=26468) INFO:     127.0.0.1:51496 - "POST /v1/chat/completions HTTP/1.1" 200 OK
[...]

PR #275, opened 2026-04-15
Flatten Gemma4 text_config on mlx-lm load

ricky-chaoju:
Summary
- To restore the nested `text_config` flatten on the mlx-lm `.args` load path.
- To unblock Gemma4 variants whose mlx-lm `ModelArgs` expose only `{model_type, text_config, vocab_size}` at the top level.

Problem:
After #271 extracted `_extract_model_args` into `ModelLifecycle`, the `.args` branch early-returned without running the nested
`text_config` setdefault merge that both paths previously shared. Gemma4 variants then die at startup in `resolve_model_dims()` with
`Cannot resolve model dimensions: num_layers, num_kv_heads, head_dim. Available keys: ['model_type', 'text_config', 'vocab_size']`.

Fix:
- Both `.args` and `.config` branches now fall through to the existing bottom `setdefault` merge, flattening nested `text_config` keys
onto the top level.
- Load-path behavior is identical to pre-#271 on both paths.
- This is a regression fix only; it does not yet enable Gemma4 26B/31B on paged attention (per-layer KV cache shapes will land in
follow-ups). After this fix, 31B loads successfully and then hits the existing `require_uniform_kv_heads` guard, which is the expected
boundary.

Related: #274

### DIFF
--- a/tests/test_model_lifecycle.py
+++ b/tests/test_model_lifecycle.py
@@ -161,6 +161,42 @@ class TestModelLifecycle:
         assert runner.num_layers == 32
         assert runner.head_dim == 128
 
+    def test_load_merges_nested_text_config_from_mlx_lm_args(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        """mlx-lm Gemma4 exposes .args with dims nested inside text_config.
+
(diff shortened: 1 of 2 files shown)


LxYuan0420 (approved):
Thanks for the fix!

Issue #276, opened 2026-04-15
[Gemma4] Full 26B/31B paged attention support


ricky-chaoju:
**Goal:** Run Gemma4 26B/31B on the paged-attention path with per-layer KV shapes, matching upstream vLLM's heterogeneous spec pattern.

**Problem:** Gemma4 26B/31B mix sliding attention layers (`num_kv_heads=16, head_dim=256`) with full-attention layers (`num_kv_heads=4, head_dim=512`). Our `MetalPagedKVCache` only supports a single uniform `(num_kv_heads, head_dim)` across all layers and rejects this layout via `require_uniform_kv_heads`. Forcing the uniform max wastes ~2.18× KV memory and OOMs under sustained serving (#274).

### Plan

- [x] Unblock loading — restore nested `text_config` flatten on the mlx-lm `.args` load path #275
- [x] Per-layer KV cache shapes — `MetalPagedKVCache` + `ModelCachePolicy` + `sdpa_forward` accept per-layer `(num_kv_heads, head_dim)`; uniform models reduce to the current path unchanged #277 
- [x] Enable 26B/31B — `DefaultModelAdapter.build_per_layer_kv_shapes()`, drop `require_uniform_kv_heads`, add a 31B golden test against `mlx_lm` #279 

### Upstream reference

- `vllm/model_executor/models/gemma4.py:446-484` — per-layer `Gemma4Attention` picks `(num_kv_heads, head_dim)` from `layer_types`
- `vllm/v1/core/kv_cache_utils.py:1107-1124` — `UniformTypeKVCacheSpecs` path allocates one tensor per layer when specs share type but differ in shape
- `vllm/v1/worker/gpu_model_runner.py:6503-6574` — `_reshape_kv_cache_tensors()` views each layer's raw buffer under its own `(num_kv_heads, head_dim)`

dschulmeist:
Picked this up on top of #277. Two stacked PRs:

- #278 — clears `multimodal_config` in `MetalPlatform.check_and_update_config` for `_TEXT_BACKBONE_OVERRIDE_TYPES` so the 31B MLX checkpoint (which ships without `preprocessor_config.json` / feature-extractor) can actually initialize. Independent bugfix, targets `main` directly.
- #279 — the rest of the #276 plan: `DefaultModelAdapter.build_per_layer_kv_shapes()`, `require_uniform_kv_heads` → no-op, dual-golden deterministic test across E2B + 26B + 31B.

[...]

PR #277, opened 2026-04-16
Per-layer KV cache shapes for heterogeneous attention

ricky-chaoju:
Summary
- To let `MetalPagedKVCache` allocate each layer with its own `(num_kv_heads, head_dim)` shape.
- To prepare the paged-attention pipeline for Gemma4 26B/31B, which mix sliding layers (16 heads, 256 dim) with full-attention layers (4 heads, 512 dim).
- To keep every existing model on the identical uniform code path with identical memory numbers.

Problem:
`MetalPagedKVCache` allocated all layers with a single `(num_kv_heads, head_dim)` pair. Gemma4 26B/31B need per-layer shapes, and
forcing the uniform max wastes ~2.18× KV memory. This PR adds the mechanism; a follow-up PR wires the Gemma4 adapter and removes the
`require_uniform_kv_heads` guard.

Fix:
- `MetalPagedKVCache` accepts optional keyword-only `kv_heads_per_layer` / `head_dim_per_layer` lists. When omitted, derives uniform
lists from the scalar params — MLA, Hybrid, and all existing MHA callers are unchanged.
- `MHAPagedAttentionBackend` propagates per-layer lists to the cache.
- `sdpa_forward` reads `kv_cache.kv_heads_per_layer[layer_idx]` / `kv_cache.head_dim_per_layer[layer_idx]` instead of scalar globals.
- `ModelCachePolicy.get_cache_block_size_bytes()` / `estimate_one_sequence_kv_bytes()` sum per-layer `kv_heads × head_dim` via a shared
`_kv_layer_size_sum()` helper. For uniform models this reduces to the existing product formula.
- `ModelCachePolicy.get_kv_cache_spec()` emits per-layer `FullAttentionSpec` from runner attributes when available.
- `MetalModelRunner` gains `kv_heads_per_layer` / `head_dim_per_layer` (default `None`, populated by the model adapter in the follow-up
PR).

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -41,6 +41,8 @@ def make_stub_runner(
         "_pending_output": None,
         "_execute_model_state": None,
         "_model_adapter": DefaultModelAdapter(),
+        "kv_heads_per_layer": None,
+        "head_dim_per_layer": None,
         "use_async_scheduling": True,
         "device": torch.device("cpu"),
         "_sampler": None,
--- a/tests/test_per_layer_kv_cache.py
+++ b/tests/test_per_layer_kv_cache.py
@@ -108,7 +108,9 @@ class TestMHABackendPerLayer:
         assert cache.key_caches[0].shape[-2:] == (16, 256)
         assert cache.key_caches[1].shape[-2:] == (4, 512)
 
-    def test_warm_up_uses_layer_zero_shape(self, monkeypatch: pytest.MonkeyPatch) -> None:
+    def test_warm_up_uses_layer_zero_shape(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
         """Kernel warm-up should compile against the concrete layer-0 shape."""
         cache = MetalPagedKVCache(
             num_layers=2,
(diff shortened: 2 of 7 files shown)


dschulmeist:
Heads-up @ricky-chaoju — I've built on top of this PR for the #276 follow-up in #279 (which populates `kv_heads_per_layer` / `head_dim_per_layer` from Gemma4's `layer_types` + `global_head_dim`, drops `require_uniform_kv_heads`, and adds a dual-golden test for E2B + 31B). The `kv_heads_per_layer[:num_cache_layers]` slicing you landed on turned out to line up cleanly with Gemma4's YOCO layout — first-N-unique-then-shared — so no further plumbing changes were needed.

Thanks for the groundwork here; I'll rebase #279 after this lands.

LxYuan0420 (approved):
Approving. The per-layer cache-shape plumbing looks sound and the backward-compatible path stays clean. One non-blocking follow-up: the layer-0 overwrite of `self.num_kv_heads` / `self.head_dim` in `MetalPagedKVCache` is a bit fragile, and hybrid + per-layer would be better rejected at a single early boundary rather than only in `_kv_layer_size_sum()`. Not blocking this PR, but I think it would make the contract clearer

ricky-chaoju:
Thanks @LxYuan0420 !

PR #278, opened 2026-04-16
Clear multimodal_config for text-backbone-override models

dschulmeist:
## Summary

Gemma4 MLX checkpoints are registered as multimodal in the HF config but are served via the text-only `mlx_lm` path by vllm-metal. vLLM's `InputProcessor` eagerly instantiates the multimodal feature extractor during engine startup and crashes on MLX checkpoints that ship neither `preprocessor_config.json` nor a `feature_extractor` section in `processor_config.json` — which is the case for the [Gemma4 31B MLX release](https://huggingface.co/mlx-community/gemma-4-31b-8bit).

Repro on current `main`:

```
GEMMA4_31B_MODEL_PATH=/path/to/gemma-4-31b-8bit \
  VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_MEMORY_FRACTION=0.62 \
  python -c "from vllm import LLM; LLM(model='${GEMMA4_31B_MODEL_PATH}', max_model_len=128, max_num_seqs=1)"
# OSError: Can't load feature extractor for '…/gemma-4-31b-8bit'. … directory containing a preprocessor_config.json file
```

## Changes

- `vllm_metal/platform.py:check_and_update_config`: clear `model_config.multimodal_config` for any model type in `_TEXT_BACKBONE_OVERRIDE_TYPES` so `is_multimodal_model` returns `False` before the input processor runs. Uses the same `should_force_text_backbone` predicate as `ModelLifecycle.load`, keeping the two decisions in sync.
- `tests/test_platform.py`: two new cases — `gemma4` clears the config, `qwen3_vl` preserves it.

## Test plan

- [x] `ruff check` / `ruff format --check` / `mypy vllm_metal` clean
- [x] `pytest -m "not slow" tests/` — 462 passed
- [x] Manual: `LLM(model='mlx-community/gemma-4-31b-8bit', …)` now starts cleanly (end-to-end validation covered by the companion #276 follow-up)

## Notes

This fix is a prerequisite for running the Gemma4 31B MLX checkpoint end-to-end on the paged attention path (issue #276), but it is an independent bugfix against any Gemma4 MLX checkpoint whose `processor_config.json` lacks a `feature_extractor` section.

### DIFF
--- a/tests/test_model_adapter.py
+++ b/tests/test_model_adapter.py
@@ -30,6 +30,41 @@ class TestShouldForceTextBackbone:
         assert result is False
 
 
+class TestNormalizeModelConfig:
+    """Tests for normalize_model_config()."""
+
+    def test_clears_multimodal_config_for_gemma4(self) -> None:
+        model_config = SimpleNamespace(
+            multimodal_config=SimpleNamespace(language_model_only=False),
+            hf_config=SimpleNamespace(model_type="gemma4"),
+        )
+
+        DefaultModelAdapter().normalize_model_config(model_config)
+
+        assert model_config.multimodal_config is None
+
+    def test_preserves_multimodal_config_for_other_models(self) -> None:
+        sentinel = SimpleNamespace(language_model_only=False)
+        model_config = SimpleNamespace(
+            multimodal_config=sentinel,
+            hf_config=SimpleNamespace(model_type="qwen3_vl"),
+        )
+
+        DefaultModelAdapter().normalize_model_config(model_config)
+
+        assert model_config.multimodal_config is sentinel
+
+    def test_noop_when_multimodal_config_already_none(self) -> None:
+        model_config = SimpleNamespace(
+            multimodal_config=None,
+            hf_config=SimpleNamespace(model_type="gemma4"),
+        )
+
+        DefaultModelAdapter().normalize_model_config(model_config)
+
+        assert model_config.multimodal_config is None
+
+
 class TestTextModel:
     def test_returns_language_model_when_present(self) -> None:
         language_model = object()
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -192,6 +192,8 @@ class TestMetalPlatform:
                     disable_cascade_attn=False,
                     tokenizer=None,
                     max_model_len=32768,
+                    multimodal_config=None,
+                    hf_config=SimpleNamespace(model_type="qwen3"),
                 ),
                 scheduler_config=SimpleNamespace(
                     async_scheduling=True,
@@ -240,6 +242,8 @@ class TestMetalPlatform:
                     disable_cascade_attn=False,
                     tokenizer=None,
                     max_model_len=32768,
+                    multimodal_config=None,
+                    hf_config=SimpleNamespace(model_type="qwen3"),
                 ),
                 scheduler_config=SimpleNamespace(
                     async_scheduling=True,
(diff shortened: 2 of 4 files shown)


LxYuan0420 (commented):
Thanks. It seem to be the right fix, but I don’t think the model-specific rule should live inline in `platform.py`.

The policy already belongs to `DefaultModelAdapter`, so I’d prefer the platform layer to call an adapter method early, and keep the model-specific normalization inside the class in `vllm_metal/v1/model_adapter.py`. I’d also drop the extra `getattr` guards here and use the typed `ModelConfig` contract directly unless there is a concrete version-skew case.

something like:

```python
# vllm_metal/platform.py
from vllm_metal.v1.model_adapter import DefaultModelAdapter

# ...

if model_config is not None:
    model_config.disable_cascade_attn = True
    DefaultModelAdapter().normalize_model_config(model_config)
```

```python
# vllm_metal/v1/model_adapter.py
class DefaultModelAdapter(ModelAdapter):

    def normalize_model_config(self, model_config: "ModelConfig") -> None:
        if model_config.multimodal_config is None:
            return
        if not self.should_force_text_backbone(model_config.hf_config):
            return

        model_config.multimodal_config = None
        logger.info(
            "Metal: forcing text-only backbone for model_type=%s "
            "(cleared multimodal_config)",
            model_config.hf_config.model_type,
        )
```

This keeps the early timing in platform, but keeps ownership of the model-specific invariant inside the adapter, which should scale better if we add more text-backbone override models later.

LxYuan0420:
Could you also share at least one actual generation output (after the fix), so we can verify the text-backbone path is producing sensible output rather than gibberish?

dschulmeist:
Thanks for the review @LxYuan0420, both points adressed in the latest push.

**Refactor**: moved the multimodal-config clearing into `DefaultModelAdapter.normalize_model_config()` like you sketched. The platform layer now just calls it right after setting `disable_cascade_attn`, and the model-specific policy lives next to `should_force_text_backbone` which it reuses. Also dropped the defensive `getattr` guards on `multimodal_config` / `hf_config` since `ModelConfig` is a typed contract anyway. Added 3 unit cases on the new method in `tests/test_model_adapter.py`, the exiting platform integration cases still verify the end-to-end flow.

**Sample 31B output** (after the fix, paged path, `mlx-community/gemma-4-31b-8bit`):

```
PROMPT: 'The capital of France is'
OUTPUT: ' a city of romance, art, and culture. It is also a city of great
         food and wine. Paris is home to some of the best restaurants in
         the world, and it is also a great'

PROMPT: 'Q: What is photosynthesis?\nA:'
OUTPUT: ' The process by which plants use sunlight, water, and carbon
         dioxide to create oxygen and energy in the form of sugar.\n\n
         Q: What is the chemical equation for photosynthesis?\nA: 6'

PROMPT: "Translate to French: 'Good morning, how are you?'"
OUTPUT: '\n\nTranslate to French: \'I am fine, thank you.\'\n\nTranslate
         to French: \'What is your name?\'\n\nTranslate to French: \'My
         name is...\''
```

Output is coherent factual continuation on prompts 1 and 2. Prompt 3 is base-model few-shot continuation (this is the base, not the instruct checkpoint).

Companion #279 also force-pushed, same refactor on its first commit, Gemma4 work intact on top.

LxYuan0420 (approved):
LGTM! Thanks for the fix

PR #279, opened 2026-04-16
[Gemma4] Enable 26B/31B heterogeneous paged attention

### LINKED ISSUE #276: [Gemma4] Full 26B/31B paged attention support

dschulmeist:
Closes #276.

## Summary

Completes the plan for #276 on top of merged #277 and #278. Populates per-layer KV cache shapes for Gemma4 26B/31B so the paged path stops rejecting variable KV-head counts and stops wasting cache memory on the uniform-max shape.

Issue #276's explicit plan:

> - Per-layer KV cache shapes — `MetalPagedKVCache` + `ModelCachePolicy` + `sdpa_forward` accept per-layer `(num_kv_heads, head_dim)` — **done in #277**
> - Enable 26B/31B — `DefaultModelAdapter.build_per_layer_kv_shapes()`, drop `require_uniform_kv_heads`, add a 31B golden test against `mlx_lm` — **this PR**

### Adapter (`vllm_metal/v1/model_adapter.py`)

- `build_per_layer_kv_shapes()` returns `(kv_heads_per_layer, head_dim_per_layer)` derived from `layer_types` + `global_head_dim` + `num_global_key_value_heads`. Returns `None` for uniform models so existing backends keep the scalar path.
- **Edge case**: E2B-style checkpoints set only `global_head_dim` and omit `num_global_key_value_heads`. Falling back to uniform would under-size full-attention layers (sliding cache slot is 256, full layer writes 512 → kernel head_dim overflow). We fall back to the sliding-layer KV-head count for full layers instead.
- `require_uniform_kv_heads()` is now a documented no-op. The paged path supports heterogeneous KV end-to-end, so the historical raise-on-mismatch would block the very models this PR enables.

### Lifecycle (`vllm_metal/v1/model_lifecycle.py`)

- `resolve_model_dims()` populates `runner.kv_heads_per_layer` and `runner.head_dim_per_layer` via the adapter, using the pre-`resolve_max_head_dim` sliding `head_dim` so sliding layers keep their true 256 while the cache still allocates at the max (512).
- Guards `int(base_head_dim)` against `None` — the missing-check above already raises, but belt-and-suspenders so `int()` never receives `None` in any future refactor.

### SDPA dispatch (`vllm_metal/metal_kernel_backend/attention_sdpa.py`)

- `is_sdpa()` no longer requires `v_proj`. Gemma4 K-eq-V layers (26B/31B full-attention drop `v_proj` and reuse `keys` for values) now route through SDPA via the hybrid dispatcher. Not exercised today (the MHA dispatcher wraps every attention module unconditionally), but the hybrid dispatcher classifies via `is_sdpa` — fixes a latent misrouting bug. `prepare_sdpa_qkv` already handles the `values = keys` branch internally.

### Tests

- `tests/test_model_adapter.py` — `TestRequireUniformKvHeads` rewritten to lock in the no-op contract; new `TestBuildPerLayerKVShapes` with 5 cases.
- `tests/test_model_lifecycle.py` — three new `TestResolveModelDims` cases (uniform / 31B layout / E2B layout).
- `tests/test_attention_dispatch.py` — new `test_gemma4_k_eq_v_attention_detected_as_sdpa` against a real `mlx_lm.gemma4_text.Attention` with `attention_k_eq_v=True`.
- `tests/test_gemma4_golden.py` — dual-golden (mlx_lm + paged) deterministic test, variants for E2B + 26B-A4B + 31B. Module-level invariant rejects silent golden drift (common-prefix ≥ 6 / 10 tokens). One variant runs per pytest invocation (MLX keeps model buffers at process scope; two Gemma4 checkpoints exceed typical 64 GB unified memory).
- `tools/gen_gemma4_golden.py` — adds `--paged` flag for regenerating the in-tree paged golden alongside the mlx_lm reference.

## Test plan

- [x] `ruff check` / `ruff format --check` clean
- [x] `mypy vllm_metal` — no issues in 53 source files
- [x] `pytest -m "not slow" tests/` — 482 passed, 1 skipped (golden test skips correctly without env vars)
- [x] `GEMMA4_MODEL_PATH=/path/to/gemma-4-e2b-it-MLX-8bit pytest tests/test_gemma4_golden.py -m slow` — 5/5 passed
- [x] `GEMMA4_31B_MODEL_PATH=/path/to/gemma-4-31b-8bit pytest tests/test_gemma4_golden.py -m slow` — 5/5 passed (M5 Max, 64 GB unified memory)

31B goldens: 4/5 prompts match `mlx_lm` bit-for-bit; the 5th (`"The weather today is not"`) diverges at token 7 where the top-2 logits are ~0.6 apart — a classic fp accumulation tie-break, not a correctness bug. The dual-golden pattern mirrors the precedent in `tests/test_paged_deterministic.py`.

## Memory

The 26B and 31B variants require a Mac with ≥ 64 GB of unified memory. E2B runs on any Apple Silicon Mac. Documented in the test file's docstring.

## Out of scope

**26B goldens deferred.** 26B is MoE (128 experts, top-8). The paged Metal kernel differs from `mx.fast.scaled_dot_product_attention` (used by `mlx_lm`) by ~1e-6 cos per layer on identical Q/K/V — below ulp for continuous ops, but enough for the router's discrete top-k selection to flip an expert near a score boundary and cascade through subsequent layers.

Measured on `mlx-community/gemma-4-26b-a4b-it-8bit` (M5 Max, 64 GB):

- 3/5 test prompts match `mlx_lm` bit-exact
- 2/5 diverge at token 1-2 (`"The weather today is not"`, `"Water boils at a temperature of"`)
- Expert-set agreement for the weather prompt: 8/8 at layers 0-7, first flip at layer 8, 2/8 at layer 17
- Residual cos drops below 0.99 at layer 13; reaches 0.38 at layer 17
- Swapping the paged kernel for `mx.fast.SDPA` at prefill restores 5/5 first-token parity

Per-layer kernel is verifiably correct (paged vs `mx.fast.SDPA` cos = 0.99999898 on identical Q/K/V at layer 0, cache scatter bit-exact). 31B is unaffected (dense, no discrete routing to amplify drift).

26B golden dicts are intentionally empty so the test auto-skips instead of violating `_MIN_COMMON_PREFIX = 6`. Follow-up in #281 proposes reading K/V out of the paged cache per-request and dispatching `mx.fast.SDPA` on contiguous tensors — keeps paged-cache memory management and prefix caching, gives bit-exact parity with `mlx_lm` for MoE models.

**Non-paged MLX path:** unchanged.

### DIFF
--- a/tests/test_gemma4_golden.py
+++ b/tests/test_gemma4_golden.py
@@ -1,29 +1,67 @@
 # SPDX-License-Identifier: Apache-2.0
 """Deterministic golden-token test for Gemma4 on the paged attention path.
 
-Verifies that vllm-metal's paged attention implementation produces the same
-greedy-decoded token IDs as running the same Gemma4 checkpoint through
-mlx_lm directly.  This catches regressions in YOCO sharing, K-eq-V fallback,
-v_norm, and variable head_dim padding — the paths exercised by Gemma4 that
-are not stressed by other models.
+Verifies that vllm-metal's paged attention implementation produces coherent
+greedy-decoded token IDs on Gemma4 checkpoints, exercising the paths that
+are unique to Gemma4: YOCO KV sharing, K-eq-V fallback, v_norm, variable
+head_dim padding, and heterogeneous per-layer KV cache shapes.
 
-Golden token IDs are generated once offline via ``tools/gen_gemma4_golden.py``
-using ``mlx_lm.stream_generate(..., sampler=temp=0)`` with the tokenizer's
-EOS set cleared so every sequence is exactly ``MAX_TOKENS`` long.
+Two golden sets are kept per model variant:
 
-Run:
-    GEMMA4_MODEL_PATH=/path/to/gemma-4-E2B-it \
+- ``GOLDEN_MLX_LM``: tokens produced by running the model directly through
+  ``mlx_lm.stream_generate`` (greedy, EOS disabled). The independent
+  reference, captured outside vllm-metal.
+- ``GOLDEN_PAGED``: tokens produced by running vllm-metal's paged attention
+  path on the same model/prompts. Captured here so small floating-point
+  tie-break drifts (top-2 logits within ~1 logit at one step) don't cause
+  spurious failures.
+
+This dual-golden pattern mirrors ``tests/test_paged_deterministic.py``.
+A module-level invariant enforces that the two sets agree on at least
+``MAX_TOKENS - 1`` of ``MAX_TOKENS`` token IDs for every prompt — i.e.
+the paged path cannot drift from mlx_lm by more than a single token. That
+limits how far the in-tree golden can diverge from the independent
+reference before the test starts accepting suspect outputs.
+
+Supported variants (each enabled by setting its env var to a local MLX
+checkpoint directory):
+
+- E2B:    ``GEMMA4_MODEL_PATH``          — 8-bit MLX E2B checkpoint.
+- 26B:    ``GEMMA4_26B_MODEL_PATH``      — 8-bit MLX 26B-A4B checkpoint.
+- 31B:    ``GEMMA4_31B_MODEL_PATH``      — 8-bit MLX 31B checkpoint
+                                           (target of issue #276).
+
+Run each variant in its own pytest invocation.  MLX holds model buffers
+at process scope, so loading two Gemma4 checkpoints in the same process
+typically exceeds 64 GB unified memory.  If several env vars are set in
+one invocation, the larger-is-better variant wins (31B > 26B > E2B).
+The 31B and 26B variants require a Mac with ≥ 64 GB unified memory.
+
+Regenerate goldens with:
+    # mlx_lm reference (no env vars required)
+    python tools/gen_gemma4_golden.py <model-path>
+    # paged-path reference (engine must run)
+    VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_USE_PAGED_ATTENTION=1 \\
+        python tools/gen_gemma4_golden.py --paged <model-path>
+
+Run tests:
+    # E2B only
+    GEMMA4_MODEL_PATH=/path/to/gemma-4-E2B-it \\
+        pytest tests/test_gemma4_golden.py -v -s -m slow
+    # 31B only (issue #276 regression test)
+    GEMMA4_31B_MODEL_PATH=/path/to/gemma-4-31b-8bit \\
         pytest tests/test_gemma4_golden.py -v -s -m slow
 """
 
 from __future__ import annotations
 
+import gc
 import os
+from dataclasses import dataclass
 
 import pytest
 from vllm import LLM, SamplingParams
 
-MODEL_ENV = "GEMMA4_MODEL_PATH"
 MAX_TOKENS = 10
 
 PROMPTS = [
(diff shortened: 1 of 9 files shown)


ricky-chaoju:
@dschulmeist Both #277 and #278 are merged, please rebase onto main.

dschulmeist:
Rebased onto main, #277 and #278 commits dropped. Single commit now.

ericcurtin:
int(base_head_dim) TypeError if base_head_dim is None for models deriving head_dim from hidden_size; 26B goldens are empty placeholders

dschulmeist:
Thanks for the review.

**1. `int(base_head_dim)`** — guarded. The missing-check above raises when `head_dim` and `hidden_size / num_attention_heads` are both absent, but `int(None)` being reachable is still smelly. Adapter call now runs only when `base_head_dim is not None`.

**2. 26B goldens** — the 26B is MoE (`enable_moe_block=True`, 128 experts, top-8 routed per token); 31B is dense. MoE top-k routing is a discrete decision, so the ~1e-6 per-layer numerical difference between the paged kernel and `mx.fast.scaled_dot_product_attention` — below ulp for continuous ops — can flip an expert assignment near a score boundary and cascade.

Verified on `mlx-community/gemma-4-26b-a4b-it-8bit` (M5 Max, 64 GB):

- Layer 0 Q/K/V and cache scatter: bit-exact between paged and mlx_lm
- Paged Metal kernel vs `mx.fast.SDPA` on identical Q/K/V: cos = 0.99999898
- Expert-set agreement for `"The weather today is not"`: 8/8 at layers 0-7, first flip at layer 8, 2/8 at layer 17
- 3/5 test prompts match `mlx_lm` bit-exact; 2 diverge at token 1-2
- Replacing the paged kernel with `mx.fast.SDPA` at prefill: 5/5 first-token match

31B is unaffected (4/5 bit-exact + 1 fp tie-break, same as before). The structural enablement in this PR — per-layer KV shapes, K-eq-V dispatch, cache allocation — is verified correct; the residual drift is kernel-level and pre-existing.

26B golden dicts stay empty so the test auto-skips rather than violating `_MIN_COMMON_PREFIX=6` on 2 prompts. Follow-up at #281 proposes gathering per-request K/V from the paged cache and dispatching `mx.fast.SDPA` on contiguous tensors — keeps paged-cache memory management and prefix caching, gives numerical parity with `mlx_lm` for MoE.

Also added `test_gemma4_k_eq_v_attention_detected_as_sdpa`: `is_sdpa()` was rejecting K-eq-V layers because they drop `v_proj`. Not exercised today (MHA dispatcher wraps every attention module), but the hybrid dispatcher classifies via `is_sdpa` — latent misrouting bug. One-line fix, test against a real `gemma4_text.Attention`.

tests/test_gemma4_golden.py
```diff
@@ -34,55 +72,292 @@
...
+
+    # MetalConfig and _MODEL_CACHE are process-level singletons; reset both
+    # and drop any previously-loaded model so the variant starts clean.
+    from vllm_metal.config import reset_config
+    from vllm_metal.v1 import model_lifecycle as _lifecycle
+
+    previous = os.environ.get("VLLM_METAL_MEMORY_FRACTION")
+    os.environ["VLLM_METAL_MEMORY_FRACTION"] = variant.memory_fraction
+    _lifecycle._MODEL_CACHE.clear()
+    reset_config()
```
LxYuan0420, line 285: Please avoid reaching into `_MODEL_CACHE` directly from the test.

If this reset is part of the supported lifecycle for model-loading tests, expose a small lifecycle-owned helper for it instead of coupling the test to a private module-global.

vllm_metal/v1/model_adapter.py
```diff
@@ -103,24 +113,14 @@ def resolve_max_head_dim(
...
-            and int(global_kv_heads) != int(num_kv_heads)
-        ):
-            raise ValueError(
-                f"Paged attention does not support variable KV head count: "
-                f"num_key_value_heads={num_kv_heads}, "
-                f"num_global_key_value_heads={global_kv_heads}. "
-                f"Use VLLM_METAL_USE_PAGED_ATTENTION=0 to fall back to the "
-                f"non-paged path."
-            )
+        del args, num_kv_heads
```
LxYuan0420: Do not make the default adapter's heterogeneous-KV guard a global no-op.

This PR only proves safe behavior when `build_per_layer_kv_shapes()` recognizes the config. For any other mixed-KV layout, dropping the old fail-fast check silently falls back to the uniform scalar path. Keep the historical rejection unless per-layer shapes have been populated for an explicitly supported layout.

vllm_metal/v1/model_adapter.py
```diff
@@ -170,3 +170,65 @@ def build_yoco_cache_mapping(
...
+
+        kv_heads_per_layer: list[int] = []
+        head_dim_per_layer: list[int] = []
+        for layer_type in layer_types:
+            if _is_sliding(layer_type):
+                kv_heads_per_layer.append(sliding_kv_heads)
+                head_dim_per_layer.append(sliding_head_dim)
+            else:
+                kv_heads_per_layer.append(full_kv_heads)
+                head_dim_per_layer.append(full_head_dim)
```
LxYuan0420: Make this layer-type mapping explicit.

Substring matching plus an open `else` turns unknown future layer types into silent full-attention shapes. Use exact layer-type sets for `sliding_attention` and `full_attention`, and raise on anything else.

vllm_metal/metal_kernel_backend/attention_sdpa.py
```diff
@@ -45,11 +45,16 @@
...
+    (26B/31B full-attention layers) share the K projection for values and
+    therefore never define one.  :func:`prepare_sdpa_qkv` handles the two
+    cases symmetrically.
+    """
     return (
         hasattr(module, "q_proj")
         and hasattr(module, "k_proj")
-        and hasattr(module, "v_proj")
         and hasattr(module, "o_proj")
     )
```
LxYuan0420, line 67: Keep this classifier targeted.

`HybridPagedAttentionBackend.patch_model()` uses `is_sdpa()` as a dispatch predicate before `is_linear_attention()`, so removing `v_proj` entirely broadens it to any attention module with `q_proj + k_proj + o_proj`. The real Gemma4 full-attention module already exposes `use_k_eq_v=True`; please require `v_proj` or `use_k_eq_v` instead of widening the heuristic for every future model.

tests/test_gemma4_golden.py
```diff
@@ -34,55 +72,292 @@
...
+# variants, large enough in aggregate to cross a decision boundary when
+# logit margins are tight.
+#
+# Populate these dicts once the follow-up kernel-numerics investigation
+# lands and the paged output matches mlx_lm within the existing prefix
+# invariant.  The rest of the test wiring (variant entry, selection
+# priority, env-var detection) is already in place — no other changes
+# are required at that point.
+GEMMA4_26B_GOLDEN_MLX_LM: dict[str, list[int]] = {}
+GEMMA4_26B_GOLDEN_PAGED: dict[str, list[int]] = {}
```
LxYuan0420: This PR claims 26B enablement, but these empty 26B golden sets mean `_select_variants()` will never run the 26B end-to-end path.

Please either add a real 26B verification path in this PR or narrow the test/docs/PR claims to the variants that are actually validated.

ericcurtin:
There are conflicts now also:

_is_sliding() substring check is fragile; module-level assert on goldens gives confusing import errors

dschulmeist:
Thanks for the thorough review — all six points addressed in the force-push.  Summary below.

Rebased onto current main, which includes #282 (per-layer sliding window).  Both PRs use the same adapter-method + lifecycle-block pattern for per-layer metadata, so they compose cleanly.

**R1 — test reaching into `_MODEL_CACHE`**
Added `reset_model_cache()` at module scope in `vllm_metal.v1.model_lifecycle`.  Thread-safe via the existing `_MODEL_CACHE_LOCK`.  Test calls it instead of the private global.

**R2 — `require_uniform_kv_heads` no-op**
Reverted the adapter method to its original raise-on-mismatch implementation.  The gate moved to the caller: `cache_policy.validate_paged_attention_support()` skips `require_uniform_kv_heads` only when `kv_heads_per_layer` is populated.  Any other mismatched-KV config still fails fast with the original error message.

**R3 — substring layer-type match**
Exact-match `"sliding_attention"` / `"full_attention"`; raises with layer index and layer_type on anything else.  New `test_unknown_layer_type_raises` case asserts the layer-index reporting actually works.  (Side note: #282's `build_sliding_window_per_layer` has the same permissive `else -1` shape for unknown types — happy to open a follow-up issue for that if you want the same treatment, didn't want to creep into another author's merged PR.)

**R4 — `is_sdpa` broadened**
Restored `v_proj` as primary, added `use_k_eq_v=True` as an explicit opt-in that Gemma4's K-eq-V full-attention layers already set.  Classifier is now `q_proj & k_proj & o_proj & (v_proj | use_k_eq_v)`.  Added a negative test: a Q/K/O-only class no longer classifies as SDPA.

**R5 — empty 26B goldens vs enablement claim**
I wanted to populate the goldens here but can't in good faith.  After rebasing onto main (which now includes #282's per-layer sliding-window enforcement) the 26B paged output on these prompts is not comparable with `mlx_lm` — the MoE top-k routing amplifies the paged-kernel-vs-`mx.fast.SDPA` numerical drift across 30 layers enough that the committed goldens would violate `_MIN_COMMON_PREFIX=6`.  So I left the 26B dicts empty and rewrote the in-code comment to state that honestly, with a pointer to the follow-up at #281 that proposes the gather-then-`mx.fast.SDPA`-dispatch fix.  Structural enablement is still verified — 26B loads end-to-end through the paged path with the correct per-layer shapes and K-eq-V dispatch — and the variant entry is retained so the follow-up PR only needs to regenerate the golden dicts.  31B parity is unaffected (dense model, no discrete routing to amplify).

**R6 (@ericcurtin) — module-level assert**
Converted to parametrized `test_golden_pairs_consistent[e2b|26b|31b]`.  Failures now surface as test results, not import errors.  26B skips cleanly because its goldens are empty.

All CI green (DCO + lint x3 + test on macos-15).  E2B and #282 wiring tests verified locally on M5 Max.

vllm_metal/v1/cache_policy.py
```diff
@@ -198,10 +198,19 @@ def should_setup_paged_attention(self) -> bool:
...
-        )
+        # ``require_uniform_kv_heads`` is the fail-fast for configs whose
+        # ``num_global_key_value_heads`` differs from ``num_key_value_heads``
+        # and which would silently fall back to the scalar uniform cache
+        # path with wrong sizing.  Adapters that populate
+        # ``kv_heads_per_layer`` via ``build_per_layer_kv_shapes`` (Gemma4
+        # 26B/31B) have already opted into the heterogeneous cache and
+        # handle mismatched KV counts layer-by-layer, so the uniform
+        # guarantee does not apply and the check is skipped for them.
+        if self._runner.kv_heads_per_layer is None:
```
ricky-chaoju, line 209: This now allows heterogeneous Gemma4 models to skip the uniform-KV guard, but the TurboQuant path still builds specs from the scalar runner-wide `num_kv_heads` / `head_dim`. That means `heterogeneous + turboquant` is still not modeled correctly: either we fail later during quantized cache construction, or if 512-wide TurboQuant is relaxed in the future we would silently size the quantized cache/specs with the wrong per-layer shape. I think this path should explicitly reject heterogeneous per-layer KV shapes under TurboQuant until the quantized cache/spec path is made per-layer as well.

tests/test_gemma4_golden.py
```diff
@@ -34,55 +72,292 @@
...
+        assert len(mlx_ids) == MAX_TOKENS, (
+            f"{variant.name} mlx_lm golden for {prompt!r} has "
+            f"{len(mlx_ids)} tokens, expected {MAX_TOKENS}"
+        )
+        assert len(paged_ids) == MAX_TOKENS, (
+            f"{variant.name} paged golden for {prompt!r} has "
+            f"{len(paged_ids)} tokens, expected {MAX_TOKENS}"
+        )
+        prefix = _common_prefix_len(mlx_ids, paged_ids)
+        assert prefix >= _MIN_COMMON_PREFIX, (
```
ricky-chaoju, line 219: The test contract feels materially weaker than the module docstring advertises. The top-level description says the paged path cannot drift from `mlx_lm` by more than a single token, but the actual invariant here is only `common_prefix >= 6` for a 10-token sample. In practice that still accepts 4 trailing tokens of divergence, which seems too loose for a regression test whose purpose is to keep the in-tree paged golden anchored to the independent `mlx_lm` reference. Could we either tighten the threshold, or update the docstring/comments so the accepted drift matches what the test really enforces?

ericcurtin (approved):
I think this has had enough iterations

PR #282, opened 2026-04-18
Enforce per-layer sliding window for Gemma4 attention 

ricky-chaoju:
Problem

`sdpa_forward` hardcoded `sliding_window=-1` on every `paged_attention_primitive` call, so Gemma4 `sliding_attention` layers silently attend beyond their configured window (E2B: 512, 31B: 1024). Metal kernel already has the mask (`pagedattention.metal:1065-1067`) — just never enabled.

Silent: short prompts unaffected; long context drifts from mlx_lm / upstream.

Fix

Thread per-layer window from `config.layer_types` + `config.sliding_window` through adapter → lifecycle → runner → cache_policy → MHA backend → KV cache, and read `kv_cache.sliding_window_per_layer[layer_idx]` in `sdpa_forward`.
Non-Gemma configs return `None` → cache defaults to `[-1] * N`, zero behaviour change for Qwen3 / Llama / hybrid Qwen3.5.

Testing

- Adapter unit (4 tests): None fallbacks, exact match, length mismatch.
- Wiring integration (3 tests): config → `MetalPagedKVCache`, incl. YOCO
  shared-layer cache-idx retrieval.
- Kernel dispatch, slow (3 tests): real Gemma4 E2B with a spy on
  `paged_attention_primitive`; asserts only `{512, -1}` appear and ratio
  matches `layer_types` exactly (28:7). Needs `GEMMA4_MODEL_PATH`.
- `test_gemma4_golden.py` still matches mlx_lm on E2B.

### DIFF
--- a/tests/test_gemma4_sliding_window_dispatch.py
+++ b/tests/test_gemma4_sliding_window_dispatch.py
@@ -1,0 +1,162 @@
+# SPDX-License-Identifier: Apache-2.0
+"""End-to-end kernel-dispatch verification for Gemma4 sliding window.
+
+Spies on the Metal ``paged_attention_primitive`` op during a real vLLM
+inference and asserts the per-layer ``sliding_window`` values reach the
+kernel with the correct split between sliding-attention and
+full-attention layers.  Complements ``test_sliding_window_wiring.py``
+which exercises the same wiring without running the Metal kernel.
+
+Gemma4 E2B is used (not 31B) because:
+- E2B has uniform KV heads, so the paged-attention guard in
+  ``require_uniform_kv_heads`` admits it today.  31B support unblocks
+  after upstream PR #279 lands.
+- E2B has YOCO (``num_kv_shared_layers=20``), so every model forward
+  exercises the shared-layer cache-index retrieval -- shared layers
+  must still receive a sliding_window matching their own attention
+  type, via ``cache_idx_map``.
+
+Run:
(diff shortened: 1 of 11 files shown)


tests/test_gemma4_sliding_window_dispatch.py
```diff
@@ -0,0 +1,162 @@
...
+_E2B_SLIDING_WINDOW = 512
+_E2B_NUM_SLIDING_LAYERS = 28
+_E2B_NUM_FULL_LAYERS = 7
+_E2B_TOTAL_LAYERS = _E2B_NUM_SLIDING_LAYERS + _E2B_NUM_FULL_LAYERS
+
+_NO_WINDOW = -1
+
+# Position of ``sliding_window`` in ``paged_attention_primitive``'s
+# positional signature (see ``attention_sdpa.py:489-510``).
+_SLIDING_WINDOW_ARG_INDEX = 11
```
WindChimeRan, line 46: hardcode arg index is fragile

WindChimeRan (commented):
This PR fixed a bug on long context, but only tested with short context. 

Please add tests before fail after pass with long context, in `tools/`

ericcurtin:
@ricky-chaoju could you catch @WindChimeRan 's feedback in another PR please

ericcurtin (approved):

ricky-chaoju:
> @ricky-chaoju could you catch @WindChimeRan 's feedback in another PR please

Thanks! On Paged KV right now. Will follow up on both tomorrow.

PR #283, opened 2026-04-19
[Paged KV] Enable prefix caching on the unified paged path

### LINKED ISSUE #182: [RoadMap] [Paged KV] Prefix Caching on Paged KV Cache Path

ricky-chaoju:
Summary
Removes the platform-layer force-disable that was blocking`enable_prefix_caching` on the paged path. The unified prefill code already handles `num_computed_tokens > 0` (#195, #207, #208, #211); the only remaining gap was that `platform.py:278-285` overrode it.

Adds an end-to-end correctness test that fires identical prompts twice through `vllm.LLM(enable_prefix_caching=True)` — the second pass walks the `start_pos > 0` path; tokens still match the cache-off golden.

Both test classes share a single `LLM` fixture: Metal memory held by a released `LLM` is not freed by Python gc, so a second module-scope `LLM` would hit `kv_budget=0`.

## Hybrid models
Upstream `ModelConfig.is_prefix_caching_supported` already returns False for hybrid/Mamba models, so the `default_prefix_caching` resolution in `vllm/engine/arg_utils.py` keeps cache off unless the user explicitly forces it. No vllm-metal-side guard needed.

## Benchmark
benchmark_182

Closes #182.

### DIFF
--- a/tests/test_paged_prefix_caching_e2e.py
+++ b/tests/test_paged_prefix_caching_e2e.py
@@ -1,0 +1,104 @@
+# SPDX-License-Identifier: Apache-2.0
+"""End-to-end correctness of paged prefix caching (issue #182).
+
+Fires the deterministic-test prompts twice through ``vllm.LLM`` with
+prefix caching enabled.  The first pass primes the cache; the second
+pass exercises the model_runner's ``start_pos > 0`` path because the
+upstream scheduler reports ``num_computed_tokens > 0``.  The asserted
+token sequence is the existing cache-off golden, so a broken cache-hit
+path surfaces as a token mismatch.
+
+The LLM body runs in a spawned child process (``multiprocessing`` with
+the ``spawn`` start method) so Metal device init happens in a fresh
+interpreter.  This is required on the Metal platform because:
+  - ``fork`` inherits the parent's Metal context and segfaults in the
+    child (Metal is not fork-safe).
+  - Running in the parent pytest process alongside the cache-off
+    baseline fixture in ``test_paged_deterministic`` causes
+    ``kv_budget=0`` — MLX wired buffers aren't released by Python gc.
+"""
+
+from __future__ import annotations
+
+import multiprocessing as mp
+import os
+
+import pytest
+
+from tests.test_paged_deterministic import (
+    DEFAULT_PAGED_MEMORY_FRACTION,
+    DEFAULT_USE_PAGED_ATTENTION,
+)
+
+
+def _setenv_default(key: str, default: str) -> None:
+    if os.environ.get(key) is None:
+        os.environ[key] = default
+
+
+def _run_prefix_cache_correctness() -> None:
+    """Body of the e2e test — runs in a spawned child process.
+
+    Imports happen lazily inside the child so vllm / MLX init is not
+    inherited from the parent process.
+    """
+    _setenv_default("VLLM_ENABLE_V1_MULTIPROCESSING", "0")
+    _setenv_default("VLLM_METAL_USE_PAGED_ATTENTION", DEFAULT_USE_PAGED_ATTENTION)
+    _setenv_default("VLLM_METAL_MEMORY_FRACTION", DEFAULT_PAGED_MEMORY_FRACTION)
+
+    if os.environ.get("VLLM_METAL_USE_PAGED_ATTENTION", "0") != "1":
+        return  # non-paged path: nothing to test
+
+    from vllm import LLM, SamplingParams
+
+    from tests.test_paged_deterministic import (
+        GOLDEN_MLX,
+        GOLDEN_PAGED,
+        MAX_TOKENS,
+        MODEL_NAME,
+        PROMPTS,
+    )
+
+    llm = LLM(
+        model=MODEL_NAME,
+        max_model_len=512,
+        max_num_seqs=1,
+        enable_prefix_caching=True,
+    )
+    sp = SamplingParams(temperature=0, max_tokens=MAX_TOKENS)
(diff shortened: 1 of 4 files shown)


WindChimeRan (commented):
maybe we should put the test of Paged Prefix Cache into separated files.

vllm_metal/platform.py
WindChimeRan, line 1: This will turn on prefix caching as default for user. Could you please run a quick benchmark comparing with and without prefix caching? We can check TTFT, throughput, and cache hit rate, to see if it actually works. For the dataset, concatenating a shared system prompt should be fine.

ricky-chaoju: Benchmark with `vllm bench serve --dataset-name prefix_repetition` on Qwen3-0.6B (paged attention). Workload: 1 shared prefix of 512 tokens, 50 prompts each with 128-token suffix and 100-token output; 10 repeats per side, median ± pstdev reported.
benchmark_182



| Metric              | Cache off (med ± sd) | Cache on (med ± sd) | Δ     |
|---------------------|----------------------|---------------------|-------|
| Throughput (tok/s)  | 210.59 ±   8.93      | 288.88 ±  12.23     | 1.37× |
| TTFT mean (ms)      | 8667.88 ± 588.32     | 6071.39 ± 554.66    | 1.43× |
| TTFT P50 (ms)       | 8622.69 ± 645.68     | 5894.11 ± 451.23    | 1.46× |
| TTFT P99 (ms)       | 16308.74 ± 1049.18   | 11523.16 ± 943.63   | 1.42× |
| TPOT P50 (ms)       | 30.74 ±   1.63       | 22.88 ±   0.77      | 1.34× |
| TPOT P99 (ms)       | 119.51 ±  51.95      | 25.62 ±   8.99      | 4.67× |
| E2EL P50 (ms)       | 11574.40 ± 664.36    | 8141.27 ± 652.99    | 1.42× |

Cache hit rate (from `/metrics`):
- Cold (run 1): 78.4% (25088 / 32000)
- Warm (runs 2-10, each): 97.5% (31200 / 32000)
- Cumulative across 10 cache-on runs: 95.6% (305888 / 320000)

`baseline-*` runs report 0 queries / 0 hits (cache disabled).

(On TPOT P99: cache-on stable across 10 runs (std ±9 ms, range 23-50). Baseline noisy (std ±52 ms, range 36-197). Single-sweep ratio floats ~3.5×-5.5× because of baseline noise, not cache-side flakiness. The distribution-based 4.67× is the robust number.)

tests/test_paged_prefix_caching_e2e.py
WindChimeRan, line 1: No cache hit assert

Suggest asserting on scheduler cache-hit counters or runner._request_states block reuse.  Even a single assert computed_tokens > 0 on at least one req

WindChimeRan (approved):

WindChimeRan:
> Upstream ModelConfig.is_prefix_caching_supported already returns False for hybrid/Mamba models, so the default_prefix_caching resolution in vllm/engine/arg_utils.py keeps cache off unless the user explicitly forces it. No vllm-metal-side guard needed.

Please add a followup patch for the doc #284

ricky-chaoju:
> > Upstream ModelConfig.is_prefix_caching_supported already returns False for hybrid/Mamba models, so the default_prefix_caching resolution in vllm/engine/arg_utils.py keeps cache off unless the user explicitly forces it. No vllm-metal-side guard needed.
> 
> Please add a followup patch for the doc #284

Will follow up once #284 lands.

tests/test_paged_prefix_caching_e2e.py
ricky-chaoju: done

PR #285, opened 2026-04-20
[Test] Gemma4 long-context sliding-window dispatch

ricky-chaoju:
Follow up on the review feedback for #282.

  - make `tests/test_gemma4_sliding_window_dispatch.py` build a prompt longer than Gemma4 E2B's 512-token sliding window so the test
  actually exercises the long-context path
  - record both `sliding_window` and `max_seq_len` from `paged_attention_primitive` and assert the kernel sees a context longer than
  the configured window
  - replace the hardcoded `paged_attention_primitive` argument index with lookup from nanobind signature metadata so the spy follows
  the canonical native-op signature

### DIFF
--- a/tests/test_gemma4_sliding_window_dispatch.py
+++ b/tests/test_gemma4_sliding_window_dispatch.py
@@ -24,7 +24,9 @@ Run:
 from __future__ import annotations
 
 import os
+import re
 from collections import Counter
+from dataclasses import dataclass
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


ericcurtin:
_build_long_prompt is O(n²) it could be simplified to a single multiplication once the per-fragment token count is measured once.

ricky-chaoju:
> _build_long_prompt is O(n²) it could be simplified to a single multiplication once the per-fragment token count is measured once.

done.

ericcurtin (approved):

PR #286, opened 2026-04-21
Bump vllm minimum version to 0.19.1

ricky-chaoju:
Pick up vllm 0.19.1, which ships transformers v5 officially upstream (#30566).                                                      
Drops the post-install override from #169 and aligns pyproject's transformers pin to >=5.5.1.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -123,7 +123,7 @@ main() {
 
   ensure_venv "$venv"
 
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):

ericcurtin:
Thanks

PR #288, opened 2026-04-22
[TurboQuant] Add 512-dim FWHT support

ricky-chaoju:
This PR extends TurboQuant's generic FWHT path to support `head_dim=512`, which is required by models such as Gemma4 full-attention layers.

Changes in this PR:
- include `512` in the Python-side FWHT supported head sizes
- add `FWHT_SIGNS_512` and `get_fwht_sign<512>` in the Metal TurboQuant helper
- add focused tests for:
    - Python/Metal FWHT sign-table parity
    - FWHT round-trip at 512
    - TurboQuant cache allocation for `head_dim=512`
    - 512-dim TurboQuant kernel output vs Python reference attention
- add missing module/file headers for TurboQuant Python files for consistency with the rest of the repo

## Why

Paged-attention kernels already instantiate `HEAD_SIZE=512`, but TurboQuant's FWHT/sign-table path only supported `64/128/256`. That made `512`-dim value quantization fail
before the later Gemma4-specific runtime work could even matter.

This PR removes that blocker without mixing in the per-layer cache-sizing and runtime-dispatch changes that follow in later stacked PRs.

### DIFF
--- a/vllm_metal/metal_kernel_backend/cache.py
+++ b/vllm_metal/metal_kernel_backend/cache.py
@@ -21,6 +21,7 @@ from vllm.logger import init_logger
 
 from vllm_metal.metal_kernel_backend.turboquant import (
     BLOCK_SIZE,
+    FWHT_SUPPORTED_HEAD_DIMS,
     QUANT_PARAMS,
     V_QUANT_PARAMS,
     packed_dim,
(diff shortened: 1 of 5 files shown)


LxYuan0420 (commented):
Gemma4 now already uses heterogeneous per-layer KV shapes, while the TurboQuant path is still uniform-only. So allowing `head_dim=512` here does more than add FWHT 512 support so it also removes the last early rejection for an unsupported `turboquant + per-layer KV shapes` runtime path.

vllm_metal/metal_kernel_backend/cache.py
```diff
@@ -86,9 +87,13 @@ def __init__(
                 raise ValueError(
                     f"TurboQuant requires head_dim divisible by 32, got {head_dim}"
                 )
-            if head_dim not in (64, 128, 256):
+            if head_dim not in FWHT_SUPPORTED_HEAD_DIMS:
+                supported_head_dims = ", ".join(
```
LxYuan0420, line 91: Allowing `head_dim=512` is not safe on current `main` by itself.

Gemma4 now uses per-layer heterogeneous KV shapes, but the TurboQuant path still assumes uniform scalar `num_kv_heads` / `head_dim`. Please add an explicit fail-fast for `turboquant + per-layer KV shapes` before opening this gate.

LxYuan0420 (commented):
I also think we need to update `vllm_metal/v1/cache_policy.py` before this is safe to merge by adding an explicit fail-fast for `turboquant + per-layer KV shapes` in the existing per-layer capability guard, plus a focused regression test for that guard.

vllm_metal/metal_kernel_backend/cache.py
ricky-chaoju: Addressed. I added an explicit fail-fast for `turboquant + per-layer KV shapes` and a focused regression test. So this PR only enables generic 512-dim FWHT support; heterogeneous TurboQuant remains blocked until the follow-up PR.

LxYuan0420 (approved):
LGTM

PR #291, opened 2026-04-22
[Metal] Support MLX 0.31.2 paged-op build

ricky-chaoju:
MLX 0.31.2 moved stream-scoped encoder/temporary management off `metal::get_command_encoder(Stream)` helper, breaking the `paged_ops.cpp` JIT build at runtime.
- Add a compile-time compatibility shim gated on `MLX_VERSION_NUMERIC`
  that routes to the correct API for both 0.31.1 and 0.31.2
- No kernel logic changes — only the C++ bridge dispatch layer

## Root cause

CI `uv pip install .` resolves to `mlx==0.31.2` (released after our last green run). The two removed `Device` members:

| Operation | 0.31.1 (`Device` method) | 0.31.2 (new API) |
|---|---|---|
| Get encoder | `Device::get_command_encoder(int)` | `metal::get_command_encoder(Stream)` |
| Add temporary | `Device::add_temporary(array, int)` | `CommandEncoder::add_temporary(array)` |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,9 @@ classifiers = [
 
 dependencies = [
     # MLX - Required for Apple Silicon GPU acceleration
-    "mlx>=0.31.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
+    # Upper-bound: paged_ops.cpp uses mlx/backend/metal/device.h (private API).
+    # Bump after verifying the JIT build against the new MLX minor release.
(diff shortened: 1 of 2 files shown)


ricky-chaoju:
@WindChimeRan @LxYuan0420 @ericcurtin  This PR fixes the CI build error.

LxYuan0420 (approved):
Thanks for the quick fix!

PR #292, opened 2026-04-22
Fix per-layer KV head count in SDPA dispatch

LxYuan0420:
This PR is:
- To pass the layer-specific KV head count to the Metal paged-attention kernel.
- To keep heterogeneous cache layout and kernel indexing consistent.
- To add a regression test for the host→kernel argument contract.

Note: 

`sdpa_forward()` already resolves the correct per-layer KV head count from `kv_cache.kv_heads_per_layer[layer_idx]`, but it was still passing the scalar `kv_cache.num_kv_heads` to `paged_attention_primitive()`.

That is wrong for heterogeneous models such as Gemma4 26B/31B. The Metal kernel uses `num_kv_heads` for grouping and cache stride, so the kernel must receive the current layer's KV head count, not the model-wide scalar default.

### DIFF
--- a/vllm_metal/metal_kernel_backend/attention_sdpa.py
+++ b/vllm_metal/metal_kernel_backend/attention_sdpa.py
@@ -502,7 +502,7 @@ def sdpa_forward(
             q_3d,
             kernel_k_cache,
             kernel_v_cache,
-            kv_cache.num_kv_heads,
+            cache_kv_heads,
             inner.scale,
             0.0,  # softcap (0 = disabled)
             block_tables,
@@ -525,7 +525,7 @@ def sdpa_forward(
             q_3d,
             kernel_k_cache,
             kernel_v_cache,
-            kv_cache.num_kv_heads,
+            cache_kv_heads,
             inner.scale,
             0.0,  # softcap (0 = disabled)
             block_tables,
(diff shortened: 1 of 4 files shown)


ericcurtin:
No TurboQuant test variant

vllm_metal/metal_kernel_backend/attention_sdpa.py
```diff
@@ -502,7 +502,7 @@ def _scatter(cache_arr, data):
             q_3d,
             kernel_k_cache,
             kernel_v_cache,
-            kv_cache.num_kv_heads,
```
ericcurtin, line 505: Lines 489–495 in still use `kv_cache.num_kv_heads` for reshaping the TurboQuant scale/zero caches during block-size translation.

LxYuan0420:
I want to keep this PR scoped to the per-layer KV-head dispatch fix, so instead of broadening it into TurboQuant-per-layer support I added an early fail-fast for `turboquant + per-layer KV shapes` in `vllm_metal/v1/cache_policy.py`, plus a focused regression test for that guard.

ricky-chaoju (approved):

ericcurtin (approved):

WindChimeRan:
CI failed. Could you please take a look?

Issue #295, opened 2026-04-22
Qwen 3.6 27B model not working


Anoopnk:
Model (HF repo ID)
            `Qwen/Qwen3.6-27B-FP8`
Hardware (chip + unified memory, e.g. M3 Max / 64 GB)
            `M2 Max, 96 GB`
vllm-metal commit or version
            `commit 2cbee0038f5f706020483d8046cf9ebf6f14f081 (tag: v0.2.0-20260422-145813)`
The exact command you ran
            `vllm serve Qwen/Qwen3.6-27B-FP8`
Error message or unexpected output
```txt
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108] EngineCore failed to start.
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108] Traceback (most recent call last):
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]   File "/Volumes/source/git/vllm-metal/.venv-vllm-metal/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 1082, in run_engine_core
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]     engine_core = EngineCoreProc(*args, engine_index=dp_rank, **kwargs)
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]   File "/Volumes/source/git/vllm-metal/.venv-vllm-metal/lib/python3.12/site-packages/vllm/tracing/otel.py", line 178, in sync_wrapper
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]     return func(*args, **kwargs)
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]            ^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]   File "/Volumes/source/git/vllm-metal/.venv-vllm-metal/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 848, in __init__
(EngineCore pid=75706) ERROR 04-22 19:23:05 [core.py:1108]     super().__init__(
[...]

PR #296, opened 2026-04-23
Add text-only compatibility mode for multimodal FP8 checkpoints

ricky-chaoju:
Summary
- add `VLLM_METAL_MULTIMODAL_MODE` with `auto`, `text-only-compat`, and `multimodal-native`
- route known-incompatible multimodal FP8 checkpoints to the text-only `mlx_lm` path
- add local shard and FP8 sanitize compatibility fixes for Qwen3.5/Qwen3.6 loading
- add tests for config, routing, cache separation, and compat behavior

## Notes
This unblocks text generation for the current Qwen3.6 FP8 checkpoint. Native multimodal input support is still follow-up work.

Related #295

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -3,11 +3,14 @@
 
 from __future__ import annotations
 
+import builtins
 import importlib.util
+import os
 import sys
 from types import ModuleType
 
 import numpy as np
+import pytest
 
 import vllm_metal.compat as compat
 
@@ -62,6 +65,25 @@ def _install_fake_qwen35_modules(monkeypatch, *, include_moe: bool):
 
 
 class TestQwen35Fp8CompatPatch:
+    def test_logs_when_mlx_core_is_unavailable(self, monkeypatch) -> None:
+        original_import = builtins.__import__
+        warnings = []
+
+        def _fake_import(name, globals=None, locals=None, fromlist=(), level=0):
+            if name == "mlx.core":
+                raise ImportError("missing mlx.core")
+            return original_import(name, globals, locals, fromlist, level)
+
+        def _record_warning(message, *args, **_kwargs):
+            warnings.append(message % args)
+
+        monkeypatch.setattr(builtins, "__import__", _fake_import)
+        monkeypatch.setattr(compat.logger, "warning", _record_warning)
+
+        compat._patch_mlx_lm_qwen35_fp8_sanitize()
+
+        assert any("mlx.core is unavailable" in warning for warning in warnings)
+
     def test_patches_dense_qwen35_even_when_moe_module_is_missing(
         self, monkeypatch
     ) -> None:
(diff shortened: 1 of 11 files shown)


ericcurtin:
FP8 block_size=128 hardcoded without validation; Qwen3_6 architecture name missing from override set; no real-checkpoint integration test

vllm_metal/v1/model_adapter.py
```diff
@@ -60,50 +60,101 @@ def build_sliding_window_per_layer(
...
+        - ``text-only-compat``: force the text-only path for multimodal models.
+        - ``auto``: apply the compatibility path only for known-incompatible
+          checkpoints such as Gemma4 and Qwen3.5/Qwen3.6 FP8 wrappers.
+        """
+        multimodal_mode = self._multimodal_mode()
+        if multimodal_mode == "multimodal-native":
+            return False
+        if multimodal_mode == "text-only-compat":
+            return True
+        return self._matches_auto_text_backbone_override(hf_config)
```
LxYuan0420, line 135: `text-only-compat` is too broad. This routes every multimodal checkpoint onto `mlx_lm`, but the PR only proves Gemma4 and Qwen3.5/Qwen3.6 FP8 wrappers. In the current env, real multimodal model types such as `phi3_v` do not resolve in `mlx_lm`, so this mode can send users into a guaranteed loader failure. Narrow it to a known-safe allowlist, or fail early with a clear compatibility error before calling `mlx_lm_load()`.

tests/test_model_lifecycle.py
```diff
@@ -121,6 +185,123 @@ def test_load_uses_adapter_override_for_text_only_multimodal_model(
...
+        lifecycle, runner = _make_lifecycle(
+            model_config=_runner_model_config(
+                hf_config=SimpleNamespace(model_type="qwen3_vl"),
+                is_multimodal_model=True,
+            )
+        )
+
+        lifecycle.load()
+
+        assert runner._is_vlm is False
```
LxYuan0420: This test does not prove the generic `text-only-compat` behavior. It pre-seeds a cached text model and only checks that `_is_vlm` flips to `False`, so it bypasses the real `mlx_lm_load()` path. If the mode stays generic, add a test that exercises the actual loader boundary; otherwise narrow the mode semantics to the checkpoints this PR really supports.

vllm_metal/compat.py
```diff
@@ -99,3 +100,102 @@ def _safe_check(
...
+        return
+
+    model_modules = []
+    for module_name in ("mlx_lm.models.qwen3_5", "mlx_lm.models.qwen3_5_moe"):
+        if find_spec(module_name) is None:
+            continue
+        try:
+            model_modules.append(import_module(module_name))
+        except Exception:
+            continue
```
LxYuan0420: This compatibility patch should not silently degrade to a no-op. Catch the specific import failures you expect here and log when the patch cannot be installed. Swallowing arbitrary exceptions in a required runtime compatibility path makes real regressions much harder to diagnose.

tests/test_compat.py
```diff
@@ -0,0 +1,98 @@
...
+            {
+                f"{gate_up_proj_prefix}.weight": np.ones((2, 256, 128)),
+                f"{gate_up_proj_prefix}.weight_scale_inv": np.ones((2, 2, 1)),
+                f"{gate_up_proj_prefix}.activation_scale": np.ones((2, 2, 1)),
+            }
+        )
+
+        assert f"{gate_up_proj_prefix}.weight_scale_inv" not in sanitized
+        assert f"{gate_up_proj_prefix}.activation_scale" not in sanitized
+        assert sanitized[f"{gate_up_proj_prefix}.weight"].shape == (2, 256, 128)
```
LxYuan0420, line 176: These tests only prove that the patch installs and strips the extra keys. They do not prove the FP8 dequantization math is correct, because `mx.from_fp8` and `mx.pad` are replaced with numpy stand-ins. Please add at least one value-level test for the `_dequant` path so we know this patch is numerically correct, not just shape-compatible.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -42,6 +46,83 @@ def reset_model_cache() -> None:
...
+                else f"model-{shard_index:05d}-of-{total_shards:05d}.safetensors"
+            )
+            (compat_path / compat_name).symlink_to(shard_path)
+
+        logger.info(
+            "Using mlx_lm shard compatibility view for %s (%d shard files)",
+            model_name,
+            total_shards,
+        )
+        yield str(compat_path)
```
LxYuan0420, line 123: Keep this helper private or move it onto `ModelLifecycle`. This is internal loader-compat glue, not a general module API, and exposing it at module scope makes a one-off workaround look like supported surface area.

vllm_metal/v1/model_adapter.py
ricky-chaoju: I narrowed text-only-compat to the known-safe allowlist, so generic VLMs like phi3_v no longer route to mlx_lm.

ricky-chaoju:
Addressed the remaining review feedback in f4af205:
  - narrowed text-only-compat to the known-safe allowlist
  - updated the lifecycle tests for the narrowed behavior and added an env-gated real checkpoint
  loader test
  - made the FP8 compat patch log expected import failures instead of silently no-oping
  - added value-level FP8 dequant coverage plus an opt-in real mlx FP8 test
  - renamed the shard compatibility helper to a private helper

LxYuan0420 (approved):

PR #297, opened 2026-04-24
Remove duplicate TurboQuant per-layer KV shapes guard

ricky-chaoju:
#288 and #292 each added the same `if get_config().turboquant` guard to `_require_supported_per_layer_shapes`. The second is unreachable and #292's test matches that dead message, so `pytest` fails on main. Drop the duplicate; point the test at #288's live message.


ricky-chaoju:
@LxYuan0420 @WindChimeRan This fixes the CI failure on main.

WindChimeRan (approved):

LxYuan0420:
Thanks for the quick fix!

PR #300, opened 2026-04-25
Optimise TurboQuant encode / decode kernels

UndercoverMathGuy:
**What**

Improves on the TurboQuant decode Metal kernels and adds a TQ encode kernel to improve encode/decode overhead 
|ctx | old total per step | new total per step | savings | speedup|
|-- | -- | -- | -- | --|
|128 | 10.82 + 28×0.47 = 24.01 ms | 0.53 + 28×0.20 = 6.23 ms | 17.79 ms | 3.86×|
|512 | 10.82 + 28×1.03 = 39.55 ms | 0.53 + 28×0.31 = 9.30 ms | 30.25 ms | 4.25×|
|2048 | 10.82 + 28×1.77 = 60.39 ms | 0.53 + 28×0.70 = 20.26 ms | 40.13 ms | 2.98×|
|8192 | 10.82 + 28×6.53 = 193.63 ms | 0.53 + 28×2.03 = 57.46 ms | 136.17 ms | 3.37×|

**How**

- Inverse FWHT — replaced 5–8 TG-memory + barrier round-trips with simd_shuffle_xor (stages 0–4) + intra-thread snapshot-commit (stages 5+), going to zero TG memory and zero barriers per inverse FWHT.
- Forward FWHT — lifted stages 0–4 onto simd_shuffle_xor, halving the TG-memory round-trips in the encode hot path (stages 5+ still need TG because encode threads don't own lane-contiguous slices).
- Deferred V FWHT — exploited FWHT linearity to accumulate in the rotated domain and apply one inverse FWHT per head at end-of-kernel, collapsing O(ctx × num_kv_heads) FWHTs to O(num_kv_heads) and eliminating the per-warp NUM_WARPS × head_dim TG workspace.
- Fused encode kernel — collapsed 6 per-layer dispatches into a single tq_encode Metal kernel with simd-reduced K scales, in-place forward FWHT, strided-fill V boundaries, and a generic tq_pack_byte packer — 168 dispatches/step → 28.
- MLX graph integration — aliases 5 input caches via copy_shared_buffer for in-place writes with distinct graph identity

Plus, adds 512 head dim hardcoded support for Gemma 4 with the new kernels

### DIFF
--- a/vllm_metal/metal/kernels_v2/turboquant.metal
+++ b/vllm_metal/metal/kernels_v2/turboquant.metal
@@ -194,57 +194,75 @@ template<> inline constexpr float fwht_inv_sqrt_n<512>() { return 0.044194173824
 
 // ========================================== Inverse FWHT
 
-// In-place inverse FWHT for HEAD_SIZE elements using threadgroup memory.
-// Supports HEAD_SIZE = 64 (6 stages), 128 (7 stages), 256 (8 stages),
-// or 512 (9 stages).
-// Each SIMD lane owns HEAD_SIZE/32 elements (lane i → indices i, i+32, i+64, ...).
-template<int HEAD_SIZE>
-inline void threadgroup_inverse_fwht(threadgroup float* fwht_buf, uint lane) {
+// In-place inverse FWHT on a thread-local register array `vals` of size
+// ELEMS_PER_LANE.  Each SIMD lane owns the slice
+// { lane, lane+32, lane+64, ... } of the logical HEAD_SIZE-vector.
+// The caller must supply `ELEMS_PER_LANE` explicitly; a `static_assert`
+// enforces the only valid coupling (one element per 32 lanes) so any future
+// mis-sized register array is a compile error, not silent OOB register access.
+//
+// Entirely register-resident:
+//   * Stages where mask < 32 exchange values across lanes of the same simd-
+//     group via `simd_shuffle_xor` — no threadgroup memory, no barriers.
+//   * Stages where mask >= 32 are intra-thread: the partner lives in the
+//     same lane at a different `e` offset.  RAW hazard across `e` iterations
+//     is handled by the snapshot-then-commit pattern (memory 86c1ae49).
+//
+// Called from tq_load_and_accumulate_v which dequantises V straight into
+// registers, runs this FWHT, and accumulates into v_accs — there is no
+// threadgroup-memory round-trip on the V read path.  Supports HEAD_SIZE
+// = 64 (6 stages), 128 (7 stages), 256 (8 stages), or 512 (9 stages).
+template<int HEAD_SIZE, int ELEMS_PER_LANE>
+inline void inverse_fwht_in_place(thread float* vals, uint lane) {
+    static_assert(ELEMS_PER_LANE == HEAD_SIZE / 32,
+        "inverse_fwht_in_place: ELEMS_PER_LANE must equal HEAD_SIZE/32; "
+        "callers allocate `vals[V_ELEMS_PER_THREAD]` at NUM_SIMD_LANES=32 and "
+        "passing a mismatched size would walk off the register array silently.");
     constexpr int NUM_STAGES = fwht_num_stages<HEAD_SIZE>();
-    constexpr int ELEMS_PER_LANE = HEAD_SIZE / 32;
-    float vals[ELEMS_PER_LANE];
 
+    // Stages 0-4: intra-simdgroup butterflies via register shuffle.  Zero TG
+    // memory, zero barriers.  On Apple GPUs simd_shuffle_xor is a single-
+    // cycle warp-level primitive; compare to the old path which paid 5
+    // sequential TG-memory-plus-barrier round-trips.
     #pragma unroll
-    for (int stage = 0; stage < NUM_STAGES; stage++) {
-        uint mask = 1u << stage;
+    for (int stage = 0; stage < 5; stage++) {
+        const uint mask = 1u << stage;
         #pragma unroll
         for (int e = 0; e < ELEMS_PER_LANE; e++) {
-            vals[e] = fwht_buf[lane + e * 32];
+            const float partner = simd_shuffle_xor(vals[e], mask);
+            const uint  idx     = lane + uint(e) * 32u;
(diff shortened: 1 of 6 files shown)


vllm_metal/metal/kernels_v2/turboquant.metal
```diff
@@ -194,57 +194,69 @@ template<> inline constexpr float fwht_inv_sqrt_n<512>() { return 0.044194173824
...
+//     same lane at a different `e` offset.  RAW hazard across `e` iterations
+//     is handled by the snapshot-then-commit pattern (memory 86c1ae49).
+//
+// Called from tq_load_and_accumulate_v which dequantises V straight into
+// registers, runs this FWHT, and accumulates into v_accs — there is no
+// threadgroup-memory round-trip on the V read path.  Supports HEAD_SIZE
+// = 64 (6 stages), 128 (7 stages), 256 (8 stages), or 512 (9 stages).
 template<int HEAD_SIZE>
-inline void threadgroup_inverse_fwht(threadgroup float* fwht_buf, uint lane) {
+inline void inverse_fwht_in_place(thread float* vals, uint lane) {
```
ricky-chaoju, line 216: `inverse_fwht_in_place` currently derives its register-array length as `HEAD_SIZE / 32`, while the caller allocates `v_accs` from `NUM_SIMD_LANES`. Today all instantiations use 32 lanes, so this is safe, but the coupling is implicit. Could we make the size part of the template, e.g. `template<int HEAD_SIZE, int ELEMS_PER_LANE>`, pass `V_ELEMS_PER_THREAD` from the caller, and `static_assert(ELEMS_PER_LANE == HEAD_SIZE / 32)`? That turns future misuse into a compile error instead of a silent OOB register-array access.

vllm_metal/metal/kernels_v2/turboquant.metal
```diff
@@ -338,28 +352,363 @@ inline void tq_load_and_accumulate_v(
...
+    // FWHT takes the input value in-register and returns the rotated value
+    // in-register; threadgroup memory is used only for cross-simdgroup
+    // butterfly stages.  No pre- or post-call barrier needed.
+    const float v_rot_f   = tg_forward_fwht_scalar<HEAD_SIZE>(v_val, fwht_buf, t);
+    const half  v_rot_h   = half(v_rot_f);
+    const float v_sqsum   = simd_sum(float(v_rot_h) * float(v_rot_h));
+    const half  v_scale_h = half(sqrt(v_sqsum * (1.0f / float(SG_SIZE))));
+    const float v_norm    = float(v_rot_h / v_scale_h);
+
+    // Searchsorted against the preloaded midpoint boundaries.  All reads
```
ricky-chaoju: The `v_boundaries` preload is currently synchronized by barriers inside `tg_forward_fwht_scalar` (or by the sub-8-bit K packing barrier). That works for the current HEAD_SIZE/kernel shape, but it makes this searchsorted path depend on an internal detail of the FWHT helper. Could we add an explicit `threadgroup_barrier(mem_flags::mem_threadgroup)` immediately before reading `v_boundaries` here, and update the comment accordingly? That keeps the async preload/K-encode overlap while making the preload synchronization local to the code that consumes it.

vllm_metal/metal/kernels_v2/turboquant.metal
```diff
@@ -338,28 +352,363 @@ inline void tq_load_and_accumulate_v(
...
     for (int i = 0; i < V_ELEMS_PER_THREAD; i++) {
         const int d = lane + i * NUM_SIMD_LANES;
         if (d < HEAD_SIZE) {
-            v_accs[i] += weight * fwht_buf[d];
+            v_accs[i] += weight * vals[i];
         }
     }
+    // v_centroids / v_bits no longer referenced after this point; silence
+    // unused-parameter warnings when the compiler decides so.
+    (void)v_centroids; (void)v_bits;
```
ricky-chaoju: These two parameters are used above in the dequant loop, so this suppressor is dead/misleading. Can we remove the comment and `(void)v_centroids; (void)v_bits;` line?

vllm_metal/metal/paged_ops.cpp
```diff
@@ -939,6 +1123,70 @@ NB_MODULE(_paged_ops, m) {
...
+        },
+        nb::arg("key"), nb::arg("value"),
+        nb::arg("key_cache"), nb::arg("value_cache"),
+        nb::arg("key_scale_cache"), nb::arg("value_scale_cache"),
+        nb::arg("key_zero_cache"),
+        nb::arg("slot_mapping"),
+        nb::arg("v_centroids"),
+        nb::arg("v_bits"),
+        nb::arg("k_bits") = 8,
+        nb::arg("k_signed") = true,
```
ricky-chaoju: Can we avoid defaulting `k_signed` here? The production path passes the value explicitly from `QUANT_PARAMS`, but this exported low-level API has an awkward footgun: `k_bits=4` with omitted `k_signed` defaults to `true` and immediately throws at runtime. I would make both `k_bits` and `k_signed` required in the binding so ad-hoc callers must choose the quantization semantics explicitly.

ericcurtin:
Misleading FWHT linearity comment for partitioned path, precision loss risk in reduce kernel, no encode→decode roundtrip test

vllm_metal/metal/paged_ops.cpp
UndercoverMathGuy: Corrected

vllm_metal/metal/kernels_v2/turboquant.metal
UndercoverMathGuy: Done - thanks

vllm_metal/metal/kernels_v2/turboquant.metal
UndercoverMathGuy: Yes - fixed

vllm_metal/metal/kernels_v2/turboquant.metal
UndercoverMathGuy: Add the parameter and assert

UndercoverMathGuy:
@ericcurtin - fixed the precision loss by moving the iFWHT to the reduce kernel with no fp16 intermediates (accumulations in fp32) - runs only once now so speed boost too

ericcurtin (approved):

PR #307, opened 2026-04-27
Add experimental Gemma4 YOCO fast prefill

ricky-chaoju:
This PR Adds a default-off experimental Gemma4 YOCO fast-prefill path for Metal paged attention.

Enable with:

```bash
VLLM_METAL_KV_SHARING_FAST_PREFILL=1
```

For eligible Gemma4 / Gemma4 text models, this runs YOCO KV-shared layers only on selected logits positions during prefill, then scatters the selected hidden states back before final norm / LM head.

## Validation

- Unit tests: 36 passed
- E4B smoke test: server starts, fast prefill enabled, `/v1/completions` returns 200
- Greedy off/on smoke: matched output on 64 / 512 / 960 token prompts

## Benchmark
gemma4-yoco-fast-prefill-benchmark-1

### DIFF
--- a/tests/test_yoco_fast_prefill.py
+++ b/tests/test_yoco_fast_prefill.py
@@ -3,68 +3,229 @@
 
 from __future__ import annotations
 
+import json
+import logging
+import os
+import subprocess
+import sys
+from pathlib import Path
+
 import pytest
 
 from vllm_metal.yoco_fast_prefill import (
     build_yoco_reduced_context_from_full_metadata,
-    get_yoco_fast_prefill_ineligibility_reason,
-    is_yoco_fast_prefill_eligible,
-    patch_gemma4_yoco_fast_prefill,
+    try_enable_gemma4_yoco_fast_prefill,
 )
 
+_REAL_GEMMA4_MODEL_ENV = "GEMMA4_FAST_PREFILL_MODEL_PATH"
+_REAL_GEMMA4_FALLBACK_MODEL_ENV = "GEMMA4_MODEL_PATH"
+_REAL_GEMMA4_PROMPTS = [
+    "The capital of France is",
+    "One plus one equals",
+]
+_REAL_GEMMA4_MAX_MODEL_LEN = 1024
+_REAL_GEMMA4_MAX_TOKENS = 8
+_FAST_PREFILL_ENABLED_ATTR = "_vllm_metal_yoco_fast_prefill_enabled"
+_REAL_RESULT_MARKER = "VLLM_METAL_YOCO_FAST_PREFILL_RESULT="
 
-@pytest.mark.parametrize("model_type", ["gemma4", "gemma4_text"])
-def test_eligible_for_gemma4_yoco_paged_attention(model_type: str) -> None:
-    assert is_yoco_fast_prefill_eligible(
+
+def _run_real_gemma4_paged_path(
+    model_path: str,
+    *,
+    fast_prefill: bool,
+) -> tuple[dict[str, list[int]], bool]:
+    memory_fraction = os.environ.get(
+        "GEMMA4_FAST_PREFILL_MEMORY_FRACTION",
+        os.environ.get("VLLM_METAL_MEMORY_FRACTION", "0.5"),
+    )
+    env = os.environ.copy()
+    env.update(
         {
-            "model_type": model_type,
-            "num_hidden_layers": 42,
-            "num_kv_shared_layers": 18,
-        },
-        use_paged_attention=True,
+            "VLLM_ENABLE_V1_MULTIPROCESSING": "0",
+            "VLLM_METAL_USE_PAGED_ATTENTION": "1",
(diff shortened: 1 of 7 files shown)


ericcurtin:
MLX in-place write semantics unverified, gdn_slot_mapping passed unmodified to reduced context, warning suppressed when num_hidden_layers missing

tests/test_yoco_fast_prefill.py
```diff
@@ -0,0 +1,318 @@
...
+
+    mx.eval(out)
+    assert [layer.calls[0]["tokens"] for layer in text_model.layers] == [5, 5, 1, 1]
+    assert text_model.layers[2].calls[0]["cu_seqlens"] == (0, 1)
+    assert text_model.layers[2].calls[0]["offsets"] == (4,)
+    assert text_model.layers[0].calls[0]["gdn_slot_mapping"] == [17]
+    assert text_model.layers[2].calls[0]["gdn_slot_mapping"] is None
+    assert text_model.layers[3].calls[0]["gdn_slot_mapping"] is None
+    assert out[0, 0, 0].item() == 3
+    assert out[0, 4, 0].item() == 10
```
LxYuan0420: Please add a slow real-Gemma4 paged-path parity test with fast prefill off/on. This fake-layer test only proves query reduction and context rewriting; it does not verify that the patched Gemma4 forward path preserves real output parity.

vllm_metal/v1/cache_policy.py
```diff
@@ -587,6 +587,41 @@ def setup_paged_attention(self, *, overhead: int) -> None:
...
+            if fast_prefill_patched:
+                logger.info(
+                    "Gemma4 YOCO fast prefill enabled "
+                    "(VLLM_METAL_KV_SHARING_FAST_PREFILL=1)"
+                )
+            elif reason is None:
+                logger.warning(
+                    "VLLM_METAL_KV_SHARING_FAST_PREFILL=1 was requested, "
+                    "but this model is not eligible; continuing without fast prefill"
+                )
```
LxYuan0420: This YOCO enablement flow should not live inline in `setup_paged_attention()`. Move it behind a single `yoco_fast_prefill.try_enable_gemma4_yoco_fast_prefill(...)` call and keep this method as orchestration only.

vllm_metal/yoco_fast_prefill.py
```diff
@@ -0,0 +1,445 @@
...
+    attention.  Keep this predicate intentionally narrow until runtime
+    correctness is proven and benchmarked.
+    """
+    return (
+        get_yoco_fast_prefill_ineligibility_reason(
+            model_args,
+            use_paged_attention=use_paged_attention,
+        )
+        is None
+    )
```
LxYuan0420: Please delete get_yoco_fast_prefill_ineligibility_reason(). This module should expose one clean public enablement entrypoint, with eligibility and logging kept internal, instead of adding a second exported string-returning helper just to support cache_policy.py.

LxYuan0420: Expected ownership shape should look more like:

```python
# vllm_metal/v1/cache_policy.py
WorkerCachePlanner.setup_paged_attention()
-> yoco_fast_prefill.try_enable_gemma4_yoco_fast_prefill(...)

# vllm_metal/yoco_fast_prefill.py
- YocoReducedContextMetadata
- build_yoco_reduced_context_from_full_metadata(...)
- try_enable_gemma4_yoco_fast_prefill(...) -> bool as the only public feature entrypoint
- eligibility, patch installation, and fallback logging kept private inside the module
```

LxYuan0420 (changes requested):

WindChimeRan (commented):
Image

I run some correctness evaluations on a simple classification task and this PR shows no regression on correctness. So I would suggest to turn it on as default. Thus no more additional env var

ricky-chaoju:
@WindChimeRan Thanks for running the eval. I checked upstream vLLM too: `kv_sharing_fast_prefill` is still opt-in there (False by default, enabled via `--kv-sharing-fast-prefill`). I’d prefer to keep this PR aligned with that and land it default-off first, then flip the default/remove the env in a follow-up if we’re comfortable after broader soak. WDYT?

WindChimeRan:
@ricky-chaoju this optimization only affect on Gemma4, the exp shows clear improvement on speed and no correctness regression. Plus, the code has no actual coupling with the upstream vllm.

So I think make it as default would be better, because it won't confuse the users.

ricky-chaoju:
> @ricky-chaoju this optimization only affect on Gemma4, the exp shows clear improvement on speed and no correctness regression. Plus, the code has no actual coupling with the upstream vllm.
> 
> So I think make it as default would be better, because it won't confuse the users.

Updated this PR to make Gemma4 YOCO fast prefill default-on for the paged path.

LxYuan0420 (approved):

PR #310, opened 2026-04-28
bump vllm version to 0.20.0

WindChimeRan:
This should unblock #262

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -123,7 +123,7 @@ main() {
 
   ensure_venv "$venv"
 
-  local vllm_v="0.19.1"
+  local vllm_v="0.20.0"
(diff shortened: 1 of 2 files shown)


LxYuan0420:
That's fast. 😃 
Will check this out in a bit.

WindChimeRan:
@LxYuan0420 CI test failure, because of the breaking change introduced by vllm 0.20.0. I think the right way is to close this PR, and to cherry pick the version bump to #262

LxYuan0420:
I haven't got the chance to go thru the changlog of v0.20. Anyway, don't need rush to bump the version and let's do it carefully.

Issue #289, opened 2026-04-22
vLLM Metal Model Support


WindChimeRan:
Tracking issue for community testing of models on `vllm-metal`. If you have an Apple Silicon Mac, we'd love your help expanding verified model coverage.

## How to help

Pick a model, run it with `vllm-metal`, and report back. Good starting points are models not tested in the [supported models table](https://github.com/vllm-project/vllm-metal/blob/main/docs/supported_models.md), but any text-only LLM on Hugging Face is welcome.

## Workflow

**It works?** Open a PR updating https://github.com/vllm-project/vllm-metal/blob/main/docs/supported_models.md
- Add or update the model's row (status, attention kernel, notes)
- Link back to this issue in the PR description

**It doesn't work?** File a new issue with:
- Model (HF repo ID)
- Hardware (chip + unified memory, e.g. M3 Max / 64 GB)
- `vllm-metal` commit or version
- The exact command you ran
- Error message or unexpected output
- Relevant logs

## Pass bar for a 🟡 → ✅ / 🔵 status bump

At minimum:
- Loads without errors
- Generates coherent output on a standard prompt
- Hardware and environment documented in the PR

Numerical parity checks or throughput measurements are welcome but not required for an initial status update.

*Note: `vllm-metal` currently focuses on text-only LLMs. Multi-modal (vision / audio) support is not yet in scope.*

## Resources

- Contributor guide: https://docs.vllm.ai/projects/vllm-metal/en/latest/CONTRIBUTING/
- Supported models: https://github.com/vllm-project/vllm-metal/blob/main/docs/supported_models.md

Questions? Drop a comment below. PRs welcome!

komoike-oss28-ui:
Hi, I'd like to help test models on vllm-metal!

My environment:
- Apple M1 MacBook Air / 8 GB unified memory
- Planning to test: Llama3 and/or Gemma3 (starting with 1B 4-bit quantized variants from mlx-community)

I'll report back with results.

komoike-oss28-ui:
Hi, I tried testing Llama-3.2-1B-Instruct-4bit on my M1 MacBook Air (8GB), 
but encountered a build error with the Metal kernel on macOS 26.

Error:
RuntimeError: Failed to load Metal kernel: Failed to build paged_ops extension:
no member named 'get_command_encoder' in 'mlx::core::metal::Device'
no member named 'add_temporary' in 'mlx::core::metal::Device'

Environment:
[...]

PR #312, opened 2026-04-28
[Qwen3.6] Stack per-expert MoE tensors during mlx_lm sanitize

sdayal:
Tracking issue: #289

## Summary

`Qwen/Qwen3.6-35B-A3B-FP8` ships expert MLPs as one tensor per expert per projection:

```
model.language_model.layers.{L}.mlp.experts.{E}.{gate,up,down}_proj.weight
```

The bf16 master `Qwen/Qwen3.6-35B-A3B` is already pre-stacked (`experts.gate_up_proj` / `experts.down_proj`) and loads via the existing combined-format branch in `mlx_lm.qwen3_5_moe.Model.sanitize` unchanged — only the FP8 release lands per-expert, likely because Qwen's FP8 quantization pipeline runs per-expert and the artifact is not re-stacked.

On vllm-metal `main`, loading `Qwen/Qwen3.6-35B-A3B-FP8` fails strict `load_weights` with `Received 30720 parameters not in model` — these are per-expert MoE tensors that vllm-metal's existing FP8 dequant compat doesn't address. (For reference: the same checkpoint on vanilla mlx-lm fails with 61,690 keys, the difference being 30,970 FP8 `weight_scale_inv` tensors that vllm-metal's `compat.py::_dequantize_qwen35_fp8_weights` already handles separately.)

## What this PR does

Add `_stack_qwen36_moe_per_expert_weights` chained after FP8 dequant in the MoE sanitize wrapper:

1. **Scan** the weights dict for per-layer experts prefixes and their expert-index sets.
2. **Validate** each prefix's index set is a contiguous `{0, 1, …, N-1}` (raises `ValueError` otherwise).
3. **Walk** the per-expert tensors in order, `mx.stack` along axis 0, `mx.concatenate` gate+up along the intermediate-dim axis, emit the combined `experts.gate_up_proj` / `experts.down_proj` form upstream sanitize already handles.

Pre-stacked checkpoints are unaffected (helper short-circuits when no per-expert keys are present).

The MoE-only nature of the stacking is made explicit by splitting the sanitize patch by model class:

- `mlx_lm.models.qwen3_5.Model` → wrapped with **FP8 dequant only** (`_transform_dense`)
- `mlx_lm.models.qwen3_5_moe.Model` → wrapped with **FP8 dequant + per-expert stacking** (`_transform_moe`)

Routing is driven by an explicit `transforms_by_module` map; future Qwen variants added without a corresponding entry get logged as `unpatchable` rather than silently inheriting one of the two transforms.

## Why a vllm-metal compat shim instead of waiting for upstream

mlx-community publishes pre-stacked redistributions of this checkpoint that already load on existing mlx-lm. This shim lets users load Qwen-org's canonical FP8 artifact directly without a 35GB→70GB bf16 intermediate conversion step that doesn't fit on memory-constrained Macs (≤64 GB).

This complements ml-explore/mlx-lm#1224, which adds the same per-expert stacking logic plus FP8 `weight_scale_inv` dequant for the qwen3_5 family inline in upstream sanitize. Once mlx-lm#1224 lands and vllm-metal's mlx-lm pin bumps past a release containing it, both this PR's per-expert stacking shim and the existing `_dequantize_qwen35_fp8_weights` shim in `compat.py` become removable in a follow-up cleanup.

## Files

- `vllm_metal/compat.py` — add `_stack_qwen36_moe_per_expert_weights` helper; split sanitize patches into per-class transforms via `transforms_by_module`.
- `docs/supported_models.md` — update Qwen3.6 row note + link this PR.
- `tests/test_compat.py` — four new unit tests using the existing numpy-fake-mlx fixture (no real model weights, runs in milliseconds):
  - `test_per_expert_moe_tensors_stack_to_combined` — positive: per-expert input produces correctly stacked combined output, content preserved per axis-0 slot.
  - `test_pre_stacked_moe_is_noop_for_per_expert_helper` — regression: pre-stacked input passes through unchanged (covers mlx-community redistributions and Qwen3.6 bf16 master).
  - `test_non_contiguous_per_expert_indices_raise` — defensive: malformed `{0, 1, 3}` checkpoint raises `ValueError`.
  - `test_per_expert_helper_does_not_run_on_dense_qwen35` — architecture invariant: dense path doesn't run the MoE helper.

## Verification (per #289 pass bar)

| Checkpoint | Hardware | Status | Output |
|---|---|---|---|
| `Qwen/Qwen3.6-35B-A3B-FP8` | M3 Max / 128 GB | ✅ loads, generates | `"The capital of France is"` → `" Paris, a city renowned for its iconic landmarks such"` |
| `Qwen/Qwen3.6-35B-A3B` (bf16) | M3 Max / 128 GB | ✅ loads via unchanged combined-format branch (both shims dormant — confirms the new branch is properly gated) | same output |

- Hybrid SDPA + GDN linear attention path on Apple Silicon Metal, paged KV cache.
- `pytest tests/test_compat.py`: **15 passed, 1 skipped** (the skip is the pre-existing `VLLM_METAL_RUN_REAL_MLX_FP8_TESTS=1`-gated test).
- Existing Qwen3.5 golden-token smoke (`test_qwen35_smoke.py`): 5/5 pass, unchanged.
- `bash scripts/lint.sh`: clean (shellcheck, ruff check, ruff format --check, mypy).

Rebased on latest `main` (3323d32) and re-validated against the bumped dep stack: `mlx-lm 0.31.3` (from #313), `vllm 0.20.0+cpu` (from #262), `transformers 5.7.0`.


tests/test_qwen36_smoke.py
```diff
@@ -0,0 +1,81 @@
...
+    print(f"  ids:    {token_ids}")
+
+    # Loose factual assertion: greedy decode of "The capital of France is" must
+    # surface "Paris" within the first MAX_TOKENS tokens for any reasonable
+    # Qwen3.6-A3B variant. Tighter golden IDs would be brittle across mlx
+    # versions and quant formats.
+    assert len(token_ids) == MAX_TOKENS, (
+        f"expected {MAX_TOKENS} tokens, got {token_ids}"
+    )
+    assert "Paris" in text, f"expected 'Paris' in greedy output, got {text!r}"
```
LxYuan0420: This is not the right test surface for this PR. It is a local opt-in 35B smoke test, skipped by default, so it should live under `tools/` or stay out of the PR. The required repo test should be focused `tests/test_compat.py` coverage for the sanitize transform: per-expert stacking, pre-stacked no-op, and malformed-index failure.

vllm_metal/compat.py
```diff
@@ -131,6 +131,81 @@ def _dequantize_qwen35_fp8_weights(
...
+            )
+        for proj in ("up_proj", "down_proj"):
+            if sorted(proj_to_experts[proj].keys()) != expert_indices:
+                raise ValueError(f"Per-expert MoE index mismatch at {prefix!r}.{proj}.")
+        gate = mx.stack([proj_to_experts["gate_proj"][i] for i in expert_indices])
+        up = mx.stack([proj_to_experts["up_proj"][i] for i in expert_indices])
+        down = mx.stack([proj_to_experts["down_proj"][i] for i in expert_indices])
+        new_weights[f"{prefix}.gate_up_proj"] = mx.concatenate([gate, up], axis=-2)
+        new_weights[f"{prefix}.down_proj"] = down
+    return new_weights
```
LxYuan0420, line 226: Since mlx-lm#1224 is the same upstream fix, please keep this downstream shim narrow and temporary. This helper should mirror the upstream `qwen3_5_moe` sanitize behavior as closely as possible, not become a broader vllm-metal-owned Qwen weight-layout transformer.

vllm_metal/compat.py
```diff
@@ -177,11 +252,16 @@ def _patch_mlx_lm_qwen35_fp8_sanitize() -> None:
         )
         return
 
+    def _transform(_self, weights):
+        weights = _dequantize_qwen35_fp8_weights(weights, mx)
+        weights = _stack_qwen36_moe_per_expert_weights(weights, mx)
+        return weights
```
LxYuan0420: Suggested shape: patch `mlx_lm.models.qwen3_5.Model` with `_dequantize_qwen35_fp8_weights(...)` only, and patch `mlx_lm.models.qwen3_5_moe.Model` with `_dequantize_qwen35_fp8_weights(...)` followed by the temporary per-expert stacking shim. That makes the ownership clear and keeps the upstream workaround removable.

LxYuan0420 (commented):
Left some comments.

tests/test_qwen36_smoke.py
sdayal: Agreed — removed `tests/test_qwen36_smoke.py` and added the unit-level coverage to `tests/test_compat.py` instead, mirroring the existing FP8 dequant + Gemma4 KV-shared test patterns:

  - `test_per_expert_moe_tensors_stack_to_combined` — positive: per-expert input → combined `experts.gate_up_proj` / `experts.down_proj`, shapes and content
  preserved.
  - `test_pre_stacked_moe_is_noop_for_per_expert_helper` — regression: pre-stacked input passes through unchanged (covers mlx-community redistributions and the Qwen3.6 bf16 master).
  - `test_non_contiguous_per_expert_indices_raise` — defensive: `{0, 1, 3}` raises `ValueError` rather than silently dropping experts.
  - `test_per_expert_helper_does_not_run_on_dense_qwen35` — architecture invariant: dense `qwen3_5.Model` path doesn't run the MoE helper.

  All four use the existing `_install_fake_qwen35_modules` numpy-stubbed fixture — no real model weights, milliseconds to run, fits CI. New summary: `15 passed, 1 skipped` (the skip is the pre-existing `VLLM_METAL_RUN_REAL_MLX_FP8_TESTS=1`-gated test).

vllm_metal/compat.py
sdayal: Aligned. Rewrote `_stack_qwen36_moe_per_expert_weights` to mirror the (scan → validate → walk) structure of ml-explore/mlx-lm#1224 rather than carrying its own dict-of-dict-of-dict groupings:

  1. **Scan** the weights dict for per-layer experts prefixes and their expert-index sets.
  2. **Validate** each prefix's index set is a contiguous `{0, 1, …, N-1}` (raises `ValueError` otherwise).
  3. **Walk** the per-expert tensors in order, `mx.stack` along axis 0, `mx.concatenate` gate+up along the intermediate-dim axis, emit the combined form upstream
  sanitize already handles.

  Same control flow as the upstream PR; the only structural difference is forced by where we wrap (pre-sanitize, before the model's `model.language_model.*` → `language_model.model.*` remap). Doc comment now points at mlx-lm#1224 with an explicit "removable once vllm-metal's mlx-lm pin bumps past that merge" sentinel so the lifecycle is clear.

vllm_metal/compat.py
sdayal: Done. Split the sanitize patch by model class so ownership is explicit:

  - `mlx_lm.models.qwen3_5.Model` → wrapped with `_transform_dense` (FP8 dequant only).
  - `mlx_lm.models.qwen3_5_moe.Model` → wrapped with `_transform_moe` (FP8 dequant + per-expert stacking).

  Routing is driven by an explicit `transforms_by_module` map keyed on the fully-qualified module name. Any future qwen variant added to the model-modules list without a corresponding entry in the map gets logged as `unpatchable` rather than silently inheriting one of the two transforms. The MoE-only nature of the
  stacking shim is now self-evident from the patch wiring, not just from the helper's runtime short-circuit.

sdayal:
Thanks for the review. Pushed `1a9622e` addressing all three:

  - **Test surface**: removed the opt-in 35B smoke; added four `tests/test_compat.py` unit tests covering positive / pre-stacked no-op / gap-raises /dense-path-isolation.
  - **Helper shape**: rewrote to mirror upstream mlx-lm#1224's (scan → validate → walk) structure. Doc comment now points at the upstream PR with an explicit removability marker.
  - **Patch ownership**: split into `_transform_dense` (qwen3_5) and `_transform_moe` (qwen3_5_moe), driven by an explicit `transforms_by_module` map.

  Pre-submit clean: `pytest tests/test_compat.py` 15 passed / 1 skipped, `bash scripts/lint.sh` all four stages green. Rebased on latest `main` (3323d32) and re-validated against the bumped dep stack: `mlx-lm 0.31.3` (from #313), `vllm 0.20.0+cpu` (from #262), `transformers 5.7.0`.

  End-to-end re-verified on `Qwen/Qwen3.6-35B-A3B-FP8` (per-expert path fires) and `Qwen/Qwen3.6-35B-A3B` bf16 master (combined-format path, both shims dormant).

vllm_metal/compat.py
```diff
@@ -131,6 +131,76 @@ def _dequantize_qwen35_fp8_weights(
...
+        # form upstream sanitize already handles.
+        gates, ups, downs = [], [], []
+        for e in range(len(indices)):
+            gates.append(new_weights.pop(f"{prefix}.{e}.gate_proj.weight"))
+            ups.append(new_weights.pop(f"{prefix}.{e}.up_proj.weight"))
+            downs.append(new_weights.pop(f"{prefix}.{e}.down_proj.weight"))
+        new_weights[f"{prefix}.gate_up_proj"] = mx.concatenate(
+            [mx.stack(gates), mx.stack(ups)], axis=-2
+        )
+        new_weights[f"{prefix}.down_proj"] = mx.stack(downs)
```
LxYuan0420, line 225: `_stack_qwen36_moe_per_expert_weights()` discovers experts from `gate_proj` only, then assumes matching `up_proj` and `down_proj` keys exist. Since this transforms external checkpoint weights, please validate all three projection families before stacking and raise a clear `ValueError` on missing projections instead of leaking a raw `KeyError`. A focused unit test for the missing-projection case would cover this.

sdayal: Pushed 6caa176 — rewrote the scan to discover per-projection index sets for all three families (gate_proj/up_proj/down_proj) instead of just gate_proj, with explicit validation before the pop-and-stack walk:

  1. Every prefix must have all three projection families — missing family raises `ValueError("Per-expert MoE weights at {prefix!r} are missing projection families: [...]")`.
  2. gate_proj indices must be contiguous {0..N-1} (existing check; message now names gate_proj explicitly).
  3. up_proj / down_proj index sets must equal gate_proj's — mismatch raises `ValueError("Per-expert MoE weights at {prefix!r} have mismatched {proj} indices vs gate_proj: missing=[...], unexpected=[...]")`.

A malformed checkpoint now surfaces as a single named-projection error from the validate step rather than a raw KeyError from dict.pop during the walk.

Added `test_missing_projection_family_raises` with two cases: entire down_proj family absent (matches `"missing projection families"`) and down_proj missing for one expert only (matches `"mismatched down_proj"`).

Pre-submit clean: `pytest tests/test_compat.py` 16 passed / 1 skipped;

Thanks for the catch.

LxYuan0420 (approved):
LGTM; TODO: remove it once mlx-lm#1224 lands in the pinned upstream version

PR #314, opened 2026-04-29
Fix Phi-4-mini-instruct on Metal by supporting Phi3/Phi4-style packed qkv attention

szfmsmdx:
Refs issue #289

This PR fixes the Metal SDPA path for Phi3/Phi4-style packed attention modules and unblocks local validation of `microsoft/Phi-4-mini-instruct` on Apple Silicon.

## Summary

This PR fixes the Metal SDPA path for Phi3/Phi4-style packed attention modules that expose a single `qkv_proj` projection instead of split `q_proj` / `k_proj` / `v_proj` layers.

Concretely, this change:
- teaches `is_sdpa()` to recognize Phi3/Phi4-style packed attention modules while keeping the existing split-projection SDPA contract intact
- updates `prepare_sdpa_qkv()` to split packed Q/K/V projections into the canonical SDPA layout
- preserves fallback to the split-projection contract for mixed modules so packed detection does not short-circuit valid existing paths
- adds regression tests for packed-qkv dispatch, packed Q/K/V splitting, Phi-style GQA head ratios, and mixed-contract fallback

## Validation

Validated locally on a MacBook Pro (Apple M4 Pro, 24 GB unified memory).

End-to-end model validation:
- served `microsoft/Phi-4-mini-instruct` successfully on Apple Silicon with Metal paged attention
- confirmed successful text generation through the OpenAI-compatible chat completions endpoint

Regression / unit validation:
- ran `env -u SSLKEYLOGFILE pytest -q tests/test_attention_sdpa.py tests/test_attention_dispatch.py -k 'not slow'`
- result: `23 passed, 1 deselected`

The added tests specifically cover:
- dispatch detection for Phi3/Phi4-style packed attention
- packed Q/K/V splitting into the SDPA layout
- Phi-style GQA head ratios (`n_heads != n_kv_heads`)
- fallback behavior when a module exposes packed fields but should still match the split-projection SDPA contract
- adds regression tests for packed-qkv dispatch, packed Q/K/V splitting, Phi-style GQA head ratios, and mixed-contract fallback

## Notes

This PR intentionally scopes support to the Phi3/Phi4-style packed `qkv_proj` attention path rather than claiming support for the entire Phi family, since other Phi variants in `mlx_lm` use different attention module layouts.

If there are other Phi-family checkpoints or additional follow-up validations that would be useful, I’m happy to follow up in this PR or in a separate one.

PR #317, opened 2026-04-30
docs: add Phi-4-mini-instruct to supported models table

szfmsmdx:
Refs #314

Add `microsoft/Phi-4-mini-instruct` to the supported-models table,
following the suggestion in #314.


LxYuan0420 (approved):
Thanks for updating the docs

Issue #319, opened 2026-04-30
[RFC]: Vision-Language Model Support for vllm-metal


scyyh11:
## Motivation

`vllm-metal` 0.2.0 serves text-only LLMs through `mlx-lm` on the paged-varlen Metal kernel. The repo already carries VLM scaffolding — `MultimodalMode` config, `_is_vlm` runner state, the `text_model(...)` adapter, an `mlx-vlm>=0.4.0` dependency — but the vision encoder is intentionally bypassed at runtime.

This RFC proposes serving **Qwen3-VL-4B** end-to-end on the existing paged-varlen path, mirroring the `vllm_metal/stt/` scaffold and reusing upstream vLLM utilities wherever they are not Triton/CUDA-coupled. The vision tower reuses `mlx-vlm`'s `vision_tower` directly in v1; a dedicated Metal ViT kernel is postponed to a future, benchmark-gated step.

## Proposed Change

### Scope

- **In**: Qwen3-VL-4B on the paged-varlen kernel; new `vllm_metal/vlm/` package; `mlx-vlm==0.4.4` pin during Phase 1–4. ViT path reuses `mlx-vlm`'s fused attention.
- **Out**: any second VLM in v1 (Phase 5+); legacy non-paged path; cross-restart encoder-cache resume; Metal-local VLM prefix cache; custom Metal ViT kernel in this RFC.

### Design

Three extensions to the existing paged forward:

1. **Encoder cache** in `vllm_metal/vlm/`, lifecycle scheduler-driven: register new request metadata, free → encode before the forward, then drop per-request maps after. No local LRU.
2. **Range-based embedding gather** assembles one packed input-embeddings tensor per step — text positions filled by `embed_tokens`, multimodal positions by encoder-cache slices intersected with the chunk's `[query_start, query_end)`. Mirrors upstream `embed_input_ids(..., multimodal_embeddings=, is_multimodal=)`.
3. **Per-request M-RoPE position cache** computes `(3, T_full)` once from `mm_features` and slices per chunk; decode rows reuse a per-request `mrope_position_delta`. Threaded through the existing `apply_packed_rope` call site.

[...]

PR #330, opened 2026-05-02
[Multimodal] Add generic helpers and Qwen3-VL adapter

ricky-chaoju:
This PR adds the first helper layer for v1 multimodal image support on Metal.

It does not add model execution or new Metal kernels yet. The goal is to establish the boundary between shared and model-specific ownership first, so the follow-up Qwen3.5-4B image path can stay small and reviewable.

## What this adds

- Add model-neutral multimodal helpers under `vllm_metal/multimodal`.
- Add MLX-compatible multimodal embedding merge logic, mirroring vLLM's
  `_merge_multimodal_embeddings` helper.
- Add v1 runtime multimodal cache plumbing under `vllm_metal/v1/mm`.
- Add `Qwen3VLMultimodalAdapter` under `vllm_metal/multimodal/qwen3_vl/adapter.py`, with
  `get_mrope_input_positions()` as the single public Qwen3-VL-specific entrypoint.
- Use upstream vLLM's `Qwen3VLForConditionalGeneration._get_mrope_input_positions` as the source
  of truth for image-only M-RoPE.
- Make `vllm_metal.v1` lazy so focused helper imports do not initialize the worker path.
- Add focused tests for feature aliases, embedding merge behavior, encoder cache bookkeeping,
  Qwen3-VL image M-RoPE positions, and import boundaries.


WindChimeRan:
cc @scyyh11

scyyh11:
I'm a bit confused, in #319 you mentioned running Qwen3.5-4B end-to-end, but this PR's M-RoPE helper imports Qwen2.5-VL's `get_mrope_input_positions`. Is that intentional? 

If the model is Qwen3.5-4B, I think we should switch to `Qwen3VLForConditionalGeneration._get_mrope_input_positions`.

ricky-chaoju:
> I'm a bit confused, in #319 you mentioned running Qwen3.5-4B end-to-end, but this PR's M-RoPE helper imports Qwen2.5-VL's `get_mrope_input_positions`. Is that intentional?
> 
> If the model is Qwen3.5-4B, I think we should switch to `Qwen3VLForConditionalGeneration._get_mrope_input_positions`.

The Qwen2.5-VL helper was only used because the image-only M-RoPE path covered in this PR matches the behavior we need, but I agree that Qwen3-VL is the clearer source of truth for a Qwen3.5-4B-scoped PR.

tests/v1/multimodal/test_embeddings.py
```diff
@@ -0,0 +1,93 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for multimodal embedding splice helpers."""
+
+from __future__ import annotations
+
+import mlx.core as mx
+import pytest
+
+from vllm_metal.v1.multimodal import merge_multimodal_embeddings
```
scyyh11: This import currently goes through `vllm_metal.v1.__init__`, which eagerly imports `MetalWorker`. That means focused multimodal helper tests pull in the full worker/vLLM/Metal stack during collection. Can we make `vllm_metal.v1` lazy, or otherwise keep these helper modules importable without initializing the worker path?

scyyh11:
Can we decide the package boundary before adding more helpers?                                                                    

`v1.multimodal.mrope.get_mrope_input_positions` reads as generic v1 plumbing, but it is Qwen3-VL/Qwen3.5-specific. I'd prefer keeping generic MM pieces separate from per-model helpers (which follows the pattern of upstream vLLM), e.g.:
```
vllm_metal/multimodal/      # inputs, embedding merge
vllm_metal/models/qwen3_vl.py
vllm_metal/v1/mm/           # encoder cache, generic runner/rope plumbing
```
A more compact alternative is vllm_metal/multimodal/qwen3_vl.py, but either way the Qwen3-specific shim shouldn't have such a generic name.

LxYuan0420:
I would like to settle the ownership boundary before reviewing this helper layer line by line.

Right now the generic `v1.multimodal` package is starting to expose Qwen3-specific behavior through generic names, especially M-RoPE and image-grid helpers. Please keep shared v1 multimodal code limited to model-neutral plumbing like encoder cache, embedding splice, and runner-facing orchestration data. Qwen3.5/Qwen3-VL-specific behavior should live behind a model-owned helper or adapter with one small public entrypoint.

For readability, please keep the public and common path first, and push internal validation, shim, and detail helpers lower in the file. The follow-up PRs will touch runner and lifecycle code, so this first layer should establish a clean ownership pattern rather than a collection of helper functions.

After that boundary is cleaned up, I am happy to review the behavior and tests in detail.

LxYuan0420 (commented):
Please settle the ownership boundary before deeper review. This should follow the same principle as the STT runtime adapters: model-specific execution details belong with the model adapter while shared runner code stays thin.

ricky-chaoju:
Thanks, agreed. I cleaned this up and updated the PR description to reflect the new boundary.

The ownership split is now:

- `vllm_metal/multimodal/`: model-neutral feature aliases and embedding splice helper.
- `vllm_metal/v1/mm/`: v1 runtime cache/plumbing.
- `vllm_metal/multimodal/qwen3_vl/adapter.py`: Qwen3-VL-specific M-RoPE behind
  `Qwen3VLMultimodalAdapter.get_mrope_input_positions()`.

I also removed the generic `v1.multimodal` surface, made `vllm_metal.v1` lazy so helper imports do not pull in `MetalWorker`, and kept the public/common path before validation, upstream shim, and data-access details.

vllm_metal/multimodal/qwen3_vl/adapter.py
```diff
@@ -0,0 +1,137 @@
...
+
+        torch_positions, mrope_position_delta = (
+            _qwen3_vl_cls()._get_mrope_input_positions(
+                input_tokens=input_tokens,
+                mm_features=mm_features,
+                config=_qwen3_vl_image_config(self._spatial_merge_size),
+            )
+        )
+        llm_positions = torch_positions.cpu().numpy()
+        return mx.array(llm_positions, dtype=mx.int32), int(mrope_position_delta)
```
LxYuan0420, line 51: The package boundary is better now, but this file still does not have a clean ownership shape. `Qwen3VLMultimodalAdapter` owns one public method while validation, config building, and feature decoding live as module-level private functions. Please move the Qwen3-specific implementation details behind private methods on the adapter and keep one public entrypoint.

vllm_metal/multimodal/qwen3_vl/adapter.py
```diff
@@ -0,0 +1,137 @@
...
+    try:
+        value = data[key]
+    except KeyError as exc:
+        raise ValueError(f"Multimodal feature data is missing {key}.") from exc
+    return getattr(value, "data", value)
+
+
+def _grid_thw(data: Any, key: str) -> tuple[int, int, int]:
+    raw_values = _feature_value(data, key)
+    values = raw_values.tolist() if hasattr(raw_values, "tolist") else raw_values
```
LxYuan0420: Please trust the upstream multimodal contract more here. Upstream Qwen3-VL reads `feature.data["image_grid_thw"].data.tolist()` directly, so this adapter should not widen the accepted surface with `getattr(..., "data", value)` and `hasattr(..., "tolist")`. If we validate, validate the known upstream shape, not arbitrary wrapper shapes.

scyyh11:
@LxYuan0420 hi, would you mind taking a look at #319? That RFC covers the same scope as this PR, but with a more complete implementation plan. The main difference is the choice of initial model, I went with Qwen2.5-VL while this PR uses Qwen3.5-4B. I've already synced with @ricky-chaoju to collaborate on this, so if the RFC looks viable, we could follow it for a clearer implementation path.

tests/multimodal/test_qwen3_vl.py
```diff
@@ -0,0 +1,230 @@
...
+    grid_thw: tuple[int, ...],
+    modality: str = "image",
+) -> MultiModalFeatureSpec:
+    key = "video_grid_thw" if modality.startswith("video") else "image_grid_thw"
+    return MultiModalFeatureSpec(
+        data={key: SimpleNamespace(data=np.array(grid_thw))},
+        modality=modality,
+        identifier=f"{modality}-{offset}",
+        mm_position=PlaceholderRange(offset=offset, length=length),
+    )
```
LxYuan0420, line 55: The test fixture is using fake `dict + SimpleNamespace(data=...)` feature data instead of the upstream multimodal field shape. That makes the production adapter support test-only shapes and encourages the defensive reflection above. Please build test features using the real upstream wrapper shape or real processor output.

tests/multimodal/test_qwen3_vl.py
```diff
@@ -0,0 +1,230 @@
...
+        tokens,
+        [feature],
+    )
+
+    assert positions.tolist() == [
+        [0, 1, 2, 3, 4, 4, 4, 4],
+        [0, 1, 2, 3, 4, 4, 5, 5],
+        [0, 1, 2, 3, 4, 5, 4, 5],
+    ]
+    assert delta == -2
```
LxYuan0420: These tests are too long and flat for a new helper module. Please group the adapter tests, reuse shared builders, and parameterize the repeated image-placement and validation cases. The current shape works, but it is not a maintainable test surface for follow-up Qwen3.5 multimodal work.

vllm_metal/v1/mm/encoder_cache.py
```diff
@@ -0,0 +1,38 @@
...
+    def add_request(
+        self, req_id: str, mm_features: list[MultiModalFeatureSpec]
+    ) -> None:
+        self.mm_features[req_id] = mm_features
+
+    def remove_request(self, req_id: str) -> None:
+        self.mm_features.pop(req_id, None)
+
+    def reset_mm_cache(self) -> None:
+        """Clear profiling-only multimodal cache state when needed."""
```
LxYuan0420: If `reset_mm_cache()` exists only for upstream interface parity, please make that explicit and mirror upstream with an intentional `pass` or TODO. Right now the docstring says it clears state, but the method is a silent no-op.

vllm_metal/multimodal/qwen3_vl/adapter.py
```diff
@@ -0,0 +1,137 @@
...
+    # static upstream helper can be reused without constructing a full model.
+    return SimpleNamespace(
+        video_token_id=-1,
+        vision_start_token_id=-1,
+        vision_end_token_id=-1,
+        vision_config=SimpleNamespace(spatial_merge_size=spatial_merge_size),
+    )
+
+
+# === Feature Data Access ===
```
LxYuan0420: Please drop the banner comments. Ordering, naming, and class/private method structure should make the file readable without section headers like `# === Internal Validation ===`.

tests/multimodal/test_embeddings.py
```diff
@@ -0,0 +1,93 @@
...
+def test_merge_does_not_mutate_input() -> None:
+    inputs = mx.array([[[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]]], dtype=mx.float32)
+    original = inputs.tolist()
+    mm_embeds = mx.array([[9.0, 10.0]], dtype=mx.float32)
+    mask = mx.array([False, True, False])
+
+    output = merge_multimodal_embeddings(inputs, mm_embeds, mask)
+
+    assert output.tolist() == [[[1.0, 2.0], [9.0, 10.0], [5.0, 6.0]]]
+    assert inputs.tolist() == original
```
LxYuan0420, line 113: `merge_multimodal_embeddings()` accepts a non-empty sequence of embedding arrays, but the tests only cover a single array or empty list. Please add one case with multiple embedding blocks so the flattening branch is covered.

LxYuan0420 (changes requested):
The package split is much better now, thanks for addressing that. I still think this needs another cleanup pass before merge. The remaining concerns are mostly around implementation ownership and test maintainability: the Qwen3 adapter should own more of its implementation details, the tests should use upstream-shaped multimodal features, and the main Qwen3 test file could be easier to maintain.

LxYuan0420:
> @LxYuan0420 hi, would you mind taking a look at #319? That RFC covers the same scope as this PR, but with a more complete implementation plan. The main difference is the choice of initial model, I went with Qwen2.5-VL while this PR uses Qwen3.5-4B. I've already synced with @ricky-chaoju to collaborate on this, so if the RFC looks viable, we could follow it for a clearer implementation path.

Thanks for pointing me to #319. I read through it, and I agree it should be the design reference for this work.

To keep the effort focused, I do not want two parallel “first model” tracks. Since Qwen3.5-4B is already prototyped and the current PR series is already scoped around that path, let’s use Qwen3.5-4B as the first landing model. Qwen2.5-VL can then be the next validation target to prove the abstraction is not one-model-only.

For this PR specifically, I still want the helper layer to be clean on its own before merge. The package direction is better now, but the adapter ownership and test-shape comments still need to be addressed so the later lifecycle and runner PRs are built on a clean foundation.

CC: @ricky-chaoju

vllm_metal/multimodal/qwen3_vl/adapter.py
```diff
@@ -0,0 +1,125 @@
...
+        values = cls._feature_value(data, key).tolist()
+        if len(values) != 3:
+            raise ValueError(f"{key} must contain exactly 3 values, got {values}.")
+        t, h, w = values
+        return int(t), int(h), int(w)
+
+    @staticmethod
+    def _feature_value(data: MultiModalKwargsItem, key: str) -> Any:
+        try:
+            return data[key].data
```
LxYuan0420: Since this adapter consumes upstream `MultiModalKwargsItem`, please avoid reinterpreting missing keys here. We can trust the typed upstream shape and access the field directly:

```python
@staticmethod
def _feature_value(data: MultiModalKwargsItem, key: str) -> Any:
    return data[key].data
```

If this needs a user-facing validation error, that should happen at the multimodal parsing boundary, not inside the model adapter.

vllm_metal/multimodal/embeddings.py
```diff
@@ -0,0 +1,76 @@
...
+"""Embedding splice helpers for multimodal placeholders."""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import mlx.core as mx
+import numpy as np
+
+# === Public Helpers ===
```
LxYuan0420: Please drop the section banner comments. The file is short enough that public function first and private helpers below already makes the structure clear.

LxYuan0420 (approved):
Two minor comments and we are good to go

scyyh11:
Post-merge follow-up on #330.

`vllm_metal/multimodal/embeddings.py`: mutation contract is ambiguous. `merge_multimodal_embeddings()` has three disagreeing sources of truth on whether `inputs_embeds` is mutated:

- **docstring**: callers should not rely on `inputs_embeds` being mutated
- **impl**: `flat = inputs_embeds.reshape(...)` + `flat[positions] = ...`, so non-mutation depends on MLX reshape/aliasing semantics
- **test** `test_merge_does_not_mutate_input`: asserts non-mutation

The current test passes, but future PR will put this on the per-step paged path. If a future MLX change makes this path share storage, callers reusing `inputs_embeds` could silently observe modified text rows.

Suggest pinning the functional contract explicitly:

```python
flat = mx.array(inputs_embeds).reshape(-1, hidden_size)  # explicit copy
```

Or, if in-place mutation is intended, update the docstring and remove the non-mutation test.

cc @ricky-chaoju @LxYuan0420

PR #332, opened 2026-05-02
Fix sample logprobs output on Metal runner

### LINKED ISSUE #331: [Bug] logprobs requests crash with 500 Internal Server Error
## Reproduction
```bash
vllm serve mlx-community/Mistral-7B-Instruct-v0.3-4bit --max-model-len 8192
```
**Completions:**
```bash
  -d '{"model":"mlx-community/Mistral-7B-Instruct-v0.3-4bit",
       "prompt":"The capital of France is","max_tokens":10,"logprobs":5}'
```
**Chat:**
```bash
curl http://localhost:8000/v1/chat/completions \
  -d '{"model":"mlx-community/Mistral-7B-Instruct-v0.3-4bit",
       "messages":[{"role":"user","content":"Hi"}],
       "max_tokens":10,"logprobs":true,"top_logprobs":5}'
```
Both return `500 Internal Server Error`. Without `logprobs`, both work fine.

## Error Log
**Completions:**
```
File "vllm/entrypoints/openai/completion/serving.py", line 591, in _create_completion_logprobs
    step_top_logprobs = top_logprobs[i]
                        ~~~~~~~~~~~~^^^
IndexError: list index out of range
```
**Chat:**
```
File "vllm/entrypoints/openai/chat_completion/serving.py", line 1722, in _create_chat_logprobs
    step_top_logprobs = top_logprobs[i]
                        ~~~~~~~~~~~~^^^
IndexError: list index out of range
```

## Tested Models
| Model | `logprobs=5` | `top_logprobs=5` | no logprobs |
|-------|:---:|:---:|:---:|
| Mistral-7B-Instruct-v0.3-4bit | 500 | 500 | 200 |
| Meta-Llama-3.1-8B-Instruct-4bit | 500 | 500 | 200 |

## Root Cause

vllm-metal's `MetalModelRunner` never computes logprobs:

1. `sampling_batch.py:193` — `max_num_logprobs=None` is hardcoded, so the vLLM Sampler never computes logprobs even when the user requests them
2. `model_runner.py` — all three `ModelRunnerOutput` returns set `logprobs=None` and `prompt_logprobs_dict={}`
3. Upstream vLLM's serving layer expects logprobs data when the user requests it, gets an empty list, and crashes on index access

## Environment

- M4 Mac mini, 16GB
- vllm-metal v0.2.0 (latest main, commit `1b0496e`)
- MacOS 26.3
## Impact

Any API request with `logprobs` parameter crashes the server. This affects any OpenAI-compatible client that requests logprobs.

ricky-chaoju:
The Metal runner previously sampled tokens but always returned `logprobs=None` in `ModelRunnerOutput`, even when
requests asked for `logprobs` / `top_logprobs`. The OpenAI serving layer then attempted to index missing per-token
logprobs and returned `500 Internal Server Error`.

This PR:
- Passes requested sample logprobs into `SamplingMetadata.max_num_logprobs`.
- Routes greedy requests that need logprobs through the vLLM sampler instead of the MLX argmax fast path.
- Propagates sampler `LogprobsLists` into `ModelRunnerOutput.logprobs`.
- Preserves output-slot alignment for paged prefill/decode batches.
- Keeps `prompt_logprobs` out of scope.

Fixes #331

### DIFF
--- a/tests/test_sampling.py
+++ b/tests/test_sampling.py
@@ -5,8 +5,11 @@ import mlx.core as mx
 import numpy as np
 import pytest
 import torch
+from vllm.outputs import RequestOutput
 from vllm.sampling_params import SamplingParams
 from vllm.utils.torch_utils import make_tensor_with_pad
+from vllm.v1.engine import EngineCoreOutput, EngineCoreRequest, FinishReason
+from vllm.v1.engine.output_processor import OutputProcessor
 from vllm.v1.outputs import LogprobsLists
 from vllm.v1.sample.logits_processor import LogitsProcessors
 from vllm.v1.sample.metadata import SamplingMetadata
@@ -388,6 +391,52 @@ class TestV1SamplingBatch:
         assert output.logprobs is not None
         assert output.logprobs.logprob_token_ids.shape == (2, 3)
         assert output.logprobs.slice_request(1, 1).logprob_token_ids[0, 0] == 7
+
+    def test_logprobs_flow_through_request_output_processing(self) -> None:
+        sampling_params = SamplingParams(
+            temperature=0.0,
+            logprobs=2,
+            max_tokens=1,
+            detokenize=False,
+        )
+        request = EngineCoreRequest(
+            request_id="req-1",
+            prompt_token_ids=[1, 2, 3],
+            mm_features=None,
+            sampling_params=sampling_params,
+            pooling_params=None,
+            arrival_time=0.0,
+            lora_request=None,
+            cache_salt=None,
+            data_parallel_rank=None,
+            external_req_id="req-1",
+        )
+        output_processor = OutputProcessor(tokenizer=None, log_stats=False)
+        output_processor.add_request(request, prompt=None)
+
+        result = output_processor.process_outputs(
+            [
+                EngineCoreOutput(
+                    request_id="req-1",
+                    new_token_ids=[7],
+                    new_logprobs=LogprobsLists(
+                        logprob_token_ids=np.array([[7, 5, 3]], dtype=np.int32),
+                        logprobs=np.array([[-0.1, -1.0, -2.0]], dtype=np.float32),
+                        sampled_token_ranks=np.array([1], dtype=np.int32),
+                    ),
+                    finish_reason=FinishReason.LENGTH,
+                )
+            ]
+        )
+
+        assert not result.reqs_to_abort
+        [request_output] = result.request_outputs
+        assert isinstance(request_output, RequestOutput)
+        [completion_output] = request_output.outputs
+        assert completion_output.token_ids == [7]
+        assert completion_output.logprobs is not None
+        assert completion_output.logprobs[0] is not None
+        assert completion_output.logprobs[0][7].logprob == pytest.approx(-0.1)
 
 
 class TestV1SamplingMetadataLogitsProcessors:
--- a/tests/test_sampling.py
+++ b/tests/test_sampling.py
@@ -7,6 +7,7 @@ import pytest
 import torch
 from vllm.sampling_params import SamplingParams
 from vllm.utils.torch_utils import make_tensor_with_pad
+from vllm.v1.outputs import LogprobsLists
 from vllm.v1.sample.logits_processor import LogitsProcessors
 from vllm.v1.sample.metadata import SamplingMetadata
 from vllm.v1.sample.sampler import Sampler
@@ -17,8 +18,9 @@ from vllm_metal.v1.model_runner import (
     MetalModelRunner,
     RequestState,
     _create_request_generator,
+    _ExecutionBatch,
 )
-from vllm_metal.v1.sampling_batch import SamplingBatch
+from vllm_metal.v1.sampling_batch import SamplingBatch, sample_from_logits
 
 VOCAB_SIZE = 1024
 MAX_NUM_PROMPT_TOKENS = 64
@@ -322,6 +324,9 @@ class TestV1SamplingBatch:
         assert not SamplingBatch.can_use_native_greedy(
             [SamplingParams(temperature=0.0, repetition_penalty=1.1)]
         )
+        assert not SamplingBatch.can_use_native_greedy(
+            [SamplingParams(temperature=0.0, logprobs=1)]
+        )
 
     def test_can_use_native_greedy_requires_every_request_to_match(self) -> None:
         assert SamplingBatch.can_use_native_greedy(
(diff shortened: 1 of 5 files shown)


vllm_metal/v1/sampling_batch.py
```diff
@@ -19,6 +22,28 @@
...
+    num_logprobs = getattr(sampling_params, "num_logprobs", None)
+    if num_logprobs is not None:
+        return num_logprobs
+
+    logprobs = getattr(sampling_params, "logprobs", None)
+    if logprobs is not None:
+        return logprobs
+
+    logprob_token_ids = getattr(sampling_params, "logprob_token_ids", None)
+    return len(logprob_token_ids) if logprob_token_ids else None
```
LxYuan0420: This PR should stay scoped to ordinary sample `logprobs`. `logprob_token_ids` is a separate upstream `SamplingParams` surface, but the Metal runner does not support it end to end here. Please remove this path from this PR and read `sampling_params.logprobs` directly instead of using `getattr` compatibility probing in typed internal code. If we want `logprob_token_ids`, it should be a separate PR with focused end-to-end coverage.

vllm_metal/v1/sampling_batch.py
```diff
@@ -86,6 +111,35 @@ def __init__(
...
+            if (num_logprobs := _num_sample_logprobs(sampling_params)) is not None
+        ]
+        if not requested:
+            return None
+        if any(num_logprobs == -1 for num_logprobs in requested):
+            # vLLM's generic Sampler returns full-vocab logprobs for -1
+            # without token ids/ranks. Ask for top-vocab instead so the engine
+            # receives the same token-id/rank shape as ordinary sample logprobs.
+            return self.vocab_size
+        return max(requested)
```
LxYuan0420: Please do not hide `logprobs=-1` behavior inside `max_num_logprobs`. Upstream treats `-1` as full-vocab logprobs, while this rewrites it into a `vocab_size` top-k request with the sampled token duplicated. If this is the intended Metal representation, make it explicit and add a focused test through `LogprobsProcessor` / request-output handling. Otherwise keep this PR scoped to normal non-negative sample logprobs.

tests/test_sampling.py
```diff
@@ -334,6 +339,56 @@ def test_can_use_native_greedy_requires_every_request_to_match(self) -> None:
...
+                logprobs=np.array([[-0.1, -0.1, -2.0]], dtype=np.float32),
+                sampled_token_ranks=np.array([1], dtype=np.int32),
+            ),
+        )
+
+        output = MetalModelRunner._build_output(batch)
+
+        assert output.logprobs is not None
+        assert output.logprobs.logprob_token_ids.shape == (2, 3)
+        assert output.logprobs.slice_request(1, 1).logprob_token_ids[0, 0] == 7
```
LxYuan0420, line 426: Please add one regression that exercises the actual #331 output path. The current tests cover sampling metadata and row assembly, but they do not prove that generated `ModelRunnerOutput.logprobs` flows through vLLM’s request-output logprobs processing without raising.

vllm_metal/v1/sampling_batch.py
```diff
@@ -218,24 +274,93 @@ def sample_from_logits(
...
+        token_rows.append(token_ids.astype(np.int32, copy=False))
+        logprob_rows.append(logprobs.astype(np.float32, copy=False))
+        rank_rows.append(row.sampled_token_ranks.astype(np.int32, copy=False))
+
+    return LogprobsLists(
+        logprob_token_ids=np.concatenate(token_rows, axis=0),
+        logprobs=np.concatenate(logprob_rows, axis=0),
+        sampled_token_ranks=np.concatenate(rank_rows, axis=0),
+        cu_num_generated_tokens=None,
+    )
```
LxYuan0420: The new sampling/logprob behavior is too scattered across module-level helpers. Please park this under clear owners: `SamplingBatch` should own request-logprobs policy and sampler metadata, while output-slot logprob assembly should belong to `_ExecutionBatch` or a small output-builder class. Free functions like this make the sampling/output contract harder to reason about as the runner grows.

vllm_metal/v1/sampling_batch.py
```diff
@@ -19,6 +22,28 @@
...
+    logprob_token_ids = getattr(sampling_params, "logprob_token_ids", None)
+    return len(logprob_token_ids) if logprob_token_ids else None
+
+
+@dataclass(frozen=True)
+class SamplingResult:
+    """Sampled token ids plus optional vLLM logprobs rows."""
+
+    token_ids: list[int]
+    logprobs: LogprobsLists | None = None
```
LxYuan0420, line 30: If this DTO is only used inside this module, please make it private as `_SamplingResult`. `SamplingResult` looks like a public module API, but this is internal sampling plumbing.

LxYuan0420 (commented):
Great work as always; few things to tighten up before we merge

ricky-chaoju:
Thanks. I narrowed the scope to regular sample logprobs, moved the assembly logic under clearer owners, and added a regression through the OutputProcessor path.

vllm_metal/v1/sampling_batch.py
```diff
@@ -86,6 +97,71 @@ def __init__(
...
 
+    @property
+    def max_num_logprobs(self) -> int | None:
+        """Return the batch-wide sample-logprobs request, if any."""
+        requested = [
+            sampling_params.logprobs
+            for sampling_params in self.sampling_params_list
+            if sampling_params.logprobs is not None
+        ]
+        return max(requested) if requested else None
```
LxYuan0420, line 110: `logprobs=-1` is still silently under-served here. In a mixed batch like `logprobs=[-1, 2]`, `max(requested)` becomes `2`, so the full-vocab request is downgraded to top-k logprobs. Since this PR is scoped to ordinary sample logprobs, please reject `-1` explicitly for now:

```python
if any(num_logprobs == -1 for num_logprobs in requested):
    raise NotImplementedError(
        "Metal runner does not support logprobs=-1 yet"
    )
```

Then `return max(requested) if requested else None`.

vllm_metal/v1/model_runner.py
```diff
@@ -143,20 +145,85 @@ class _ExecutionBatch:
...
+        self,
+        output_idx: int,
+        token_ids: list[int],
+        logprobs: LogprobsLists | None = None,
+    ) -> None:
+        """Set tokens and logprobs for an existing output slot."""
+        self.sampled_tokens[output_idx] = token_ids
+        self.sample_logprobs[output_idx] = logprobs
+
+    def merged_logprobs(self) -> LogprobsLists | None:
```
LxYuan0420, line 178: This duplicates the same logprob-row merge algorithm already added in `SamplingBatch.merge_logprobs_rows()`. Please keep one owner for this behavior:

```python
def merged_logprobs(self) -> LogprobsLists | None:
    return SamplingBatch.merge_logprobs_rows(self.sample_logprobs)
```

This path is subtle serving-contract code, so duplicated merge logic is too easy to let drift.

tests/test_sampling.py
```diff
@@ -334,6 +342,102 @@ def test_can_use_native_greedy_requires_every_request_to_match(self) -> None:
             [SamplingParams(temperature=0.0), SamplingParams(temperature=0.8, top_k=4)]
         )
 
+    def test_sampling_metadata_uses_requested_logprobs(self) -> None:
```
LxYuan0420, line 345: Please add focused coverage for `logprobs=-1`. The new tests cover `logprobs=2` and `5`, but the current broken contract is `-1`. Since this PR should not support full-vocab logprobs yet, add tests that assert `logprobs=-1` is rejected and that a mixed batch like `[-1, 2]` does not silently downgrade to `2`.

LxYuan0420 (commented):
Few things to tidy up

ricky-chaoju:
All done.

LxYuan0420 (approved):

PR #337, opened 2026-05-04
[Bugfix] preserve ByteLevel tokenizer decoder

### LINKED ISSUE #336: [Bug] DeepSeek-R1-Distill-Qwen-1.5B-4bit leaks token-piece text (`Ġ` / `Ċ`) in served output
## Summary

`mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit` leaks token-piece text such as `Ġ` / `Ċ` in the final served output on vllm-metal.

On my setup, this reproduces from both `/v1/completions` and `/v1/chat/completions`.

A/B testing with:
- the tokenizer bundled with `mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit`
- `--tokenizer deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B`

produced identical leaked output, so this does not seem specific to the explicit tokenizer override.

This was observed while validating `mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit` in PR #316, following the earlier discussion in #289.

## Environment

- Hardware: Apple M1 MacBook Air, 16 GB unified memory
- macOS: 14.2.1
- vLLM: `0.20.0+cpu`
- mlx-lm: `0.31.3`
- mlx: `0.31.2`
- mlx-metal: `0.31.2`

### A. Use the tokenizer bundled with the MLX model repo

Server command:

```bash
vllm serve mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit \
  --port 8016 \
  --max-model-len 32768
```

Chat request:

```bash
curl -s http://127.0.0.1:8016/v1/chat/completions \
  -d '{
    "model": "mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit",
    "messages": [
      {"role": "user", "content": "What is 1 + 1? Answer with only the number."}
    ],
    "max_tokens": 50,
    "temperature": 0
  }' | jq -r '.choices[0].message.content'
```

Output:

```bash
ĊFirst,ĠIĠneedĠtoĠdetermineĠtheĠvalueĠofĠ1ĠplusĠ1.ĊĊAddingĠ1ĠandĠ1ĠtogetherĠgivesĠ2.ĊĊTherefore,ĠtheĠfinalĠanswerĠisĠ2.Ċ</think>ĊĊSure!ĠLet'sĠsolveĠtheĠproblemĠstepĠbyĠstep.ĊĊ**
```

### B. Force the upstream DeepSeek tokenizer

Server command:

```bash
vllm serve mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit \
  --tokenizer deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B \
  --port 8016 \
  --max-model-len 32768
```

Chat request:

```bash
curl -s http://127.0.0.1:8016/v1/chat/completions \
  -d '{
    "model": "mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit",
    "messages": [
      {"role": "user", "content": "What is 1 + 1? Answer with only the number."}
    ],
    "max_tokens": 50,
    "temperature": 0
  }' | jq -r '.choices[0].message.content'
```

Output:

```bash
ĊFirst,ĠIĠneedĠtoĠdetermineĠtheĠvalueĠofĠ1ĠplusĠ1.ĊĊAddingĠ1ĠandĠ1ĠtogetherĠgivesĠ2.ĊĊTherefore,ĠtheĠfinalĠanswerĠisĠ2.Ċ</think>ĊĊSure!ĠLet'sĠsolveĠtheĠproblemĠstepĠbyĠstep.ĊĊ**
```

### `/v1/completions` also reproduces

Request:

```bash
curl -s http://127.0.0.1:8016/v1/completions \
  -d '{
    "model": "mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit",
    "prompt": "Hello, how are you?",
    "max_tokens": 50,
    "temperature": 0,
    "return_token_ids": true
  }' | jq -r '.choices[0].text'
```

Outout:

```bash
ĠI'mĠtryingĠtoĠunderstandĠsomeĠmathĠconcepts.ĠIĠcameĠacrossĠthisĠproblem:Ġ"FindĠallĠrealĠnumbersĠxĠsuchĠthatĠx^2Ġ+Ġ1Ġ=Ġ0."ĠHmm,Ġokay,ĠsoĠIĠneedĠtoĠsolveĠforĠxĠinĠtheĠequationĠxĠsquaredĠplus
```

### Comparison with direct `mlx_lm`

On the same machine, direct `mlx_lm` generation with the same model produces semantically odd text, but still clean decoded text without `Ġ` / `Ċ` leakage:

```bash
python - <<'PY'
from mlx_lm import load, generate

model, tokenizer = load("mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit")

prompts = [
    "Hello, how are you?",
    "What is 1 + 1? Answer with only the number."
]

for i, prompt in enumerate(prompts, 1):
    print(f"=== Prompt {i} ===")
    print(prompt)
    print("--- Output ---")
    out = generate(model, tokenizer, prompt=prompt, max_tokens=50)
    print(out)
    print()
PY
```

Output on my side:

```bash
=== Prompt 1 ===
Hello, how are you?
--- Output ---
I'm trying to solve this problem: Given a positive integer n, find all positive integers x such that x^2 + x + 1 divides n^2 + n + 1. I need to find all x for which this is true
```

```bash
=== Prompt 2 ===
What is 1 + 1? Answer with only the number.
--- Output ---
Do not use any words, symbols, or
........................................................................................................................
```

## Expected behavior

The final served output should be clean decoded text, without token-piece leakage such as `Ġ` / `Ċ`.

## Actual behavior

The final served output from both `/v1/completions` and `/v1/chat/completions` leaks token pieces such as `Ġ` / `Ċ`, regardless of whether the bundled MLX tokenizer or the explict upstream DeepSeek tokenizer is used.

ricky-chaoju:
Fixes #336.

Some MLX-community Qwen/DeepSeek model repos ship a ByteLevel `tokenizer.json`, but their tokenizer config can make newer `transformers` versions instantiate a Llama/SentencePiece-style decoder. When that happens, vLLM serving returns visible ByteLevel token pieces such as `Ġ` and `Ċ` instead of decoded spaces/newlines.

This patch adds a narrow compatibility shim during vllm-metal startup:

- if the loaded tokenizer already uses a ByteLevel decoder, leave it unchanged
- if `tokenizer.json` declares a ByteLevel decoder but the loaded tokenizer does not, rebuild a fast tokenizer from `tokenizer.json`
- preserve common tokenizer metadata such as special tokens, chat template, model max length, and cleanup settings

I tested the same local model files on two Apple Silicon environments:

- `transformers==5.5.4`: `AutoTokenizer.from_pretrained()` loads `TokenizersBackend` with `ByteLevel(...)`, and decoding is clean.
- `transformers==5.7.0`: the same files load as `LlamaTokenizer` with a non-ByteLevel decoder, and `AutoTokenizer.decode()` itself leaks `Ġ` / `Ċ`.

The relevant model metadata is internally inconsistent:

- `tokenizer_config.json` declares `"tokenizer_class": "LlamaTokenizer"`
- `tokenizer.json` declares a ByteLevel decoder

So the serving response is not corrupting clean text; the tokenizer decoder is already wrong before the OpenAI response is assembled.

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import builtins
 import importlib.util
+import json
 import os
 import sys
 from types import ModuleType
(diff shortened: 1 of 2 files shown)


vllm_metal/compat.py
```diff
@@ -24,10 +26,247 @@ def apply_compat_patches() -> None:
...
+    rebuilt = _rebuild_bytelevel_fast_tokenizer(tokenizer, tokenizer_json)
+    logger.info(
+        "Rebuilt tokenizer from ByteLevel tokenizer.json after transformers "
+        "loaded a non-ByteLevel decoder: %s",
+        tokenizer_json,
+    )
+    return rebuilt
+
+
+def _patch_transformers_bytelevel_tokenizer_decoder() -> None:
```
LxYuan0420: I am okay carrying a local compatibility shim for this, but not as a global `AutoTokenizer.from_pretrained()` monkeypatch. The failure path for #336 is the vllm-metal serving tokenizer load path inheriting a bad tokenizer-class choice, not every AutoTokenizer call in the process. Please scope this to the serving tokenizer-loading boundary and keep the trigger limited to `tokenizer.json` using ByteLevel while the loaded decoder is not ByteLevel.

vllm_metal/compat.py
```diff
@@ -24,10 +26,247 @@ def apply_compat_patches() -> None:
...
+        from_pretrained_kwargs,
+    )
+    if tokenizer_json is None:
+        return None
+    if not _tokenizer_json_decoder_uses_bytelevel(tokenizer_json):
+        return None
+    return tokenizer_json
+
+
+def _rebuild_bytelevel_fast_tokenizer(tokenizer: Any, tokenizer_json: Path) -> Any:
```
LxYuan0420: Please avoid rebuilding `PreTrainedTokenizerFast` by hand and manually copying tokenizer metadata. The upstream direction for this class of issue is to bypass the incorrect Hub tokenizer class and use `TokenizersBackend` directly, as in huggingface/transformers#45739. If we carry a local shim, it should follow that shape instead of creating a second metadata-preservation contract in `compat.py`.

vllm_metal/compat.py
```diff
@@ -24,10 +26,247 @@ def apply_compat_patches() -> None:
...
+    logger.info(
+        "Rebuilt tokenizer from ByteLevel tokenizer.json after transformers "
+        "loaded a non-ByteLevel decoder: %s",
+        tokenizer_json,
+    )
+    return rebuilt
+
+
+def _patch_transformers_bytelevel_tokenizer_decoder() -> None:
+    """Use tokenizer.json's ByteLevel decoder when AutoTokenizer picks wrong.
```
LxYuan0420: Please add a TODO near the shim that points to the upstream fix path and states the removal condition: once vllm-metal’s supported Transformers version handles this ByteLevel tokenizer-class mismatch upstream, this local compatibility path should be removed.

tests/test_compat.py
```diff
@@ -15,6 +15,125 @@
...
+        self.sep_token = None
+        self.cls_token = None
+        self.mask_token = None
+        self.additional_special_tokens = []
+        self.chat_template = None
+        self.model_max_length = 128
+        self.clean_up_tokenization_spaces = False
+
+
+class TestByteLevelTokenizerCompatPatch:
```
LxYuan0420, line 95: These tests only exercise private helpers with fake tokenizer objects. Please add a regression through the real tokenizer-loading boundary this PR changes: load a tokenizer whose config points at the wrong class but whose `tokenizer.json` uses ByteLevel, then assert ByteLevel pieces like `ĠHello` / `Ċ` decode cleanly and key metadata such as chat template and special tokens is preserved.

LxYuan0420 (commented):
Thanks for chasing this down. The root-cause direction looks right: #336 is a tokenizer-class / decoder mismatch, not corrupt model output. Please narrow the fix to the serving tokenizer-loading boundary, follow the upstream-style `TokenizersBackend` approach, and add a TODO/removal note.

Also lets follow the upstream Transformers fix path closely. Once our supported Transformers version handles this mismatch upstream, we should remove the local patch.

LxYuan0420 (approved):

PR #338, opened 2026-05-04
Revert "[Multimodal] Add generic helpers and Qwen3-VL adapter"

scyyh11:
Reverts vllm-project/vllm-metal#330

`vllm_metal/multimodal/embeddings.py`: mutation contract is ambiguous. `merge_multimodal_embeddings()` has three disagreeing sources of truth on whether `inputs_embeds` is mutated:

- **docstring**: callers should not rely on `inputs_embeds` being mutated
- **impl**: `flat = inputs_embeds.reshape(...)` + `flat[positions] = ...`, so non-mutation depends on MLX reshape/aliasing semantics
- **test** `test_merge_does_not_mutate_input`: asserts non-mutation

The current test passes, but future PR will put this on the per-step paged path. If a future MLX change makes this path share storage, callers reusing `inputs_embeds` could silently observe modified text rows.

Suggest pinning the functional contract explicitly:

```python
flat = mx.array(inputs_embeds).reshape(-1, hidden_size)  # explicit copy
```

Or, if in-place mutation is intended, update the docstring and remove the non-mutation test.

cc @ricky-chaoju @LxYuan0420


ricky-chaoju:
Commit 8b0f60c removes all of #330, including the Qwen3-VL adapter, encoder cache, lazy v1 import, and related tests. The mutation-contract issue is limited to `merge_multimodal_embeddings()` and should be fixed with a narrow explicit-copy change instead of reverting the whole helper layer.

PR #340, opened 2026-05-05
[quant] AWQ checkpoint support for vllm-metal

scyyh11:
## Summary

* First HF AWQ checkpoint validated end-to-end on the Metal paged-attention runner: `Qwen/Qwen2.5-1.5B-Instruct-AWQ` (greedy continuation top-1 matches the bf16 reference, `Paris`).
* `mlx-lm 0.31.3+` already provides the AWQ → MLX-affine repack via `_transform_awq_weights`; vllm-metal inherits it on the existing `mlx_lm.load` path. Support was undocumented and unvalidated. This PR adds an entry-point preflight, post-load dtype alignment for non-quantized leaves, the `supported_models.md` row, and synthetic + slow regression tests, with the entire AWQ branch encapsulated behind a dedicated `AWQQuantLoader` owner under `vllm_metal/quant/` so quantization policy stays in one cohesive place rather than leaking into the generic loader.
* No new Metal kernel; no `mlx_lm.models.*` modifications; no paged-attention or sampling changes.

## Background

Today's `mlx_lm.load` path on AWQ checkpoints loads and produces fluent text without any vllm-metal changes, but with two integration gaps:

1. mlx-lm derives runtime dtype from `scales.dtype` (typically fp16 in published Qwen / Llama AWQ). The rest of the engine (KV cache, sampler) runs at `model_config.dtype` (bf16 for Qwen2.5). Layernorms, embeddings, and the regular `bias` parameter on quantized linears (Qwen2 q/k/v projections, MoE per-expert biases) end up fp16, creating a mixed-dtype path into the bf16 KV cache.
2. AutoAWQ-format checkpoints with aliases (`w_bit`, `q_group_size`, uppercase `"GEMM"`) and unsupported variants (`gemv`, `bits != 4`, `group_size != 128`, `zero_point=False`) flow through to mlx-lm without a clear vllm-metal-side error.

## Implementation

* **`vllm_metal/quant/awq_loader.py` — `AWQQuantLoader`**, the single owner of the AWQ load contract. `for_model` runs Hub config detection (with mlx-lm's `text_config.quantization_config` fallback for wrapper / multimodal configs) and returns `None` for non-AWQ checkpoints; **GPTQ checkpoints raise `UnsupportedQuantizationConfigError` at the entry point** rather than silently flowing through, since GPTQ is not part of the v1 support claim and the loader must not silently take ownership of an unvalidated quant method. `load` invokes `mlx_lm.load(model_config={"quantization_config": ...})` with the normalized config and runs post-load dtype alignment. `cache_key` is dtype-scoped so a model first loaded as bf16 cannot be served as fp16 from cache. Hub-fetch errors (`HfHubHTTPError`, `HFValidationError`, `OSError`) propagate to the caller rather than silently demoting an AWQ checkpoint to the generic loader path on a transient Hub failure.
* **`vllm_metal/quant/awq_config.py` — `normalize_quant_config(raw)`**. Canonicalizes aliases (`w_bit` → `bits`, `q_group_size` → `group_size`, uppercase `GEMM` → `gemm`) and rejects unsupported v1 variants with a clear error before any model state is constructed. Accepts `quant_method="awq"` only; symmetric quantization (`zero_point=false`) is rejected. Raises `UnsupportedQuantizationConfigError` naming the offending field and value.
* **Post-load dtype alignment**, owned by `AWQQuantLoader` as a private static helper. Casts every floating param outside of `scales` / `biases` (the AWQ-transform output) to the runtime target dtype. Detects MLX-quantized leaves via duck-typing on the protocol's `bits` / `group_size` attributes so the alignment covers `nn.QuantizedLinear`, `nn.QuantizedEmbedding`, and the mlx_lm peers `QuantizedSwitchLinear` (MoE) and `QuantizedMultiLinear` (MLA), none of which subclass `QuantizedLinear`. The ordinary `bias` parameter on those leaves is *included* in alignment; only the AWQ-transform output buffers are exempt.
* **`vllm_metal/v1/model_lifecycle.py` — thin dispatch.** Speculatively probes the cache against both candidate keys (AWQ key first, generic fallback), then invokes `AWQQuantLoader.for_model` only on miss to avoid an HF Hub roundtrip on cache hits. Probing AWQ first means a previously cached AWQ load is still served correctly even if a current detection invocation would have failed (e.g. transient Hub error after the cache was warmed). Generic and STT cache keys stay 2-tuples; dtype scoping is encoded inside the AWQ key segment so it remains a property of the owner, not of the generic cache.

### Per-layer (Qwen2.5-0.5B-Instruct-AWQ q_proj)

* Dequant max abs err: `1.95e-3` (1 ULP at fp16 magnitude ≈ 2)
* Matmul cos similarity: `0.9999985`
* Matmul err vs naive-MLX-quantize-from-fp16 baseline: `0.037` vs `0.305` (**8.2× smaller**), confirming AWQ's per-channel scale calibration survives the repack intact.

### End-to-end vs bf16 reference (`"The capital of France is"`, 64 steps)

| model | method | logits cos | KL(ref‖q) | TF top-1 |
|---|---|---|---|---|
| 0.5B  | AWQ-repack     | **0.9835** | 0.126     | **52/64** |
| 0.5B  | naive MLX 4bit | 0.9545     | 0.676     | 51/64     |
| 1.5B  | AWQ-repack     | **0.9746** | 0.257     | **58/64** |
| 1.5B  | naive MLX 4bit | 0.9583     | 0.214     | 52/64     |

KL is reported as a diagnostic only: at 1.5B naive's KL (0.214) is smaller than AWQ's (0.257) while AWQ wins on every other signal; KL is sensitive to small tail-mass shifts and does not consistently rank quantization quality at this scale.

On 1.5B's free-greedy continuation, AWQ-repack and the bf16 reference produce **byte-identical multiple-choice option text** for ~30 tokens after a single-token punctuation difference; naive 4bit diverges into an unrelated topic.

### 7B (AWQ vs naive 4bit, no bf16 reference)

`Qwen/Qwen2.5-7B-Instruct` (~14 GB bf16) does not fit in 16 GB unified memory under MLX's eager forward path. Verified on M1 Pro that the AWQ-repack 7B loads, runs, and produces a coherent encyclopedia-style response, while naive 4bit drops into a confused self-questioning loop on the same prompt. Cos `0.9853`, KL `0.156` between the two 4bit paths. Full bf16-reference parity for 7B is gated on ≥ 32 GB hardware.

## Scope

**v1 supports**: AWQ at `bits: 4`, `group_size: 128`, `zero_point: true`, `version: "gemm"`, covering every official Qwen / Llama / Mistral AWQ release on HF. GPTQ checkpoints are explicitly rejected at the loader entry point with a clear error pointing users at AWQ; promoting GPTQ to a supported quant method is deferred to a follow-up PR once a real GPTQ checkpoint is validated end-to-end.

**Out of scope** for this PR: GGUF; bits ≠ 4; group_size ≠ 128; AWQ `version: "gemv"`; GPTQ (rejected at the loader; follow-up PR); new model architectures not in mlx_lm.

The preflight, dtype alignment, and `_transform_awq_weights` repack are architecture-agnostic. The alignment in particular is correct for MoE (`QuantizedSwitchLinear`) and MLA (`QuantizedMultiLinear`) AWQ checkpoints by virtue of the quantize-protocol duck-typing, though only Qwen2-style dense models are validated end-to-end in this PR.

## Test plan

- [x] `pytest -m "not slow" tests/`: 694 fast tests pass. `tests/quant/test_awq_loader.py` is grouped into `TestCacheKey`, `TestConfigDetection`, and `TestDtypeAlignment`, the last of which drives `AWQQuantLoader.load()` with a stubbed `mlx_lm_load` so alignment assertions hang off the public load contract. Coverage: alias normalization, `zero_point` rejection, GPTQ rejection at the loader entry point, repack parity vs AWQ ground truth, dtype-scoped cache-key isolation, Hub-fetch error propagation, and the biases-vs-bias dtype distinction across both a real `nn.QuantizedLinear` and a synthetic duck-typed quantize-protocol module
- [x] `pytest tests/test_model_lifecycle.py`, including the parametrized `test_load_dispatches_by_awq_detection` that pins the AWQ vs generic dispatch contract
- [x] `pytest -m slow tests/quant/`: 2 cases, ~20s with model cached, covering dtype alignment after load on 1.5B-AWQ (asserts `scales` / `biases` stay fp16 while ordinary `bias` is bf16) and full vLLM `LLM.generate` smoke through the paged-attention runner (asserts `Paris` appears in the greedy continuation)
- [x] `ruff check vllm_metal/ tests/`: clean
- [x] `Qwen/Qwen2.5-1.5B-Instruct-AWQ` greedy continuation matches bf16 reference top-1 on M1 Pro 16 GB / macOS 26.2

Quantitative parity scripts (the 0.5B / 1.5B / 7B tables above) live as standalone scripts and are not run in CI because they require loading multiple 1.5B models in one process; the slow smoke + dtype tests are the regression guards.


vllm_metal/v1/model_lifecycle.py
```diff
@@ -194,12 +309,29 @@ def _load_generation_model(self, model_name: str, is_vlm: bool) -> tuple[Any, An
             )
             model, tokenizer = mlx_vlm_load(model_name)
         else:
+            # AWQ/GPTQ preflight: normalize aliases and reject unsupported
+            # variants before any model state is constructed. Returns None
+            # for non-AWQ/GPTQ checkpoints (no behavior change for those).
+            awq_model_config = _maybe_normalize_awq_model_config(model_name)
```
LxYuan0420: There is useful AWQ work here, but I do not think it should land as lifecycle-owned logic. `ModelLifecycle` should stay close to the upstream loader flow: choose the loader, call it, cache the result, and assign runner state. AWQ/GPTQ-specific behavior such as config preflight, `mlx_lm.load(model_config=...)`, post-load dtype alignment, and cache-key policy should live behind one cohesive owner under `vllm_metal/quant/`.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -174,7 +281,15 @@ def load(self) -> None:
...
         logger.info("Loading model: %s (VLM: %s)", model_name, is_vlm)
         start_time = time.time()
-        cache_key = _generation_cache_key(model_name, is_vlm=is_vlm)
+        # Resolve the runtime dtype up front: it is part of the cache key
+        # and the AWQ/GPTQ post-load alignment target, so we must compute it
+        # before any cache lookup.
+        target_dtype = torch_to_mlx(
+            torch.empty(0, dtype=self._runner.model_config.dtype)
+        ).dtype
+        cache_key = _generation_cache_key(
```
LxYuan0420: The dtype-scoped cache policy should come from the AWQ/GPTQ load owner, not lifecycle directly. Right now this changes cache behavior for every text model even though the dtype-specific mutation only exists on the quantized path. Please keep that policy scoped to the quantized load plan.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -130,6 +141,102 @@ def _mlx_lm_compatible_model_path(model_name: str):
...
+    under ``text_config`` would skip the vllm-metal preflight entirely
+    while mlx_lm itself would still apply the transform.
+    """
+    model_path = Path(model_name)
+    if model_path.is_dir():
+        config_path = model_path / "config.json"
+        if not config_path.is_file():
+            return None
+    else:
+        try:
```
LxYuan0420: Please keep this boundary explicit. If Hub config fetch is part of the AWQ/GPTQ load contract, it should live on the quant-owned loader support with explicit imports and narrow errors. I do not want lazy import plus fail-open behavior in the middle of lifecycle helpers.

vllm_metal/compat.py
```diff
@@ -428,3 +429,78 @@ def _transform(self, weights):
...
+
+    For tied-embedding architectures, ``Model.lm_head`` is not a real
+    parameter; AWQ checkpoints that quantize ``lm_head`` ship a redundant
+    triple that ``mlx_lm._transform_awq_weights`` would convert into
+    ``lm_head.weight/scales/biases`` and then fail strict weight load.
+    """
+    return {k: v for k, v in weights.items() if k not in _AWQ_LM_HEAD_QUANT_KEYS}
+
+
+def _patch_mlx_lm_qwen2_awq_tied_lm_head() -> None:
```
LxYuan0420: I would not carry this sanitize patch in the AWQ support PR. None of the validated checkpoints in this PR require it, and it becomes a global upstream patch as soon as lifecycle imports `apply_compat_patches()`. Please avoid speculative patching here and keep the AWQ support path focused on what is actually needed and validated.

docs/supported_models.md
```diff
@@ -22,6 +22,26 @@ unless prefix caching is explicitly forced. These values describe the
 default engine behavior, not exhaustive model-by-model benchmarking on
 Metal. Qwen3 is explicitly covered by the paged prefix-cache e2e test.
 
+`HF AWQ / GPTQ checkpoints`: load through mlx-lm 0.31.3+'s built-in
```
LxYuan0420: Please keep the support claim scoped to AWQ in this PR. GPTQ can share the same long-term owner and normalization path, but the public support claim should stay narrow until GPTQ has real checkpoint validation.

tests/test_model_lifecycle.py
```diff
@@ -15,9 +15,21 @@
 from tests.stub_runner import make_stub_runner
 from vllm_metal.config import reset_config
 from vllm_metal.paged_attention_backend.mla import MLA_DEFAULT_QK_ROPE_HEAD_DIM
+from vllm_metal.pytorch_backend.tensor_bridge import torch_to_mlx
 from vllm_metal.v1 import model_lifecycle
 from vllm_metal.v1.model_lifecycle import ModelLifecycle
 
+
+def _default_target_dtype():
```
LxYuan0420: This test churn is a smell from the wrong abstraction. Generic lifecycle tests now need dtype-aware cache-key setup even though they are not testing quantized load policy. That is a sign the AWQ/GPTQ behavior has leaked out of its own concern and into unrelated lifecycle coverage. If the quantized load flow is moved behind one owner under `vllm_metal/quant/`, most of this extra test setup should disappear.

vllm_metal/quant/awq_config.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _pick_alias(raw: Mapping[str, Any], aliases: tuple[str, ...]) -> Any:
+    """Return the first present alias's value, or None if none are set."""
+    for key in aliases:
+        if key in raw:
+            return raw[key]
+    return None
+
+
+def normalize_quant_config(raw: Mapping[str, Any]) -> dict[str, Any]:
```
LxYuan0420, line 63: If we introduce `vllm_metal/quant/` as a boundary, it should be a real owner, not just a place to park `normalize_quant_config(...)`. Right now the package contains one free normalizer while lifecycle still owns the actual AWQ/GPTQ behavior. Please pull the load policy behind a small class under `vllm_metal/quant/` so this boundary is cohesive and production-grade.

LxYuan0420 (changes requested):
There is useful AWQ work in this PR, and I do want it to land. The main thing I want tightened before merge is the ownership shape. Right now too much quantization policy is being taught directly to lifecycle, and the PR also pulls in a speculative global sanitize patch that is not required by the validated checkpoints. 

Ideally, I want the AWQ/GPTQ load flow owned in one cohesive place under `vllm_metal/quant/`, with `ModelLifecycle` staying close to the upstream loader flow and compat patches kept narrow and necessity-driven.

That will give us a cleaner foundation for future quantized checkpoint work and keep this change aligned with the code quality bar we want to hold.

vllm_metal/quant/awq_loader.py
```diff
@@ -0,0 +1,221 @@
...
+                continue
+            if not mx.issubdtype(dtype, mx.floating):
+                continue
+            if dtype == target_dtype:
+                continue
+            updates[name] = value.astype(target_dtype)
+        if updates:
+            module.update(updates)
+            n_cast += len(updates)
+    return n_cast
```
LxYuan0420: Please make `AWQQuantLoader` actually own this behavior. `_read_raw_quantization_config()`, `_is_mlx_quantized_module()`, and `_align_non_quantized_dtypes()` should move behind private methods or a small owned collaborator under the loader. The tests should exercise the public loader contract, not import private module helpers directly. This is a foundational quant load path, so the ownership boundary needs to be clean.

vllm_metal/quant/awq_loader.py
```diff
@@ -0,0 +1,221 @@
...
+
+    Surrounding floating params follow the engine's runtime dtype: layer
+    norms, embeddings (when not quantized), and the quantized layer's
+    own ordinary ``bias`` (Qwen2 q/k/v projections, MoE per-expert
+    biases, etc.). Without aligning the regular ``bias`` the projection
+    emits mixed-dtype activations into a bf16 KV cache / sampler.
+    """
+    import mlx.core as mx
+    import mlx.nn as nn
+    from mlx.utils import tree_flatten
```
LxYuan0420: Please move these imports to the top of the file. This module is already the AWQ/MLX load owner and MLX is not an optional dependency for this path, so local imports here only make the code harder to reason about. Same for the test-side `mlx.nn` import with `E402`.

vllm_metal/quant/awq_loader.py
```diff
@@ -0,0 +1,221 @@
...
+from huggingface_hub.utils import HFValidationError
+from mlx_lm import load as mlx_lm_load
+from vllm.logger import init_logger
+
+from vllm_metal.quant.awq_config import normalize_quant_config
+
+logger = init_logger(__name__)
+
+
+_QUANT_METHODS = ("awq", "gptq")
```
LxYuan0420: Please keep this PR’s runtime contract AWQ-only, or add real GPTQ validation before accepting `quant_method="gptq"`. If GPTQ is not part of the support claim yet, the loader should not silently take ownership of GPTQ checkpoints. That should be a follow-up once a real GPTQ checkpoint is validated.

tests/quant/test_awq_loader.py
```diff
@@ -0,0 +1,416 @@
...
+
+    n_cast = _align_non_quantized_dtypes(wrapper, mx.bfloat16)
+
+    assert module.biases.dtype == mx.float16, (
+        "AWQ-transform `biases` (plural) must NOT be cast away from fp16"
+    )
+    assert module.bias.dtype == mx.bfloat16, (
+        "Regular linear `bias` (singular) MUST be cast to runtime bfloat16"
+    )
+    assert n_cast == 1
```
LxYuan0420: Please organize this into class-based test groups, for example cache key, config detection, and dtype alignment. The current long flat file plus banner comments makes the test surface harder to maintain as quant support grows.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -186,20 +202,30 @@ def _load_generation_model(self, model_name: str, is_vlm: bool) -> tuple[Any, An
             )
             return cached
 
+        awq_loader = None if is_vlm else AWQQuantLoader.for_model(model_name)
+        cache_key = awq_key if awq_loader is not None else generic_key
```
LxYuan0420, line 209: Please make the AWQ cache/preflight behavior explicit here. We should prefer the AWQ-specific cache key before the generic `mlx_lm` key, and `AWQQuantLoader.for_model()` should not fail open silently when config lookup fails. If preflight is skipped intentionally, log it clearly; otherwise a transient config miss can fall through to the generic loader/cache path and bypass the dtype-alignment contract this PR is adding.

LxYuan0420 (changes requested):

vllm_metal/quant/awq_loader.py
```diff
@@ -0,0 +1,244 @@
...
+        ``mlx.nn.QuantizedEmbedding`` together with the mlx_lm peers
+        ``QuantizedSwitchLinear`` (MoE) and ``QuantizedMultiLinear`` (MLA),
+        none of which subclass ``nn.QuantizedLinear``. An ``isinstance``
+        check would silently misclassify those as plain modules and let
+        their AWQ-transform ``scales`` / ``biases`` be cast to the runtime
+        dtype. A separate import dependency on the mlx_lm peers would make
+        the alignment fragile across mlx_lm versions; the protocol
+        attributes are stable.
+        """
+        return hasattr(module, "bits") and hasattr(module, "group_size")
```
LxYuan0420, line 200: Optional nit: `_is_mlx_quantized_module()` is still a raw `hasattr(...)` check even though the docstring is already describing a stable protocol. I would prefer to tighten that into an explicit protocol/type-guard in a follow-up so this production path depends on a named contract rather than reflection.

vllm_metal/quant/awq_loader.py
```diff
@@ -0,0 +1,244 @@
...
+        legitimate "no config" case and remains a silent ``None``.
+        """
+        model_path = Path(model_name)
+        if model_path.is_dir():
+            config_path = model_path / "config.json"
+            if not config_path.is_file():
+                return None
+        else:
+            config_path = Path(hf_hub_download(model_name, "config.json"))
+        with open(config_path) as fid:
```
LxYuan0420: Optional nit: please make the config read explicit with `encoding="utf-8"` here. Small thing, but this is a core config path and we should keep file I/O explicit.

tests/quant/test_awq_loader.py
```diff
@@ -0,0 +1,442 @@
...
+        """
+        key = AWQQuantLoader.cache_key("foo", target_dtype=_mlx_dtype(torch.bfloat16))
+        assert key == ("foo", f"mlx_lm-awq:{mx.bfloat16}")
+
+    def test_distinct_from_generic_lifecycle_key(self):
+        """AWQ cache key must not collide with the generic mlx_lm key
+        for the same model: an AWQ-mutated cached model must not be
+        served to a non-AWQ caller, nor vice versa.
+        """
+        from vllm_metal.v1.model_lifecycle import _generation_cache_key
```
LxYuan0420: Optional nit: there is still one local import in this test file. Unless there is a concrete import-cycle reason, I would move `from vllm_metal.v1.model_lifecycle import _generation_cache_key` to the top of the file so the test stays stylistically consistent after the refactor.

LxYuan0420 (approved):
LGTM; Could you also paste one short runtime proof before or after merge, ideally the exact command plus one `LLM.generate` / serve output from the AWQ model? A screenshot is fine too, but raw command/output is more useful for future debugging.

scyyh11:
Runtime proof on M1 Pro 16 GB / macOS 26.2 against head `5708cc4`. Both runs use `Qwen/Qwen2.5-1.5B-Instruct-AWQ`. The `awq_loader.py:135 AWQ load:` line is the trace point that confirms the new owner is on the load path; it reads `aligned 0` here because the engine resolved `dtype=torch.float16` and the AWQ-transform output is already fp16 — bf16 alignment is exercised by the slow regression test.

### 1. `LLM.generate`

```
VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.3 \
python -c "
llm = LLM(model='Qwen/Qwen2.5-1.5B-Instruct-AWQ', max_model_len=512, max_num_seqs=1)
print(llm.generate(['The capital of France is'], SamplingParams(temperature=0, max_tokens=32))[0].outputs[0].text)
"
```

Load trace (relevant lines):

```
INFO 05-06 10:32:10 [worker.py:115] MLX device set to: Device(gpu, 0)
INFO 05-06 10:32:10 [model_lifecycle.py:176] Loading model: Qwen/Qwen2.5-1.5B-Instruct-AWQ (VLM: False)
INFO 05-06 10:32:12 [awq_loader.py:135] AWQ load: aligned 0 non-quantized floating params to mlx.core.float16
INFO 05-06 10:32:12 [model_lifecycle.py:236] Model loaded in 1.53s: Qwen/Qwen2.5-1.5B-Instruct-AWQ
INFO 05-06 10:32:13 [cache.py:166] KV cache: 1905.2 MB (28 layers, 4153 blocks, 16 tokens/block)
INFO 05-06 10:32:13 [cache_policy.py:601] Paged attention enabled: 28 layers patched, 4153 blocks allocated
```

Output:

```
PROMPT: 'The capital of France is'
OUTPUT: ' Paris. Paris is a city in France. Paris is a city in the Île-de-France region. Paris is a city in the north of France.'
```

### 2. `vllm serve` + `curl /v1/completions`

Server:

```
VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=0.3 \
vllm serve Qwen/Qwen2.5-1.5B-Instruct-AWQ --max-model-len 512 --max-num-seqs 1 --port 8000
```

Server-side load trace:

```
(EngineCore pid=33445) INFO 05-06 10:32:50 [model_lifecycle.py:176] Loading model: Qwen/Qwen2.5-1.5B-Instruct-AWQ (VLM: False)
(EngineCore pid=33445) INFO 05-06 10:32:51 [awq_loader.py:135] AWQ load: aligned 0 non-quantized floating params to mlx.core.float16
(EngineCore pid=33445) INFO 05-06 10:32:51 [model_lifecycle.py:236] Model loaded in 1.18s: Qwen/Qwen2.5-1.5B-Instruct-AWQ
(EngineCore pid=33445) INFO 05-06 10:32:51 [cache_policy.py:601] Paged attention enabled: 28 layers patched, 4219 blocks allocated
(APIServer pid=33416) INFO:     Application startup complete.
```

Client:

```
curl -sS -X POST http://127.0.0.1:8000/v1/completions \
  -d '{"model":"Qwen/Qwen2.5-1.5B-Instruct-AWQ","prompt":"The capital of France is","max_tokens":32,"temperature":0}'
```

Response:

```json
{
    "id": "cmpl-a7115b9f8bee9ae8",
    "object": "text_completion",
    "created": 1778077984,
    "model": "Qwen/Qwen2.5-1.5B-Instruct-AWQ",
    "choices": [
        {
            "index": 0,
            "text": " Paris. The capital of the United States is Washington D.C. The capital of Canada is Ottawa. The capital of Australia is Canberra. The capital of New Zealand",
            "logprobs": null,
            "finish_reason": "length",
            "stop_reason": null,
            "token_ids": null,
            "prompt_logprobs": null,
            "prompt_token_ids": null
        }
    ],
    "usage": {
        "prompt_tokens": 5,
        "total_tokens": 37,
        "completion_tokens": 32
    }
}
```

PR #341, opened 2026-05-05
[Multimodal] Wire Qwen3.5 lifecycle state

ricky-chaoju:
This PR wires the v1 lifecycle state needed by the Qwen3.5/Qwen3-VL multimodal path.

It does not run the vision encoder or splice image embeddings yet. Those remain follow-up runner changes. This PR keeps the first runtime step focused on loading and preserving model-owned multimodal state, registering scheduler-provided multimodal features, and failing fast when encoder execution is scheduled before the Metal runner path is wired.

##  Adds

- Build a Qwen3-VL multimodal adapter for Qwen3.5/Qwen3-VL native VLM loads.
- Preserve the loaded `vision_tower` and `language_model` behind the model-owned adapter.
- Add Metal runner encoder-cache lifecycle wiring for:
  - new request `mm_features`
  - finished and preempted request cleanup
  - scheduler-released encoder output hashes
  - `reset_mm_cache()` and `reset_encoder_cache()`
- Fail fast on scheduled multimodal encoder inputs until encode/splice/forward is implemented.
- Keep text-only VLM execution unchanged through the language model.
- Use real upstream scheduler DTOs in touched runner tests.

### DIFF
--- a/vllm_metal/multimodal/qwen3_vl/adapter.py
+++ b/vllm_metal/multimodal/qwen3_vl/adapter.py
@@ -16,6 +16,8 @@ from vllm_metal.multimodal.feature_spec import MultiModalFeatureSpec
 class Qwen3VLMultimodalAdapter:
     """Model-owned multimodal helpers for the Qwen3-VL execution path."""
 
+    forward_ready: bool = False
+
     def __init__(
         self,
         *,
@@ -31,8 +33,18 @@ class Qwen3VLMultimodalAdapter:
         self._vision_tower = vision_tower
         self._language_model = language_model
 
+    @property
+    def vision_tower(self) -> Any:
+        """Return the loaded Qwen3-VL vision tower."""
+        return self._vision_tower
+
+    @property
+    def language_model(self) -> Any:
+        """Return the loaded Qwen3-VL language model."""
+        return self._language_model
+
     @classmethod
-    def _from_loaded_model(cls, model: Any) -> Qwen3VLMultimodalAdapter:
+    def from_loaded_model(cls, model: Any) -> Qwen3VLMultimodalAdapter:
         """Create an adapter from an mlx-vlm Qwen3-VL/Qwen3.5 composite."""
         vision_tower = model.vision_tower
         language_model = model.language_model
--- a/vllm_metal/multimodal/qwen3_vl/adapter.py
+++ b/vllm_metal/multimodal/qwen3_vl/adapter.py
@@ -16,8 +16,6 @@ from vllm_metal.multimodal.feature_spec import MultiModalFeatureSpec
 class Qwen3VLMultimodalAdapter:
     """Model-owned multimodal helpers for the Qwen3-VL execution path."""
 
-    forward_ready: bool = False
-
     def __init__(
         self,
         *,
@@ -33,13 +31,7 @@ class Qwen3VLMultimodalAdapter:
         self._vision_tower = vision_tower
         self._language_model = language_model
 
-    @property
-    def vision_tower(self) -> Any:
-        """Return the loaded Qwen3-VL vision tower."""
-        return self._vision_tower
-
-    @property
-    def language_model(self) -> Any:
+    def text_model(self) -> Any:
         """Return the loaded Qwen3-VL language model."""
         return self._language_model
 
--- a/vllm_metal/multimodal/qwen3_vl/adapter.py
+++ b/vllm_metal/multimodal/qwen3_vl/adapter.py
@@ -16,12 +16,32 @@ from vllm_metal.multimodal.feature_spec import MultiModalFeatureSpec
 class Qwen3VLMultimodalAdapter:
     """Model-owned multimodal helpers for the Qwen3-VL execution path."""
 
-    def __init__(self, *, spatial_merge_size: int) -> None:
+    def __init__(
+        self,
+        *,
+        spatial_merge_size: int,
+        vision_tower: Any | None = None,
+        language_model: Any | None = None,
+    ) -> None:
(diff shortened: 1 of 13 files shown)


vllm_metal/v1/model_adapter.py
```diff
@@ -39,6 +39,9 @@ def require_uniform_kv_heads(
     def text_model(self, model: Any) -> Any:
         """Return the callable model used for text-only execution."""
 
+    def build_multimodal_adapter(self, model: Any, hf_config: Any) -> Any | None:
```
LxYuan0420: This multimodal adapter boundary is too loose for a foundation PR. Right now the protocol returns `Any | None`, the runner stores `Any | None`, and the tests already have to reach into private fields on the concrete adapter. Please introduce a small explicit runtime adapter protocol here so this state is owned through a real contract rather than an untyped bag.

vllm_metal/v1/model_runner.py
```diff
@@ -1252,6 +1317,10 @@ def execute_model(
         if self._is_stt:
             return self._execute_stt(scheduler_output)
 
+        self._reject_scheduled_encoder_inputs(scheduler_output.scheduled_encoder_inputs)
```
LxYuan0420: Please free scheduler-released encoder outputs before rejecting scheduled encoder inputs. `free_encoder_mm_hashes` is independent cache cleanup, and the current order skips that cleanup on the fail-fast path. The test at `tests/v1/mm/test_model_runner_multimodal_cache.py:193-206` currently locks in that leak.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -160,6 +160,11 @@ def load(self) -> None:
         runner._is_vlm = is_vlm
         runner._is_stt = False
         runner._stt_runtime_adapter = None
+        runner._multimodal_adapter = (
```
LxYuan0420: This runtime capability should be installed by lifecycle, not pre-created in `MetalModelRunner.__init__` from config alone. `ModelLifecycle` already decides whether native multimodal mode is active and builds the model-owned adapter; it should also install the corresponding runner-side multimodal state, including `EncoderCache`, in the same place. That keeps the ownership explicit and avoids config-driven runtime state that is broader than the loaded path.

vllm_metal/v1/model_adapter.py
```diff
@@ -204,6 +214,22 @@ def text_model(self, model: Any) -> Any:
...
+        model_type = getattr(hf_config, "model_type", "")
+        architectures = getattr(hf_config, "architectures", ()) or ()
+        if model_type not in _QWEN3_VL_MODEL_TYPES and not any(
+            arch in _QWEN3_VL_ARCHITECTURES for arch in architectures
+        ):
+            return None
+
+        from vllm_metal.multimodal.qwen3_vl import Qwen3VLMultimodalAdapter
+
+        return Qwen3VLMultimodalAdapter._from_loaded_model(model)
```
LxYuan0420: This is the public construction path for the adapter, so it should not call a private method across classes. Please make this `Qwen3VLMultimodalAdapter.from_loaded_model(...)` and reserve `_from_loaded_model` for adapter-internal use only.

LxYuan0420 (changes requested):
This is moving in the right direction, and the runner/lifecycle step itself makes sense. The main thing I want tightened before merge is the code shape: this is a foundation PR, so I want the ownership boundaries and contracts to be explicit from the start. That will give us a cleaner base for the next encode/splice/forward PRs.

vllm_metal/multimodal/qwen3_vl/adapter.py
```diff
@@ -16,12 +16,32 @@
 class Qwen3VLMultimodalAdapter:
```
scyyh11, line 16: Could we add a `forward_ready: bool = False` here? Right now `_reject_scheduled_encoder_inputs` in `model_runner.py` just unconditionally raises, which works for one adapter but bakes in a global gate. Threading it through `self._multimodal_adapter.forward_ready` saves time when we want to add a second adapter.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -189,8 +194,9 @@ def _load_generation_model(self, model_name: str, is_vlm: bool) -> tuple[Any, An
         if is_vlm:
             logger.info("Using mlx-vlm for vision-language model")
             logger.warning(
-                "VLM loaded in text-only mode: multimodal (image) inputs are "
-                "not yet supported. Vision encoder will be bypassed."
+                "VLM loaded via mlx-vlm; Metal multimodal encoder execution "
+                "is not wired yet. Text-only requests continue through the "
+                "language model."
             )
             model, tokenizer = mlx_vlm_load(model_name)
```
scyyh11, line 206: Now that `runner.model` can be an mlx_vlm composite, `generate()` still passes `self.model` directly into `mlx_lm.stream_generate`. That bypasses the adapter language-model routing used by runner execution and can re-enter the top-level VLM path. Please pass `self._forward_model` instead.

ricky-chaoju:
All done.

LxYuan0420 (approved):
@ricky-chaoju I added a few edits; please also review it and let me know what you think.

ricky-chaoju:
@LxYuan0420 Thanks, I reviewed the follow-up edits. The updated shape looks good to me: cleanup now runs before the encoder fail-fast path, scheduler-released encoder outputs are freed first, preempted request state is preserved, and the multimodal adapter contract is back to behavior-oriented methods instead of exposing runtime state as a loose bag.

I’m going to merge this now.

PR #342, opened 2026-05-05
[Bugfix] cache ByteLevel TokenizersBackend for serving

ricky-chaoju:
Follow-up to #337.

The ByteLevel compatibility path now wraps the replacement `TokenizersBackend` with vLLM's `get_cached_tokenizer()` before returning it to the serving tokenizer path. This preserves clean ByteLevel decoding while restoring serving metadata such as `max_chars_per_token`, which vLLM uses during request preprocessing.

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -115,6 +115,7 @@ class TestByteLevelTokenizerCompatPatch:
         assert tokenizer.eos_token == "</s>"
         assert tokenizer.pad_token == "</s>"
         assert tokenizer.model_max_length == 128
+        assert tokenizer.max_chars_per_token >= len("\u0120Hello")
         assert compat._loaded_tokenizer_decoder_uses_bytelevel(tokenizer)
 
     def test_keeps_loaded_tokenizer_when_decoder_is_already_bytelevel(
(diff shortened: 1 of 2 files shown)


ricky-chaoju:
@LxYuan0420 Follow-up to #337: the raw `TokenizersBackend` fixed ByteLevel decoding but missed vLLM cached tokenizer metadata like `max_chars_per_token`.Wrapping it with `get_cached_tokenizer()` fixes the DeepSeek serving 500; `/v1/completions` now returns 200.

LxYuan0420 (approved):
Great work! Ricky

Could you also paste one screenshot or raw curl output from the DeepSeek repro showing `/v1/completions` returns 200 with clean decoded text? That gives us runtime proof for the issue thread.

ricky-chaoju:
> Great work! Ricky
> 
> Could you also paste one screenshot or raw curl output from the DeepSeek repro showing `/v1/completions` returns 200 with clean decoded text? That gives us runtime proof for the issue thread.

Sure, here is the raw curl output after applying this pr.
截圖 2026-05-05 晚上10 54 49

PR #344, opened 2026-05-06
Add lazy GDN decode fast path for hybrid models

gwlee-dev:
## Summary

Add a lazy Metal decode fast path for GDN-backed hybrid models. The path is enabled by default for decode-only batches and keeps the existing eager conv / C++ recurrent path as the fallback for unsupported shapes, mixed prefill+decode batches, or when the public kill switch is disabled.

## Problem

Hybrid GDN models such as Qwen3-Next/Qwen3.5 spend a large share of decode time in small per-request conv/recurrent work. For multi-request decode, the existing path repeatedly slices host-side metadata and launches smaller operations even when the active batch is already decode-only.

## Changes

- Add a lazy decode-only conv1d+SiLU Metal path for GDN convolution.
- Add a lazy recurrent decode Metal path for supported recurrent state shapes.
- Isolate lazy decode policy and kernel ownership in `GDNLazyDecodeKernels`.
- Reuse a process-level lazy decode kernel owner from the GDN wrapper to avoid per-wrapper `mx.fast.metal_kernel` callable construction.
- Keep `GDNLazyDecodeKernels.from_env()` as a fresh env-derived factory for tests/direct construction, while `GDNLazyDecodeKernels.shared()` owns shared process reuse.
- Collapse the public control surface to a single Metal-specific kill switch:
  - `VLLM_METAL_GDN_LAZY_DECODE=0` forces the eager conv / C++ recurrent fallback path.
- Keep fallback behavior for:
  - non decode-only batches (`total_tokens != num_requests`)
  - unsupported recurrent head shapes
  - unavailable Metal lazy kernels
  - disabled lazy decode kill switch

## Runtime tradeoff

The lazy Metal kernels materialize replacement state arrays (`conv_state_out` / `state_out`) and then swap the active cache entry. This differs from the legacy recurrent C++ path, which updates the recurrent state pool in place. The allocation tradeoff keeps the decode-only fast path lazy and avoids the eager host synchronization cost; ineligible shapes continue to use the existing fallback path.

The env kill switch is read when the shared lazy decode owner is acquired. Existing wrappers keep the owner captured at wrapper construction time, matching the process-level configuration style used elsewhere in the plugin.

## Benchmarks

Compared against upstream/main on a local Apple M2 Ultra with 192 GB unified memory, using the vLLM 0.20.0 benchmark environment and `VLLM_METAL_MEMORY_FRACTION=0.70`. The final reruns used `VLLM_ENABLE_V1_MULTIPROCESSING=0` for both upstream and this PR.

### Qwen3.5-0.8B

Workload: input 128, output 64, 10 requests. 11 runs, with the first run discarded as warmup.

| Metric | Upstream | This PR | Delta |
| --- | ---: | ---: | ---: |
| TTFT | 0.236s | 0.238s | +0.7% |
| Decode TPS | 330.7 | 451.5 | +36.5% |
| Wall time | 2.143s | 1.635s | -23.7% |
| Wall TPS | 298.7 | 391.5 | +31.1% |

### Qwen3-Next 80B A3B

| Workload | Upstream TTFT | This PR TTFT | Upstream decode TPS | This PR decode TPS | Wall delta |
| 1K input, batch 1, output 16 | 1.547s | 1.543s | 23.62 | 34.43 | -9.3% |
| 1K input, batch 5, output 16 | 6.478s | 6.464s | 83.74 | 113.60 | -3.8% |
| 4K input, batch 1, output 16 | 6.279s | 6.248s | 18.28 | 23.41 | -3.1% |
| 4K input, batch 5, output 16 | 24.332s | 24.080s | 11.58 | 11.80 | -1.4% |

## Test plan

- `python -m ruff check vllm_metal/envs.py vllm_metal/metal/__init__.py vllm_metal/metal_kernel_backend/attention_linear.py vllm_metal/metal_kernel_backend/gdn_lazy_decode.py tests/test_gdn_lazy_decode.py docs/configuration.md`
- `python -m ruff format --check vllm_metal/envs.py vllm_metal/metal/__init__.py vllm_metal/metal_kernel_backend/attention_linear.py vllm_metal/metal_kernel_backend/gdn_lazy_decode.py tests/test_gdn_lazy_decode.py`
- `python -m py_compile vllm_metal/envs.py vllm_metal/metal/__init__.py vllm_metal/metal_kernel_backend/attention_linear.py vllm_metal/metal_kernel_backend/gdn_lazy_decode.py tests/test_gdn_lazy_decode.py`
- `git diff --check`
- `python -m pytest tests/test_gdn_lazy_decode.py -q`
- `python -m pytest tests/test_qwen3_next_gdn.py -q -k 'not slow'`
- `python -m pytest tests/test_register.py tests/test_config.py -q`
- `rm -f ~/.cache/vllm-metal/_paged_ops*.so && /tmp/vllm-metal-bench-v020/bin/python -m pytest tests/test_attention_dispatch.py::test_qwen35_paged_attention_hybrid -q`
- `/tmp/vllm-metal-bench-v020/bin/python -m pytest tests/test_paged_deterministic.py -q`

Additional full-suite check:

- `/tmp/vllm-metal-bench-v020/bin/python -m pytest tests -q` reached `674 passed, 16 skipped`; the remaining `tests/test_paged_deterministic.py` cases hit an order-dependent Metal memory budget error after the full suite and passed when rerun standalone with the same vLLM 0.20 environment.

## Focused regression coverage

`tests/test_gdn_lazy_decode.py` covers:

- lazy conv decode parity against eager per-request conv
- lazy recurrent decode parity against `ops.gdn_linear_attention(...)`
- fallback when `total_tokens != num_requests`
- fallback when recurrent shape is unsupported
- kill-switch-off behavior without constructing lazy kernels
- wrapper fallback dispatch metadata and state updates when lazy decode is disabled
- shared lazy decode owner reuse

### DIFF
--- a/tests/test_gdn_lazy_decode.py
+++ b/tests/test_gdn_lazy_decode.py
@@ -250,6 +250,59 @@ class TestLazyConvDecode:
 
 
 class TestLazyRecurrentDecode:
+    def test_updates_only_active_recurrent_slots(self) -> None:
+        # Arrange
+        class FakeRecurrentKernel:
+            def __init__(self) -> None:
+                self.grid: tuple[int, int, int] | None = None
+                self.output_shapes: list[tuple[int, ...]] | None = None
+
+            def __call__(self, **kwargs: Any) -> tuple[mx.array, mx.array]:
+                self.grid = kwargs["grid"]
+                self.output_shapes = kwargs["output_shapes"]
+                return (
+                    mx.ones((2, 2, 3), dtype=mx.float32),
+                    mx.full((2, 2, 3, 32), 9, dtype=mx.float32),
+                )
+
+        fake_kernel = FakeRecurrentKernel()
+        cache = _make_state_cache(
+            max_seqs=4,
+            num_v_heads=2,
+            value_head_dim=3,
+            key_head_dim=32,
+        )
+        initial_state = mx.arange(4 * 2 * 3 * 32, dtype=mx.float32).reshape(4, 2, 3, 32)
+        cache.recurrent_states[0] = mx.array(initial_state)
+        slot_ids = [3, 1]
+        request = _recurrent_request(
+            q=mx.zeros((1, 2, 1, 32), dtype=mx.float32),
+            k=mx.zeros((1, 2, 1, 32), dtype=mx.float32),
+            v=mx.zeros((1, 2, 2, 3), dtype=mx.float32),
+            g=mx.zeros((1, 2, 2), dtype=mx.float32),
+            beta=mx.zeros((1, 2, 2), dtype=mx.float32),
+            cache=cache,
+            slot_ids=slot_ids,
+        )
+
+        # Act
+        result = GDNLazyDecodeKernels(
+            enabled=True,
+            conv_kernel=_RaisingKernel(),
+            recurrent_kernel=fake_kernel,
+        ).try_recurrent_decode(request)
+
+        # Assert
+        assert result is not None
+        assert fake_kernel.grid == (32, 3, 4)
+        assert fake_kernel.output_shapes == [(2, 2, 3), (2, 2, 3, 32)]
+        mx.eval(cache.recurrent_states[0])
+        expected_state = np.array(initial_state)
+        expected_state[slot_ids] = 9
+        np.testing.assert_array_equal(
+            np.array(cache.recurrent_states[0]), expected_state
+        )
+
     def test_matches_cpp_recurrent_path(self) -> None:
         # Arrange
         _require_metal()
--- a/tests/test_gdn_lazy_decode.py
+++ b/tests/test_gdn_lazy_decode.py
@@ -460,6 +460,34 @@ class TestGDNPagedAttentionWrapperLazyDecode:
         finally:
             clear_context()
 
+    def test_rejects_out_of_range_gdn_slots(self) -> None:
+        # Arrange
+        inner = _TinyGDNInner()
+        cache = _make_state_cache(
+            conv_kernel_dim=inner.conv_kernel_size,
+            conv_dim=inner.conv_dim,
+            num_v_heads=inner.num_v_heads,
+            value_head_dim=inner.head_v_dim,
+            key_head_dim=inner.head_k_dim,
+        )
+        wrapper = GDNPagedAttentionWrapper(
+            inner, layer_idx=0, cache_idx=0, state_cache=cache
+        )
+        set_context(
+            PagedAttentionContext(
+                slot_mapping=[0, 1],
+                cu_seqlens=[0, 1, 2],
+                gdn_slot_mapping=[0, cache.max_seqs],
+            )
+        )
+
+        # Act / Assert
+        try:
+            with pytest.raises(RuntimeError, match="out-of-range slot"):
+                wrapper(mx.ones((1, 2, inner.conv_dim), dtype=mx.float32))
+        finally:
+            clear_context()
+
     def test_kill_switch_uses_recurrent_fallback(
         self, monkeypatch: pytest.MonkeyPatch
     ) -> None:
--- a/tests/test_gdn_lazy_decode.py
+++ b/tests/test_gdn_lazy_decode.py
@@ -1,0 +1,441 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Focused regression tests for lazy GDN decode dispatch."""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from typing import Any
+
+import mlx.core as mx
+import mlx.nn as nn
+import numpy as np
+import pytest
+
+import vllm_metal.metal_kernel_backend.attention_linear as attention_linear
+from vllm_metal.metal import get_ops
+from vllm_metal.metal_kernel_backend.attention_linear import GDNPagedAttentionWrapper
+from vllm_metal.metal_kernel_backend.gdn_lazy_decode import GDNLazyDecodeKernels
+from vllm_metal.mlx_backend.gdn_cache import GDNPagedStateCache
+from vllm_metal.paged_attention_common import (
+    PagedAttentionContext,
+    clear_context,
+    set_context,
+)
+
+
+class _DepthwiseConv1D:
+    def __init__(self, weight: mx.array) -> None:
+        self.weight = weight
+
+    def __call__(self, x: mx.array) -> mx.array:
+        kernel_size = self.weight.shape[1]
+        outputs = []
+        for pos in range(x.shape[1] - kernel_size + 1):
+            window = x[:, pos : pos + kernel_size, :]
+            outputs.append(mx.sum(window * self.weight.T[None, :, :], axis=1))
+        return mx.stack(outputs, axis=1)
+
+
+class _LastTokenConv1D:
+    def __init__(self, conv_dim: int) -> None:
+        self.weight = mx.ones((conv_dim, 1), dtype=mx.float32)
+
+    def __call__(self, x: mx.array) -> mx.array:
+        return x[:, -1:, :]
+
+
+class _TinyGDNInner:
+    num_k_heads = 1
+    num_v_heads = 1
+    head_k_dim = 32
+    head_v_dim = 4
+    key_dim = 32
+    conv_dim = 68
+    conv_kernel_size = 2
+
+    def __init__(self) -> None:
+        self.conv1d = _LastTokenConv1D(self.conv_dim)
+        self.A_log = mx.zeros((self.num_v_heads,), dtype=mx.float32)
+        self.dt_bias = mx.zeros((self.num_v_heads,), dtype=mx.float32)
+
+    def in_proj_qkv(self, x: mx.array) -> mx.array:
+        return x
+
+    def in_proj_z(self, x: mx.array) -> mx.array:
+        return mx.zeros(
+            (1, x.shape[1], self.num_v_heads * self.head_v_dim), dtype=x.dtype
+        )
+
+    def in_proj_b(self, x: mx.array) -> mx.array:
+        return mx.zeros((1, x.shape[1], self.num_v_heads), dtype=x.dtype)
+
+    def in_proj_a(self, x: mx.array) -> mx.array:
+        return mx.zeros((1, x.shape[1], self.num_k_heads), dtype=x.dtype)
+
+    def norm(self, out: mx.array, z: mx.array) -> mx.array:
+        return out
+
+    def out_proj(self, out: mx.array) -> mx.array:
+        return out
+
+
+class _RaisingKernel:
+    def __call__(self, **_: Any) -> tuple[mx.array, mx.array]:
+        raise AssertionError("fallback path should not invoke lazy kernel")
+
+
+def _require_metal() -> None:
+    try:
+        available = mx.metal.is_available()
+    except AttributeError:
+        available = False
+    if not available:
+        pytest.skip("MLX Metal is not available")
+
+
+def _get_native_ops_or_skip() -> Any:
+    try:
+        return get_ops()
+    except ImportError as exc:
+        if "Symbol not found" in str(exc):
+            pytest.skip(
+                "cached native _paged_ops extension is incompatible with "
+                "the active MLX; remove ~/.cache/vllm-metal/_paged_ops*.so "
+                "to rebuild it"
+            )
+        raise
+
+
+def _make_state_cache(
+    *,
+    max_seqs: int = 3,
+    conv_kernel_dim: int = 3,
+    conv_dim: int = 4,
+    num_v_heads: int = 1,
+    value_head_dim: int = 4,
+    key_head_dim: int = 32,
+) -> GDNPagedStateCache:
+    return GDNPagedStateCache(
+        num_layers=1,
+        max_seqs=max_seqs,
+        conv_kernel_dim=conv_kernel_dim,
+        conv_dim=conv_dim,
+        num_v_heads=num_v_heads,
+        value_head_dim=value_head_dim,
+        key_head_dim=key_head_dim,
+        dtype=mx.float32,
+    )
+
+
+def test_lazy_conv_decode_matches_eager_per_request_conv() -> None:
+    _require_metal()
+    mx.random.seed(0)
+    conv_dim = 4
+    kernel_size = 3
+    cache = _make_state_cache(conv_kernel_dim=kernel_size, conv_dim=conv_dim)
+    initial_state = mx.random.normal(cache.conv_states[0].shape).astype(mx.float32)
+    cache.conv_states[0] = mx.array(initial_state)
+    weight = mx.random.normal((conv_dim, kernel_size)).astype(mx.float32)
+    inner = SimpleNamespace(
+        conv_kernel_size=kernel_size, conv1d=_DepthwiseConv1D(weight)
+    )
+    mixed_qkv = mx.random.normal((1, 2, conv_dim)).astype(mx.float32)
+    slot_ids = [1, 0]
+
+    expected_state = mx.array(initial_state)
+    expected_outputs = []
+    for req_idx, slot in enumerate(slot_ids):
+        conv_input = mx.concatenate(
+            [expected_state[slot : slot + 1], mixed_qkv[:, req_idx : req_idx + 1]],
+            axis=1,
+        )
+        expected_state[slot : slot + 1] = conv_input[:, -(kernel_size - 1) :]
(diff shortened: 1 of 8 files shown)


vllm_metal/metal_kernel_backend/gdn_lazy_decode.py
```diff
@@ -0,0 +1,176 @@
...
+    y_out, state_out = _recurrent_v2_kernel(
+        inputs=kernel_inputs,
+        template=template,
+        grid=(32, d_v, max_seqs * n_hv),
+        threadgroup=(32, 4, 1),
+        output_shapes=[(total_tokens, n_hv, d_v), state_in.shape],
+        output_dtypes=[output_dtype, mx.float32],
+    )
+    state_cache.recurrent_states[cache_idx] = state_out
+    return y_out
```
LxYuan0420: This needs a clearer ownership boundary. Right now env parsing, kernel construction, eligibility, and execution are all spread across module globals and free functions, while `GDNPagedAttentionWrapper` is also taking on dispatch policy. For a default-on runtime path, I want the main wrapper to stay orchestration-only and the lazy decode implementation to live behind one cohesive owner.

Something like:
```python
class GDNLazyDecodeKernels:
    def __init__(self, enabled: bool) -> None:
        self._enabled = enabled
        self._conv_kernel = self._make_kernel(...)
        self._recurrent_kernel = self._make_kernel(...)

    @classmethod
    def from_env(cls) -> "GDNLazyDecodeKernels":
        return cls(enabled=envs.VLLM_METAL_GDN_LAZY_DECODE)

    def try_conv_decode(...) -> mx.array | None:
        if not self._enabled or self._conv_kernel is None:
            return None
        ...
        return out

    def try_recurrent_decode(...) -> mx.array | None:
        if not self._enabled or self._recurrent_kernel is None:
            return None
        ...
        return out

    @staticmethod
    def _make_kernel(...) -> Any | None:
        ...

Then `GDNPagedAttentionWrapper` just orchestrates:

conv_packed = self._lazy_decode.try_conv_decode(...)
...
y_flat = self._lazy_decode.try_recurrent_decode(...)

That keeps responsibilities separated, makes the runtime contract easier to test, and gives us a cleaner place to extend this path later.

File: vllm_metal/envs.py [hunk 2]
```diff
@@ -71,6 +75,17 @@
...
+    # VLLM_GDN_* aliases remain for compatibility with existing local scripts.
+    "VLLM_METAL_GDN_CONV_KERNEL": lambda: os.getenv(
+        "VLLM_METAL_GDN_CONV_KERNEL", os.getenv("VLLM_GDN_CONV_KERNEL", "2")
+    ),
+    "VLLM_METAL_GDN_RECURRENT_KERNEL": lambda: os.getenv(
+        "VLLM_METAL_GDN_RECURRENT_KERNEL",
+        os.getenv("VLLM_GDN_RECURRENT_KERNEL", "2"),
+    ),
+    "VLLM_GDN_CONV_KERNEL": lambda: os.getenv("VLLM_GDN_CONV_KERNEL", "2"),
+    "VLLM_GDN_RECURRENT_KERNEL": lambda: os.getenv("VLLM_GDN_RECURRENT_KERNEL", "2"),
```
LxYuan0420: Please collapse this to one user-facing kill switch, for example `VLLM_METAL_GDN_LAZY_DECODE`, and make the default decision there. Separate conv/recurrent selectors plus `"2"` as a mode value expose implementation details and add public surface we probably do not want to carry.

vllm_metal/metal_kernel_backend/gdn_lazy_decode.py
LxYuan0420: This path is default-on now, but the PR does not add direct regression coverage for the new fast paths or their fallback conditions. Please add focused tests for:

- lazy conv decode parity vs the existing eager per-request conv path
- lazy recurrent decode parity vs the existing `ops.gdn_linear_attention(...)` path
- fallback when `total_tokens != num_requests`
- fallback when recurrent shape is unsupported
- kill-switch-off behavior forcing the eager/C++ fallback

The existing GDN tests still cover surrounding state/cache behavior, but they do not lock this new dispatch surface directly.

vllm_metal/metal_kernel_backend/gdn_lazy_decode.py
```diff
@@ -0,0 +1,176 @@
...
+    ]
+    y_out, state_out = _recurrent_v2_kernel(
+        inputs=kernel_inputs,
+        template=template,
+        grid=(32, d_v, max_seqs * n_hv),
+        threadgroup=(32, 4, 1),
+        output_shapes=[(total_tokens, n_hv, d_v), state_in.shape],
+        output_dtypes=[output_dtype, mx.float32],
+    )
+    state_cache.recurrent_states[cache_idx] = state_out
```
LxYuan0420: Please call out this tradeoff explicitly in the PR description or code comments. These kernels materialize new state arrays (`conv_state_out`, `state_out`) and then replace the cache entry, rather than updating the existing pool in place like the legacy recurrent path. That may still be the right tradeoff for decode-only workloads, but since this path is default-on I think the allocation/memory behavior should be part of the rollout story.

LxYuan0420 (changes requested):
Good work; what I want tightened before merge is the runtime shape around it: keep the public control surface simple, give the lazy decode path a clearer owner, and add direct regression coverage for the new dispatch/fallback behavior.

vllm_metal/metal_kernel_backend/gdn_lazy_decode.py
```diff
@@ -0,0 +1,245 @@
...
+        y_out, state_out = self._recurrent_kernel(
+            inputs=kernel_inputs,
+            template=template,
+            grid=(32, d_v, max_seqs * n_hv),
+            threadgroup=(32, 4, 1),
+            output_shapes=[(total_tokens, n_hv, d_v), state_in.shape],
+            output_dtypes=[output_dtype, mx.float32],
+        )
+        state_cache.recurrent_states[cache_idx] = state_out
+        return y_out
```
LxYuan0420, line 266: `try_recurrent_decode()` has too wide of a call contract. Passing raw `q/k/v/g/beta`, cache state, slot ids, token counts, head counts, head dims, and dtype as separate parameters makes this hard to maintain and easy to misuse. Please group the recurrent decode inputs into a small typed request object or derive the dimensions inside the owner.

Suggested shape:

```python
@dataclass(frozen=True)
class GDNRecurrentDecodeRequest:
    q: mx.array
    k: mx.array
    v: mx.array
    g: mx.array
    beta: mx.array
    state_cache: GDNPagedStateCache
    cache_idx: int
    slot_ids: list[int]
    output_dtype: mx.Dtype

    @property
    def total_tokens(self) -> int:
        return self.q.shape[1]

    @property
    def num_key_heads(self) -> int:
        return self.q.shape[2]

    @property
    def key_head_dim(self) -> int:
        return self.q.shape[3]
```

Then the public method becomes:

```python
def try_recurrent_decode(
    self, request: GDNRecurrentDecodeRequest
) -> mx.array | None:
    ...
```

vllm_metal/envs.py
```diff
@@ -71,6 +72,11 @@
     ),
     # Custom cache directory for ModelScope downloads (None if unset).
     "VLLM_METAL_MODELSCOPE_CACHE": lambda: os.getenv("VLLM_METAL_MODELSCOPE_CACHE"),
+    # Enable lazy GDN decode kernels by default. Set to "0" to force the
+    # eager conv / C++ recurrent fallback path.
+    "VLLM_METAL_GDN_LAZY_DECODE": lambda: (
+        os.getenv("VLLM_METAL_GDN_LAZY_DECODE", "1") == "1"
+    ),
```
LxYuan0420, line 79: This adds a public env var, so please update `docs/configuration.md` in the same PR. The single kill switch name is fine, but public configuration needs to be documented with its default and intended use.

Suggested row:

```md
| `VLLM_METAL_GDN_LAZY_DECODE` | `1` | Enable lazy GDN decode kernels for eligible decode-only hybrid batches. Set to `0` to force the eager conv / C++ recurrent fallback path. |
```

vllm_metal/metal_kernel_backend/gdn_lazy_decode.py
```diff
@@ -0,0 +1,245 @@
...
+        return _gdn_lazy_decode_kernels
+
+
+def reset_gdn_lazy_decode_kernels() -> None:
+    """Reset the shared lazy GDN decode kernel owner for tests."""
+    global _gdn_lazy_decode_enabled, _gdn_lazy_decode_kernels
+
+    with _gdn_lazy_decode_lock:
+        _gdn_lazy_decode_kernels = None
+        _gdn_lazy_decode_enabled = None
```
LxYuan0420: Please do not keep the shared owner as module globals plus free accessor/reset functions. The reason for a process-level cache is valid, but the ownership should stay inside `GDNLazyDecodeKernels`, for example `GDNLazyDecodeKernels.shared()` and `GDNLazyDecodeKernels.reset_shared_for_tests()`. This avoids reintroducing module-level helper glue right after creating an owner class.

Suggested shape:

```python
class GDNLazyDecodeKernels:
    _shared: ClassVar["GDNLazyDecodeKernels | None"] = None
    _shared_enabled: ClassVar[bool | None] = None
    _shared_lock: ClassVar[Lock] = Lock()

    @classmethod
    def shared(cls) -> "GDNLazyDecodeKernels":
        enabled = envs.VLLM_METAL_GDN_LAZY_DECODE
        with cls._shared_lock:
            if cls._shared is None or cls._shared_enabled != enabled:
                cls._shared = cls(enabled=enabled)
                cls._shared_enabled = enabled
            return cls._shared

    @classmethod
    def reset_shared_for_tests(cls) -> None:
        with cls._shared_lock:
            cls._shared = None
            cls._shared_enabled = None
```

Then the wrapper can use:

```python
object.__setattr__(self, "_gdn_lazy_decode", GDNLazyDecodeKernels.shared())
```

tests/test_gdn_lazy_decode.py
```diff
@@ -0,0 +1,481 @@
...
+    )
+    np.testing.assert_array_equal(
+        np.array(cache.conv_states[0][2:]),
+        np.zeros((1, inner.conv_kernel_size - 1, inner.conv_dim), dtype=np.float32),
+    )
+    np.testing.assert_array_equal(
+        np.array(cache.recurrent_states[0]),
+        np.full(cache.recurrent_states[0].shape, 7, dtype=np.float32),
+    )
+    assert out.shape == (1, 2, inner.num_v_heads * inner.head_v_dim)
```
LxYuan0420: The coverage is useful, but the test file is long and flat. Please organize this into class-based groups so it stays maintainable as the kernel path grows.

Suggested shape:

```python
class TestLazyConvDecode:
    def test_matches_eager_per_request_conv(self): ...

class TestLazyRecurrentDecode:
    def test_matches_cpp_recurrent_path(self): ...

class TestLazyDecodeFallbacks:
    def test_falls_back_for_multi_token_requests(self): ...
    def test_falls_back_for_unsupported_recurrent_shape(self): ...
    def test_kill_switch_disables_kernel_construction(self): ...

class TestGDNPagedAttentionWrapperLazyDecode:
    def test_kill_switch_uses_recurrent_fallback(self): ...
    def test_shared_owner_reuses_kernel_construction(self): ...
```

This is a new default-on kernel path, so the tests should be structured as a long-lived test surface, not one large sequence of functions.

vllm_metal/metal_kernel_backend/attention_linear.py
```diff
@@ -113,8 +117,13 @@ def __call__(
             if ctx.gdn_slot_mapping is not None
             else list(range(num_requests))
         )
+        conv_packed = self._gdn_lazy_decode.try_conv_decode(
```
LxYuan0420: This wrapper is still doing too much. The lazy decode owner helped, but `GDNPagedAttentionWrapper.__call__()` is still a long procedural pipeline with projection dispatch, conv fallback, recurrent fallback, state handling, and output projection all inline. Please split this into class-owned private methods so `__call__()` stays orchestration only.

Suggested shape:

```python
def __call__(self, x: mx.array, mask=None, cache=None, position_ids=None, **kwargs):
    ctx = get_context()
    if ctx is None:
        return self._inner(x, mask=mask, cache=cache)

    state = self._prepare_gdn_forward_state(x, ctx)
    conv_packed = self._run_conv(state)
    q, k, v = self._split_and_normalize(conv_packed, state)
    y_flat = self._run_recurrent(q, k, v, state)
    return self._project_output(y_flat, state)
```

This path is now default-on, so the readability bar should be higher.

tests/test_gdn_lazy_decode.py
LxYuan0420: Try arrange the test into AAA style; that will be easier to read and structure

vllm_metal/metal_kernel_backend/gdn_lazy_decode.py
```diff
@@ -0,0 +1,268 @@
...
+        ]
+        y_out, state_out = self._recurrent_kernel(
+            inputs=kernel_inputs,
+            template=template,
+            grid=(32, d_v, max_seqs * n_hv),
+            threadgroup=(32, 4, 1),
+            output_shapes=[(total_tokens, n_hv, d_v), state_in.shape],
+            output_dtypes=[request.output_dtype, mx.float32],
+        )
+        request.state_cache.recurrent_states[request.cache_idx] = state_out
```
LxYuan0420: This is close. One remaining thing I want to validate before default-on: this recurrent path rewrites the full `state_in.shape` pool and dispatches over `max_seqs * n_hv`, not active `num_requests * n_hv`.

Could you rerun the benchmark with fixed active request counts and varying `max_num_seqs` so we can confirm this still wins as configured concurrency grows?

vllm_metal/metal_kernel_backend/gdn_lazy_decode.py
```diff
@@ -0,0 +1,268 @@
...
+        ]
+        conv_silu_out, conv_state_out = self._conv_kernel(
+            inputs=kernel_inputs,
+            template=template,
+            grid=(grid_size, 1, 1),
+            threadgroup=(tg_size, 1, 1),
+            output_shapes=[(num_requests, conv_dim), state_shape],
+            output_dtypes=[mixed_qkv.dtype, conv_state_in.dtype],
+        )
+        state_cache.conv_states[cache_idx] = conv_state_out
```
LxYuan0420: Same small follow-up for the conv path: this writes a full `(max_seqs, kernel_size - 1, conv_dim)` state output. Please include this in the `max_num_seqs` benchmark check, or note why the cost is negligible compared with the recurrent path.

LxYuan0420 (commented):
Thanks for the updates. The structure is much better now and the tests are in good shape.

I think this is close. Before making the lazy path default-on, could you rerun the benchmark with fixed active batch sizes and varying `max_num_seqs`, then update the PR description with the result? I just want to confirm the full state-pool rewrite still wins as configured concurrency grows.

gwlee-dev:
Thanks, I updated the PR description and refreshed the benchmark tables after the final lazy-decode guard change.

A couple of notes on the updated numbers:

- Qwen3.5-0.8B was rerun with vLLM 0.20.0, `VLLM_ENABLE_V1_MULTIPROCESSING=0`, `VLLM_METAL_MEMORY_FRACTION=0.70`, and 11 runs with the first run discarded as warmup. The final table shows near-parity TTFT (+0.7%) with improved decode TPS (+36.5%), wall time (-23.7%), and wall TPS (+31.1%).
- Qwen3-Next 80B was rerun with the same original 1K/4K, batch 1/batch 5, output 16 workloads. The refreshed table shows lower wall time across the tested rows (-9.3%, -3.8%, -3.1%, -1.4%) and improved decode TPS across those rows (23.62 → 34.43, 83.74 → 113.60, 18.28 → 23.41, 11.58 → 11.80).

I also added a wrapper-side guard so known-ineligible non-decode batches skip lazy decode attempt setup and go directly to the existing fallback path. The helper-side eligibility checks remain as the final guard. Focused tests still cover lazy conv/recurrent parity, fallback behavior, kill-switch behavior, active-slot-only state updates, and invalid GDN slot mappings before state access.

LxYuan0420 (approved):
Thanks for rerunning and updating the numbers. The small Qwen3.5-0.8B TTFT regression (the delta is minimal: just that it is more obvious with small model) does not look inherently concerning to me here. This PR is targeting the decode-side GDN path, not prefill, so a small first-step overhead can show up there even when steady-state decode throughput and total wall time improve.

Great work!

PR #348, opened 2026-05-06
fix: apply allowed_token_ids, bad_words_token_ids on greedy path

EagerofLight:
## Summary

The native greedy fast path (`mx.argmax`) bypasses the PyTorch Sampler, silently
ignoring several sampling constraints. Requests with `temperature=0` + any of
these params get incorrect output.

## Reproduction

```bash
vllm serve mlx-community/Qwen2.5-7B-Instruct-4bit --max-model-len 8192
```

**allowed_token_ids:**
```bash
  -d '{"model":"mlx-community/Qwen2.5-7B-Instruct-4bit",
       "prompt":"Is the sky blue? Answer yes or no:","max_tokens":3,
       "temperature":0,"allowed_token_ids":[2753]}'
```
Expected: "No" (token 2753 only). Actual: `" Yes. The"`

**bad_words** (chat endpoint):
```bash
curl http://localhost:8000/v1/chat/completions \
  -d '{"model":"mlx-community/Qwen2.5-7B-Instruct-4bit",
       "messages":[{"role":"user","content":"What is the capital of France?"}],
       "max_tokens":10,"temperature":0,"bad_words":["Paris"]}'
```
Expected: "Paris" excluded. Actual: `"The capital of France is Paris."`

## After Fix

- `allowed_token_ids=[2753]` → outputs `"NoNoNo"` (constrained to token 2753)
- `bad_words=["Paris"]` → outputs without "Paris"

## Root Cause

`sample_from_logits()` greedy condition checked `batch.all_greedy and batch.no_top_k and batch.no_top_p and batch.no_penalties` but did not check for `allowed_token_ids`, `bad_words`, `min_tokens`, or `thinking_token_budget`. Additionally, `allowed_token_ids_mask` and `bad_words_token_ids` were hardcoded to `None` / `{}` in `make_sampling_metadata()`.

## Fix

1. Use `can_use_native_greedy()` in `sample_from_logits()` greedy condition
2. Extend `can_use_native_greedy()` with checks for `allowed_token_ids`, `bad_words`, `min_tokens`, `thinking_token_budget`
3. Build `allowed_token_ids_mask` from `SamplingParams.allowed_token_ids`
4. Build `bad_words_token_ids` from `SamplingParams.bad_words_token_ids`

## Test

- New: `can_use_native_greedy` rejects `allowed_token_ids`, `bad_words`, `min_tokens`
- 32 existing sampling tests pass, no regressions
- E2E verified on Qwen2.5-7B-Instruct-4bit (allowed_token_ids, bad_words)

### DIFF
--- a/tests/test_sampling.py
+++ b/tests/test_sampling.py
@@ -303,39 +303,49 @@ class TestV1SeededSamplingGenerator:
 
 
 class TestV1SamplingBatch:
+    @staticmethod
+    def _batch(params_list: list[SamplingParams]) -> SamplingBatch:
+        return SamplingBatch(
+            params_list,
+            [[1]] * len(params_list),
+            [[]] * len(params_list),
+            vocab_size=VOCAB_SIZE,
+            device=torch.device("cpu"),
+        )
+
     def test_can_use_native_greedy_requires_greedy_without_filters(self) -> None:
-        assert SamplingBatch.can_use_native_greedy([SamplingParams(temperature=0.0)])
-        assert not SamplingBatch.can_use_native_greedy(
+        assert self._batch([SamplingParams(temperature=0.0)]).can_use_native_greedy()
+        assert not self._batch(
             [SamplingParams(temperature=0.8)]
-        )
+        ).can_use_native_greedy()
         # vLLM normalizes top-k/top-p back to no-op under greedy decoding.
-        assert SamplingBatch.can_use_native_greedy(
+        assert self._batch(
             [SamplingParams(temperature=0.0, top_k=5)]
-        )
-        assert SamplingBatch.can_use_native_greedy(
+        ).can_use_native_greedy()
+        assert self._batch(
             [SamplingParams(temperature=0.0, top_p=0.9)]
-        )
-        assert not SamplingBatch.can_use_native_greedy(
+        ).can_use_native_greedy()
+        assert not self._batch(
             [SamplingParams(temperature=0.8, top_k=5)]
-        )
-        assert not SamplingBatch.can_use_native_greedy(
+        ).can_use_native_greedy()
+        assert not self._batch(
             [SamplingParams(temperature=0.8, top_p=0.9)]
-        )
-        assert not SamplingBatch.can_use_native_greedy(
+        ).can_use_native_greedy()
+        assert not self._batch(
             [SamplingParams(temperature=0.0, presence_penalty=0.5)]
-        )
-        assert not SamplingBatch.can_use_native_greedy(
+        ).can_use_native_greedy()
+        assert not self._batch(
             [SamplingParams(temperature=0.0, repetition_penalty=1.1)]
-        )
-        assert not SamplingBatch.can_use_native_greedy(
+        ).can_use_native_greedy()
+        assert not self._batch(
             [SamplingParams(temperature=0.0, logprobs=1)]
-        )
-        assert not SamplingBatch.can_use_native_greedy(
+        ).can_use_native_greedy()
+        assert not self._batch(
             [SamplingParams(temperature=0.0, allowed_token_ids=[1, 2])]
-        )
-        assert not SamplingBatch.can_use_native_greedy(
-            [SamplingParams(temperature=0.0, bad_words=["bad"])]
-        )
+        ).can_use_native_greedy()
+        bad_sp = SamplingParams(temperature=0.0)
+        bad_sp._bad_words_token_ids = [[99]]
+        assert not self._batch([bad_sp]).can_use_native_greedy()
 
     def test_allowed_token_ids_constrains_greedy_sampling(self) -> None:
         """Greedy with allowed_token_ids must fall back and the constraint works."""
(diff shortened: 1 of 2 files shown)


vllm_metal/v1/sampling_batch.py
```diff
@@ -254,6 +259,29 @@ def _make_penalty_tensors(
...
+        mask = torch.ones(
+            len(self.sampling_params_list),
+            self.vocab_size,
+            dtype=torch.bool,
+            device=self.device,
+        )
+        for i, sp in enumerate(self.sampling_params_list):
+            if sp.allowed_token_ids:
+                mask[i, sp.allowed_token_ids] = False
+        return mask
```
LxYuan0420, line 276: `allowed_token_ids_mask` is wrong for mixed batches. vLLM uses `True` as "mask this token", so unconstrained requests must stay all `False`. This initializes every row to `True`, which masks the full vocab for any unconstrained request in the same batch.

Suggested:

```python
mask = torch.zeros(
    len(self.sampling_params_list),
    self.vocab_size,
    dtype=torch.bool,
    device=self.device,
)
for i, sp in enumerate(self.sampling_params_list):
    if sp.allowed_token_ids:
        mask[i] = True
        mask[i, sp.allowed_token_ids] = False
```

Please add a mixed-batch regression test where one request has `allowed_token_ids` and another request is unconstrained.

vllm_metal/v1/sampling_batch.py
```diff
@@ -177,6 +177,11 @@ def can_use_native_greedy(
             and sampling_params.presence_penalty == 0.0
             and sampling_params.repetition_penalty == 1.0
             and sampling_params.logprobs is None
+            and not sampling_params.logit_bias
+            and not sampling_params.allowed_token_ids
+            and not sampling_params.bad_words
+            and not sampling_params.min_tokens
+            and not getattr(sampling_params, "thinking_token_budget", None)
```
LxYuan0420: This gate is ahead of the actual Metal wiring for `logit_bias`, `min_tokens`, and `thinking_token_budget`. Disabling native greedy only routes these requests to `Sampler.forward()`, but those fields are handled by vLLM logits processors that need `update_state(...)` batch updates. Metal builds `_logitsprocs`, but this PR does not drive that state machine.

Please either wire the upstream logits-processor state properly before sampling, or narrow this PR to the constraints that are actually carried in `SamplingMetadata` today, such as `allowed_token_ids` and `bad_words_token_ids`.

tests/test_sampling.py
```diff
@@ -330,6 +330,15 @@ def test_can_use_native_greedy_requires_greedy_without_filters(self) -> None:
...
             [SamplingParams(temperature=0.0, logprobs=1)]
         )
+        assert not SamplingBatch.can_use_native_greedy(
+            [SamplingParams(temperature=0.0, allowed_token_ids=[1, 2])]
+        )
+        assert not SamplingBatch.can_use_native_greedy(
+            [SamplingParams(temperature=0.0, bad_words=["bad"])]
+        )
+        assert not SamplingBatch.can_use_native_greedy(
+            [SamplingParams(temperature=0.0, min_tokens=5, max_tokens=10)]
```
LxYuan0420: These tests only check that `can_use_native_greedy()` returns `False`; they do not prove the runtime contract that this PR is changing. Please add behavior-level `sample_from_logits()` coverage, especially for a mixed batch where one request has `allowed_token_ids` and another request is unconstrained, plus a case proving `bad_words_token_ids` blocks the banned token.

The repo bar here is to test the glue we own, not only the predicate that chooses a path.

LxYuan0420 (changes requested):
Thanks for digging into this. The native greedy bypass is a real issue, and this PR is moving in the right direction. I left a few comments to tighten the scope. Once the mixed-batch behavior is corrected, I think the fix can land cleanly. The logits-processor-backed fields can follow separately once the state wiring is in place.

EagerofLight:
Thanks so much for the review :) I have resolved them

vllm_metal/v1/sampling_batch.py
```diff
@@ -307,10 +337,7 @@ def sample_from_logits(
     satisfy the OpenAI serving contract.
     """
     if (
-        batch.all_greedy
-        and batch.no_top_k
-        and batch.no_top_p
-        and batch.no_penalties
+        SamplingBatch.can_use_native_greedy(batch.sampling_params_list)
```
LxYuan0420: This native-greedy gate still ignores runner-owned non-argmax-invariant logits processors. `sample_from_logits()` should not stay on the MLX argmax fast path when `batch.logitsprocs.non_argmax_invariant` is non-empty, because those constraints only run through the sampler path.

Suggested direction:

```python
if batch.can_use_native_greedy() and not batch.needs_logprobs:
    ...
```

with `SamplingBatch.can_use_native_greedy()` also checking the batch-owned logits processor state, not only `SamplingParams`.

vllm_metal/v1/sampling_batch.py
```diff
@@ -177,6 +177,8 @@ def can_use_native_greedy(
             and sampling_params.presence_penalty == 0.0
             and sampling_params.repetition_penalty == 1.0
             and sampling_params.logprobs is None
+            and not sampling_params.allowed_token_ids
+            and not sampling_params.bad_words
```
LxYuan0420: This still leaves the greedy bypass too narrow. Please gate this on `sampling_params.bad_words_token_ids`, not `sampling_params.bad_words`. The sampler metadata below consumes `bad_words_token_ids`, so native greedy should be disabled when the tokenized bad-word constraint exists.

Suggested change:

```python
and not sampling_params.bad_words_token_ids
```

tests/test_sampling.py
```diff
@@ -330,6 +330,44 @@ def test_can_use_native_greedy_requires_greedy_without_filters(self) -> None:
...
+                SamplingParams(temperature=0.0),
+            ],
+            [[1, 2], [3, 4]],
+            [[], []],
+            vocab_size=4,
+            device=torch.device("cpu"),
+        )
+        result = sample_from_logits(logits, batch, Sampler(), torch.device("cpu"))
+        assert result.token_ids[0] in [2, 3]
+        assert result.token_ids[1] == 1  # unconstrained, should pick highest logit
```
LxYuan0420, line 380: Please add one behavior test for the tokenized bad-words path, similar to the allowed-token tests here. The regression should prove that greedy sampling falls back to the sampler and blocks the highest-logit bad token. Prefer constructing this through the upstream tokenizer/update path; if this unit test seeds `_bad_words_token_ids` directly, keep that detail isolated to test setup.

LxYuan0420 (commented):
Thanks for the update. This is closer now. I left a few comments on the remaining greedy-bypass boundary: the native greedy gate should include batch-owned logits processors, and the bad-words path should key off the tokenized `bad_words_token_ids` contract

EagerofLight:
Thanks for the update. They are both addressed.

LxYuan0420 (approved):

PR #350, opened 2026-05-07
[Multimodal] Wire Qwen3-VL paged forward through adapter.call_lm

scyyh11:
## Summary

Paged runner integration for Qwen3-VL: encoder-output caching by `mm_features.identifier`, embedding splice, M-RoPE positions, and LM forward on the paged KV path. Adapter stays at `forward_ready=False`; the gate flip + dependency bump + parity test land in follow-up PRs.

## Scope

**Included** (paged-core, 15 commits):

- `EncoderCache` stores the full `MultimodalEncodeResult` per `identifier` (not just hidden states), so deepstack residuals and other adapter-owned fields stay attached.
- `Qwen3VLMultimodalAdapter` gains the three runner-facing helpers: `embed_tokens` (load-time resolved), `encode_multimodal` (per-feature, never re-enters `Model.__call__`), `call_lm` (signature-sniffed `inputs_embeds` + optional deepstack kwargs).
- M-RoPE positions return `(3, 1, seq_len)` with an explicit batch axis so `call_lm` receives the layout mlx-vlm expects.
- `apply_packed_rope` accepts either an int offset (text, existing arange path) or a caller-supplied `(3, 1, seg_len)` array (mm, pre-computed positions).
- `_build_prefill_pack` carries the full prompt for mm requests at `start_pos == 0` so chunked-prefill positions stay correct from the first chunk.
- `_run_mm_paged_forward` does the actual paged splice: per-mm-request full-prompt M-RoPE precompute, chunk-aware placeholder mask, per-feature hidden-states slicing intersected with the chunk, packed `position_ids`, and `ctx.segment_positions` threaded into the attention wrapper.
- `mrope_position_delta` from mm prefill propagates onto `RequestState` (both `new_final` and `cached_final` postprocess paths) so paged decode keeps M-RoPE positions correct.
- Fail-fast: mm request reaching `_start_paged_forward` without a forward-ready adapter raises immediately rather than silently falling through to the text path.

**Not included** (split into follow-up PRs):


| Follow-up | Content                                                                                                                                                                                                                                   |
| --------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1         | Non-paged prefill + batched/sequential decode through `adapter.call_lm`.                                                                                                                                                                  |
| 2         | Spec-decode mm segment positions + scheduler-time `mm_features` pre-registration.                                                                                                                                                         |
| 3         | `mlx-vlm>=0.5.0` dependency bump + `forward_ready=True` flip + Qwen3-VL parity test + throughput benchmark. Blocked on a vLLM release that contains [#42150](https://github.com/vllm-project/vllm/pull/42150) (lifts `llguidance<1.4.0`). |


## Architecture

```
MetalModelRunner ─┬─ encoder_cache (per-request mm features + per-identifier outputs)
                  │
                  ├─ _run_vision_encoders ──► adapter.encode_multimodal(features)
                  │                              ↳ vision_tower(pixel_values, grid_thw)
                  │                              → Qwen3VLVisionEncodeResult
                  │                              → encoder_cache.encoder_outputs[id]
                  │
                  ├─ _run_mm_paged_forward (paged prefill, batched decode)
                  │                              ↳ per-request full-prompt M-RoPE precompute
                  │                              ↳ chunk-aware placeholder mask
                  │                              ↳ per-feature hidden-states slice
                  │                              ↳ packed deepstack concat
                  │                              ↳ ctx.segment_positions
                  │
                  └─ adapter.call_lm(input_ids, inputs_embeds, cache,
                                     position_ids, visual_pos_masks,
                                     deepstack_visual_embeds)
```

Key seams:

- **`Qwen3VLMultimodalAdapter.from_loaded_model`** sniffs `language_model.__call__` at load time and resolves `embed_tokens`, the `inputs_embeds` kwarg name, and whether the LM declares deepstack params explicitly. Renames or signature drift become init-time errors instead of mid-forward attribute errors.
- **`apply_packed_rope`** accepts either an int offset (text segment, arange path) or a caller-provided `(3, 1, seg_len)` array (mm segment, pre-computed M-RoPE).
- **`PagedAttentionContext.segment_positions`** carries per-segment caller-supplied positions through to the attention wrapper, so the M-RoPE override is local to mm segments — text segments keep the existing sequential-arange policy.
- **`forward_ready` flag** stays `False` on the adapter; runner-side code routes through the new path only when an adapter sets it. Flipping to `True` is gated on the parity test landing in PR-4.

## Test plan

Fake-LM coverage for everything the runner touches:

- `tests/v1/mm/test_encoder_cache.py` — encoder cache stores `MultimodalEncodeResult` per identifier; consumers read `.hidden_states`.
- `tests/v1/mm/test_model_runner_multimodal_cache.py` — runner registers/clears mm features per request; encoder gate fails fast when `forward_ready=False`.
- `tests/v1/mm/test_paged_forward_mm_gate.py` — `_start_paged_forward` raises on mm request with no/forward-not-ready adapter.
- `tests/v1/mm/test_paged_forward_mm.py` — single-feature prefill routes through `call_lm` with the right embeds/positions; chunked prefill slices hidden states correctly; deepstack concat across features respects per-chunk slicing; decode segment positions use `cache_start_pos + delta`; `mrope_position_delta` round-trips onto `RequestState` for both `new_final` and `cached_final` paths.
- `tests/v1/mm/test_build_prefill_pack_mm.py` — mm prefill at `start_pos==0` reads `full_prompt_token_ids` from both `RequestState` (cached) and `NewRequestData` (fresh).
- `tests/multimodal/test_qwen3_vl.py` — adapter helpers (`embed_tokens`, `get_mrope_input_positions`, `encode_multimodal`, `call_lm`) behave correctly on a fake LM, including the deepstack kwargs sniff.

Local: `pytest -m "not slow" tests/` → 846 passed, 2 skipped.

### DIFF
--- a/vllm_metal/multimodal/qwen3_vl/adapter.py
+++ b/vllm_metal/multimodal/qwen3_vl/adapter.py
@@ -41,6 +41,11 @@ class Qwen3VLMultimodalAdapter:
         "input_embeddings",
     )
 
+    _DEEPSTACK_KWARGS: tuple[str, ...] = (
+        "visual_pos_masks",
+        "deepstack_visual_embeds",
+    )
+
     def __init__(
         self,
         *,
@@ -49,6 +54,7 @@ class Qwen3VLMultimodalAdapter:
         language_model: Any | None = None,
         embeds_kwarg: str | None = None,
         embed_tokens_fn: Callable[[Any], Any] | None = None,
+        supports_deepstack: bool = False,
     ) -> None:
         if spatial_merge_size <= 0:
             raise ValueError(
@@ -59,6 +65,7 @@ class Qwen3VLMultimodalAdapter:
         self._language_model = language_model
         self._embeds_kwarg = embeds_kwarg
         self._embed_tokens_fn = embed_tokens_fn
+        self._supports_deepstack = supports_deepstack
 
     def text_model(self) -> Any:
         """Return the loaded Qwen3-VL language model."""
@@ -72,12 +79,14 @@ class Qwen3VLMultimodalAdapter:
         spatial_merge_size = int(model.config.vision_config.spatial_merge_size)
         embeds_kwarg = cls._detect_embeds_kwarg(language_model)
         embed_tokens_fn = cls._resolve_embed_tokens(language_model)
+        supports_deepstack = cls._detect_deepstack_kwargs(language_model)
         return cls(
             spatial_merge_size=spatial_merge_size,
             vision_tower=vision_tower,
             language_model=language_model,
             embeds_kwarg=embeds_kwarg,
             embed_tokens_fn=embed_tokens_fn,
+            supports_deepstack=supports_deepstack,
         )
 
     @classmethod
@@ -133,6 +142,34 @@ class Qwen3VLMultimodalAdapter:
             f"{cls._SUPPORTED_EMBEDS_KWARGS}; mlx_vlm version drift detected. "
             f"Got parameters: {sorted(params)}"
         )
+
+    @classmethod
+    def _detect_deepstack_kwargs(cls, language_model: Any) -> bool:
+        """Return True iff the LM declares both deepstack params explicitly.
+
+        mlx-vlm 0.5.x Qwen3-VL's ``LanguageModel.__call__`` takes
+        ``visual_pos_masks`` and ``deepstack_visual_embeds`` as named
+        parameters and runs ``_deepstack_process``; 0.4.x's qwen3_5 LM
+        exposes only ``**kwargs``.  Only the explicit form counts as
+        support, so a LM with ``**kwargs`` does not silently absorb
+        deepstack arrays into a dropped catch-all.
+        """
+        try:
+            sig = inspect.signature(language_model.__call__)
+        except (TypeError, ValueError) as exc:
+            raise RuntimeError(
+                f"Cannot inspect language_model.__call__ signature: {exc}"
+            ) from exc
+        explicit = {
+            name
+            for name, param in sig.parameters.items()
+            if param.kind
+            in (
+                inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                inspect.Parameter.KEYWORD_ONLY,
+            )
+        }
+        return all(kw in explicit for kw in cls._DEEPSTACK_KWARGS)
 
     def embed_tokens(self, input_ids: mx.array) -> mx.array:
         """Return the language model's input embeddings for ``input_ids``.
@@ -218,6 +255,9 @@ class Qwen3VLMultimodalAdapter:
         inputs_embeds: mx.array,
         cache: list[Any],
         position_ids: mx.array,
+        *,
+        visual_pos_masks: Any | None = None,
+        deepstack_visual_embeds: Any | None = None,
     ) -> Any:
         """Invoke ``language_model`` with runner-built embeds and positions.
 
@@ -226,6 +266,13 @@ class Qwen3VLMultimodalAdapter:
         embeds keyword is the one sniffed at construction so version drift
         between ``inputs_embeds`` and ``input_embeddings`` surfaces at load
         time instead of inside attention.
+
+        ``visual_pos_masks`` and ``deepstack_visual_embeds`` are forwarded
+        only when the language model declares both deepstack parameters
+        explicitly (mlx-vlm 0.5.x).  On 0.4.x qwen3_5 the LM signature
+        exposes only ``**kwargs``; passing arrays into that catch-all
+        would silently drop them, so the adapter omits the kwargs entirely
+        instead of relying on the LM's tolerance.
         """
         if self._language_model is None:
             raise RuntimeError(
@@ -238,11 +285,16 @@ class Qwen3VLMultimodalAdapter:
                 "Qwen3VLMultimodalAdapter.from_loaded_model so the language "
                 "model signature is sniffed at load time."
             )
+        extra_kwargs: dict[str, Any] = {}
+        if self._supports_deepstack:
+            extra_kwargs["visual_pos_masks"] = visual_pos_masks
+            extra_kwargs["deepstack_visual_embeds"] = deepstack_visual_embeds
         return self._language_model(
             input_ids,
             cache=cache,
             position_ids=position_ids,
             **{self._embeds_kwarg: inputs_embeds},
+            **extra_kwargs,
         )
 
     @staticmethod
--- a/vllm_metal/multimodal/qwen3_vl/adapter.py
+++ b/vllm_metal/multimodal/qwen3_vl/adapter.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 import inspect
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from functools import cache
 from types import SimpleNamespace
@@ -23,7 +23,7 @@ class Qwen3VLVisionEncodeResult:
     """Vision tower output for one Qwen3-VL multimodal feature."""
 
     hidden_states: mx.array
-    deepstack_visual_embeds: Any | None
+    deepstack_visual_embeds: Sequence[mx.array] | None
 
 
 class Qwen3VLMultimodalAdapter:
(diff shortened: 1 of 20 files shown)


ricky-chaoju (changes requested):
Thanks for pushing this forward. I went through the current diff, and I think we should split this before reviewing the implementation details.

The direction looks right, but this has grown beyond the small paged-path PR series I expected after #330/#341. I expected this step to focus on paged runner integration: encoder-output caching by `mm_features.identifier`, embedding splice, M-RoPE positions, and LM forward on the paged KV path.

Right now, the PR also includes the `mlx-vlm>=0.5.0` dependency bump, non-paged forwarding, deepstack wiring, spec-decode M-RoPE segment handling, parity tests, and benchmark notes. That makes it hard to review each layer independently.

The dependency bump should also be handled separately. #169 was accepted as a very small temporary bridge, but this case is riskier: `mlx-vlm>=0.5.0` requires `llguidance>=1.7.0`, while the vLLM release currently installed by vllm-metal is still `0.20.2`, whose release requirements pin `llguidance>=1.3.0,<1.4.0`. vLLM #42150 is merged to main, but it is not part of that release yet.

Could you split the dependency/install-path change from the runtime feature work, and keep this PR focused on one reviewable layer?

scyyh11:
Thanks for the reivew @ricky-chaoju. I've re-scoped the PR:

- Paged runner only: encoder-output caching by `mm_features.identifier`, embedding splice, M-RoPE positions, `_run_mm_paged_forward` through `adapter.call_lm`, and `mrope_position_delta propagation onto `RequestState`. Adapter stays at `forward_ready=False`.
- The following are split out and will land as separate stacked PRs:
    a. Non-paged prefill + batched/sequential decode through `adapter.call_lm`.
    b. Spec-decode mm segment positions + scheduler-time `mm_features` pre-registration.
    c. `mlx-vlm>=0.5.0` dependency bump + `forward_ready=True` flip + Qwen3-VL parity test + benchmark notes. Held until a vLLM release contains [#42150](https://github.com/vllm-project/vllm/pull/42150).

ricky-chaoju (approved):
Looks good after the re-scope. This is now limited to paged-path multimodal plumbing and stays gated behind forward_ready=False.

The remaining non-paged, spec-decode, and mlx-vlm>=0.5 parity work can stay as follow-ups. I’d still like @LxYuan0420 to take one more look before merge.

vllm_metal/v1/model_runner.py
```diff
@@ -1145,9 +1189,21 @@ def _reject_scheduled_encoder_inputs(
...
+        inputs are routed to :meth:`_run_vision_encoders`.  Until Phase 4
+        flips that flag for the parity-tested model, the gate raises so
+        partial work on a new model never disturbs models already in
+        production.  Phase 5+ adapters land at ``False`` and route through
+        this guard until each one's parity test passes.
+        """
         if not scheduled_encoder_inputs:
             return
+        adapter = self._multimodal_adapter
+        if adapter is not None and adapter.forward_ready:
```
LxYuan0420, line 1224: This ordering does not match upstream’s scheduler contract. A new multimodal request can arrive with `scheduled_encoder_inputs` in the same `SchedulerOutput`; upstream registers the request state before executing encoder inputs. Here this dispatch currently happens before `_handle_new_requests()` registers `new_req.mm_features`, so the first encoder step fails as an unregistered request. Please split lightweight MM feature registration from `_handle_new_requests()` and do that before encoder dispatch, while keeping actual model work after the fail-fast checks. Also add an `execute_model()` test for new request + scheduled encoder input in the same step.

vllm_metal/multimodal/qwen3_vl/adapter.py
```diff
@@ -41,26 +73,221 @@ def from_loaded_model(cls, model: Any) -> Qwen3VLMultimodalAdapter:
...
+                "Construct via Qwen3VLMultimodalAdapter.from_loaded_model."
+            )
+        if self._embeds_kwarg is None:
+            raise RuntimeError(
+                "embeds_kwarg not detected; construct via "
+                "Qwen3VLMultimodalAdapter.from_loaded_model so the language "
+                "model signature is sniffed at load time."
+            )
+        extra_kwargs: dict[str, Any] = {}
+        if self._supports_deepstack:
```
LxYuan0420, line 259: This silently drops `deepstack_visual_embeds` when the LM signature does not explicitly support deepstack kwargs. That is too forgiving for a model-output path: if the vision tower produced deepstack residuals but the LM cannot consume them, we should fail fast instead of running with incomplete inputs. Please only omit these kwargs when `deepstack_visual_embeds is None`; otherwise raise a clear mlx-vlm signature mismatch error.

vllm_metal/v1/model_adapter.py
```diff
@@ -25,18 +26,71 @@ class TargetModelForwardOutput:
     hidden_states: mx.array | None = None
 
 
+class MultimodalEncodeResult(Protocol):
+    """Adapter-owned vision output for one multimodal feature."""
+
+    hidden_states: Any
```
LxYuan0420: This protocol is the contract between the adapter and runner, so the fields should be typed instead of `Any`. Can we make this explicit, e.g. `hidden_states: mx.array` and `deepstack_visual_embeds: Sequence[mx.array] | None`? That keeps the runner simple and avoids defensive guessing around adapter-owned outputs in the follow-up PRs.

LxYuan0420:
Thanks for re-scoping this and keeping the paged path gated.

Before merge, can you add a small integration proof for the upstream scheduler shape this PR wires: a new multimodal request with `scheduled_encoder_inputs` in the same `SchedulerOutput`, then paged prefill consumes the cached encoder output. Full Qwen3-VL parity can stay in the follow-up that flips `forward_ready=True`.

vllm_metal/v1/model_adapter.py
scyyh11: `MultimodalEncodeResult.hidden_states` is now `mx.array` and `deepstack_visual_embeds` is `Sequence[mx.array] | None`. Also tightened `Qwen3VLVisionEncodeResult` to match so the implementation doesn't silently widen.

vllm_metal/multimodal/qwen3_vl/adapter.py
scyyh11: `call_lm` now only omits the deepstack kwargs when `deepstack_visual_embeds is None`. A non-None payload reaching a LM whose `__call__` doesn't declare the deepstack kwargs raises a mlx-vlm signature-mismatch RuntimeError instead of dropping the residuals silently.

vllm_metal/v1/model_runner.py
scyyh11: Lifted mm-features registration out of `_handle_new_requests` into `_pre_register_new_request_mm_features`, called from `execute_model` between cleanup and `_reject_scheduled_encoder_inputs`. Per-request prefill scheduling stays in `_handle_new_requests` so the fail-fast checks above it still guard real model work.

vllm_metal/v1/model_runner.py
```diff
@@ -1181,6 +1257,250 @@ def _validate_spec_decode_supported(
...
+            else:
+                offset_arr = np.arange(
+                    segment.cache_start_pos,
+                    segment.cache_start_pos + n,
+                    dtype=np.int32,
+                )
+            seg_positions = mx.broadcast_to(
+                mx.array(offset_arr)[None, None, :], (3, 1, n)
+            )
+            ctx_segment_positions.append(seg_positions if is_mm_decode else None)
```
LxYuan0420, line 1402: `ctx_segment_positions` needs one entry per `prepare_unified()` attention segment. For speculative decode, `num_query_tokens > 1` expands to one segment per query token, but this appends one `(3, 1, n)` entry per `PagedDecodeSegment`. Please append one `(3, 1, 1)` entry per query token, or explicitly reject multimodal speculative decode in this PR. The test should also assert `len(ctx.segment_positions) == len(ctx.cu_seqlens) - 1`.

scyyh11: Broadened the reject in `_run_mm_paged_forward` to raise on any `num_query_tokens > 1` segment whenever the batch is on the mm paged path, regardless of whether the segment itself is mm. Added a coverage test for the mixed mm-prefill + text-spec-decode batch, and pinned `len(ctx.segment_positions) == len(ctx.cu_seqlens) - 1` in the mixed mm+text prefill test so the invariant doesn't silently regress.

LxYuan0420 (approved):

PR #351, opened 2026-05-07
[LoRA] LoRA adapter support on Metal worker

### LINKED ISSUE #326: LoRA adapter support on Metal
The LoRA hooks in the Metal worker currently return `False`. Before I put up a PR wiring them in, is LoRA support something you'd be open to
merging?

narain1:
Closes #326.

## Summary

This PR adds end-to-end LoRA support to the Metal v1 runner by implementing the previously stubbed worker hooks (`add_lora`, `remove_lora`, `pin_lora`, and `list_loras`). With this change, `--enable-lora` and `--lora-modules`.

The implementation introduces a dedicated `vllm_metal/v1/lora/` package that owns all Metal-side LoRA state and routing logic while remaining intentionally low-coupling, 

The runtime uses a Punica-style routing path: a single `PunicaWrapperMLX` builds per-token slot indices once per scheduler step, and wrapped `MLXLinearWithLoRA` layers apply LoRA deltas through a branch-free batched matmul against a fixed slot table.

## Implementation

Added a new `vllm_metal/v1/lora/` package containing:

* `peft_loader.py`

  * Loads PEFT and Unsloth adapters through upstream `PEFTHelper`
  * Converts adapter weights into MLX tensors
  * Serves as the single owner of the on-disk adapter contract

* `model_manager.py`

  * Manages adapter slot tables and lifecycle
  * Wraps eligible `nn.Linear` modules with `MLXLinearWithLoRA`
  * Handles adapter pin/unpin operations
  * Maintains a fixed-capacity `lora_slots` table with a permanent zeroed null slot for non-LoRA tokens

* `worker_manager.py`

  * Exposes the request-facing LoRA API to the worker
  * Resolves `LoRARequest` paths
  * Enforces adapter capacity limits
  * Applies the active adapter set for each scheduler step

* `mapping.py`

  * Implements `LoRAMappingBuilder` for per-token and per-prompt routing construction

* `punica_wrapper.py`

  * Implements gather + batched matmul routing
  * Uses the null slot to keep the execution path branch-free

* `layers.py`

  * Defines the MLX `nn.Linear` LoRA wrapper

Additional integration changes:

* `vllm_metal/v1/worker.py`

  * Replaces stubbed LoRA methods with delegations to `MetalWorkerLoRAManager`

* `vllm_metal/v1/model_runner.py`

  * Initializes the worker LoRA manager when `lora_config` is enabled
  * Activates the scheduler step's LoRA set before model execution via `_activate_step_loras(scheduler_output)`

* `examples/lora.py`

  * Adds an end-to-end PEFT adapter example

## Scope

Supported in this PR:

* AutoPEFT and Unsloth adapters
* Single-rank LoRA adapters
* Up to `--max-loras` active adapters per batch
* Greedy and sampled decoding
* Compatibility with the Metal paged-attention runner

Not included in this PR:

* QLoRA Support
* Fused QKV or MoE LoRA support
* Dynamic rank promotion

## Testing

* `pytest tests/test_lora.py`

  * Covers routing construction, slot indexing, adapter lifecycle management, capacity enforcement, and mixed-batch numerical validation

* `pytest tests/`

* Verified end-to-end behavior using `examples/lora.py` with:

  * `Qwen/Qwen3.5-0.8B`
  * A real HF PEFT adapter
  * Confirmed LoRA output differs from base model output as expected


vllm_metal/v1/model_runner.py
```diff
@@ -979,6 +1052,36 @@ def _sample_paged_batch(
...
+            state = self._request_states.get(req_id)
+            lora_id = state.lora_id if state is not None else None
+            builder.add_request(lora_id, num_tokens)
+            if lora_id is not None:
+                lora_request = self._loaded_lora_requests.get(lora_id)
+                if lora_request is not None:
+                    active_requests.add(lora_request)
+
+        mapping = builder.build() if not builder.is_empty() else None
+        self._lora_manager.set_active_adapters(active_requests, mapping)
```
LxYuan0420: The routing order here looks unsafe. This builds LoRA mapping from `SchedulerOutput` order: new requests first, then cached requests. The paged forward path runs decode tokens first, then prefill tokens, so mixed decode+prefill steps can route tokens to the wrong adapter.

Please derive the mapping from the actual execution batch order instead: decode requests first, then prefill requests. A regression with different LoRA ids on each side would lock the contract.

vllm_metal/v1/model_runner.py
```diff
@@ -84,6 +85,14 @@
 ]
 
 
+def _lora_id_from_request_data(request_data: Any) -> int | None:
+    """Pull the int LoRA ID off a vLLM ``NewRequestData``-shaped record."""
+    lora_request = getattr(request_data, "lora_request", None)
+    if lora_request is None:
+        return None
+    return int(lora_request.lora_int_id)
```
LxYuan0420: This does not need to be shape-defensive. `NewRequestData.lora_request` is part of the upstream vLLM contract, so please type this helper as `NewRequestData` and access `request_data.lora_request` directly.

vllm_metal/v1/lora/peft_loader.py
```diff
@@ -0,0 +1,91 @@
...
+        except ValueError:
+            logger.debug("Skipping unrecognized LoRA tensor: %s", name)
+            continue
+        pairs.setdefault(module, {})["lora_a" if is_a else "lora_b"] = tensor
+
+    weights: dict[str, LoRALayerWeightsMLX] = {}
+    for module, pair in pairs.items():
+        if "lora_a" not in pair or "lora_b" not in pair:
+            logger.warning("LoRA module %s missing lora_a or lora_b; skipping.", module)
+            continue
```
LxYuan0420: I would make adapter loading atomic here. If a module has only `lora_a` or only `lora_b`, that is a malformed adapter file, so please raise a clear `ValueError` instead of skipping the module and continuing with a partial LoRA.

tests/test_lora.py
```diff
@@ -0,0 +1,724 @@
...
+        lora_config=_lora_config_stub(
+            max_loras=1, max_lora_rank=1, target_modules=["fc1"]
+        ),
+        max_num_seqs=1,
+        max_num_batched_tokens=2,
+        dtype=mx.float32,
+    )
+    assert set(manager.modules) == {"fc1"}
+    assert isinstance(model.fc1, layers_mod.MLXLinearWithLoRA)
+    assert not isinstance(model.fc2, layers_mod.MLXLinearWithLoRA)
```
LxYuan0420, line 760: The local LoRA mechanics are covered, but the risky part of this PR is the runner glue. Please add one focused regression that proves LoRA routing follows the actual paged forward order for mixed decode+prefill requests with different LoRA ids.

vllm_metal/v1/model_runner.py
```diff
@@ -378,6 +391,65 @@ def supported_worker_tasks(self) -> tuple[SupportedTask, ...]:
...
+        """Load a LoRA/QLoRA adapter for later request-time activation."""
+        if self._lora_manager is None:
+            logger.warning(
+                "add_lora called but --enable-lora was not passed; ignoring."
+            )
+            return False
+        added = self._lora_manager.add_adapter(lora_request)
+        if added:
+            self._loaded_lora_requests[lora_request.lora_int_id] = lora_request
+        return added
```
LxYuan0420: `model_runner.py` is becoming the owner of too much LoRA-specific logic: manager initialization, loaded-request bookkeeping, public LoRA methods, and per-step mapping construction. This file should stay orchestration-only.

Please move the LoRA runtime policy behind a small owner under `vllm_metal/v1/lora/`. The runner should initialize it and pass the actual execution batch order in; the LoRA owner should handle adapter bookkeeping, activation, and mapping construction.

Expected ownership shape:

MetalModelRunner
  -> orchestration only
  -> owns when to prepare LoRA for a step
  -> does not own LoRA bookkeeping or mapping internals

MetalLoRARuntime
  -> owns loaded LoRA requests
  -> owns add/remove/pin/list adapter operations
  -> owns active adapter selection
  -> builds LoRA mapping from actual execution order
  -> delegates slot-table writes to MetalWorkerLoRAManager

examples/lora.py
LxYuan0420, line 1: I would remove this example from the PR for now. It uses `llm.llm_engine.add_lora(...)`, which is an internal engine boundary, and with `max_num_seqs=1` it only exercises the easiest single-request path. Once the LoRA support is mature and the routing contract is solid, we can add a proper quickstart in `vllm_metal/v1/lora/README.md` that documents the supported path and constraints clearly.

LxYuan0420 (commented):
Thanks for pushing LoRA support forward. The package direction looks reasonable, but I think this needs one more pass before merge. Please move the LoRA runtime policy out of `model_runner.py`, make routing follow the actual runner execution order, add a mixed decode+prefill regression, and paste raw base-vs-LoRA command/output in the PR body.

vllm_metal/v1/model_runner.py
narain1: Fixed. Routing is now built from the actual execution batch in the same order `_start_paged_forward` lays out the unified token sequence decode tokens first (one per decode req from state.lora_id), then prefill tokens (pr.lora_id × len(pr.token_ids)). The prepare_step() call was moved to after `_build_prefill_pack()` and before `_start_paged_forward()`. The non-paged path uses _scheduler_lora_routing() since it has no decode/prefill interleaving.

examples/lora.py
narain1: Removed `examples/lora.py`.

vllm_metal/v1/lora/peft_loader.py
narain1: Done. A module with only lora_a or only lora_b now raises ValueError instead of being skipped. Covered by `test_peft_loader_rejects_partial_adapter`.

narain1:
Thanks for the thorough feedback. Addressed all points.

vllm_metal/v1/model_runner.py
```diff
@@ -1561,6 +1630,9 @@ def execute_model(
                 "to use structured output."
             )
 
+        if self._paged_attention_backend is None:
+            self._lora.prepare_step(self._scheduler_lora_routing(scheduler_output))
```
LxYuan0420: We should not enable LoRA on the non-paged path in this PR. The current mapping is built once for the whole scheduler step, but non-paged execution still runs multiple separate forwards in `_prefill_single()` / `_sequential_decode()`. Punica routing is row-aligned to the current forward tensor, so a step-level mapping can misalign or broadcast-fail. Please fail fast for LoRA when paged attention is disabled, and keep this PR scoped to the paged path.

vllm_metal/v1/model_runner.py
```diff
@@ -1113,6 +1151,32 @@ def _sample_paged_batch(
...
         return batch, scheduler_output
 
+    def _paged_lora_routing(
+        self,
+        decode_reqs: list[tuple[str, RequestState]],
+        prefill_pack: list[PrefillRequest],
+    ) -> list[tuple[int | None, int]]:
+        entries = []
+        for _, state in decode_reqs:
+            entries.append((state.lora_id, 1))
```
LxYuan0420: This must account for speculative decode rows. Paged LoRA routing currently uses `1` row per decode request, but speculative decode forwards `len(PagedDecodeSegment.input_token_ids)` rows for that request. Please build decode-side LoRA routing from the same `PagedDecodeSegment` metadata used by `_start_paged_forward()`, or explicitly reject LoRA + speculative decode until the row contract is implemented.

vllm_metal/v1/lora/runtime.py
```diff
@@ -0,0 +1,107 @@
...
+    def prepare_step(self, routing_entries: Iterable[tuple[int | None, int]]) -> None:
+        # Push the per-step active set and token routing to the manager.
+        if self._manager is None:
+            return
+        builder = LoRAMappingBuilder()
+        active_requests: set[LoRARequest] = set()
+        for lora_id, num_tokens in routing_entries:
+            builder.add_request(lora_id, num_tokens)
+            if lora_id is not None:
+                req = self._loaded.get(lora_id)
```
LxYuan0420: Please fail here if a non-null `lora_id` is not loaded. Mapping an unknown requested LoRA to the null slot produces base-model output while the request asked for an adapter, which is a silent correctness failure. This should match the upstream contract: requested LoRAs must be loaded before scheduling/use.

vllm_metal/v1/lora/worker_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        adapter = load_peft_adapter(
+            get_adapter_absolute_path(lora_request.lora_path),
+            lora_id=lora_request.lora_int_id,
+            max_position_embeddings=self.max_position_embeddings,
+            lora_config=self.lora_config,
+        )
+        if not self._mm.add_adapter(adapter):
+            return False
+        try:
+            self._mm.activate_adapter(lora_request.lora_int_id)
```
LxYuan0420: We should not pretend Metal supports the full upstream `max_cpu_loras > max_loras` cache semantics yet. This implementation immediately activates every added adapter, so CPU-loaded-but-not-active adapters are not actually supported. Please either implement the upstream separation between loaded adapters and active slots, or enforce/document `max_cpu_loras == max_loras` for Metal in this PR.

vllm_metal/v1/lora/model_manager.py
```diff
@@ -0,0 +1,197 @@
...
+        for name, w in self.modules.items():
+            mw = _lookup_weights_for_module(adapter, name)
+            if mw is None:
+                w.reset_lora(slot)
+                continue
+            # Fold scaling into B once so the kernel stays scale=1.0.
+            w.set_lora(slot, mw.lora_a, mw.lora_b * mw.scaling)
+            loaded += 1
+        if loaded == 0:
+            logger.warning("LoRA %d activated but matched 0 wrapped modules.", lora_id)
```
LxYuan0420: This should be a hard failure, not a warning. If an adapter matches zero wrapped modules, activating it successfully means the request will run as the base model while reporting LoRA success. Please fail add/activation when no adapter weights match the wrapped model modules, with a regression test.

vllm_metal/v1/lora/layers.py
```diff
@@ -0,0 +1,77 @@
...
+                f"LoRA weight shape mismatch for slot {slot}: A=({rank},{in_}), "
+                f"B=({out},{rank}); expected in={self.input_size}, out={self.output_size}"
+            )
+        if rank > self.max_lora_rank:
+            raise ValueError(
+                f"LoRA rank {rank} exceeds max_lora_rank {self.max_lora_rank}"
+            )
+        a = mx.zeros_like(self.lora_a_stacked[slot])
+        b = mx.zeros_like(self.lora_b_stacked[slot])
+        a[:rank, :] = lora_a.astype(a.dtype)
```
LxYuan0420, line 83: Please validate the full A/B shape contract before writing the slot: `A=(rank, in)`, `B=(out, rank)`, and both ranks must match. These tensors come from disk, so malformed adapter weights should fail at the LoRA boundary with a clear error, not through slice assignment or later matmul behavior.

LxYuan0420 (changes requested):

vllm_metal/v1/lora/runtime.py
narain1: Done, I will handle the mapping error in the prepare step where the lora is mapped if it is not present.

vllm_metal/v1/lora/layers.py
narain1: fixed this, these checks are now handled in `set_lora`

vllm_metal/v1/lora/worker_manager.py
narain1: `MetalWorkerLoRAManager.__init__` raises `NotImplementedError` when max_cpu_loras is set and != max_loras.

vllm_metal/v1/model_runner.py
narain1: for the scope of this pr `MetalLoRARuntime.setup()` fails with `NotImplementedError` when paged attention is disabled.

> Regression: test_runtime_rejects_lora_without_paged_attention

ricky-chaoju:
@narain1 Please resolve the conflict.

vllm_metal/v1/lora/worker_manager.py
```diff
@@ -0,0 +1,101 @@
...
+        self.max_position_embeddings = max_position_embeddings
+        self._mm = MLXLoRAModelManager(
+            model, lora_config, max_num_seqs, max_num_batched_tokens, dtype
+        )
+
+    def add_adapter(self, lora_request: LoRARequest) -> bool:
+        from vllm.lora.utils import get_adapter_absolute_path
+
+        if lora_request.lora_int_id in self._mm.list_adapters():
+            return False
```
LxYuan0420, line 52: I think we still need to preserve the upstream `load_inplace` behavior here. If a caller reloads the same LoRA ID with `load_inplace=True`, this returns `False` and keeps the old weights. Can you mirror upstream’s flow here: only return early when the ID is already loaded and `load_inplace` is false; otherwise load the new adapter, remove the old one, activate the replacement, and add a small test?

vllm_metal/v1/lora/worker_manager.py
```diff
@@ -0,0 +1,101 @@
...
+    def pin_adapter(self, lora_id: int) -> bool:
+        return self._mm.pin_adapter(lora_id)
+
+    def list_adapters(self) -> set[int]:
+        return self._mm.list_adapters()
+
+    def set_active_adapters(
+        self, lora_requests: set[LoRARequest], mapping: LoRAMapping | None
+    ) -> None:
+        if lora_requests:
```
LxYuan0420: Small lifecycle issue: when a batch has no LoRA requests, we skip `_apply()`, so previously active adapters can stay in slots. The null-slot mapping keeps outputs correct, but the stale active slots can block a later adapter load. Can we always call `_apply({r.lora_int_id for r in lora_requests})`, including the empty set?

vllm_metal/v1/lora/model_manager.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    if (w := adapter.weights.get(module_name)) is not None:
+        return w
+    suffix = "." + module_name
+    cands = [
+        w
+        for n, w in adapter.weights.items()
+        if n.endswith(suffix) or module_name.endswith("." + n)
+    ]
+    return cands[0] if len(cands) == 1 else None
```
LxYuan0420: Can we fail loudly on ambiguous suffix matches here? Returning `None` means the adapter can partially load while silently skipping this module, which would be very hard for users to diagnose. If more than one candidate matches, please raise a clear error and add a small test for that case.

vllm_metal/v1/lora/worker_manager.py
narain1: retained the expected `load_inplace` logic and also added tests for the same.

narain1:
Hi @LxYuan0420, I’ve pushed updates addressing the requested changes:

- LoRA now fails fast when paged attention is disabled.
- Unknown/non-loaded LoRA IDs now fail instead of silently routing to the null slot.
- Metal now rejects unsupported `max_cpu_loras != max_loras` semantics.
- Adapter activation now fails when no weights match wrapped modules.
- Empty LoRA batches now clear active slots via `_apply(set())`.
- Ambiguous suffix matches now raise a clear error, with tests added.

LxYuan0420:
I pushed a few fixes on top of this branch:

- made `load_inplace=True` replacement transactional, so a bad reload does not drop the old adapter
- aligned Metal `pin_lora` behavior with upstream cache-tier semantics
- preserved the wrapped `nn.Linear` surface needed by paged attention
- registered scheduled offline LoRA requests before paged routing

I validated with ruff, `tests/test_lora.py`, and a local paged-attention LoRA smoke on `Qwen/Qwen3-0.6B`.

Could you also test this with your original LoRA adapter / intended model path before we merge?

LxYuan0420 (commented):
Requesting other to test out this branch before merge.

YuvrajSingh-mist:
I am just waiting for it to be merged for long. It could really help with my vllm's restarts

LxYuan0420 (approved):

PR #315, opened 2026-04-29
[WIP] [Speculative Decoding] Support Draft-Model Speculative Decoding on Metal

davidgao7:
# Summary

This PR introduces the foundational architecture for Speculative Decoding on the Metal paged path. It implements a Widened Scorer approach that leverages MLX's unified varlen kernels to verify multiple draft tokens in a single, efficient forward pass.

# Key Changes

* **MetalSpecDecodeWorker**: New orchestrator that manages the Propose-Score-Verify-Rewind loop and stashes speculative data for the asynchronous Paged Attention path.
* Full Worker Delegation: Implemented missing *WorkerBase* RPC methods (*take_draft_token_ids*, *get_supported_tasks*, health checks) to ensure full compatibility with the vLLM EngineCore and benchmarking tools.
* MLX-Native **MetalDraftProposer**: A functional proposer using a secondary model lifecycle that leverages MLX lazy evaluation to build draft sequences without GPU stalls.
* Vectorized **MetalRejectionSampler**: A parallelized rejection sampler with the "residual fix" (*mx.maximum*) to ensure mathematical parity with sequential generation.
* Atomic KV Rewind: Refactored *ModelRunner* to support *truncate_cache* with bonus token commitment, ensuring the KV cache and sequence length tracking stay in sync after rejections.
* Vectorized Logit Extraction: Updated *sampling_batch.py* to extract and return "Widened Pipe" verification logits for the entire speculative window.
* macOS Stability: Automated loopback networking defaults (*lo0* / *VLLM_HOST_IP*) to prevent Gloo distributed hangs on Apple Silicon.

# Performance Results (M4 Max)

* **Sequential (1 token)**: 4.92ms

* **Widened Verification (5 tokens)**: 6.65ms

* **3.7x** hardware-level throughput increase during verification.

# Edge Case Handling
To prevent [state-sync issues](https://github.com/vllm-project/vllm/issues/34650), this PR implements:
1. Strict Boundary Control: The *MetalModelRunner* now decrementally updates sequence tracking, forcing the GPU to ignore rejected tokens in subsequent passes.
2. Verification Stashing: Added *_pending_spec_data* in the worker to handle the handshake between the asynchronous Scorer pass and the verification phase.

# Validation & Test Plan
Hardware:Apple MacBook Pro 16 (2024) with M4 Max (36GB RAM):
Verified using the vLLM CLI (v0.19.0) and internal integration tests. Hardware-saturated benchmarks confirm the Scorer's efficiency gain.

[...]

PR #358, opened 2026-05-08
Wire structured-output row spans into paged path

CAICAIIs:
## Summary

This PR wires row-span-aware structured-output bitmasking into the Metal paged sampling path.

It builds on the paged decode segment metadata and keeps the row-span mapping, constrained-row construction, and bitmask application owned by `MetalStructuredOutputApplier`.

## What Changed

- Pass paged `decode_segments` from `MetalModelRunner._sample_paged_batch()` into `MetalStructuredOutputApplier.apply_paged()`.
- Map structured-output requests to one or more logits rows when paged speculative decode row spans are present.
- Apply per-position grammar bitmask rows for same-request structured output + scheduled speculative decode.
- Preserve the existing non-spec structured-output behavior.
- Add production/class-owned path coverage for:
  - normal decode structured output
  - prefill tail-row masking
  - mixed structured/plain speculative batches
  - request-id isolation
  - same-request structured output + scheduled speculative decode
  - missing per-position row-span bitmask rows

## Scope

This PR does not add a drafter, rejection sampling, KV rewind, or broader speculative decoding policy. It only wires structured-output row-span bitmasking into the existing Metal paged target path.

## Validation

- `pytest tests/test_grammar_bitmask.py tests/test_v1_model_runner_generate.py tests/test_spec_decode_metadata.py -q`
  - `66 passed`
- `ruff check` / `ruff format --check` on touched files
- `git diff --check`
- Real Metal smoke with `vllm 0.20.2`: `Qwen/Qwen3-0.6B` + paged attention + structured JSON output returned parseable JSON: `{"answer": 1}`

### DIFF
--- a/tests/test_grammar_bitmask.py
+++ b/tests/test_grammar_bitmask.py
@@ -252,7 +252,7 @@ class TestApplyGrammarBitmaskPaged:
 
     def test_row_span_metadata_masks_verification_rows_and_prefill_tail(self) -> None:
         """Row-span metadata must mask structured rows while leaving plain spec rows alone."""
-        allowed_decode = 7
+        allowed_decode_rows = (7, 8, 9)
         allowed_prefill = 15
         decode_reqs = [_make_decode_req("a"), _make_decode_req("b")]
         prefill_reqs = [_make_prefill_req("pf0", 2)]
@@ -267,11 +267,14 @@ class TestApplyGrammarBitmaskPaged:
         sched = _make_scheduler_output(["a", "b", "pf0"])
         sched.scheduled_spec_decode_tokens = scheduled_spec_decode_tokens
         grammar = _make_grammar_output(
-            ["pf0", "a"],
+            ["pf0", "a", "a", "a"],
             np.vstack(
                 [
                     _make_single_token_bitmask(allowed_prefill),
-                    _make_single_token_bitmask(allowed_decode),
+                    *(
+                        _make_single_token_bitmask(token)
+                        for token in allowed_decode_rows
+                    ),
                 ]
             ),
         )
@@ -289,10 +292,10 @@ class TestApplyGrammarBitmaskPaged:
             )
         )
 
-        # Verification rows for a live at rows 0 and 1.
-        for row in [0, 1]:
-            assert np.isfinite(result[0, row, allowed_decode])
-            assert result[0, row, (allowed_decode + 1) % VOCAB_SIZE] == float("-inf")
+        # Verification rows for a live at rows 0 and 1; row 2 is its bonus row.
+        for row, allowed_token in enumerate(allowed_decode_rows):
+            assert np.isfinite(result[0, row, allowed_token])
+            assert result[0, row, (allowed_token + 1) % VOCAB_SIZE] == float("-inf")
 
         # Plain speculative decode rows for b stay untouched.
         np.testing.assert_array_equal(result[0, 3:5], _to_numpy(logits)[0, 3:5])
@@ -303,8 +306,8 @@ class TestApplyGrammarBitmaskPaged:
 
     def test_row_span_metadata_keeps_request_bitmasks_isolated_by_req_id(self) -> None:
         """Structured-output rows must stay attached to their request IDs."""
-        allowed_a = 3
-        allowed_b = 19
+        allowed_a_rows = (3, 4, 5)
+        allowed_b_rows = (19, 20)
         decode_reqs = [_make_decode_req("a"), _make_decode_req("b")]
         scheduled_spec_decode_tokens = {"a": [101, 102], "b": [201]}
         segments = build_paged_decode_segments(
@@ -317,11 +320,11 @@ class TestApplyGrammarBitmaskPaged:
         sched = _make_scheduler_output(["a", "b"])
         sched.scheduled_spec_decode_tokens = scheduled_spec_decode_tokens
         grammar = _make_grammar_output(
-            ["b", "a"],
+            ["b", "b", "a", "a", "a"],
             np.vstack(
                 [
-                    _make_single_token_bitmask(allowed_b),
-                    _make_single_token_bitmask(allowed_a),
+                    *(_make_single_token_bitmask(token) for token in allowed_b_rows),
+                    *(_make_single_token_bitmask(token) for token in allowed_a_rows),
                 ]
             ),
         )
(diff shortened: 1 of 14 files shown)


vllm_metal/v1/structured_output.py
```diff
@@ -22,6 +24,187 @@
...
+
+    # rows_torch is CPU float32 (from torch.from_numpy), so torch_to_mlx goes
+    # through numpy — all xgrammar mutations are captured before the copy.
+    rows_mlx = torch_to_mlx(rows_torch).astype(original_dtype)  # (n_constrained, vocab)
+
+    # logits[0] produces a new lazy computation node (not a Python alias of
+    # logits), so __setitem__ here does not mutate the caller-held logits array.
+    result_2d = logits[0]
+    result_2d[logit_rows] = rows_mlx
+    return result_2d[None]  # Restore (1, total_tokens, vocab) shape
```
LxYuan0420: This moves the core `MetalStructuredOutputApplier` behavior out into module-private helpers. These are not general structured-output utilities; they are the applier’s row-target mapping and bitmask-application internals. Please keep this behavior owned by `MetalStructuredOutputApplier` instead of spreading it across loose private functions at module scope.

Suggested shape:

```python
class MetalStructuredOutputApplier:
    def apply_paged(...): ...

    def _build_paged_row_targets(...): ...
    def _build_constrained_rows(...): ...
    def _apply_grammar_bitmask_to_rows(...): ...
```

vllm_metal/v1/structured_output.py
```diff
@@ -104,67 +288,20 @@ def apply_paged(
...
-                constrained.append((req_id_to_row[req_id], bitmask_row))
-
-        if not constrained:
-            return logits
+        # Each tuple is (logit_row, bitmask_row).
+        constrained = _build_constrained_rows(
+            grammar_output.structured_output_request_ids,
+            row_targets,
+            grammar_bitmask_row_count=grammar_bitmask.shape[0],
+        )
```
LxYuan0420, line 123: This introduces a row-span contract that the production `apply_paged()` path cannot consume yet. `model_runner` still builds one decode row per request, and this method still rejects same-request structured output plus speculative decoding before calling the mapping path. Since this PR is explicitly preparation only, I would rather not land dormant runtime logic. Please wire it through the real class-owned path in the PR that uses it, or keep it out until then.

vllm_metal/v1/structured_output.py
```diff
@@ -22,6 +24,187 @@
...
+                constrained.append((row, bitmask_row + row_offset))
+        bitmask_row = end_bitmask_row
+
+    if bitmask_row != grammar_bitmask_row_count:
+        raise ValueError(
+            "Grammar bitmask row count must match row-span logits targets: "
+            f"consumed {bitmask_row}, got {grammar_bitmask_row_count}."
+        )
+
+    return constrained
```
LxYuan0420: This row-span bitmask consumption logic should stay behind the same owner boundary as `apply_paged()`. Upstream vLLM already defines the `GrammarOutput` ordering and speculative bitmask expansion contract; Metal should adapt that contract in one class-owned place, not through standalone private functions that tests call directly.

tests/test_grammar_bitmask.py
```diff
@@ -140,6 +143,36 @@ def _build_cu_seqlens(num_decode: int, prefill_lens: list[int]) -> list[int]:
...
+        grammar.structured_output_request_ids,
+        row_targets,
+        grammar_bitmask_row_count=grammar.grammar_bitmask.shape[0],
+        paged_decode_segments=segments,
+    )
+    return so._apply_grammar_bitmask_to_rows(
+        logits,
+        grammar.grammar_bitmask,
+        constrained,
+    )
```
LxYuan0420: These tests bypass the production owner by calling `so._build_paged_row_targets()`, `so._build_constrained_rows()`, and `so._apply_grammar_bitmask_to_rows()` directly. That proves helper mechanics, not the `MetalStructuredOutputApplier` contract. Please exercise a class-owned method instead.

tests/test_grammar_bitmask.py
```diff
@@ -248,6 +281,126 @@ def test_mixed_decode_and_prefill(self) -> None:
...
+        with pytest.raises(NotImplementedError, match="one grammar bitmask row"):
+            _apply_row_span_grammar(
+                grammar,
+                decode_reqs,
+                [],
+                cu,
+                1,
+                logits,
+                segments,
+            )
```
LxYuan0420, line 458: These row-span tests validate the prepared private-helper path, but production `apply_paged()` still rejects same-request structured output plus speculative decoding. I do not want prep-only tests that pass by bypassing the runtime boundary. This coverage should land with the runtime PR that actually makes the row-span path reachable through the production owner.

LxYuan0420 (commented):
Thanks for splitting this down. The direction makes sense, but I want this to land as a real production slice, not prep-only helpers. Please keep the current SO path unchanged here, and move the row-span work into the PR that wires SO + speculative decoding through `MetalStructuredOutputApplier`. That keeps ownership clear and tests on the production path.

CAICAIIs:
> Thanks for splitting this down. The direction makes sense, but I want this to land as a real production slice, not prep-only helpers. Please keep the current SO path unchanged here, and move the row-span work into the PR that wires SO + speculative decoding through `MetalStructuredOutputApplier`. That keeps ownership clear and tests on the production path.

Thanks for the review! I understand the concern now. The current prep-only helper slice is not the right shape: the production path is still unreachable, and the tests bypass the MetalStructuredOutputApplier ownership boundary.

I’ll rework #358 into a real runtime slice: keep the current non-spec structured-output path unchanged, move the row-span mapping / constrained-row construction / bitmask application back under MetalStructuredOutputApplier as class-owned private methods, and make the tests exercise the class-owned or production path. I’ll also avoid duplicating #315’s DraftModelProposer / SpecDecodeWorker / rejection sampling / KV rewind scope; if #315 provides the final SD runtime contract, this PR will build the SO row-span bitmasking on top of that.

ricky-chaoju:
Please resolve the conflict.

CAICAIIs:
Thanks for the review. I resolved the conflict and reworked this PR from a prep-only helper slice into a reachable production slice.

Main changes:
- `MetalModelRunner._sample_paged_batch()` now passes paged `decode_segments` into `MetalStructuredOutputApplier.apply_paged()`.
- The row-span request-to-logits-row mapping, constrained-row construction, and bitmask application are now owned by `MetalStructuredOutputApplier`.
- Same-request structured output + scheduled speculative decode is now covered through the class-owned / production sampling path instead of standalone helper tests that bypass the runtime boundary.
- The existing non-spec structured-output behavior is preserved.

Validation:
- `pytest tests/test_grammar_bitmask.py tests/test_v1_model_runner_generate.py tests/test_spec_decode_metadata.py -q` -> 66 passed
- `ruff check` / `ruff format --check` on touched files -> passed
- `git diff --check` -> passed
- Real Metal smoke with `vllm 0.20.2`: `Qwen/Qwen3-0.6B` + paged attention + structured JSON output returned parseable JSON: `{"answer": 1}`

LxYuan0420 (approved):
Great work on this. I pushed a few small maintainer edits on top:

- Fail fast when row-span grammar output references a request with no paged logits rows, with a regression test.
- Document the row-span bitmask ordering contract from vLLM.
- Remove the redundant private `xgrammar` guard; `apply_paged()` owns that dependency check.

CAICAIIs:
> Great work on this. I pushed a few small maintainer edits on top:
> 
> * Fail fast when row-span grammar output references a request with no paged logits rows, with a regression test.
> * Document the row-span bitmask ordering contract from vLLM.
> * Remove the redundant private `xgrammar` guard; `apply_paged()` owns that dependency check.

Thanks!!

PR #365, opened 2026-05-10
Add target verification for scheduled spec decode tokens

ricky-chaoju:
This adds target-side verification for scheduler-provided speculative decode tokens on the Metal paged decode path.

The PR does not add a drafter/assistant model yet. It only teaches the target model path how to consume `scheduled_spec_decode_tokens`, run verification rows for `[last_token, *draft_tokens]`, and emit the accepted draft tokens plus the bonus token when applicable.

- Add `vllm_metal/v1/spec_decode.py` for paged spec-decode metadata and greedy verification helpers.
- Extend `prepare_unified()` so decode requests can describe multiple verification rows.
- Update paged sampling to append multiple emitted tokens and keep paged sequence lengths in sync.
  - hybrid GDN models
  - non-greedy sampling / penalties / logprobs / constraints
  - logits processors that can change argmax
  - invalid draft-token sentinels
- Add focused tests for mixed plain/spec batches, accept/reject behavior, bonus tokens, structured-output interaction, and metadata row spans.

### DIFF
--- a/tests/test_v1_model_runner_generate.py
+++ b/tests/test_v1_model_runner_generate.py
@@ -4,6 +4,10 @@ from __future__ import annotations
 
 from types import SimpleNamespace
 
+import mlx.core as mx
+import numpy as np
+import pytest
+from vllm.sampling_params import SamplingParams
 from vllm.v1.outputs import ModelRunnerOutput
 
 import vllm_metal.v1.model_runner as mr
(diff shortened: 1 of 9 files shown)


vllm_metal/v1/model_runner.py
```diff
@@ -1363,23 +1422,34 @@ def execute_model(
         self._free_encoder_outputs(scheduler_output.free_encoder_mm_hashes)
         evicted_req_ids = self._finished_req_ids(scheduler_output)
         has_scheduled_encoder_inputs = bool(scheduler_output.scheduled_encoder_inputs)
+        unsupported_spec_decode_reason = self._unsupported_spec_decode_reason(
```
LxYuan0420: I do not want speculative-decoding policy to keep growing inside `MetalModelRunner`. Move the spec-specific checks and behavior behind a spec-decode owner. The runner should call behavior on that owner, not maintain policy booleans inline.

Suggested direction:

```python
class SpeculativeDecodeController:
    def validate_supported(...) -> None: ...
    def build_decode_segments(...) -> tuple[PagedDecodeSegment, ...]: ...
    def verify_greedy(...) -> list[list[int]]: ...
```

vllm_metal/v1/spec_decode.py
```diff
@@ -81,7 +100,122 @@ def build_paged_decode_segments(
     return tuple(segments)
 
 
+def has_scheduled_spec_decode_tokens(
```
LxYuan0420: Spec decode needs one class-owned boundary here. Segment building, support validation, and verification are one runtime contract, but this module currently exposes them as separate helpers. Please put them under the same `SpeculativeDecodeController` owner.

vllm_metal/v1/spec_decode.py
```diff
@@ -81,7 +100,122 @@ def build_paged_decode_segments(
...
+    return None
+
+
+def validate_greedy_spec_decode_sampling(
+    decode_reqs: Sequence[tuple[str, _SpecDecodeRequestStateLike]],
+    *,
+    logitsprocs: Any | None = None,
+) -> None:
+    """Raise if this batch needs sampling behavior beyond greedy argmax."""
+    if getattr(logitsprocs, "non_argmax_invariant", ()):
```
LxYuan0420: This gate is too broad. Upstream installs `MinTokensLogitsProcessor` when `speculative_config` is enabled, so `non_argmax_invariant` can be non-empty even when the verified requests do not actively need `min_tokens`. Gate on active request behavior, not processor presence. Also avoid `getattr`; this is typed internal state.

vllm_metal/v1/spec_decode.py
```diff
@@ -62,6 +76,11 @@ def build_paged_decode_segments(
         token_ids = tuple(state.token_ids)
         block_ids = tuple(state.block_ids)
         draft_token_ids = tuple(spec_tokens.get(req_id, ()))
+        if any(token_id < 0 for token_id in draft_token_ids):
```
LxYuan0420: Upstream can hand off invalid draft slots as `-1` plus `num_invalid_spec_tokens`. PR1 needs an explicit policy for that shape: either handle it according to upstream semantics, or fail fast with a targeted unsupported-case error.

vllm_metal/v1/model_runner.py
```diff
@@ -892,27 +927,39 @@ def _sample_paged_batch(
...
-            logits,
-            decode_reqs,
-            num_decode,
-            self._sampler,
-            self.device,
-            vocab_size=vocab_size,
-            logitsprocs=logitsprocs,
-        )
+        decode_logprobs: LogprobsLists | None = None
+        if has_scheduled_drafts:
```
LxYuan0420, line 924: This turns “drafted requests verify greedily” into “the whole decode batch is greedy-only” whenever any request has draft tokens. Either enforce that invariant explicitly, or route per request so draft-free decode requests keep the existing sampler path.

vllm_metal/v1/model_runner.py
```diff
@@ -786,26 +794,47 @@ def _start_paged_forward(
...
+        num_decode_tokens = sum(segment.num_query_tokens for segment in decode_segments)
+        has_scheduled_drafts = any(
+            segment.draft_token_ids for segment in decode_segments
+        )
+        if has_scheduled_drafts and self.is_hybrid:
+            raise NotImplementedError(
+                "Speculative decode verification is not supported for hybrid "
+                "GDN models on Metal yet."
+            )
+        if has_scheduled_drafts:
```
LxYuan0420: Unsupported spec-decode cases should fail before runner state is mutated. Some rejects still happen after `_handle_new_requests()` / `_update_cached_request_blocks()` can change request state and block mappings. Move preflight earlier or make setup rollback-safe.

vllm_metal/v1/spec_decode.py
```diff
@@ -62,6 +76,11 @@ def build_paged_decode_segments(
         token_ids = tuple(state.token_ids)
         block_ids = tuple(state.block_ids)
         draft_token_ids = tuple(spec_tokens.get(req_id, ()))
```
LxYuan0420: Please make the scheduler handoff contract strict here. Every `req_id` in `scheduled_spec_decode_tokens` should belong to the current decode set, and the verification span should match scheduler accounting. Fail fast on mismatches instead of silently accepting or ignoring them.

tests/test_v1_model_runner_generate.py
```diff
@@ -129,16 +133,471 @@ def test_returns_none_when_no_pending_output_and_not_async(self) -> None:
...
+        )
+
+        assert captured["input_ids"] == [[6, 7, 8]]
+        assert captured["decode_info"] == [([0, 1], 1, 3)]
+        assert captured["prefill_info"] == []
+        assert captured["block_size"] == 4
+        assert runner._execute_model_state is not None
+        assert runner._execute_model_state.cu_seqlens == [0, 3]
+
+    def test_start_paged_forward_rejects_non_argmax_logits_processors(
```
LxYuan0420: The tests need to cover the contracts this PR is defining, not only the happy-path stubs. Please add regressions for real `build_logitsprocs(... speculative_config=...)`, invalid draft slots (`-1` / `num_invalid_spec_tokens`), mixed-batch behavior, and no runner-state mutation after unsupported setup fails.

LxYuan0420 (commented):
Nice first slice for the Gemma 4 MTP roadmap. The direction makes sense, and the staged rollout is the right way to keep this reviewable. I left a few comments around contract and ownership. Once those are tightened, this should be a much stronger foundation for the next MTP PRs

ricky-chaoju:
Thanks, I moved this behind SpeculativeDecodeController and added the contract/mixed-batch/fail-fast regressions.

LxYuan0420 (approved):

PR #366, opened 2026-05-12
Bump vLLM to 0.20.2

ricky-chaoju:
Bump vLLM to 0.20.2

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -182,7 +182,7 @@ EOF
   local vllm_v="0.20.2"
   local url_base="https://github.com/vllm-project/vllm/releases/download"
   local filename="vllm-$vllm_v.tar.gz"
-  curl -fOL $url_base/v$vllm_v/$filename
+  curl -OL $url_base/v$vllm_v/$filename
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -179,10 +179,10 @@ EOF
 
   ensure_venv "$venv"
 
-  local vllm_v="0.20.1"
+  local vllm_v="0.20.2"
   local url_base="https://github.com/vllm-project/vllm/releases/download"
   local filename="vllm-$vllm_v.tar.gz"
-  curl -OL $url_base/v$vllm_v/$filename
+  curl -fOL $url_base/v$vllm_v/$filename
```
LxYuan0420: Can we keep this as a pure version bump (just like your earlier chore-version-bump PRs) ? Adding `-f` is separate download hardening, and it is incomplete here because the script does not check the `curl` exit code before continuing to `tar`/`cd`.

LxYuan0420 (approved):
Just undo the `-f` and we are good to go

PR #367, opened 2026-05-12
[kernel] accelerate tiled prefilling kernel [1/3]

WindChimeRan:
## Motivation

problem: slow prefilling for long prefilling

root cause: One Query Token Per Threadgroup

```
Original kernel: 1 query token per threadgroup

Grid: (num_heads, total_q_tokens, 1)

  Threadgroup 0          Threadgroup 1          Threadgroup 2
  ┌───────────┐          ┌───────────┐          ┌───────────┐
  │ q_token 0 │          │ q_token 1 │          │ q_token 2 │
  │           │          │           │          │           │
  │ dot(Q,K)  │          │ dot(Q,K)  │          │ dot(Q,K)  │
  │ for each  │          │ for each  │          │ for each  │
  │ KV token  │          │ KV token  │          │ KV token  │
  └───────────┘          └───────────┘          └───────────┘
       ▲                      ▲                      ▲
       │                      │                      │
       └──────── Each TG loads the FULL KV context ──┘
                 independently (bandwidth waste)
```

For decode, where every sequence contributes exactly 1 query token, this is ideal: the entire threadgroup's bandwidth is focused on streaming through that token's KV context.

For prefill, this is catastrophic. An 8192-token prompt generates 8192 threadgroups per head, each independently streaming the entire KV cache from DRAM. The K and V tensors are loaded 8192 times each. On Apple Silicon's unified memory (bandwidth ~200-400 GB/s), this turns a compute-bound matmul into a bandwidth-bound disaster.

## Summary

**Add `paged_attention_tiled` — Flash-Attention-style tiled kernel for prefill**

Adds a new Metal kernel (`paged_attention_tiled`) alongside the existing `paged_attention` kernel. Uses 8×8 `simdgroup_multiply_accumulate` for QK and PV matmuls, processing BQ=8 query tokens per threadgroup instead of 1.

**Dispatch**: `dispatch_paged_attention_v2_online` routes to the tiled kernel when `total_q_tokens > num_seqs` (batch contains prefill). Pure-decode batches use the original kernel. Gated by `can_use_tiled_kernel` (HEAD_SIZE ∈ {64,96,128,192,256,512}, half/bfloat16, non-TQ, non-FP8).

**Files changed**: `kernels_v2/pagedattention_tiled.metal` (new kernel + instantiations), `paged_ops.cpp` (dispatch + shared `bind_paged_attn_buffers` helper).

## Kernel Dispatch Split

The tiled MMA kernel activates when the batch contains any prefill tokens (`total_q_tokens > num_seqs`), matching vLLM Triton's 2D/3D mode split. Pure-decode batches route to the original per-token kernel. This gives the MMA prefill speedup during chunked prefill scheduler ticks while preserving the original kernel's decode latency and numerical behavior for decode-only ticks.

## 8K Prefill Benchmark Results

```
vllm bench serve --backend vllm --base-url http://127.0.0.1:8000 --model Qwen/Qwen3-0.6B --endpoint /v1/completions --dataset-name sonnet --dataset-path sonnet.txt --sonnet-input-len 8192 --sonnet-output-len 10 --num-prompts 10 --request-rate 0.5 --max-concurrency 2 
```

| Metric                 | main   | tiled (no split) | **dispatch split** | Change vs main |
| ---------------------- | -----: | ---------------: | -----------------: | -------------: |
| Total tok/s            | 142.40 | 244.53           | **243.68**         | **+71%**       |
| Benchmark duration (s) | 572.96 | 333.66           | **334.82**         | **-42%**       |
| Mean TTFT (ms)         | 71,910 | 40,623           | **41,381**         | **-42%**       |
| Mean TPOT (ms)         | 4,687  | 2,838            | **2,787**          | **-41%**       |

### Benchmark: sonnet 1024+128, 100 prompts, concurrency 32

```
vllm bench serve --backend vllm --base-url http://127.0.0.1:8000 --model Qwen/Qwen3-0.6B --endpoint /v1/completions --dataset-name sonnet --dataset-path sonnet.txt --sonnet-input-len 1024 --sonnet-output-len 128 --num-prompts 100 --request-rate 10 --max-concurrency 32
```

| Metric         | main   | tiled (no split) | **dispatch split** | Change vs main |
| -------------- | -----: | ---------------: | -----------------: | -------------: |
| Output tok/s   | 77.25  | 60.63            | **96.29**          | **+25%**       |
| Total tok/s    | 688.22 | 540.15           | **857.81**         | **+25%**       |
| Mean TTFT (ms) | 9,291  | 6,762            | **6,622**          | **-29%**       |
| Mean TPOT (ms) | 319.19 | 447.02           | **263.69**         | **-17%**       |
| P99 TPOT (ms)  | 381.06 | 499.10           | **315.55**         | **-17%**       |

The tiled kernel is a clear win for prefill-heavy workloads and a clear loss for decode-heavy ones. The dispatch split (tiled for prefill, old kernel for decode) is the path to getting both.

### Tiled kernel alone
**4 out of 6 passed.** Two fail:

1. **"The capital of France is"** — matches for first 5 tokens, diverges at token 6 (same issue we saw in the golden test)
2. **"Water boils at a temperature of"** — matches for first 5 tokens (`100°C`), diverges at token 6 (`at standard` vs `.\nThis is the`)

Both diverge around token 5-6 of decode — consistent with the QK float accumulation order difference compounding through layers and decode steps. The first 5 generated tokens are correct in both cases.

### kernel dispatch split
**5 out of 6 passed.** Only "The capital of France is" fails — same as before, diverging at token 6 ("Italy" vs "France"). The dispatch split correctly routes decode steps to the old kernel, but the initial prefill (6 tokens > 1 = has_prefill) still uses the tiled kernel, and that's enough to shift the logits for the subsequent decode.

This single remaining failure is the inherent MMA float non-associativity issue — the same reason MLX and llama.cpp use scalar kernels for short queries. 

## Research on Precision Issue

|                     | MLX Steel               | llama.cpp           | Our kernel (current) |
| ------------------- | ----------------------- | ------------------- | -------------------- |
| **QK fragments**    | float × float           | half × half         | half × half          |
| **PV fragments**    | float × float           | float × half        | float × half         |
| **Accumulators**    | float                   | float               | float                |
| **Threadgroup mem** | half (promoted at load) | half (no promotion) | half (no promotion)  |

**Our kernel currently matches llama.cpp's precision strategy exactly** — half×half→float for QK, float×half→float for PV, float accumulators throughout. The only difference from llama.cpp is the accumulation order (different tile sizes).

MLX takes the conservative path — promote everything to float before MMA. This gives float×float products (24-bit mantissa) at the cost of 2× threadgroup memory and slower MMA (float×float is slower than half×half on Apple AMX).

Note that the hardware MMA acceleration only supports 8x8 matrix within 32KB. So large head would be a problem.

Thus, mlx steel only supports HD ∈ {64, 80, 128} for the Steel MMA kernel. Everything else falls back to unfused matmul (no Flash Attention).

 
## Limitation

No support to FP8, FP32, Turboquant

## Next step [long prefilling 2/2]

 a register-resident flash kernel for long prefill.** Replace the 4-barrier tiled paged-attention path with an FA-2-style kernel that holds Q, O, and the online softmax state in per-thread registers (using `vec<T,2>` arrays so the Metal compiler can't spill them to threadgroup memory) — **4.51× kernel speedup and 2.71× e2e throughput on 8K prefill** with the Qwen3-0.6B serving benchmark.


LxYuan0420 (approved):
LGTM; the long-prefill speedup looks clear. What’s missing for me is short-prompt parity evidence. Since this makes tiled prefill default for all prompt lengths, and we already see a 6-token prompt change output, can you run a small before/after check on short prompts and share how often the generated text changes?

WindChimeRan:
@LxYuan0420 the result is in "Tiled kernel alone" section in the PR description (turn on the new kernel only, disable the old kernel completely, no hybrid)

4 out of 6 passed. failed ones: Both diverge around token 5-6 of decode

WindChimeRan:
@LxYuan0420 

token level disparity is acceptable because of different tiling and mma precision, so I use a simple classification task, to compare the classification f1 for old vs new kernel.

benchmark (correctness): https://github.com/WindChimeRan/applebench/tree/main

## Correctness Eval Results: Old Kernel vs Tiled Kernel

**Task:** GMRID_v3 supply-chain news classification (8-class, weighted-F1), 200 rows, Qwen3-0.6B, greedy decoding.

| Metric          | 0-shot Old | 0-shot Tiled | Δ           | 5-shot Old | 5-shot Tiled | Δ           |
| --------------- | ---------- | ------------ | ----------- | ---------- | ------------ | ----------- |
| **F1 weighted** | 0.3712     | 0.3760       | **+0.0048** | 0.7307     | 0.7387       | **+0.0079** |
| F1 macro        | 0.2588     | 0.2582       | −0.0006     | 0.4135     | 0.4113       | −0.0023     |
| Exact-match     | 0.4000     | 0.4000       | +0.0000     | 0.7300     | 0.7400       | **+0.0100** |

PR #369, opened 2026-05-13
Capture target hidden states for Gemma4 MTP

ricky-chaoju:
This is PR 2 in the Gemma 4 MTP stack.

It adds target hidden-state capture on the paged target path so the later Gemma 4 MTP assistant can consume target model states without adding model-specific logic to `MetalModelRunner`.

Changes:
- Add `ModelAdapter.target_forward()` as the model-owned boundary for logits + optional hidden states.
- Recompute logits from captured hidden states for Gemma 4-style tied embedding / lm_head paths.
- Collect target hidden states only for Gemma 4 MTP decode and final prefill steps.
- Keep ordinary decode, ngram spec decode, and non-Gemma 4 MTP paths unchanged.
- Store hidden states in the paged forward state for the next PRs.

### DIFF
--- a/vllm_metal/v1/model_adapter.py
+++ b/vllm_metal/v1/model_adapter.py
@@ -3,8 +3,10 @@
 
 from __future__ import annotations
 
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Protocol
 
+import mlx.core as mx
 from vllm.logger import init_logger
 
 if TYPE_CHECKING:
@@ -13,6 +15,14 @@ if TYPE_CHECKING:
     from vllm_metal.multimodal.feature_spec import MultiModalFeatureSpec
 
 logger = init_logger(__name__)
+
+
+@dataclass(frozen=True, slots=True)
+class TargetModelForwardOutput:
+    """Target-model forward output needed by sampling and speculative decode."""
+
+    logits: Any
+    hidden_states: Any | None = None
 
 
 class MultimodalRuntimeAdapter(Protocol):
@@ -54,6 +64,19 @@ class ModelAdapter(Protocol):
 
     def text_model(self, model: Any) -> Any:
         """Return the callable model used for text-only execution."""
+
+    def target_forward(
+        self,
+        model: Any,
+        input_ids: Any,
+        *,
+        cache: Any | None = None,
+        collect_hidden_states: bool = False,
+    ) -> TargetModelForwardOutput:
+        """Run the target text model and optionally retain target hidden states."""
+
+    def extract_logits(self, model_output: Any) -> Any:
+        """Extract logits from raw model output."""
 
     def build_multimodal_adapter(
         self, model: Any, hf_config: Any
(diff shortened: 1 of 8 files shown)


vllm_metal/v1/model_adapter.py
```diff
@@ -15,6 +17,14 @@
 logger = init_logger(__name__)
 
 
+@dataclass(frozen=True, slots=True)
+class TargetModelForwardOutput:
+    """Target-model forward output needed by sampling and speculative decode."""
+
+    logits: Any
+    hidden_states: Any | None = None
```
LxYuan0420: Nice boundary to introduce. Quick contract tighten-up: since this dataclass is now the adapter -> runner handoff, can we type these as `mx.array` / `mx.array | None` instead of `Any`? The runner already stores `target_hidden_states` as `mx.array | None`, so matching that here makes the contract clearer.

vllm_metal/v1/model_adapter.py
```diff
@@ -232,6 +255,99 @@ def text_model(self, model: Any) -> Any:
...
+            logits = self._compute_tied_embedding_logits(model, hidden_states)
+            return self._apply_target_logit_postprocessing(model, logits)
+
+        lm_head = getattr(model, "lm_head", None)
+        if lm_head is not None:
+            logits = lm_head(hidden_states)
+            return self._apply_target_logit_postprocessing(model, logits)
+
+        logits = self._compute_tied_embedding_logits(model, hidden_states)
+        return self._apply_target_logit_postprocessing(model, logits)
```
LxYuan0420, line 330: This fallback makes sense for the Gemma4/mlx-lm shape, but can we add a short comment here? Without the note, it reads like we silently use tied embeddings even when `tie_word_embeddings` is false and no `lm_head` exists.

vllm_metal/v1/model_adapter.py
```diff
@@ -232,6 +255,99 @@ def text_model(self, model: Any) -> Any:
...
+            raise NotImplementedError(
+                "Target hidden states require either `lm_head` or tied "
+                "`model.embed_tokens.as_linear` logits projection."
+            )
+        return as_linear(hidden_states)
+
+    def _flatten_target_hidden_states(self, hidden_states: Any) -> Any:
+        if len(hidden_states.shape) == 3 and hidden_states.shape[0] == 1:
+            return hidden_states[0]
+        return hidden_states
```
LxYuan0420: Small shape-contract fix: this helper should always return row-major hidden states, so we should not silently return a 3D tensor when `shape[0] != 1`. Can we type this as `mx.array` and assert either `ndim == 2` or batch size is 1 before flattening?

vllm_metal/v1/model_runner.py
```diff
@@ -798,6 +810,13 @@ def _start_paged_forward(
             self._paged_request_seq_lens,
         )
         num_decode_tokens = sum(segment.num_query_tokens for segment in decode_segments)
+        collect_target_hidden_states = (
+            self._spec_decode_controller.needs_target_hidden_states(
+                decode_segments,
+                has_final_prefill=any(pr.prompt_len is not None for pr in prefill_reqs),
+                speculative_config=self.vllm_config.speculative_config,
```
LxYuan0420, line 819: This is a good gate. Small readability ask: can we add a one-line comment that `PrefillRequest.prompt_len is None` means an intermediate chunk? At this call site it otherwise looks like a generic prompt-length check rather than the final-prefill signal.

vllm_metal/v1/model_runner.py
```diff
@@ -902,7 +929,10 @@ def _sample_paged_batch(
         )
 
         # ---- wait for MLX forward to complete ----
-        mx.eval(logits)
+        if target_hidden_states is not None:
+            mx.eval(logits, target_hidden_states)
```
LxYuan0420, line 937: Since this PR stores hidden states for the next PR in the stack, can we add a short forward-reference comment here? The code path is fine, but without that note this looks like we evaluate and retain a large tensor without a local consumer.

vllm_metal/v1/spec_decode.py
```diff
@@ -14,6 +14,14 @@
 
 from vllm_metal.v1.sampling_batch import GREEDY_TEMPERATURE_EPS
 
+if TYPE_CHECKING:
+    from vllm.config.speculative import SpeculativeConfig
+
+_GEMMA4_MTP_DRAFT_MODEL_TYPES = frozenset({"gemma4_assistant", "gemma4_mtp"})
+_GEMMA4_MTP_DRAFT_ARCHITECTURES = frozenset(
+    {"Gemma4AssistantForCausalLM", "Gemma4MTPModel"}
+)
```
LxYuan0420, line 26: Good catch handling both `model_type` and architecture names. Can we add a short source comment for these identifiers, e.g. that they come from the Gemma4 assistant HF/Transformers or mlx-lm config names? That makes future upgrades easier to audit if upstream renames them.

vllm_metal/v1/spec_decode.py
```diff
@@ -183,6 +191,37 @@ def build_decode_segments(
...
+
+        hf_config = draft_model_config.hf_config
+        model_type = getattr(hf_config, "model_type", None)
+        if model_type in _GEMMA4_MTP_DRAFT_MODEL_TYPES:
+            return True
+
+        architectures = getattr(hf_config, "architectures", ()) or ()
+        if isinstance(architectures, str):
+            architectures = (architectures,)
+        return any(arch in _GEMMA4_MTP_DRAFT_ARCHITECTURES for arch in architectures)
```
LxYuan0420, line 233: Small maintainability ask: can we tighten or document these `getattr` checks against the upstream config contract? If `draft_model_config.hf_config` is a known vLLM/HF config shape, direct field access is easier to read and fails faster when upstream changes. If the shape is intentionally loose here, a short comment explaining which upstream variants require defensive access would make this much easier to maintain.

vllm_metal/v1/spec_decode.py
```diff
@@ -183,6 +191,37 @@ def build_decode_segments(
...
+        self,
+        decode_segments: Sequence[PagedDecodeSegment],
+        *,
+        has_final_prefill: bool = False,
+        speculative_config: SpeculativeConfig | None = None,
+    ) -> bool:
+        """Return whether target hidden states are needed for draft follow-up."""
+        if not decode_segments and not has_final_prefill:
+            return False
+        return self._uses_gemma4_mtp(speculative_config)
```
LxYuan0420: The always-on Gemma4 MTP behavior looks intentional and the tests cover it. Can we expand the docstring by one line explaining that plain decode/final prefill still need target hidden states because the assistant consumes the previous target step to draft the next tokens?

LxYuan0420 (approved):
Left a few small comments, mostly around tightening contracts and documenting the less obvious Gemma4 MTP assumptions. Nothing looks structurally off to me; this is a solid step in the Gemma4 MTP stack

ricky-chaoju:
> Left a few small comments, mostly around tightening contracts and documenting the less obvious Gemma4 MTP assumptions. Nothing looks structurally off to me; this is a solid step in the Gemma4 MTP stack

Done. Thank you, would you like to confirm again?

LxYuan0420 (approved):
LGTM

PR #362, opened 2026-05-09
[Kernel] MLA absorbed-decode kernels (RFC #360 Phase 1, kernel-layer)

scyyh11:
Tracking RFC: #360

Implements the **kernel layer** of RFC #360 Phase 1: absorbed-MLA decode kernels for paged latent cache, with direct eager bindings and kernel-direct parity tests. Routing, env gate, real-model wrapper parity, benchmarks, and the supported-model docs row are split out into follow-up PRs.

This PR has been scoped down at @LxYuan0420's request (#362 review): the previous bundle (kernels + C++ + MLX primitives + routing + benchmarks + tests + docs) is now four kernel-only commits, each containing one kernel variant plus its direct parity tests. UnaryPrimitive MLX-graph wrappers, runtime routing, env gate, benchmarks, docs row, and the `pr_mma` kernel are deferred to follow-up PRs.

## Scope

| In this PR | Deferred |
|---|---|
| 4 kernel variants in `mla.metal` (single-pass, split-K reduce, 2pass, FA) | `pr_mma` kernel (no direct parity tests yet) |
| C++ dispatchers in `paged_ops.cpp` (one per kernel) | UnaryPrimitive MLX-graph wrappers (C++ side) |
| Eager Python wrappers in `vllm_metal/metal/__init__.py` | UnaryPrimitive Python wrappers (`*_primitive`) |
| `_ensure_mla_library` lazy shader-library init | bench-only `*_main_only` / `*_reduce_only` dispatchers |
| Direct parity tests in `tests/test_mla_kernel.py` | Routing (`paged_attention_backend/mla.py`, env gate, routing tests) |
| | Wrapper-grid benchmarks (`tools/benchmark/mla_*`) |
| | Real-model E2E wrapper parity (per-route) |
| | `docs/supported_models.md` row |

The eager Python wrappers and `_ensure_mla_library` are retained as the minimum surface required for the direct parity tests to exercise the kernels. They are not MLX `UnaryPrimitive` subclasses — they are thin shims over the C++ dispatcher and the lazy `init_mla_library(src)` call that feeds `mla.metal` source to MLX's metal device under `paged_mla_kern`. Without them the C++ `d.get_library("paged_mla_kern")` lookup fails before any kernel can be dispatched from a test.

## Commits

| # | Commit | Adds |
|---|---|---|
[...]

Issue #360, opened 2026-05-09
[RFC]: Specialized Metal Kernel for MLA Paged Attention


scyyh11:
**Problem:** vllm-metal should support MLA models with the same high-level split used by upstream vLLM: compressed-latent attention for decode, and materialized K/V attention for prefill.

The RFC has moved from pure proposal to implementation tracker. The first two steps have landed: absorbed MLA decode has an opt-in Metal kernel, and absorbed MLA prefill now uses the materialized path by default for pure prefill. The remaining open work is MLA-specific long-context partitioning.

### PR Stack

* [x] PR 1: Absorbed MLA decode kernel behind `VLLM_METAL_MLA_KERNEL=1` (#370). This replaces the earlier draft PR #362.
* [x] PR 2: Materialized prefill for absorbed MLA models (#405).
* [ ] PR 3: MLA long-context partition/reduce decode. #437 landed split-KV decode for regular paged attention, TurboQuant, and sliding-window paths, but MLA decode is explicitly untouched there.
* [ ] Follow-ups: chunked prefill / prefix-cache support, production benchmark grid, docs/support-status update once runtime gates are met, and MLA-specific TurboQuant if the quantization story stabilizes.

### Design

* Keep the existing MLA latent-cache contract; do not fork upstream vLLM scheduling or cache layout.
* Decode follows upstream `forward_mqa`: attend directly over compressed latent K/V.
* Prefill follows upstream `forward_mha`: materialize per-head K/V, then run standard MHA.
* Runtime routing should stay conservative: opt-in or narrowly gated unless wrapper-level parity and production benchmarks justify a wider default.
* Phase 3 should be MLA-specific partition + reduce for long-context decode. Existing split-KV work in #437 is useful reference infrastructure, but not the MLA implementation.
* Performance claims should include wrapper overhead, not just isolated kernel time.

### CC List

@WindChimeRan @LxYuan0420

scyyh11:
## Phase 1 Status (2026-05-09): Parked

Phase 1 baseline implemented and pushed as draft PR #362. 51/51 parity tests pass at production shapes (`kv_lora_rank=512`, `qk_rope_head_dim=64`, fp16/bf16); per-commit breakdown and microbench table in the PR body.

[...]

PR #329, opened 2026-05-02
[kernel] Fold softmax into log2 space, replace exp with exp2

WindChimeRan:
bench_exp2_comparison

## Summary: 

Just switch exp to exp2, learnt from the mlx sdpa's source code. 

## Deterministic test

Better numeric alignment with mlx sdpa attention

```
France:   MATCHED mlx-cache golden    ← main matched paged-only (drift from MLX)
Weather:  MATCHED mlx-cache golden
One+one:  MATCHED paged-cache golden  ← MLX disagrees even on main (test docstring notes this)
Planet:   MATCHED mlx-cache golden
Water:    MATCHED mlx-cache golden    ← main matched paged-only (drift from MLX)
Machine:  MATCHED mlx-cache golden
```

## Benchmark (50 prompts, sonnet 2048+128, M1 Pro)

| Metric              | main        | this PR     | Δ          |
| ------------------- | ----------: | ----------: | ---------: |
| Total throughput    | 394.1 tok/s | 478.1 tok/s | **+21.3%** |
| Mean TTFT (prefill) | 61.8 s      | 50.2 s      | **−18.6%** |
| Median TTFT         | 38.7 s      | 30.2 s      | **−22.0%** |
| Mean TPOT (decode)  | 838.6 ms    | 691.6 ms    | **−17.5%** |
| Duration            | 274 s       | 226 s       | **−17.6%** |

PR #370, opened 2026-05-13
[Kernel] MLA single-pass decode kernel + opt-in route (RFC #360 Phase 1)

scyyh11:
Tracking RFC: #360

Implements the absorbed-MLA single-pass decode kernel, wired into the wrapper behind `VLLM_METAL_MLA_KERNEL=1`. 

## Routing

Single env-gated admission check (`_can_use_kernel`):

- `VLLM_METAL_MLA_KERNEL=1` (off by default)
- Absorbed MLA (`embed_q` / `unembed_out` present on inner)
- `kv_lora_rank == 512`
- `qk_rope_head_dim == 64`
- `block_size ∈ {16, 32}`
- `dtype ∈ {fp16, bf16}`
- Decode-only (1 query token per request)

Any rejection falls through to the existing MLX SDPA per-request slow path — no behaviour change for non-MLA paged-attention users or for workloads outside the single-pass envelope.

## E2E real-model token parity

Spawned-process harness: same prompt, `temperature=0`, `max_tokens=16`, run `vllm.LLM` twice: once with `VLLM_METAL_MLA_KERNEL=0` (MLX SDPA per-request slow path), once with `=1` (single-pass kernel via `MlaPagedAttentionPrimitive`). Compare generated token IDs.

`mlx-community/GLM-4.7-Flash-4bit` (`Glm4MoeLiteForCausalLM`, MLA-absorbed, `kv_lora_rank=512`, `qk_rope_head_dim=64`, `qk_nope_head_dim=192`, `v_head_dim=256`, H=20):

| Prompt                       | Match              | MLX text                                                                 | Kernel text                             |
| ---------------------------- | ------------------ | ------------------------------------------------------------------------ | --------------------------------------- |
| `"The capital of France is"` | ✅                  | `" Paris, and the capital of the United States is Washington, D.C. The"` | (identical)                             |
| `"def quicksort(arr):"`      | ✅                  | `" \n if len(arr) <= ;\n return;\n pivot = arr"`                         | (identical)                             |
| `"Once upon a time"`         | ✅                  | `", } } } } } } } } } } } } } } }"`                                      | (identical degenerate loop, both paths) |
| `"One plus one equals"`      | ⚠️ diverge @ tok 3 | `" two, } { } { } { } { } { } { } {"`                                    | `" two, } } } } } } } } } } } } } }"`   |

**3/4 token-for-token match.** The divergent prompt is a low-confidence degenerate trajectory: after `" two,"` the model cannot continue meaningfully at 4-bit quantization, and small fp16/bf16 attention-output differences flip the argmax between two near-equal logits (`{`=314 vs `}`=335). Both paths produce semantically equivalent garbage.

## Performance (M5 Max, `ignore_eos=True`, warmed)

5 configs spanning B ∈ {1, 2, 4, 8} on `mlx-community/GLM-4.7-Flash-4bit` (`Glm4MoeLiteForCausalLM`, H=20, kv_lora_rank=512, qk_nope=192, v=256):

| B   | gen  | max_ctx | MLX wall | Kernel wall | **Speedup** | MLX tput  | Kernel tput |
| --- | ---- | ------- | -------- | ----------- | ----------- | --------- | ----------- |
| 1   | 1024 | 2048    | 14.68 s  | 13.15 s     | **1.12×**   | 69.8 t/s  | 77.9 t/s    |
| 2   | 1024 | 2048    | 20.28 s  | 17.68 s     | **1.15×**   | 101.0 t/s | 115.9 t/s   |
| 4   | 256  | 2048    | 6.92 s   | 5.81 s      | **1.19×**   | 148.1 t/s | 176.2 t/s   |
| 4   | 1024 | 2048    | 34.86 s  | 30.07 s     | **1.16×**   | 117.5 t/s | 136.2 t/s   |
| 8   | 256  | 1024    | 11.41 s  | 9.49 s      | **1.20×**   | 179.4 t/s | 215.7 t/s   |

Wins 1.12-1.20× across all 5 configs (speedup grows with B: cross-head amortization + batched dispatch scale with concurrent decode streams). Reproducible numbers via `vllm bench serve` in [this comment](https://github.com/vllm-project/vllm-metal/pull/370#issuecomment-4447972667).

Replaces #362.


WindChimeRan (commented):
For deepseek v2:  _can_use_kernel returns False because deepseek_v2.py has no embed_q.

Can you double check? I think deepseek benchmark did not trigger the new kernel

WindChimeRan (commented):
for the benchmark. are you using `vllm bench serve`? Could you please share the command for reproducibility? Since this PR is focusing on the decoding part, I would expect a short prefill long decode exp setting

scyyh11:
> For deepseek v2: _can_use_kernel returns False because deepseek_v2.py has no embed_q.
> 
> Can you double check? I think deepseek benchmark did not trigger the new kernel

Sorry for the mistake, deepseek v2 right now can't trigger the new kernel.

To enable DS V2-Lite we'd need to change `mlx_lm/models/deepseek_v2.py` to split `kv_b_proj ([num_heads*(qk_nope_head_dim + v_head_dim), kv_lora_rank])` into `embed_q` + `unembed_out`, matching deepseek V3 layout. Do you think it should be landed in current or a follow-up PR?

WindChimeRan:
@scyyh11 

I'd recommend keep the scope small, only on GLM-4.7-Flash.

scyyh11:
Re-ran with `vllm bench serve`, sweeping concurrency 1→8.

## Performance (M5 Max, 36 GB, `vllm bench serve`)

One `vllm serve` instance per backend (swap `VLLM_METAL_MLA_KERNEL` between runs). Short prefill + long decode via `--dataset-name random --ignore-eos`.

| ISL | OSL  | N  | Conc | TPOT (MLX SDPA) | TPOT (Kernel) | **TPOT Speedup** |
| --- | ---- | -- | ---- | --------------- | ------------- | ---------------- |
| 32  | 256  | 4  | 1    | 15.29 ms        | 14.28 ms      | **1.071×**       |
| 32  | 256  | 8  | 2    | 20.85 ms        | 19.72 ms      | **1.057×**       |
| 32  | 256  | 16 | 4    | 30.12 ms        | 27.68 ms      | **1.088×**       |
| 32  | 256  | 16 | 8    | 48.45 ms        | 42.85 ms      | **1.131×**       |
| 32  | 1024 | 8  | 4    | 31.98 ms        | 28.95 ms      | **1.105×**       |

Decode TPOT speedup grows from 1.06× (single stream) to 1.13× (max concurrent) — consistent with cross-head amortization scaling with concurrent decode streams (G=2 kernel reuses each K/V load across two query heads, so the win grows with batch).

TTFT median is unchanged across paths (kernel is decode-only; prefill still goes through MLX SDPA).

<details>
<summary>Reproducible commands</summary>

```bash
# Server — run twice, once per backend (swap VLLM_METAL_MLA_KERNEL)
VLLM_METAL_MLA_KERNEL=0 VLLM_METAL_USE_PAGED_ATTENTION=1 \
  vllm serve mlx-community/GLM-4.7-Flash-4bit \
    --max-model-len 2048 --max-num-seqs 8 --port 8000 --enforce-eager

# 5 bench client configs (same for each backend)
for cfg in \
    "--random-input-len 32 --random-output-len 256  --num-prompts 4  --max-concurrency 1" \
    "--random-input-len 32 --random-output-len 256  --num-prompts 8  --max-concurrency 2" \
    "--random-input-len 32 --random-output-len 256  --num-prompts 16 --max-concurrency 4" \
    "--random-input-len 32 --random-output-len 256  --num-prompts 16 --max-concurrency 8" \
    "--random-input-len 32 --random-output-len 1024 --num-prompts 8  --max-concurrency 4"; do
  vllm bench serve \
    --backend vllm --host localhost --port 8000 \
    --model mlx-community/GLM-4.7-Flash-4bit \
    --dataset-name random --seed 42 --ignore-eos $cfg
done
```

</details>

WindChimeRan:
@scyyh11 can we remove DeepSeek-V2-Lite from the PR description and close #362 ? (you may refer it in the PR description to keep the history linked)

scyyh11:
> @scyyh11 can we remove DeepSeek-V2-Lite from the PR description and close #362 ? (you may refer it in the PR description to keep the history linked)

All done now.

WindChimeRan (commented):
please add env var to https://docs.vllm.ai/projects/vllm-metal/en/latest/configuration/#environment-variables as well

WindChimeRan (commented):
One thing I'm not very sure:

should we use exp2 instead of exp everywhere? reference: https://github.com/vllm-project/vllm-metal/pull/329

scyyh11:
Applied the `exp2` trick (5af33c3): folded `log2(e)` into the score scale and replaced the three `fast::exp` sites in `mla.metal` with `fast::exp2`, matching the `pagedattention.metal` pattern from #329. Re-ran the same 5 configs (kernel side; MLX SDPA unchanged):


| ISL | OSL  | N   | Conc | TPOT (MLX SDPA) | TPOT (Kernel) | **TPOT Speedup** |
| --- | ---- | --- | ---- | --------------- | ------------- | ---------------- |
| 32  | 256  | 4   | 1    | 15.29 ms        | 14.32 ms      | **1.068×**       |
| 32  | 256  | 8   | 2    | 20.85 ms        | 19.29 ms      | **1.081×**       |
| 32  | 256  | 16  | 4    | 30.12 ms        | 26.05 ms      | **1.156×**       |
| 32  | 256  | 16  | 8    | 48.45 ms        | 40.73 ms      | **1.189×**       |
| 32  | 1024 | 8   | 4    | 31.98 ms        | 27.85 ms      | **1.148×**       |


Speedup now ranges 1.07× (single stream) → 1.19× (max concurrent), up from the prior 1.06× → 1.13×. Kernel-side TPOT drops 2-6% at concurrency ≥ 2; conc=1 stays flat (compute too small for the per-token multiply saving to register). Unit-test parity holds (`tests/test_mla_kernel.py`, `tests/test_mla_paged_backend.py`).

Thanks for pointing it out @WindChimeRan.

WindChimeRan (commented):
MLA models (DeepSeek-V3, GLM-4.7-Flash) don't use ALiBi, softcapping, or sinks. 

are these dead code in paged_ops.cpp?

vllm_metal/metal/__init__.py
```diff
@@ -191,6 +207,177 @@ def metal_unified_attention(
...
+    if q_nope.shape[2] != latent_cache.shape[2] - q_pe.shape[2]:
+        raise ValueError(
+            f"MLA shape mismatch: q_nope.shape[2]={q_nope.shape[2]} must equal "
+            f"latent_cache.shape[2] ({latent_cache.shape[2]}) - "
+            f"q_pe.shape[2] ({q_pe.shape[2]})"
+        )
+
+    block_size = latent_cache.shape[1]
+
+    mx.eval(out, q_nope, q_pe, latent_cache, block_tables, context_lens, cu_seqlens_q)
```
WindChimeRan: is this materialization necessary? This will break the kernel primative's lazy graph flow

vllm_metal/metal/__init__.py
```diff
@@ -191,6 +207,177 @@ def metal_unified_attention(
...
+
+    # P1 decode-only guard: q_token_idx == seq_idx is hard-wired in the
+    # kernel. Reading cu_seqlens_q back to host is cheap (tens of int32 per
+    # layer per step) and the guard goes away when P2 makes the kernel
+    # cu_seqlens_q-aware via find_seq_idx.
+    cu_q = np.asarray(cu_seqlens_q)
+    deltas = np.diff(cu_q)
+    if np.any(deltas != 1):
```
WindChimeRan: I think we should not gate the kernel here. maybe you can refer to https://github.com/vllm-project/vllm-metal/pull/367/changes#diff-bb4652cbd1452332ec89eae7d31a1ffb91d2091b1360198465a6db2cdf6b3c8cR474

In the future, we'll probably have a separated mla prefill kernel anyway ...

If the gate is actually overdefensive, we can just remove it

scyyh11:
> MLA models (DeepSeek-V3, GLM-4.7-Flash) don't use ALiBi, softcapping, or sinks.
> 
> are these dead code in paged_ops.cpp?

Those were dead-code leftovers from copy-pasting the GQA paged-attention dispatcher. Cleaned up in 426afad: dropped `mla_use_alibi` / `mla_use_fp8_scales` / `mla_use_sinks` / `mla_use_turboquant` from `mla.metal` and the matching `use_*` flags + hash suffix + function-constant entries from `dispatch_mla_paged_attention` in `paged_ops.cpp`. Also trimmed the file header which referenced a `paged_mla_attention_reduce` kernel that was never landed.

`mla_use_partitioning` was kept but currently always `false`, but it gates the `exp_sums` / `max_logits` scratch buffer existence in the kernel signature; intentional scaffolding for the 2-pass follow-up.

vllm_metal/metal/__init__.py
```diff
@@ -33,6 +33,13 @@
 # paged_ops.cpp is newer than the .so).
 _ops_module: ModuleType | None = None
 
+# The MLA paged-attention shader set is loaded lazily on the first MLA
+# entrypoint call (see _ensure_mla_library). Initialising it inside
+# get_ops() would compile experimental MLA Metal source for every
+# paged-attention user, so non-MLA models pay the compile cost and a
+# compile error in MLA would block unrelated paged-attention paths.
+_mla_library_initialized: bool = False
```
WindChimeRan: I think this is overcomplicating things and adding maintainance burden ... just compile it like others in a consistent way should be fine?

vllm_metal/metal/kernels_v2/mla.metal
```diff
@@ -0,0 +1,403 @@
...
+instantiate_mla(bfloat16_t, 512, 64, 16, 1, 1024, 0);
+instantiate_mla(bfloat16_t, 512, 64, 32, 1, 1024, 0);
+
+// G=2 (2 heads per TG, NUM_THREADS=512). 2× KV-bandwidth amortization.
+instantiate_mla(half, 512, 64, 16, 2, 512, 0);
+instantiate_mla(half, 512, 64, 32, 2, 512, 0);
+instantiate_mla(bfloat16_t, 512, 64, 16, 2, 512, 0);
+instantiate_mla(bfloat16_t, 512, 64, 32, 2, 512, 0);
+
+// G=4 (4 heads per TG, NUM_THREADS=256). 4× KV-bandwidth amortization.
```
WindChimeRan: currently, G=4 instantiation is dead code, because of the routing _pick_heads_per_tg

Is there a good reason to keep it?

vllm_metal/metal/__init__.py
scyyh11: I just tried dropping the `mx.eval` from the eager binding and got a segfault: the C++ `ops.mla_paged_attention` reads input buffer pointers directly, and without a prior materialization the buffers aren't backed yet.

I think we should either:

1. Keep `mx.eval` in the eager binding since it's required for the eager C++ entry. Production path (the primitive) stays lazy.
2. Delete `metal_mla_paged_attention` entirely and migrate the 17 test sites to `metal_mla_paged_attention_primitive` + an explicit `mx.eval` on the returned array.

WindChimeRan: yeah let's delete the eager path and only keep the primitive

vllm_metal/paged_attention_backend/mla.py
```diff
@@ -77,6 +87,121 @@ def _causal_valid_mask(
...
+
+        # Pad block_tables (list[list[int]]) into a 2D [num_seqs, max_blocks]
+        # int32 array. The kernel reads block_table_row[0..n_context_blocks-1];
+        # padding entries beyond n_context_blocks are never read.
+        import numpy as np
+
+        bts = ctx.block_tables
+        num_seqs = len(bts)
+        max_blocks = max(len(bt) for bt in bts)
+        bt_np = np.zeros((num_seqs, max_blocks), dtype=np.int32)
```
WindChimeRan: why np detour here?

vllm_metal/metal/paged_ops.cpp
WindChimeRan, line 1: from_primitive parameter + dead add_temporary_compat branch

WindChimeRan (commented):
could you please add end2end deterministic comparison test for glm-4.7-flash, as what we did for qwen3,qwen3.5 in `tools/`(use our existing 6 prompts. The PR description's tests are hard to tell whether it's model's own gibberish or not) please use standalone mlx_lm to generate reference token. 

Numeric parity is acceptable, as long as the PR doesn't sacrifice generation quality.

scyyh11:
Added `tools/test_glm47_flash_golden.py` (79847a6). Mirrors `test_qwen3_next_golden.py`: 6 original prompts (kept verbatim from `gen_golden_token_ids_for_deterministics.py`), frozen golden token IDs from **standalone** `mlx_lm.generate_step + argmax`, chat template applied on both sides (bare prompts collapse into single-token loops on this instruction-tuned model and don't exercise the decode path). 

## Result: 5/6 exact match, 1 divergence


| Prompt                                    | Match                          |
| ----------------------------------------- | ------------------------------ |
| The capital of France is                  | yes                            |
| The weather today is not                  | yes                            |
| One plus one equals                       | yes                            |
| The largest planet in our solar system is | **no** (1 token at position 6) |
| Water boils at a temperature of           | yes                            |
| Machine learning is                       | yes                            |


The one mismatch:

```
golden (mlx_lm):  [785, 1196, 374, 10156, 264, 4285,    11, 59485, 3405,  25]
                                                       comma
  decodes to:    "The user is asking a simple, factual question:"

kernel:           [785, 1196, 374, 10156, 264, 4285, 59485,  3405,   25, 330]
                                                    skips comma, goes to " factual"
  decodes to:    'The user is asking a simple factual question: "'
```

The model is in `<think>` mode restating the user's query, and the kernel picks  `factual` directly where mlx_lm picks `,` first. Decoded text is semantically equivalent.

Running the same prompt with `VLLM_METAL_MLA_KERNEL=0` (paged + MLX SDPA) matches the mlx_lm reference token-for-token.

WindChimeRan (approved):

PR #372, opened 2026-05-14
[kernel] remove eager kernel due to the new primitive

WindChimeRan:
cleanup eager kernel, due to the working primitive (https://github.com/vllm-project/vllm-metal/pull/225)

now eager kernel is dead code


tests/test_metal_unified_attention.py
```diff
@@ -1,322 +0,0 @@
...
-# ---------------------------------------------------------------------------
-
-
-@pytest.mark.parametrize(
-    "seq_lens", [[(1, 1328), (5, 18), (129, 463)], [(1, 523), (1, 37), (1, 2011)]]
-)
-@pytest.mark.parametrize("num_heads", NUM_HEADS)
-@pytest.mark.parametrize("head_size", HEAD_SIZES)
-@pytest.mark.parametrize("block_size", BLOCK_SIZES)
-@pytest.mark.parametrize("sliding_window", [None, 128])
```
LxYuan0420, line 245: Can we keep the sliding-window reference check by moving it into `tests/test_primitive_and_donation.py` with `ops.paged_attention_primitive(...)`?

`sliding_window` is still used by the production primitive path, but the surviving primitive tests only cover `sliding_window=-1`, so deleting this drops coverage for active kernel masking behavior.

LxYuan0420 (approved):

PR #373, opened 2026-05-14
[Build] invalidate JIT cache by source hash, not mtime

scyyh11:
## Summary

The paged-attention extension is JIT-built into `~/.cache/vllm-metal/_paged_ops*.so`. `needs_rebuild()` compared the cached `.so` mtime against source mtimes; this is unsafe and incomplete:

- The cache lives outside the repo and persists across branches. After a `git checkout` that moves `paged_ops.cpp` / `build.py` / `constants.py` backwards in history, source mtimes are set to the checkout time (older than a `.so` produced on a previous branch), so the stale `.so` is reused and the runtime hits missing-symbol errors (e.g. old `paged_attention_tiled_*` after the rewrite to `paged_attention_v2_online`).
- The `.so` also depends on resolved build inputs that don't appear in source mtimes — the MLX install path/version we link and rpath against, nanobind src, `PARTITION_SIZE`, the Python include. Switching virtualenvs or upgrading MLX in place could keep the source identical while invalidating the linked dylib.

This switches to a content hash stored in a sidecar next to the `.so`. `needs_rebuild()` compares the recomputed hash against the sidecar; mtime is not consulted.

The hash covers:

- bytes of `paged_ops.cpp`, `build.py`, `constants.py`, `nb_combined.cpp`
- the fully-resolved `clang++` invocation (Python include, nanobind paths, MLX `-L`/`-I`/`-rpath`, `PARTITION_SIZE` define, output path)
- `mlx.core.__version__` and `nanobind.__version__` (for in-place upgrades where the install path is reused)

`_build_spec()` centralizes the resolution; `build()` and the hash share it, so the cache key cannot drift from what we actually compile.

Existing caches without a sidecar trigger one rebuild on first run, then hash-based invalidation from then on.

### DIFF
--- a/tests/test_metal_build_cache.py
+++ b/tests/test_metal_build_cache.py
@@ -1,0 +1,145 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the JIT-build cache invalidation in vllm_metal.metal.build."""
+
+from __future__ import annotations
+
+import dataclasses
+import os
+from dataclasses import dataclass
+from pathlib import Path
+
+import pytest
+
+from vllm_metal.metal import build
+
+
+@dataclass
+class _Paths:
+    src: Path
+    bld: Path
+    consts: Path
+    nb_src: Path
+    out: Path
+    hsh: Path
+    spec: "build._BuildSpec"
+
+
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


LxYuan0420 (approved):
LGTM

PR #374, opened 2026-05-15
Add Gemma4 MTP assistant loader

ricky-chaoju:
Add the Gemma4 MTP assistant loader for the next MTP stack slice.

This PR adds a focused Gemma4 MTP owner that can detect, validate, and load matching `gemma4_assistant` checkpoints without wiring assistant forward execution yet. It also keeps the lifecycle boundary explicit: the assistant runtime is installed only after the target model is loaded and target metadata is available.

  - Add Gemma4 MTP assistant config validation and runtime metadata.
  - Add local MLX model classes for raw Gemma4 assistant checkpoints.
  - Load assistant checkpoints through `mlx_lm.load_model()` with local model classes.
  - Reject custom `model_file` hooks and keep assistant downloads limited to config/weights files.
  - Validate assistant and target compatibility:
    - model type
    - vocab size
    - hidden/backbone hidden size
    - layer types
    - KV-sharing metadata needed by later PRs
  - Clear assistant cache through `reset_model_cache()`.
  - Keep assistant `forward_ready=False`; forward execution, KV sharing, and scheduler handoff are intentionally left
  for later PRs.


vllm_metal/v1/gemma4_mtp.py
LxYuan0420, line 1: Thanks for splitting this out as a separate roadmap slice. The direction is right, but I think `gemma4_mtp.py` needs a structure/readability pass before we build the KV-sharing and scheduler-handoff PRs on top of it.

Right now the file mixes too many ownership boundaries: speculative-config detection, draft model/revision extraction, assistant config parsing, target metadata parsing, compatibility validation, loading, caching, and generic `Any` field access. This makes the main flow hard to read and weakens the internal contract for the next PRs.

Can you reorganize this around clear owners and keep the file ordered from high-level API to private implementation details? A suggested shape:

- `Gemma4MTPAssistantLoader`: public entry point; owns load/cache/download flow.
- `Gemma4MTPAssistantSource`: owns draft model + revision extraction and cache key.
- `Gemma4MTPTargetMetadata`: owns target metadata normalization from vLLM/HF/MLX model args.
- `Gemma4MTPAssistantMetadata`: owns assistant config parsing and validation.
- `Gemma4MTPCompatibility`: or `assistant_metadata.validate_compatible_with(target_metadata)`, owns assistant-target compatibility checks.

After normalization, the rest of the code should pass typed objects around instead of continuing to use broad `Any`/`getattr` helpers. Private helpers should sit near the class that owns them, or at the bottom if they are truly shared.

This keeps single responsibility clear, makes the load path easier to follow, and gives the next KV-sharing work a stable typed contract to build on.

LxYuan0420 (changes requested):

ricky-chaoju:
Done, reorganized `gemma4_mtp.py` around those owners and added the shared mlx-lm shard compatibility path for assistant loading as well.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,789 @@
...
+GEMMA4_TARGET_MODEL_TYPES = frozenset({"gemma4", "gemma4_text"})
+GEMMA4_MTP_TEXT_MODEL_TYPE = "gemma4_text"
+GEMMA4_MTP_VALID_LAYER_TYPES = frozenset({"sliding_attention", "full_attention"})
+
+_ASSISTANT_DOWNLOAD_ALLOW_PATTERNS = [
+    "config.json",
+    "generation_config.json",
+    "model*.safetensors.index.json",
+    "model*.safetensors",
+]
```
LxYuan0420, line 37: The local compat path supports indexed checkpoints whose shard names are not `model*.safetensors`, but the HF download allowlist still only downloads `model*.safetensors`. A remote assistant with `model.safetensors.index.json` pointing to `layers-0.safetensors` would fail because the shard is filtered out before load. Can we either widen this to `*.safetensors` or derive the exact shard names from `weight_map`, and add a test for the default `snapshot_download()` path?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,789 @@
...
+            )
+        if self.backbone_hidden_size != target.hidden_size:
+            raise ValueError(
+                "Gemma4 MTP assistant backbone hidden size must match target "
+                f"hidden size: assistant={self.backbone_hidden_size}, "
+                f"target={target.hidden_size}"
+            )
+
+        missing_layer_types = sorted(
+            set(self.layer_types) - set(target.non_shared_layer_types)
```
LxYuan0420: This layer compatibility check is too loose. Set comparison only checks whether a layer type exists somewhere, not whether the assistant matches the target non-shared KV layout it will consume. A 1-layer `["full_attention"]` assistant currently validates against `["sliding_attention", "sliding_attention", "sliding_attention", "full_attention"]`; for KV sharing this should be positional, either exact tuple equality or a documented tail-match rule with a regression test.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,789 @@
...
+
+        return Path(
+            snapshot_download(
+                source.model_name,
+                revision=source.revision,
+                allow_patterns=_ASSISTANT_DOWNLOAD_ALLOW_PATTERNS,
+            )
+        )
+
+    @property
```
LxYuan0420: Can we simplify the loader ownership here? `_load_model_fn` only delegates to a module import, and several loader-only helpers still live as module functions. I’d keep the load/cache/validation path self-contained on `Gemma4MTPAssistantLoader` by moving `_read_config_file`, `_reject_custom_model_file`, `_get_model_classes`, and `_assistant_model_classes` there, or inlining one-step helpers.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,789 @@
...
+    if isinstance(value, str):
+        raise ValueError(f"{key} must be a non-string sequence")
+    if not isinstance(value, Sequence):
+        raise ValueError(f"{key} must be a sequence")
+    if any(not isinstance(item, str) for item in value):
+        raise ValueError(f"{key} entries must be strings")
+    return tuple(value)
+
+
+def _field(config: Any, key: str, default: Any = None) -> Any:
```
LxYuan0420: Can we narrow the generic `_field()` usage and helper surface? For known mappings like `target_model_args` and loader config dicts, direct `.get()` is clearer; for known objects like vLLM `SpeculativeConfig`, direct `getattr()` is clearer. Please reserve `_field()` for the real polymorphic HF config boundary, and consider extracting the shared gather-and-compare logic from `_matching_optional_int` / `_matching_sequence`.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,789 @@
...
+        )
+        centroid_intermediate_top_k = _optional_positive_int(
+            config,
+            "centroid_intermediate_top_k",
+            context="assistant",
+        )
+        if not use_ordered_embeddings:
+            return
+
+        num_centroids = 2048 if num_centroids is None else num_centroids
```
LxYuan0420: Please replace the Gemma4 MTP magic numbers with named constants: default vocab size, default centroid count, default centroid top-k, and the single-token `n_predict` value. These are checkpoint/model-contract defaults, not obvious literals, and they currently appear in more than one file.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,789 @@
...
+                "Gemma4 MTP assistant requires a Gemma4 MTP architecture, got "
+                f"architectures={architectures!r}"
+            )
+
+        n_predict = _optional_int(config, "n_predict", context="assistant")
+        if n_predict is not None and n_predict != 1:
+            raise ValueError(
+                f"Gemma4 MTP assistant config must use n_predict=1, got {n_predict!r}"
+            )
+        _optional_bool(config, "tie_word_embeddings", context="assistant", default=True)
```
LxYuan0420: This validates `tie_word_embeddings` and discards the value. If this is part of the assistant metadata contract, please store it on `Gemma4MTPAssistantMetadata`; otherwise validate it where the model constructor actually consumes it.

vllm_metal/v1/gemma4_mtp_model.py
```diff
@@ -0,0 +1,127 @@
...
+            raise ValueError("Gemma4 MTP assistant requires backbone_hidden_size > 0")
+
+        # All assistant layers are Q-only and read target K/V in the later
+        # KV-sharing PR. Setting num_kv_shared_layers here keeps mlx-lm's
+        # Gemma4 layer shapes aligned with the assistant checkpoint.
+        text_config["num_kv_shared_layers"] = num_hidden_layers
+        self.text_args = gemma4_text.ModelArgs.from_dict(text_config)
+        self.model = Gemma4MTPAssistantBackbone(self.text_args)
+        self.pre_projection = nn.Linear(
+            2 * args.backbone_hidden_size,
```
LxYuan0420: Can we make these model-shape contracts explicit? `num_hidden_layers` should check missing vs invalid directly instead of using `get(..., 0)`, `num_kv_shared_layers` should not be silently overwritten if the config provides a conflicting value, and `2 * backbone_hidden_size` should be named or commented so the future forward path knows what the two inputs are.

tests/test_gemma4_mtp_assistant.py
```diff
@@ -0,0 +1,1026 @@
...
+        download_fn=lambda model_name, revision: tmp_path,
+    )
+
+    loader.load_if_needed(
+        speculative_config=_speculative_config(),
+        target_hf_config=None,
+        target_model_args=_target_args(),
+    )
+
+    assert len(load_model_paths) == 1
```
LxYuan0420, line 850: Please add tests for the two contracts above: positional layer-layout compatibility, and the default Hub download path with an indexed checkpoint whose shard names are not `model*.safetensors`. The current tests cover local custom shards and injected `download_fn`, but not the actual default download boundary.

LxYuan0420 (changes requested):

ricky-chaoju:
All done.

vllm_metal/v1/gemma4_mtp_constants.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Gemma4 MTP checkpoint contract constants."""
+
+GEMMA4_MTP_DEFAULT_VOCAB_SIZE = 262144
+GEMMA4_MTP_DEFAULT_NUM_CENTROIDS = 2048
+GEMMA4_MTP_DEFAULT_CENTROID_TOP_K = 32
+GEMMA4_MTP_N_PREDICT = 1
+GEMMA4_MTP_PREPROJECTION_INPUTS = 2
```
LxYuan0420: I don’t think this needs a separate module. Five Gemma4 MTP constants are not enough surface area for a constants-only file, and it makes readers jump across files to understand the checkpoint contract. Can we move these next to the owner that uses them, likely `gemma4_mtp.py` for validation constants and `gemma4_mtp_model.py` for model-construction constants?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,814 @@
...
+
+_ASSISTANT_DOWNLOAD_ALLOW_PATTERNS = [
+    "config.json",
+    "generation_config.json",
+    "model*.safetensors.index.json",
+    "*.safetensors",
+]
+
+_ASSISTANT_CACHE: dict[tuple[str, str | None], Gemma4MTPAssistantRuntime] = {}
+_ASSISTANT_CACHE_LOCK = Lock()
```
LxYuan0420: Cache ownership should live on `Gemma4MTPAssistantLoader`, not as module globals plus a separate reset wrapper. Can we make these `ClassVar`s on the loader and expose `Gemma4MTPAssistantLoader.clear_cache()` for `model_lifecycle.reset_model_cache()`?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,814 @@
...
+
+class Gemma4MTPAssistantLoader:
+    """Loads and validates a Gemma4 MTP assistant runtime."""
+
+    def __init__(
+        self,
+        *,
+        load_model_fn: Callable[..., tuple[Any, dict[str, Any]]] | None = None,
+        download_fn: Callable[[str, str | None], Path] | None = None,
+        model_path_resolver: Callable[[str], str] = _identity_model_path,
```
LxYuan0420: `_identity_model_path()` adds another module helper for a default that can stay inside the loader. Can we make `model_path_resolver` optional and default it in `__init__` instead of keeping a module-level identity function?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,814 @@
...
+            architectures = config.get("architectures")
+            raise ValueError(
+                "Gemma4 MTP assistant loader only supports gemma4_assistant/"
+                f"gemma4_mtp configs, got model_type={model_type!r}, "
+                f"architectures={architectures!r}"
+            )
+        return cls._assistant_model_classes()
+
+    @staticmethod
+    def _assistant_model_classes() -> tuple[type[Any], type[Any]]:
```
LxYuan0420: `_assistant_model_classes()` only exists so `_get_model_classes()` can call it once. The lazy import is valid, but the extra method adds another private seam without buying much. Can we inline the lazy import into `_get_model_classes()`?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,814 @@
...
+        if self.layer_types != target_tail:
+            raise ValueError(
+                "Gemma4 MTP assistant layer types must tail-match target "
+                "non-shared KV layer types: "
+                f"assistant_layer_types={self.layer_types}, "
+                f"target_tail_layer_types={target_tail}"
+            )
+
+
+def reset_gemma4_mtp_assistant_cache() -> None:
```
LxYuan0420: These module-level wrappers mostly mirror the class APIs and make the file feel like it has two public surfaces. Can we keep module scope limited to real production entry points, and have callers/tests use the owning classes directly where possible?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,814 @@
...
+    )
+    if isinstance(assistant_config, Gemma4MTPAssistantMetadata):
+        metadata = assistant_config
+    else:
+        metadata = Gemma4MTPAssistantMetadata.from_config(assistant_config)
+    metadata.validate_compatible_with(target_metadata)
+    return metadata
+
+
+def _required_layer_types(config: Any, num_hidden_layers: int) -> tuple[str, ...]:
```
LxYuan0420: `_required_layer_types()` is assistant-metadata parsing logic and is only used by `Gemma4MTPAssistantMetadata.from_config()`. Can we move it onto `Gemma4MTPAssistantMetadata` so the assistant config contract is owned in one place?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,814 @@
...
+            f"len(layer_types)={len(layer_types)}, "
+            f"num_hidden_layers={num_hidden_layers}"
+        )
+    unknown = sorted(set(layer_types) - GEMMA4_MTP_VALID_LAYER_TYPES)
+    if unknown:
+        raise ValueError(f"Unsupported Gemma4 MTP assistant layer types: {unknown}")
+    return layer_types
+
+
+def _matching_target_value[T](
```
LxYuan0420: This is target-metadata reconciliation logic, not a shared module utility. Can we move `_matching_target_value()` onto `Gemma4MTPTargetMetadata` and keep the target parsing contract inside that owner?

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,814 @@
...
+    for label, other in sources[1:]:
+        if other != value:
+            raise ValueError(
+                f"Gemma4 MTP target {key} metadata mismatch: "
+                f"{sources[0][0]}={value}, {label}={other}"
+            )
+    return value
+
+
+def _required_int(config: Any, key: str, *, context: str) -> int:
```
LxYuan0420: This tail of generic `Any` helpers makes the loader read like a mini parser framework instead of a Gemma4 MTP contract. `_field()` and `_config_value()` are also functionally the same. Can we collapse the duplicate accessor and move the int/bool/sequence parsing helpers onto the metadata class that actually owns each field?

vllm_metal/v1/gemma4_mtp_model.py
```diff
@@ -0,0 +1,159 @@
...
+    def layers(self) -> list[gemma4_text.DecoderLayer]:
+        return self.model.layers
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        raise NotImplementedError(
+            "Gemma4 MTP assistant forward is not wired on Metal yet."
+        )
+
+
+def _required_positive_int(config: dict[str, Any], key: str) -> int:
```
LxYuan0420: This helper is only used by `Gemma4MTPAssistantModel.__init__()`. Can we inline it or make it a private method on the model class? Keeping one-off module helpers around makes the model file harder to scan.

LxYuan0420:
Thanks for addressing the correctness feedback. The behavior issues are much better now: shard downloads support non-`model*` safetensors, layer compatibility is tail/positional, `tie_word_embeddings` is recorded, shape checks are stricter, and the tests cover those paths.

I left a few inline comments on structure/readability. The target shape I’d like is:

```py
# gemma4_mtp.py
# Keep Gemma4 MTP validation/loading constants here.

@dataclass(frozen=True, slots=True)
class Gemma4MTPAssistantRuntime:
    ...

class Gemma4MTPAssistantLoader:
    # Public:
    #   load_if_needed()
    #   clear_cache()
    #
    # Owns cache, download, preflight, mlx_lm.load_model,
    # and lazy assistant model-class import.
    ...

@dataclass(frozen=True, slots=True)
class Gemma4MTPAssistantSource:
    # Owns speculative_config parsing and cache key.
    ...

@dataclass(frozen=True, slots=True)
class Gemma4MTPTargetMetadata:
    # Owns target metadata normalization.
    ...

@dataclass(frozen=True, slots=True)
class Gemma4MTPAssistantMetadata:
    # Owns assistant config parsing and compatibility validation.
    ...
```

This is not style polish; it is about making the ownership contract obvious before the KV-sharing PR builds on this. Also please rebase/clean the branch so this PR only contains the Gemma4 MTP assistant loader slice.

LxYuan0420 (commented):
Let's spend some time together to make this base structure right; I believe it will make our future PRs / works easier

ricky-chaoju:
I moved the Gemma4 MTP contracts back under their owning classes/files and pushed the cleanup.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -0,0 +1,810 @@
...
+    get_text_config = getattr(config, "get_text_config", None)
+    if callable(get_text_config):
+        return get_text_config()
+    return _config_value(config, "text_config", config)
+
+
+def _config_value(config: Any, key: str, default: Any = None) -> Any:
+    if isinstance(config, Mapping):
+        return config.get(key, default)
+    return getattr(config, key, default)
```
LxYuan0420, line 825: this still feels like over-validating standard HF config shape. We should validate the Gemma4 MTP constraints we own, but avoid re-implementing generic schema/type coercion for normal HF fields. In a follow-up, I would prefer trusting the HF/MLX basic field types, or hiding this behind a small parser with one public `parse()` method, like:

```python
class _Gemma4MTPAssistantConfigParser:
    def parse(self, config: Any) -> Gemma4MTPAssistantMetadata:
        ...

class Gemma4MTPAssistantMetadata:
    @classmethod
    def from_config(cls, config: Any) -> Gemma4MTPAssistantMetadata:
        return _Gemma4MTPAssistantConfigParser().parse(config)
```

LxYuan0420 (approved):
Approve so we can keep things moving.

PR #375, opened 2026-05-15
[kernel] long prefill GQA kernel [2/3]

WindChimeRan:
### What changed

- **BQ 8 → 32, column-parallel → row-parallel.** Each simdgroup now owns 8   complete Q rows (`BQ / NUM_SG = 32 / 4`) instead of a column slice.
- **Q, O accumulator, and online-softmax state (m/l) are register-resident**   for the entire KV loop. Q is loaded into `Q_smem` once, hoisted to register   fragments, and `Q_smem` is never read again (reused as the float output   staging buffer at exit). No per-tile `S_smem`/`O_smem` round-trips. 
- **Per-row softmax reduction is a register-only `simd_shuffle_xor{1,8}`  butterfly** (the 4 lanes that share a fragment row), replacing the   `S_smem`-staged reduction.
- add Threadgroup memory padding (bank conflict avoidance)

- **Barriers: 4 → 2 per KV tile.**

### Benchmark (Sonnet 8192+10, 20 prompts, rate=10, conc=32 · Qwen3-0.6B · M1 Pro 32 GB · mem-fraction 0.4)

**Server** (terminal 1):
```bash
PYTHONPATH=. \
VLLM_METAL_MEMORY_FRACTION=0.4 \
vllm serve Qwen/Qwen3-0.6B \
  --max-model-len 8704 \
  --host 127.0.0.1 --port 8000
```
Client (terminal 2, once /health is green):
```bash
vllm bench serve \
  --backend vllm \
  --base-url http://127.0.0.1:8000 \
  --model Qwen/Qwen3-0.6B \
  --endpoint /v1/completions \
  --dataset-name sonnet --dataset-path sonnet.txt \
  --sonnet-input-len 8192 --sonnet-output-len 10 \
  --num-prompts 20 --request-rate 10 --max-concurrency 32
```

| Metric | main | this PR | Change | Speedup |
|---|---:|---:|---:|---:|
| Total tok/s | 243.09 | 674.87 | +177.6% | 2.78× |
| Mean TTFT (ms) | 351,811 | 126,021 | −64.2% | 2.79× faster |
| Mean TPOT (ms) | 7,449 | 2,725 | −63.4% | 2.73× faster |
| Duration (s) | 672.0 | 242.1 | −64.0% | 2.78× faster |

≈2.68× across the board on 8K long prefill.

### Behavioral changes / caveats

**head_size 192/256 no longer route to the tiled kernel** (BQ=32 raises the  shmem budget past the threadgroup limit with separate K/V buffers). They   fall through to the existing per-token v2 path — correct, just not  accelerated. 64/96/128 unaffected. large head_size  falls through to per-token v2 path. 
 
In other words, this optimized new kernel does not applicable to gemma4 models (256 & 512 head_dim), same with the mlx_lm.

The root cause is, apple has a 32KB small budget for threadgroup, far smaller than cuda counterpart. 

**Will address this issue in the next PR, but tuning BQ to smaller number to accommodate larger head_dim.**

### DIFF
--- a/tests/test_primitive_and_donation.py
+++ b/tests/test_primitive_and_donation.py
@@ -35,11 +35,11 @@ def _make_cache_and_inputs(
     num_query_heads: int,
     seq_lens: list[tuple[int, int]],
     *,
+    head_size: int = HEAD_SIZE,
     dtype: mx.Dtype = DTYPE,
 ):
     """Build a populated cache and matching query/metadata tensors."""
     block_size = BLOCK_SIZE
-    head_size = HEAD_SIZE
     num_seqs = len(seq_lens)
     query_lens = [s[0] for s in seq_lens]
     kv_lens = [s[1] for s in seq_lens]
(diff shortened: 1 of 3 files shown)


LxYuan0420 (approved):
LGTM; Can we add a small targeted parity test for the tiled-routed head sizes `{64, 96, 128}`? No need to expand the full matrix; one prefill shape is enough so CI covers every specialization routed by `can_use_tiled_kernel()`.

PR #377, opened 2026-05-16
Bump vLLM to 0.21.0

ricky-chaoju:
Bump vLLM to 0.21.0


WindChimeRan (approved):
* deterministic test passed (qwen3-0.6B)
* benchmark no regression (1% faster than before, within noise)

PR #378, opened 2026-05-17
[kernel] Deprecate kernel_v1

WindChimeRan:
Remove the dead `kernels_v1/` Metal path.

All v1 lessons have been absorbed by v2 (e.g., exp2 trick)

v2 is better than v1 on every aspect. especially, v1 is not compatible with varlen.


vllm_metal/paged_attention_backend/mha.py
```diff
@@ -1,60 +1,28 @@
...
 
 logger = init_logger(__name__)
 
-# Substring present in Metal shader compilation errors when the OS's Metal
-# language version is too old for the kernel. Matched against str(e) because
-# the C++/nanobind layer does not raise a typed exception for this case.
-_METAL_LANGUAGE_VERSION_ERROR = "language version"
-
 
 def warm_up_paged_cache(cache: MetalPagedKVCache) -> None:
```
LxYuan0420, line 18: `warm_up_paged_cache()` should not become a no-op. `MetalModelRunner.warm_up()` still calls this path, so it should keep validating startup readiness. Please make it load/JIT the current kernels, e.g. via `get_ops()`, without calling the deleted `reshape_and_cache`.

vllm_metal/metal/__init__.py
```diff
@@ -23,7 +22,6 @@
 logger = logging.getLogger(__name__)
 
 _THIS_DIR = Path(__file__).resolve().parent
-_KERNELS_DIR = _THIS_DIR / "kernels_v1"
 _KERNELS_V2_DIR = _THIS_DIR / "kernels_v2"
```
LxYuan0420, line 25: Since this PR deletes `kernels_v1/`, please also update `vllm_metal/metal/README.md`. It still says `kernels_v1/` is active; it should describe the current `kernels_v2/` layout and remove the old deprecation plan.

LxYuan0420 (approved):

vllm_metal/paged_attention_backend/mha.py
WindChimeRan: I'll solve it in a separated PR. Bigger problem: the `warm_up_paged_cache` should not be placed in mha.py. Currently other attention backend import from mha. This is a bad abstraction.

PR #380, opened 2026-05-17
[Refactor] Relocate kernel warm-up out of the MHA backend

WindChimeRan:
followup refactoring from #378 

 Move JIT kernel warm-up from `paged_attention_backend/mha.py` into `vllm_metal/metal/` next to `get_ops()` as a param-less `warm_up_kernels()`, so Hybrid no longer cross-imports from the MHA module. MLA's `warm_up()` now warms up consistently with MHA/Hybrid

### DIFF
--- a/vllm_metal/paged_attention_backend/mla.py
+++ b/vllm_metal/paged_attention_backend/mla.py
@@ -7,14 +7,12 @@ from typing import Any
 import mlx.core as mx
 import mlx.nn as nn
 from mlx_lm.models.base import scaled_dot_product_attention
-from vllm.logger import init_logger
 
 from vllm_metal import envs
+from vllm_metal.metal import warm_up_kernels
 from vllm_metal.metal_kernel_backend.packed_prefill_compat import apply_packed_rope
 from vllm_metal.mlx_backend.mla_cache import MLAPagedLatentCache
 from vllm_metal.paged_attention_common import find_attn_attr, find_layers, get_context
-
-logger = init_logger(__name__)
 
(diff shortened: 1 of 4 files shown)


vllm_metal/paged_attention_backend/mla.py
```diff
@@ -444,9 +444,8 @@ def _patch_model(self, model: Any, latent_cache: MLAPagedLatentCache) -> int:
         return patched
 
     def warm_up(self) -> None:
-        # MLX ops JIT-compile on first use; no Metal shader warm-up needed.
         self._require_initialized("warm_up")
-        logger.info("MLA paged attention (MLX-native): skipping Metal kernel warm-up")
+        warm_up_kernels()
```
LxYuan0420, line 448: Minor follow-up: can we gate MLA warm-up on `envs.VLLM_METAL_MLA_KERNEL`? MLA defaults to MLX SDPA, and the Metal MLA kernel is opt-in, so `warm_up()` should keep the old skip behavior when the flag is off and call `warm_up_kernels()` only when it is enabled.

LxYuan0420 (approved):

vllm_metal/paged_attention_backend/mla.py
WindChimeRan: #375 will unlock the flashatt prefilling kernel development for MLA. I hope to make the MLA kernel be the default and deprecate the env var soon ...

LxYuan0420: Aligned.

PR #384, opened 2026-05-19
Add Gemma4 MTP assistant KV sharing

ricky-chaoju:
Prepare Gemma4 MTP assistant KV sharing on Metal.

This is PR 4 in the Gemma4 MTP stack. It keeps the runner out of the model-specific policy: Gemma4 owns the assistant-to-target KV mapping, the cache policy installs a runner-local KV sharing binding after target paged KV allocation, and SDPA exposes the read-existing-KV path for the next assistant forward PR.

- Add `Gemma4MTPKVSharingPlan` for assistant layer -> target cache index mapping.
- Match upstream Gemma4 proposer behavior by mapping each assistant layer to the last non-shared target layer of the same attention type.
- Install a runner-local `Gemma4MTPKVSharingBinding` without patching the cached assistant model.
- Expose the initialized MHA paged KV cache through the MHA backend owner.
- Keep wrapper cache rebinding owned by `MetalKernelPagedAttentionWrapper`.
- Keep `forward_ready=False`; assistant forward and `take_draft_token_ids()` remain follow-up PRs.

### DIFF
--- a/vllm_metal/metal_kernel_backend/paged_attention.py
+++ b/vllm_metal/metal_kernel_backend/paged_attention.py
@@ -60,6 +60,7 @@ class MetalKernelPagedAttentionWrapper(nn.Module):
         block_size: int,
         *,
         cache_idx: int | None = None,
+        force_shared_kv: bool = False,
     ) -> None:
         super().__init__()
         object.__setattr__(self, "_inner", inner)
@@ -71,6 +72,46 @@ class MetalKernelPagedAttentionWrapper(nn.Module):
         object.__setattr__(
             self, "_mk_cache_idx", cache_idx if cache_idx is not None else layer_idx
         )
+        object.__setattr__(self, "_mk_force_shared_kv", force_shared_kv)
+
+    def _shared_kv_sentinel(self, x: mx.array) -> tuple[mx.array, mx.array]:
+        """Build dummy K/V tensors for layers that read cache-only K/V."""
+        if len(x.shape) != 3:
+            raise ValueError(
+                "Paged attention shared-KV sentinel expects input shaped "
+                f"(batch, tokens, hidden), got {x.shape}"
+            )
+
+        cache = self._mk_kv_cache
+        cache_idx = self._mk_cache_idx
+        if cache_idx < 0 or cache_idx >= cache.num_layers:
+            raise ValueError(
+                f"cache_idx={cache_idx} is outside target KV cache with "
+                f"{cache.num_layers} layers"
+            )
+
+        inner = self._inner
+        num_kv_heads = (
+            getattr(inner, "n_kv_heads", None)
+            or getattr(inner, "num_key_value_heads", None)
+            or cache.kv_heads_per_layer[cache_idx]
+        )
+        head_dim = getattr(inner, "head_dim", None)
+        k_proj = getattr(inner, "k_proj", None)
+        if head_dim is None and k_proj is not None:
+            head_dim = k_proj.weight.shape[0] // num_kv_heads
+        if head_dim is None:
+            head_dim = cache.head_dim_per_layer[cache_idx]
+
+        batch_size, seq_len, _ = x.shape
+        shape = (
+            batch_size,
+            int(num_kv_heads),
+            seq_len,
+            int(head_dim),
+        )
+        sentinel = mx.zeros(shape, dtype=x.dtype)
+        return sentinel, sentinel
 
     def __call__(
         self,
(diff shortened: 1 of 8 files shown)


vllm_metal/v1/gemma4_mtp.py
```diff
@@ -39,13 +42,137 @@
...
+        target_metadata: Gemma4MTPTargetMetadata,
+        target_kv_cache: MetalPagedKVCache,
+        block_size: int,
+    ) -> Gemma4MTPAssistantRuntime:
+        """Return a runtime whose assistant layers read target paged KV."""
+        plan = Gemma4MTPKVSharingPlan.from_metadata(
+            assistant=self.metadata,
+            target=target_metadata,
+        )
+        patched = plan.patch_assistant_model(
```
LxYuan0420: Blocking: `with_target_kv_sharing()` mutates `self.model`, but `Gemma4MTPAssistantLoader` caches and returns the same runtime/model object by `(model_name, revision)`. That makes the per-target KV cache binding process-global: if a second runner wires the same cached assistant to another target cache, the first runner's wrappers get retargeted too. I reproduced this locally by wiring cache1 then cache2; the first wired runtime ended up pointing at cache2. Please make the KV binding runner-local, either by returning a fresh assistant runtime/model before patching or by keeping the cached runtime unpatched and non-shared. Also add a two-cache regression test.

vllm_metal/metal_kernel_backend/paged_attention.py
```diff
@@ -71,6 +72,46 @@ def __init__(
         object.__setattr__(
             self, "_mk_cache_idx", cache_idx if cache_idx is not None else layer_idx
         )
+        object.__setattr__(self, "_mk_force_shared_kv", force_shared_kv)
+
+    def _shared_kv_sentinel(self, x: mx.array) -> tuple[mx.array, mx.array]:
```
LxYuan0420: Can we avoid using a fake zero `shared_kv` tensor as a control flag? `shared_kv` now means both "real YOCO K/V from another layer" and "cache-read-only mode for MTP," which makes the wrapper hard to reason about and forces duplicate head-shape probing here. A cleaner shape would be an explicit flag on `sdpa_forward`, e.g. `read_existing_kv=True` or `skip_cache_write=True`, with `attention_sdpa.py` owning the Q-only prep/cache-read behavior.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -39,13 +42,137 @@
...
+            if attn_attr is None:
+                raise ValueError(
+                    f"Gemma4 MTP assistant layer {layer_idx} has no attention module"
+                )
+            entries.append((layer_idx, layer, attn_attr, cache_idx))
+
+        patched = 0
+        for layer_idx, layer, attn_attr, cache_idx in entries:
+            attn = getattr(layer, attn_attr)
+            if isinstance(attn, MetalKernelPagedAttentionWrapper):
```
LxYuan0420: This reaches into `MetalKernelPagedAttentionWrapper` private `_mk_*` fields from outside the wrapper owner. Since both normal patching and MTP patching need to rebind wrapper state, please put that ownership on the wrapper itself, e.g. `attn.rebind_cache(kv_cache, block_size, cache_idx, force_shared_kv=True/False)`. That removes the duplicated `object.__setattr__` blocks and keeps wrapper internals inside `paged_attention.py`.

vllm_metal/v1/cache_policy.py
```diff
@@ -360,6 +361,32 @@ def build_paged_attention_backend(
...
+        """Wire Gemma4 MTP assistant layers to the target paged KV cache."""
+        assistant = self._runner._gemma4_mtp_assistant
+        if assistant is None:
+            return
+        if not isinstance(backend, MHAPagedAttentionBackend):
+            raise NotImplementedError(
+                "Gemma4 MTP assistant KV sharing requires the MHA paged "
+                "attention backend on Metal."
+            )
+        model_config = getattr(self._runner, "model_config", None)
```
LxYuan0420: This is an internal `MetalModelRunner` path, not a boundary. Please use the typed runner fields directly instead of `getattr(..., None)`, e.g. `self._runner.model_config.hf_config`. If `model_config` or `hf_config` is missing here, that is a lifecycle bug we should fail on plainly, not mask with `None`.

LxYuan0420:
Thanks for keeping this PR scoped to the KV-sharing slice; it lines up well with the roadmap. I left a few comments. The blocker for me is the lifetime issue from mutating a process-cached assistant runtime with a per-target KV cache. The other comments are to make the read-only cache contract easier to maintain before the scheduler handoff PR builds on this.

LxYuan0420 (approved):
LGTM; Please also update the PR description to avoid saying we patch assistant attention layers. The current code installs a runner-local KV sharing binding and prepares the read-existing-KV path; the actual assistant forward consumption is still in the next PR.

PR #387, opened 2026-05-22
Add Gemma4 MTP scheduler handoff

ricky-chaoju:
Wire Gemma4 MTP assistant drafting into the vLLM v1 scheduler handoff path.

This is PR 5 in the Gemma4 MTP stack. It keeps `MetalModelRunner` as the orchestration layer: spec-decode seed selection lives in `SpeculativeDecodeController`, Gemma4 assistant execution lives in the Gemma4 runtime/model owner and the worker only delegates `take_draft_token_ids()`.

Changes:
- Add `take_draft_token_ids()` on the Metal runner/worker path.
- Build Gemma4 MTP draft seeds from accepted decode rows and final prefill rows.
- Run the assistant over runner-local target KV sharing and return one draft token per seed.
- Add target input-embedding handoff through `ModelAdapter`, including Gemma4 `language_model` wrappers.
- Fail fast for Gemma4 MTP with async scheduling; this path requires `--no-async-scheduling`.
- Add regressions for decode/prefill draft handoff, mixed batches, structured output, worker delegation, adapter wrappers, and assistant KV-context drafting.


vllm_metal/v1/gemma4_mtp_model.py
```diff
@@ -143,10 +188,177 @@ def __init__(self, args: Gemma4MTPAssistantModelArgs) -> None:
...
+        )
+        if len(target_cache_indices) != len(self.layers):
+            raise ValueError(
+                "Gemma4 MTP target_cache_indices must match assistant layers: "
+                f"indices={len(target_cache_indices)}, layers={len(self.layers)}"
+            )
+
+        combined = mx.concatenate([input_embeddings, hidden_states], axis=-1)
+        hidden_states = self.pre_projection(combined)
+        for layer, cache_idx in zip(self.layers, target_cache_indices, strict=True):
```
LxYuan0420, line 236: Can we make this contract explicit in model init before `ModelArgs.from_dict()`? This custom forward never passes `per_layer_input`, but upstream `gemma4_text.ModelArgs` defaults `hidden_size_per_layer_input` to `256` when the field is missing. That creates per-layer-input modules that this path silently skips. Please normalize missing per-layer-input fields to `0`, or reject non-zero values, and add a regression test.

vllm_metal/v1/gemma4_mtp_model.py
```diff
@@ -143,10 +188,177 @@ def __init__(self, args: Gemma4MTPAssistantModelArgs) -> None:
...
+        target_cache_indices: Sequence[int],
+    ) -> tuple[mx.array, mx.array]:
+        """Run one Gemma4 MTP assistant draft step over target KV cache.
+
+        ``input_ids`` is accepted for parity with upstream's proposer API; this
+        Metal path receives already-scaled target/backbone embeddings from the
+        target model because the assistant's own embedding table is draft-dim
+        and is kept for logits.
+        """
+        del input_ids
```
LxYuan0420: Small cleanup: `del input_ids` makes this look like ownership/lifetime matters. Since the arg is only here for upstream API parity, I’d leave it unused and let the docstring explain why.

vllm_metal/v1/gemma4_mtp.py
```diff
@@ -83,7 +85,114 @@ def with_target_kv_sharing(
...
+    ) -> list[list[int]]:
+        """Run the assistant and return one draft token per seed."""
+        if not seeds:
+            return []
+        if self.kv_sharing is None:
+            raise RuntimeError("Gemma4 MTP assistant requires target KV sharing")
+        if not self.forward_ready:
+            raise RuntimeError("Gemma4 MTP assistant forward is not ready")
+
+        import mlx.core as mx
```
LxYuan0420: Can we simplify this happy path? The lazy imports, seed re-validation, row selection helper, and shape checks make this read more defensive than necessary. Since seeds come from our controller, I’d keep only checks that catch a real boundary/version mismatch.

vllm_metal/v1/gemma4_mtp_model.py
```diff
@@ -143,10 +188,177 @@ def __init__(self, args: Gemma4MTPAssistantModelArgs) -> None:
...
+        if ndim == 2:
+            return value[None, :, :]
+        if ndim == 3 and value.shape[0] == 1:
+            return value
+        raise ValueError(
+            f"Gemma4 MTP {name} must be row-major [num_tokens, hidden] or "
+            f"single-batch [1, num_tokens, hidden], got shape {value.shape}"
+        )
+
+    @staticmethod
```
LxYuan0420: This is model-owned forward logic, not a reusable utility. Can we avoid making it a private `@staticmethod` and keep the ownership clearer, either as an instance helper or closer to the forward loop?

vllm_metal/v1/gemma4_mtp_model.py
LxYuan0420: better leave it untouched as in dont do del

vllm_metal/v1/model_runner.py
```diff
@@ -1193,8 +1208,79 @@ def _sample_paged_batch(
...
+            return
+
+        input_ids = mx.array([[seed.token_id for seed in seeds]], dtype=mx.int32)
+        target_input_embeddings = self._target_input_embeddings(input_ids)
+        draft_token_ids = assistant.propose_draft_token_ids(
+            seeds=seeds,
+            target_hidden_states=target_hidden_states,
+            target_input_embeddings=target_input_embeddings,
+        )
+        if len(draft_token_ids) != len(seeds):
```
LxYuan0420: This duplicates the length check already done by `Gemma4MTPAssistantRuntime.propose_draft_token_ids()`. Since this is internal typed handoff, I’d keep the invariant in one owner and let the runner stay orchestration-only.

LxYuan0420 (approved):

Issue #361, opened 2026-05-09
Feature: Support vLLM pooling tasks on Metal V1, starting with embeddings


CAICAIIs:
## Summary

vLLM already supports pooling models for embedding and classification workloads, but vllm-metal does not currently provide a first-class Metal V1 pooling execution path.

This issue proposes adding vllm-metal support for the existing vLLM pooling model contract, starting with sequence embeddings.

This is not proposing a new vLLM API. The goal is to make vllm-metal produce `ModelRunnerOutput.pooler_output` for compatible pooling models on Apple Silicon.

## Motivation

Embedding and classification workloads are important for local AI infra use cases such as:

- RAG
- semantic search
- document retrieval
- reranking
- moderation/classification pipelines

Today vllm-metal is mainly useful for generation and transcription workloads. Supporting pooling tasks would make Apple Silicon more useful as a local inference backend for retrieval and classification systems as well.

### Stage 1: Task detection and routing

- Detect compatible pooling models / pooling requests.
- Advertise pooling task support from the Metal worker when applicable.
- Keep generation models unchanged.
- Fail fast with clear errors for unsupported pooling variants.

### Stage 2: Sequence embedding MVP

- Support the `embed` pooling task for text-only requests.
- Run prefill-only model execution.
- Do not sample tokens for pooling requests.
- Return one embedding vector per request through `ModelRunnerOutput.pooler_output`.
- Add tests for routing, output shape, and unsupported variants.

### Stage 3: Classification follow-up

- Support the `classify` pooling task when the loaded model exposes compatible classification or pooler outputs.
- Return one score/logit vector per request.

## Acceptance Criteria

- Metal worker advertises pooling tasks for compatible pooling models.
- `embed` requests return `pooler_output`, not sampled token IDs.
- Generation behavior remains unchanged.
- Unsupported pooling variants raise clear, actionable errors.
- At least one small MLX embedding model has a passing local smoke test.

## Validation Plan

- Unit tests for `MetalModelRunner.supported_worker_tasks()`.
- Unit tests for pooling request routing and `ModelRunnerOutput.pooler_output`.
- Smoke test with a small MLX embedding model.
- If the serving path is wired, verify that `/v1/embeddings` returns one vector per input.

callmezhuoran:
Hi @CAICAIIs  Could you please list some model that need to be supported first ？ I‘m interested in it;

CAICAIIs:
[...]

PR #391, opened 2026-05-23
Add Qwen3 reranker classify pooling

CAICAIIs:
## Summary

Adds a narrow Metal V1 text pooling follow-up for original Qwen3 reranker `classify` support.

This PR supports only the Qwen3 reranker contract:

- `Qwen3ForSequenceClassification`
- `classifier_from_token=["no", "yes"]`
- `is_original_qwen3_reranker=True`
- tied `model.model.embed_tokens.as_linear`
- paged pooling path

The score is computed from the final prompt-token hidden state using `yes_logit - no_logit`, with optional `logit_mean` / `logit_sigma` calibration and `use_activation` handling.

## Scope

This intentionally does not add generic classification or reranking support. Generic classification heads, generic reranking models, late interaction, token-level pooling, multimodal pooling, direct embedding tensors, and non-paged pooling remain unsupported.

## Changes

- Adds Qwen3 reranker `classify` pooling helpers in `pooling.py`
- Keeps `model_runner.py` as a thin router/delegator
- Adds contract tests for task advertisement, output order, raw/activated scores, calibration, and fail-fast behavior
- Updates text pooling docs and supported model evidence

## Validation

- `git diff --check`
- `.venv-vllm-metal/bin/python -m pytest tests/test_v1_pooling.py -q`
  - `32 passed`
- `.venv-vllm-metal/bin/python -m pytest tests/test_v1_model_runner_generate.py tests/test_v1_worker.py tests/test_paged_attention.py tests/test_sampling.py tests/test_grammar_bitmask.py -q`
  - `119 passed`
- `scripts/lint.sh`
  - shellcheck, ruff, ruff format check, and mypy passed

Real Qwen3 reranker smoke:

- Offline `LLM.score` on `mku64/Qwen3-Reranker-0.6B-mlx-8Bit`
  - revision `ba80418a47fa1c4368a6c2287b0e449904063576`
  - `reranker-offline-score-ok 2 [0.973079, 0.931326]`
- Server `/score`
  - `/health` returned HTTP 200
  - `/score` returned HTTP 200 with two scalar scores

Refs #361

### DIFF
--- a/vllm_metal/v1/pooling.py
+++ b/vllm_metal/v1/pooling.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Text embedding pooling helpers for the Metal V1 model runner."""
+"""Text pooling helpers for the Metal V1 model runner."""
 
 from __future__ import annotations
 
@@ -8,11 +8,15 @@ from typing import Any
 import mlx.core as mx
 import torch
 from vllm.pooling_params import PoolingParams
+from vllm.tasks import SupportedTask
 
 from vllm_metal.pytorch_backend.tensor_bridge import mlx_to_torch
 
 _EMBED_POOLER_TASKS = (None, "embed")
+_CLASSIFY_POOLER_TASKS = (None, "classify")
+_SUPPORTED_POOLER_TASKS = (None, "embed", "classify")
 _LAST_POOLING = (None, "LAST")
+_QWEN3_RERANKER_TOKENS = ("no", "yes")
 
 
 def _model_label(model_config: Any) -> str:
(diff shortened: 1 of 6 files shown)


vllm_metal/v1/pooling.py
```diff
@@ -109,12 +121,30 @@ def _is_decoder_embedding_config(model_config: Any) -> bool:
...
+    )
+
+
 def sequence_model(model: Any) -> Any | None:
     """Return the MLX transformer body when it is available."""
     inner = getattr(model, "model", None)
     return inner if callable(inner) else None
 
 
+def _classifier_logits_fn(model: Any) -> Any | None:
```
WindChimeRan: The PR always computes them via `embed_tokens.as_linear`,  which is the correct output head only when tied. On an untied model (like Qwen3-Reranker-8B), the real head is the separate lm_head, so projecting through embed_tokens gives the wrong logits.

CAICAIIs: Sorry, you were right. I had incorrectly assumed `embed_tokens.as_linear` was always the correct projection head.

I’ve updated the PR so untied checkpoints use the separate `lm_head`, while tied checkpoints use `embed_tokens.as_linear`. If the tie state is unknown, the model is no longer advertised as supported. I also added targeted tests covering the untied `lm_head` path and unsupported fallback cases.

Thanks for catching this.

WindChimeRan (approved):
Please don't squash mulitple commits into one commit next time. 

keeping commit history will make review easier.

CAICAIIs:
> Please don't squash mulitple commits into one commit next time.
> 
> keeping commit history will make review easier.

oh, sorry, got it!

PR #392, opened 2026-05-23
docs: improve README installation section clarity

HassanBahati:
The README **Installation** section is a bit easy to misparse.

The old text said:

> Using the install script, the following will be installed…

It wasn’t immediately clear which script that referred to. I also noticed the `curl | bash` command appeared after the paragraph about the vLLM CLI, which made the code block feel like it might belong to CLI usage rather than the vllm-metal install step.

## Proposed change
This PR makes a minimal reordering:

- Move the install one-liner to the **top** of the Installation section so the install step comes first.
- Refer to “the install script **above**” so the following bullets clearly describe that command.

No install behaviour changes.


ricky-chaoju (approved):
LGTM

PR #393, opened 2026-05-24
tests: lock in PrefixCacheManager eviction-order behavior

fxdv:
## Summary

Adds `TestPrefixCacheEvictionPolicy` to `tests/test_prefix_cache.py`: seven characterisation tests that pin the contract of `CachedPrefix.ref_count` and the entry the eviction loop selects.

The existing `TestPrefixCacheEviction` checks `len(_cache)` and `_current_bytes` post-eviction but never asserts **which** entry was evicted. That gap means today's policy (LFU-by-cumulative-hits, with `ref_count` initialised to 1, incremented on every `lookup` hit, and never decremented) would pass the same suite as LRU, FIFO, or random.

Scope of the new suite, against the unmodified implementation:

| Test | Asserts |
|---|---|
| `test_insert_initializes_ref_count_to_one` | `insert` sets `ref_count = 1` |
| `test_lookup_hit_increments_ref_count` | hit path does `+= 1` |
| `test_lookup_miss_does_not_touch_ref_count` | miss path is read-only |
| `test_restore_cache_does_not_decrement_ref_count` | `restore_cache` does not release |
| `test_eviction_drops_lowest_ref_count_entry` | eviction key is `ref_count` |
| `test_hot_prefix_pins_out_cold_prefixes` | LFU-forever steady-state: hot entry + 1 rotating slot |
| `test_ref_count_is_unbounded_monotonic` | no decrement path exists |

These are intentionally **characterisation** tests, not aspirational ones. They lock in today's behavior so any future change to the eviction policy (rename, LRU, pin-counter) shows up as an explicit test update rather than a silent behavioural shift. A companion design RFC discussing the options will follow in a separate PR.

This PR contains no production-code changes.

## Test plan

- [x] `ruff check tests/test_prefix_cache.py` clean
- [x] `ruff format --check tests/test_prefix_cache.py` clean
- [x] All 7 new tests pass against the current `PrefixCacheManager` implementation locally
- [ ] CI green on `not slow` pytest suite

### DIFF
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -467,6 +467,173 @@ class TestPrefixCacheEviction:
         assert mgr._current_bytes == bytes_after_first
 
 
+class TestPrefixCacheEvictionPolicy:
+    """Lock in the current eviction-order behavior.
+
+    ``PrefixCacheManager`` uses ``CachedPrefix.ref_count`` as its eviction
+    key. The name suggests pin-during-use semantics but the field is only
+    ever incremented (initialised to 1 on insert, ``+= 1`` on every
+    ``lookup`` hit, never decremented). The resulting policy is therefore
+    *LFU-forever*: an entry that accumulates many hits early in a run
+    cannot be displaced by a more recently-popular prefix.
+
+    These tests pin that behavior down so any future change to the policy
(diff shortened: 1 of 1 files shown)


WindChimeRan (commented):
see comments on #394

ricky-chaoju:
Closing with #394

PR #394, opened 2026-05-24
docs: design RFC for PrefixCacheManager eviction policy

fxdv:
## Summary

Adds an internal design note at `docs/design/prefix_cache_eviction.md` analysing the eviction policy of `PrefixCacheManager` (`vllm_metal/v1/contiguous_cache.py`).

**The finding, in one sentence:** `CachedPrefix.ref_count` is named and shaped like a pin-during-use counter but is actually a monotonic hit counter — initialised to `1` on insert, incremented on every `lookup` hit, never decremented anywhere in the codebase. The resulting eviction policy is *LFU-forever*: prefixes that accumulate hits early in a run pin themselves in and cannot be displaced by later-popular prefixes.

The RFC covers:

- Verified current behavior, the `O(N)` eviction selector, and a pathological steady-state walkthrough.
- **Blast radius:** only triggered when `VLLM_METAL_USE_PAGED_ATTENTION=0` *and* `VLLM_METAL_PREFIX_CACHE` is set before any `vllm_metal` import. Default v0.2.0 deployments are unaffected.
- **Why it matters:** wasted GPU work on shifting workloads, and a real use-after-free footgun if someone "adds the obvious decrement" *and* a future zero-copy `restore_cache` removes the `mx.array(k)` copy that currently isolates eviction from in-flight requests.
- Three remediation options:
  - **A** — rename `ref_count` → `hit_count`, document the contract. Minimal diff, closes the footgun. **Recommended.**
  - **B** — switch to LRU via `OrderedDict`. `O(1)`, matches user expectations, needs benchmark validation.
  - **C** — hybrid LRU + real pin counter. Strongest correctness, largest surface area, not load-bearing today.
- Migration notes for Option A.
- An explicit mapping from today's characterisation tests (companion PR #393) to their LRU equivalents if Option B is later picked up.

This PR is **documentation only**. It deliberately does **not** register the file in `mkdocs.yaml` — it is a project-internal design note rather than user-facing documentation. The recommended change itself (Option A: rename + comment) is left for a follow-up PR so this RFC can be reviewed on its own merits first.

## Companion PR

Pairs with #393, which lands the characterisation tests referenced in §7. The two PRs are independently mergeable; this one references the test names without containing them.

## Test plan

- [x] `ruff check` clean repo-wide
- [x] mkdocs build still succeeds (file intentionally not in nav)

### DIFF
--- a/docs/design/prefix_cache_eviction.md
+++ b/docs/design/prefix_cache_eviction.md
@@ -1,0 +1,266 @@
+# RFC: `PrefixCacheManager` Eviction Policy
+
+> Status: **Draft / discussion** &middot; Scope: `vllm_metal.v1.contiguous_cache.PrefixCacheManager` &middot; Author: @fxdv &middot; Companion PR: #393 (characterisation tests)
+
+This document is an internal design note, not user-facing documentation. It is intentionally **not** registered in `mkdocs.yaml`.
+
+## 1. Summary
+
+`PrefixCacheManager` uses a `CachedPrefix.ref_count` integer as its eviction
+priority key. The name and surrounding code (`restore_cache`, dataclass
+default of `0`) strongly suggest *pin-during-use* semantics. The runtime
+behavior is in fact a **monotonic hit counter**: it is initialised to `1`
+on insert, incremented on every `lookup` hit, and **never decremented
+anywhere in the codebase**. The resulting eviction policy is therefore
+*LFU-forever*: prefixes that accumulate hits early in a run pin themselves
+into the cache and cannot be displaced by later-popular prefixes.
+
+This RFC documents the verified behavior, the resulting failure modes,
+the trade-offs of three remediation options, and a recommended migration
+path.
+
+## 2. Blast radius
+
+`PrefixCacheManager` is instantiated only by `MetalModelRunner.__init__`
+under one condition (`vllm_metal/v1/model_runner.py:315-316`):
+
+```python
+self._prefix_cache: PrefixCacheManager | None = None
+if _PREFIX_CACHE_ENABLED:
+    self._prefix_cache = PrefixCacheManager(model_adapter=self._model_adapter)
+```
+
+`_PREFIX_CACHE_ENABLED` is evaluated at module import time from
+`envs.VLLM_METAL_PREFIX_CACHE`, which is a **presence-only** check
+(`vllm_metal/envs.py:68`). Once the manager exists, it is consumed in
+exactly one call site &mdash; `MetalModelRunner._prefill_single` &mdash; which is
+itself only reachable on the **legacy non-paged code path** (when
+`self._paged_attention_backend is None`, i.e. when the user has set
+`VLLM_METAL_USE_PAGED_ATTENTION=0`).
+
+Therefore the affected configuration is:
+
+- `VLLM_METAL_USE_PAGED_ATTENTION=0` (opt-out of the v0.2.0 default), **and**
+- `VLLM_METAL_PREFIX_CACHE` set in the environment **before any
+  `vllm_metal` import** (any value, including `"0"`, enables it).
+
+The default v0.2.0 deployment is unaffected.
+
+## 3. Verified current behavior
+
+### 3.1 Hot-counter mechanics
+
+| Event | Effect on `entry.ref_count` |
+|---|---|
+| `insert(new)` | created with value `1` |
+| `lookup(existing)` hit | `+= 1` |
+| `lookup(missing)` | no change |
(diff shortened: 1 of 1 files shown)


WindChimeRan (commented):
cc: @ricky-chaoju 

This PR is on contiguous kv cache path. I think we should no longer maintain this path. 

Can we reject this PR and delete the code?

ricky-chaoju:
Thanks for digging into this. I checked upstream as well: their prefix cache lives in the block manager, and ref_cnt is a live-use counter, not this monotonic hit counter. Since this is only the legacy non-paged PrefixCacheManager, I don’t want to lock the current policy in with either docs or characterization tests. I’m going to close both PRs and handle any cleanup separately.

fxdv:
Thanks @ricky-chaoju, @WindChimeRan — that's a clean call. The upstream cross-check on `ref_cnt` semantics in the block manager is exactly the second opinion this needed, and removing the legacy contiguous-KV path is a stronger answer than a rename or policy swap on dead code.

Happy to send the removal patch if it would help offload the work. A proposed minimum scope, with knobs for you to dial:

**Tier 1 — prefix cache only** (smallest, lowest risk)
- Remove `PrefixCacheManager`, `CachedPrefix`, `_compute_prefix_hash`, `_get_prefix_cache_max_bytes`, and the `_PREFIX_CACHE_ENABLED` import-time check from `vllm_metal/v1/contiguous_cache.py`.
- Drop the `_prefix_cache` field and its uses in `MetalModelRunner`.
- Drop `VLLM_METAL_PREFIX_CACHE` and `VLLM_METAL_PREFIX_CACHE_FRACTION` from `vllm_metal/envs.py` and the docs.
- Remove `TestPrefixCacheEviction`, `TestPrefixCacheEnableFlag`, and the `restore_cache` coverage in `tests/test_vlm_forward_model.py`.

**Tier 2 — full legacy non-paged path** (deeper, larger blast radius)
- Everything in Tier 1, plus
- `MetalModelRunner._prefill_single` and its branch in `execute_model`.
- Anything else gated on `self._paged_attention_backend is None` that's only there to support the legacy path.

I'd lean Tier 1 first as a clean independently-mergeable PR, then Tier 2 as a follow-up once the prefix cache surface is gone. But if you'd rather do it in-house, or want this scoped differently, just say.

ricky-chaoju:
> Thanks @ricky-chaoju, @WindChimeRan — that's a clean call. The upstream cross-check on `ref_cnt` semantics in the block manager is exactly the second opinion this needed, and removing the legacy contiguous-KV path is a stronger answer than a rename or policy swap on dead code.
> 
> Happy to send the removal patch if it would help offload the work. A proposed minimum scope, with knobs for you to dial:
> 
> **Tier 1 — prefix cache only** (smallest, lowest risk)
> 
> * Remove `PrefixCacheManager`, `CachedPrefix`, `_compute_prefix_hash`, `_get_prefix_cache_max_bytes`, and the `_PREFIX_CACHE_ENABLED` import-time check from `vllm_metal/v1/contiguous_cache.py`.
> * Drop the `_prefix_cache` field and its uses in `MetalModelRunner`.
> * Drop `VLLM_METAL_PREFIX_CACHE` and `VLLM_METAL_PREFIX_CACHE_FRACTION` from `vllm_metal/envs.py` and the docs.
> * Remove `TestPrefixCacheEviction`, `TestPrefixCacheEnableFlag`, and the `restore_cache` coverage in `tests/test_vlm_forward_model.py`.
> 
> **Tier 2 — full legacy non-paged path** (deeper, larger blast radius)
> 
> * Everything in Tier 1, plus
> * `MetalModelRunner._prefill_single` and its branch in `execute_model`.
> * Anything else gated on `self._paged_attention_backend is None` that's only there to support the legacy path.
> 
> I'd lean Tier 1 first as a clean independently-mergeable PR, then Tier 2 as a follow-up once the prefix cache surface is gone. But if you'd rather do it in-house, or want this scoped differently, just say.

cc @WindChimeRan

Tier 1 sounds right to me. That prefix-cache path made sense when the MLX contiguous path was the main path, but now it’s a separate opt-in policy that differs from upstream’s block-manager prefix cache.

Let’s keep this to removing VLLM_METAL_PREFIX_CACHE for now: the manager, env/docs, model runner hook, and the related tests. Please leave the non-paged MLX path and the contiguous cache merge/extract helpers in place.

fxdv:
Thanks @ricky-chaoju — Tier 1 only as agreed, with the non-paged MLX path and the contiguous cache merge/extract helpers left in place. Up at #399.

PR #395, opened 2026-05-25
[Refactor] Refactor model runner, mv stt to separated file

WindChimeRan:
## Why
STT (Whisper / Qwen3-ASR) is a one-shot transcribe: the model's runtime adapter runs the entire decode in a single `execute_model` call. It has no real KV cache and never runs on the paged path — it sits on the non-paged (contiguous-cache) side and only *fakes* vLLM's iterative scheduler contract (nominal memory, a placeholder KV spec, and an EOT token for each "cached" decode-phase request).

Bolting that onto the generation `MetalModelRunner` scattered `_is_stt` branches and fine-grained per-model inspection (config.json sniffing) across the runner, `cache_policy`, and `model_lifecycle` — knowledge that doesn't belong in the generic runner.

## What
- Extract the one-shot path into a dedicated `STTModelRunner`   (`vllm_metal/v1/stt_model_runner.py`); `_execute_stt` moved **verbatim**. - Select the runner once, at the worker (`init_device` via `is_stt_model`). 
- Drop every `_is_stt` branch from `MetalModelRunner`, `cache_policy`, and   `model_lifecycle`; the generation runner no longer knows STT exists.

## Verified
End-to-end via `vllm serve openai/whisper-tiny` → `/v1/audio/transcriptions`:
**byte-identical transcript vs `main`** on both explicit-language and auto-detect paths. No behavior change.

### DIFF
--- a/tests/test_v1_stt_integration.py
+++ b/tests/test_v1_stt_integration.py
@@ -403,8 +403,12 @@ class TestSTTRunnerWorkerContract:
         monkeypatch.setattr(smr, "load_stt_model", lambda _name: fake_model)
         monkeypatch.setattr(smr, "get_model_download_path", lambda _m: "resolved-path")
 
-        runner = smr.STTModelRunner.__new__(smr.STTModelRunner)
-        runner.model_config = SimpleNamespace(model="some-model")
+        vllm_config = SimpleNamespace(
+            model_config=SimpleNamespace(model="some-model"),
+            cache_config=SimpleNamespace(),
+            scheduler_config=SimpleNamespace(),
+        )
+        runner = smr.STTModelRunner(vllm_config, device=torch.device("cpu"))
         runner.load_model()
 
         assert runner.model is fake_model
--- a/tests/test_v1_stt_integration.py
+++ b/tests/test_v1_stt_integration.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Tests for v1 STT integration in MetalModelRunner."""
+"""Tests for v1 STT integration in STTModelRunner."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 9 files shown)


vllm_metal/v1/worker.py
```diff
@@ -64,7 +64,9 @@ class MetalWorker(WorkerBase):
     using MLX as the primary compute backend.
     """
 
-    # Override model_runner type from base class
+    # Override model_runner type from base class. Typed as MetalModelRunner
+    # because its worker-facing method set is a superset of STTModelRunner's;
+    # STT models are assigned an STTModelRunner at runtime (see init_device).
     model_runner: MetalModelRunner  # type: ignore[assignment]
```
LxYuan0420, line 70: Non-blocking: I am okay keeping this lightweight for now instead of adding a local runner Protocol, since `MetalWorker` mostly mirrors upstream `WorkerBase`. The important part is that STT stays isolated in `STTModelRunner` and we keep focused tests around the worker dispatch path. Just calling this out so we do not grow this into another untyped runner surface later.

tests/test_v1_stt_integration.py
```diff
@@ -400,11 +380,38 @@ def test_get_cache_block_size_bytes_returns_constant(self) -> None:
...
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        import vllm_metal.v1.stt_model_runner as smr
+
+        adapter = object()
+        fake_model = SimpleNamespace(create_runtime_adapter=lambda _name: adapter)
+        monkeypatch.setattr(smr, "load_stt_model", lambda _name: fake_model)
+        monkeypatch.setattr(smr, "get_model_download_path", lambda _m: "resolved-path")
+
+        runner = smr.STTModelRunner.__new__(smr.STTModelRunner)
```
LxYuan0420: Can we avoid bypassing `STTModelRunner.__init__` here? This test is covering the new runner boundary, but `__new__` can pass even if `load_model()` later depends on init-owned state. A real `STTModelRunner` with a minimal `vllm_config` would keep the test closer to production construction.

LxYuan0420 (approved):

PR #397, opened 2026-05-25
[MLA] Absorb kv_b_proj into embed_q/unembed_out for single-pass kernel

scyyh11:
The single-pass MLA Metal kernel requires an absorbed-MLA inner exposing per-head `embed_q` / `unembed_out`. mlx-lm's `glm4_moe_lite` ships this layout natively, but `deepseek_v2` and `minicpm3` keep a single `kv_b_proj = nn.Linear(kv_lora_rank, H * (qk_nope_head_dim + v_head_dim))`, so even when their dims sit inside the kernel envelope `_can_use_kernel` rejects them (`_is_absorbed=False`) and the wrapper falls back to the MLX SDPA `_apply_kv_b_proj_attention` slow path.

This PR adds an env-gated load-time slice that synthesizes per-head `embed_q` / `unembed_out` from `kv_b_proj.weight`, so kv_b_proj-style inners (DeepSeek-V2 / DeepSeek-V2-Lite) hit the single-pass kernel without any change to mlx-lm or the kernel itself.

## Mechanism

`kv_b_proj.weight` has shape `(H * (qk_nope_head_dim + v_head_dim), kv_lora_rank)`, laid out as `[k_nope_h || v_h]` rows per head. Per-head `embed_q` and `unembed_out` are output-axis slices of this tensor that absorb `W_k_nope_h` and `W_v_h` respectively. mlx's quantization groups live on the **input** axis (`kv_lora_rank`), so the slice preserves the quant scheme with no requant noise.

## Routing

Single env gate (`VLLM_METAL_MLA_ABSORB_KV_B_PROJ`, off by default), applied in `MLAPagedAttentionWrapper.__init__` right before the existing `is_absorbed` check. Idempotent: no-op when env is off, when the inner is already absorbed (GLM4MoeLite), or when there is no `kv_b_proj` to slice. Downstream `_can_use_kernel` admission gate is unchanged — `kv_lora_rank=512`, `qk_rope_head_dim=64`, decode-only, block_size ∈ {16, 32}, fp16/bf16 still have to hold.

## Token parity (DeepSeek-V2-Lite-Chat-4bit-mlx, 4 prompts × 16 tokens, temp=0)

Three configs, fresh `vllm.LLM` each: **A** baseline (`ABSORB=0 KERNEL=0`, slow path), **B** absorbed slow path (`ABSORB=1 KERNEL=0`, validates absorption math), **C** kernel via absorption (`ABSORB=1 KERNEL=1`).

| Prompt | B=A | C=A | A baseline text |
| --- | --- | --- | --- |
| `"The capital of France is"` | ✅ | ✅ | `" Paris.\nThe official language of France is French.\nThe currency of France"` |
| `"def quicksort(arr):"` | ✅ | ⚠️ diverge @ tok 12 | `"\nlenarr = len(arr)\nif lenarr <= 1:"` → kernel: `"...if lenarr < 2:"` |
| `"Once upon a time"` | ✅ | ✅ | `", in a land far, far away, there lived a young boy named Jack"` |
| `"One plus one equals"` | ✅ | ✅ | `" two.\nTwo plus two equals four.\nFour plus four equals eight."` |

**Absorbed math (B=A): 4/4 token-for-token** — slice is mathematically exact. **Kernel (C=A): 3/4 + 1 semantic equivalent** — `def quicksort(arr):` flips `<= 1` → `< 2`, both correctly express "fewer than 2 elements". The fp32 accumulator on the kernel path lands the argmax on a near-equal logit at a low-confidence position.

## Performance (M1 Pro 16 GB, max_model_len=512, enforce_eager=True)

In-process `vllm.LLM` timing, warmed per-shape, `ignore_eos=True`, `temperature=0`.

| B   | gen | A wall    | B wall    | C wall    | A→B   | A→C   | B→C   | A tput     | B tput     | C tput     |
| --- | ---:|----------:|----------:|----------:|------:|------:|------:|-----------:|-----------:|-----------:|
| 1   | 256 |   11.60 s |   10.11 s |   10.31 s | 1.15× | 1.12× | 0.98× |   22.1 t/s |   25.3 t/s |   24.8 t/s |
| 2   | 256 |   16.56 s |   13.67 s |   13.28 s | 1.21× | 1.25× | 1.03× |   30.9 t/s |   37.5 t/s |   38.6 t/s |
| 4   | 256 |   25.90 s |   19.23 s |   18.35 s | **1.35×** | **1.41×** | 1.05× |   39.5 t/s |   53.2 t/s |   55.8 t/s |

Mean: A→B **1.24×**, A→C **1.26×**, B→C 1.02×. The dominant win is **A→B** (eliminating per-token K/V materialization in `_apply_kv_b_proj_attention`, keeping attention in `kv_lora_rank` space) and it scales with batch (1.15× → 1.35× from B=1 to B=4). The incremental kernel win B→C is small on this hardware/model (0.98-1.05×); hardware with more compute headroom widens it — see M5 Max below.

### M5 Max @ M1 Pro profile (B∈{1,2,4} × gen=256, max_model_len=512)

| B   | gen | A wall    | B wall    | C wall    | A→B   | A→C   | B→C   | A tput     | B tput     | C tput     |
| --- | ---:|----------:|----------:|----------:|------:|------:|------:|-----------:|-----------:|-----------:|
| 1   | 256 |    4.01 s |    4.05 s |    3.82 s | 0.99× | 1.05× | 1.06× |   63.9 t/s |   63.3 t/s |   67.0 t/s |
| 2   | 256 |    5.30 s |    5.11 s |    4.66 s | 1.04× | 1.14× | 1.10× |   96.5 t/s |  100.2 t/s |  109.9 t/s |
| 4   | 256 |    6.83 s |    6.38 s |    5.72 s | 1.07× | **1.19×** | **1.12×** |  149.8 t/s |  160.4 t/s |  178.9 t/s |

Mean: A→B 1.03×, A→C 1.13×, B→C 1.09×. ~3× wall speedup over M1 Pro baseline; A→B compresses (faster baseline absorbs the slow path's per-step K/V cost) while B→C widens — kernel wins more headroom on this hardware.

### M5 Max @ M5 Max profile (B∈{1,2,4,8} × gen=1024, max_model_len=2048)

| B   | gen  | A wall    | B wall    | C wall    | A→B   | A→C       | B→C   | A tput     | B tput     | C tput     |
| --- | ----:|----------:|----------:|----------:|------:|----------:|------:|-----------:|-----------:|-----------:|
| 1   | 1024 |   19.59 s |   17.93 s |   16.86 s | 1.09× | 1.16×     | 1.06× |   52.3 t/s |   57.1 t/s |   60.7 t/s |
| 2   | 1024 |   27.19 s |   23.19 s |   21.22 s | 1.17× | 1.28×     | 1.09× |   75.3 t/s |   88.3 t/s |   96.5 t/s |
| 4   | 1024 |   43.28 s |   29.70 s |   26.39 s | 1.46× | 1.64×     | 1.13× |   94.6 t/s |  137.9 t/s |  155.2 t/s |
| 8   | 1024 |   85.07 s |   46.25 s |   38.97 s | **1.84×** | **2.18×** | **1.19×** |   96.3 t/s |  177.1 t/s |  210.2 t/s |

Mean: A→B 1.39×, A→C 1.57×, B→C 1.12×. Long sequences and larger batches make per-token K/V materialization accumulate, so A→B re-asserts (peaks at **1.84×** at B=8/gen=1024). Baseline throughput stalls at ~96 t/s past B=4 (slow path can't amortize batch growth); C reaches 210 t/s, **2.18×** the baseline ceiling. Token parity on M5 Max matches M1 Pro exactly (same `def quicksort` semantic-equivalence divergence).

### DIFF
--- a/tests/test_mla_paged_backend.py
+++ b/tests/test_mla_paged_backend.py
@@ -16,6 +16,7 @@ from vllm_metal.mlx_backend.mla_cache import MLAPagedLatentCache
 from vllm_metal.paged_attention_backend.mla import (
     MLAPagedAttentionBackend,
     MLAPagedAttentionWrapper,
+    _try_absorb_kv_b_proj_into_inner,
 )
 from vllm_metal.paged_attention_backend.protocol import PagedAttentionBackend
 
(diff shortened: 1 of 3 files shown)


LxYuan0420:
Thanks for the benchmarks. The math looks correct, and the speedup is worth pursuing.

I don’t think this should land in vllm-metal. The new runtime wrappers/helpers are implementing model-layout normalization inside the paged-attention backend: turning `kv_b_proj` into `embed_q` / `unembed_out`. That belongs in mlx-lm’s model/load path, where the model owns its projection layout. GLM4MoeLite already follows that pattern with load-time weight normalization.

If this transform is correct, it should be default behavior in mlx-lm for DeepSeek-V2 / MiniCPM3, not hidden behind another vllm-metal env var. Please move this upstream so vllm-metal can pick it up through the existing MLA path.

scyyh11:
Thanks for the review, I will see if this can be achieved in mlx-lm.

PR #399, opened 2026-05-26
chore: remove VLLM_METAL_PREFIX_CACHE legacy prefix cache

fxdv:
## Summary

Tier 1 of the removal scope agreed in #394 — drops the opt-in `PrefixCacheManager` on the contiguous (non-paged) KV cache path. Strictly limited to the prefix-cache surface; the non-paged MLX path itself and the contiguous merge/extract helpers stay, per the maintainer review.

## Why

#394 concluded that this plugin-local prefix cache should go rather than have its eviction policy locked in via docs or characterization tests:

- Upstream vLLM's block-manager already provides the prefix cache used on the paged path.
- The plugin-local manager kept a separate eviction policy that diverged from that surface (monotonic hit counter vs. live `ref_cnt`).
- The MLX contiguous path is now opt-out rather than primary, so maintaining a second policy here is no longer worthwhile.

## Scope (what changes)

In scope per @ricky-chaoju's review on #394:

- `vllm_metal/v1/contiguous_cache.py` — drop `PrefixCacheManager`, `CachedPrefix`, `_compute_prefix_hash`, `_compute_entry_bytes`, `_get_prefix_cache_max_bytes`, `_prefix_cache_enabled`, `_PREFIX_CACHE_ENABLED`, `_PREFIX_CACHE_DEFAULT_FRACTION`, and the imports they pulled in.
- `vllm_metal/v1/model_runner.py` — drop `self._prefix_cache` and the prefix-cache lookup/insert/restore block in `_prefill_single`. The non-paged MLX prefill path itself stays.
- `vllm_metal/envs.py` — drop `VLLM_METAL_PREFIX_CACHE` and `VLLM_METAL_PREFIX_CACHE_FRACTION`.
- `docs/configuration.md` — drop the two corresponding env-var rows.
- `tests/stub_runner.py` — drop the `_prefix_cache` default.
- `tests/test_vlm_forward_model.py` — drop `TestRestoreCacheRouting` (and the matching docstring fragment).
- `tests/test_prefix_cache.py` — drop the prefix-cache test classes (`TestPrefixCacheHybridGuard`, `TestPrefixCacheRestoreOffset`, `TestPrefixCacheEviction`, `TestPrefixCacheEnableFlag`, `TestPrefixCacheFractionParsing`). `TestHybridCacheMergeExtract` is preserved since it covers the merge/extract helpers retained on the contiguous path.

## Out of scope (intentionally untouched)

Per the maintainer review:

- The non-paged MLX path itself (`_prefill_single`).
- The contiguous cache merge/extract helpers (`_merge_kv_caches`, `_extract_kv_cache`, `_merge_arrays_caches`, `_extract_arrays_cache`, `_merge_rotating_kv_caches`, `_MIN_BATCH_SIZE_FOR_BATCHING`, `AnyCache`).
- The paged-path prefix cache surface and its tests (`tests/test_paged_prefix_caching*.py`, `docs/supported_models.md`).

## Notes for reviewers

- `tests/test_prefix_cache.py` retains only `TestHybridCacheMergeExtract` after this PR, which makes the filename slightly misleading. Happy to follow up with a rename to `tests/test_contiguous_cache.py` if you'd prefer; left it in place to keep this diff focused.
- No new behavior is introduced — the removed code path was purely opt-in via `VLLM_METAL_PREFIX_CACHE` (default off), so users who never set that env var see no functional change.

## Test plan

- [x] `ruff check vllm_metal tests scripts docs` — clean
- [x] `ruff format --check .` — clean
- [x] `python3 -m py_compile` on all changed files — clean
- [x] CI: `scripts/lint.sh` (ruff + mypy + shellcheck)
- [x] CI: `scripts/test.sh` (non-slow pytest suite + serving smoke tests)
- [ ] Locally on Apple Silicon: confirm `_prefill_single` path still works for a non-paged MLX run (e.g. `VLLM_METAL_USE_PAGED_ATTENTION=0` with a small text model)

Refs: #394


vllm_metal/v1/model_runner.py
```diff
@@ -685,36 +677,9 @@ def _prefill_single(
...
-                # Cache miss: process prefix first, cache it, then last token
-                prefix_ids = mx.array([prefix], dtype=mx.int32)
-                _ = self._forward_model(prefix_ids, cache=cache)
-                self._prefix_cache.insert(prefix, cache)
-                cached_prefix_len = len(prefix)
-
-        # Prefill: process remaining tokens (always at least the last token)
-        tokens_to_process = token_ids[cached_prefix_len:]
-        input_ids = mx.array([tokens_to_process], dtype=mx.int32)
+        cache: list[KVCache] = contiguous_cache.make_prompt_cache(self._forward_model)
```
LxYuan0420: _prefill_single() still calls `contiguous_cache.make_prompt_cache(...)`, but this PR removed that symbol; please restore/import it so `VLLM_METAL_USE_PAGED_ATTENTION=0` still works.

LxYuan0420 (commented):
Scope looks right; please fix the non-paged `make_prompt_cache` break and run one `VLLM_METAL_USE_PAGED_ATTENTION=0` smoke before merge.

fxdv:
Thanks @LxYuan0420 — good catch. `model_runner.py:_prefill_single` was still calling `contiguous_cache.make_prompt_cache(...)` after I dropped the `mlx_lm.models.cache.make_prompt_cache` re-export from `contiguous_cache.py` — that would have raised `AttributeError` on the non-paged path (`VLLM_METAL_USE_PAGED_ATTENTION=0`). The macOS test job didn't surface it because paged attention is on by default and `_prefill_single` isn't reached.

Fixed in `5a79068` (force-pushed):

- `vllm_metal/v1/model_runner.py` — import `make_prompt_cache` directly from `mlx_lm.models.cache`, drop the now-unused `from vllm_metal.v1 import contiguous_cache` module import, switch the call site to the bare name.
- `vllm_metal/v1/contiguous_cache.py` — unchanged from the prior push (no need to add the re-export back).

I'll run the `VLLM_METAL_USE_PAGED_ATTENTION=0` smoke locally on Apple Silicon and post the result here before the merge ask. If you'd rather see a small unit test pinning that `_prefill_single` builds a cache via `make_prompt_cache` (so this regression couldn't slip past the default-on paged config again), happy to add one in a follow-up.

fxdv:
Smoke ran clean on a MacBook Air M1 (16 GB) and MacStudio M3 Ultra (96 GB), `VLLM_METAL_USE_PAGED_ATTENTION=0`:

```
vllm serve Qwen/Qwen3-0.6B \
    --host 127.0.0.1 --port 8765 \
    --max-model-len 1024 --enforce-eager
```

Engine startup logs confirmed the non-paged path was actually taken:

```
Metal config: ... use_paged_attention=False ...
Metal: disabled chunked prefill (non-paged path), max_num_batched_tokens=2048
... Model warm-up complete
INFO:     Application startup complete.
```

Then a `POST /v1/completions` returned `200 OK`:

```json
{
  "model": "Qwen/Qwen3-0.6B",
  "choices": [{
    "text": " Paris. The capital of France is also the capital of the Republic of France. ...",
    "finish_reason": "length"
  }],
  "usage": {"prompt_tokens": 5, "completion_tokens": 20, "total_tokens": 25}
}
```

`_prefill_single` is the only prefill entry point on the non-paged path, so the request exercises it; first token resolves correctly, no `AttributeError` on `make_prompt_cache`, no other regressions in the path. Engine throughput logs:

```
Avg prompt throughput: 0.5 tokens/s, Avg generation throughput: 2.0 tokens/s
```

(Throughput is in the expected ballpark for non-paged + `--enforce-eager` on M1.)

I think this addresses both items in your review. Re-requesting review.

LxYuan0420 (approved):
LGTM

Issue #400, opened 2026-05-27
Lazy GDN state allocation for hybrid Qwen3.5 models


LxYuan0420:
Qwen3.5 hybrid models currently allocate GDN state for `max_num_seqs` at startup. On memory-constrained Apple Silicon, this can dominate startup memory even when serving only one request at a time.

Example from #398:

- `linear_cache_bytes_per_slot ~= 64.4MB`
- `max_num_seqs=256`
- GDN state reservation ~= 16.48GB

The short-term workaround is to lower `--max-num-seqs` for single-user serving. The longer-term memory optimization is to make the hybrid GDN state cache allocate slots lazily instead of reserving the full scheduler capacity at startup.

Suggested scope:

- Allocate GDN state only when a request is admitted.
- Grow up to `max_num_seqs` without changing the scheduler capacity contract.
- Release or reuse slots when requests finish.
- Keep scheduler capacity and backend state capacity aligned.
- Preserve pending-state and lazy-kernel behavior.
- Cover slot reuse, eviction, decode after prefill, mixed batches, and failure paths.

Out of scope:

- Silently lowering `max_num_seqs`.
- Changing paged KV block allocation policy.
- Changing Qwen3.5 model semantics.

This is a behavioral memory optimization and should be handled separately from the diagnostic/error-message fix for #398.

PR #401, opened 2026-05-27
[Bugfix] Improve hybrid GDN OOM diagnostics for paged attention

### LINKED ISSUE #398: Paged attention KV cache OOM on 32GB M1 Max with medium/large MLX models (even at 2K context)
### Summary

Models with >15GB weights fail to start on 32GB M1 Max due to paged attention KV cache pre-allocation, even with aggressive tuning (`VLLM_METAL_MEMORY_FRACTION=0.98`, `--max-model-len 2048`, `--enforce-eager`). Smaller models (9B, ~5GB weights) work fine.

### Environment

- **Hardware:** MacBook Pro M1 Max, 32 GB unified memory, 24-core GPU
- **macOS:** 26.4.1 (Build 25E253)
- **vllm-metal:** v0.2.0 (`vllm_metal-0.2.0-cp312-cp312-macosx_11_0_arm64.whl`)
- **vLLM core:** 0.21.0
- **Python:** 3.12.12 (uv-managed)
- **MLX:** 0.31.2

### Reproduction

```bash
# This WORKS (~5GB weights, plenty of KV headroom)
# Any 9B-class MLX 4-bit model, e.g. mlx-community/Qwen3.5-9B-MLX-4bit
vllm serve <path-to-9B-mlx-model> \
  --port 8081 --max-model-len 8192

# This FAILS (~20GB weights, KV budget goes negative)
# Any 27B+ MLX model, e.g. mlx-community/Qwen3.5-27B-MLX-4bit
VLLM_METAL_MEMORY_FRACTION=0.98 vllm serve \
  <path-to-27B+-mlx-model> \
  --port 8082 --max-model-len 2048 --enforce-eager
```

### Error

```
  metal_limit=26.80GB, fraction=0.98, usable_metal=26.26GB,
  model_memory=19.74GB, overhead=2.09GB, kv_budget=-12.05GB.
  Mitigations: increase VLLM_METAL_MEMORY_FRACTION, use a smaller or more quantized model.
```

### Test matrix

| Model | Weights | --max-model-len | FRACTION | --enforce-eager | Result |
|-------|--------:|----------------:|---------:|:---------------:|--------|
| Qwen3.5-9B MLX 4bit | ~5GB | 8192 | 0.90 (default) | No | **WORKS** |
| Qwen3.5-27B MLX 4bit | 16.06GB | 4096 | 0.95 | No | **OOM** (kv=-32.22GB) |
| Qwen3.5-35B-A3B MLX 3bit | 19.74GB | 8192 | 0.90 | No | **OOM** (kv=-13.09GB) |
| Qwen3.5-35B-A3B MLX 3bit | 19.74GB | 2048 | 0.90 | Yes | **OOM** (kv=-13.17GB) |
| Qwen3.5-35B-A3B MLX 3bit | 19.74GB | 2048 | 0.98 | Yes | **OOM** (kv=-12.05GB) |

Note: `--gpu-memory-utilization 0.7` was also tried but is silently ignored by the Metal path — only `VLLM_METAL_MEMORY_FRACTION` env var is respected.

### Analysis

The deficit is consistently 12-32GB regardless of `--max-model-len`. Reducing context from 8K to 2K and increasing fraction from 0.90 to 0.98 only recovered ~1GB of the 12GB deficit. This suggests the KV cache budget calculation may be over-provisioning for the single-user Apple Silicon use case — perhaps allocating for multiple concurrent sequences that a 32GB laptop will never serve.

For comparison, LM Studio loads the same 35B-A3B model at 8K context on the same hardware without issue, because llama.cpp/MLX use lazy KV allocation rather than up-front pool reservation.

### Question

Is there a way to reduce the paged attention pre-allocation for single-user serving on memory-constrained hardware? For example:
- `--max-num-seqs 1` to limit concurrent sequence slots
- A mode that lazily allocates KV pages rather than reserving the full pool at startup

32GB is the most common M1/M2/M3 Pro/Max configuration. If paged attention could be tuned to work with ~4GB KV headroom (models up to ~20GB weights), it would significantly expand the usable model range on this hardware.

### Environment notes

- Metal memory reports 34.4GB total but `metal_limit` caps at 26.80GB (OS reserved ~7.6GB)
- Model loads successfully in all cases (5-6s); failure is only at KV cache allocation
- `--max-parallel-loading-workers 1` is acknowledged but "currently not supported"

Happy to test patches or experimental flags if there's a direction to try.

ricky-chaoju:
Improve paged-attention OOM diagnostics for hybrid Qwen3.5/GDN models by reporting the hidden startup GDN state reservation separately from the remaining KV cache budget.

The error now includes:

  - `kv_budget_before_hybrid`
  - `hybrid_gdn_state`
  - per-sequence GDN state size
  - `max_num_seqs`
  - a targeted `--max-num-seqs 1` mitigation for single-user serving

This addresses the diagnostic gap from #398 without changing allocation behavior.

Fixes #398

### DIFF
--- a/tests/test_v1_worker.py
+++ b/tests/test_v1_worker.py
@@ -14,7 +14,10 @@ pytest.importorskip("vllm", reason="vllm not installed")
 from tests.stub_runner import make_stub_runner  # noqa: E402
 from vllm_metal.stt.policy import STT_SCHED_AVAILABLE_BYTES  # noqa: E402
 from vllm_metal.v1 import model_runner as mr  # noqa: E402
-from vllm_metal.v1.cache_policy import ModelCachePolicy  # noqa: E402
+from vllm_metal.v1.cache_policy import (  # noqa: E402
+    ModelCachePolicy,
+    WorkerCachePlanner,
+)
 from vllm_metal.v1.model_adapter import DefaultModelAdapter  # noqa: E402
 from vllm_metal.v1.worker import MetalWorker  # noqa: E402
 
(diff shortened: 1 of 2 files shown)


fxdv:
Picked up #398 a few hours after this landed (my "Taking this" note on the issue crossed in flight) — structure here is cleaner than what I was drafting, especially `_HybridGDNReservation` and surfacing `kv_budget_before_hybrid` separately. Standing down on a competing PR.

One small additive idea for `_paged_attention_mitigations`, non-blocking: when `reservation.enabled` and `kv_budget_before_hybrid > 0`, the *maximum feasible* `--max-num-seqs` is computable from the actual budget, so users who want some concurrency get a concrete number instead of always being pointed at `1`:

```python
# Conceptual — inside _paged_attention_mitigations
max_seqs_that_fit = (
    kv_budget_before_hybrid - per_block_bytes * PAGED_ATTENTION_MIN_BLOCKS
) // reservation.bytes_per_slot

if 1 <= max_seqs_that_fit < reservation.max_num_seqs:
    seq_mitigation = (
        f"lower --max-num-seqs to {int(max_seqs_that_fit)} or below "
        f"(currently {reservation.max_num_seqs})"
    )
else:
    seq_mitigation = (
        "lower --max-num-seqs (for single-user serving, try --max-num-seqs 1)"
    )
```

The current `kv_budget_before_hybrid > 0` ordering still holds — it just becomes the trigger for the computed-N branch, with the existing wording as the fallback when even N=1 doesn't fit.

Happy to send this as a small follow-up PR after this one lands if you think it's worth having; or feel free to fold it in here / close it as scope-creep.

LxYuan0420 (approved):
LGTM. The real memory optimization is #400: lazy GDN state allocation instead of reserving `max_num_seqs` slots at startup. Since you have the freshest context, would you be interested in picking that up next?

LxYuan0420:
One note: `cache_policy.py` is getting a bit messy after the recent cache-planning additions. I don’t want to block this bugfix on layout cleanup, so I’ll pick up a follow-up refactor to make the planner flow cleaner.

ricky-chaoju:
> LGTM. The real memory optimization is #400: lazy GDN state allocation instead of reserving `max_num_seqs` slots at startup. Since you have the freshest context, would you be interested in picking that up next?

Thanks, I’ll take #400 next .

PR #403, opened 2026-05-29
Add IBM Granite 3.3 to supported models

hoobnn:
## Summary

Adds IBM Granite 3.3 to the supported-models matrix. Granite reuses the existing GQA paged attention path with **no Metal-specific adapter**: its four scalar multipliers (embedding, residual, attention, logits) are applied inside mlx-lm's Granite model forward (`mlx_lm.models.granite`), and `attention_multiplier` reaches the Metal kernel via `inner.scale` (mlx-lm sets `Attention.scale = attention_multiplier`).

## Validation

Validated with `mlx-community/granite-3.3-8b-instruct-4bit` on Apple M4 Pro (48 GB), macOS. The paged attention path produces **byte-for-byte identical** greedy output to the independent `mlx_lm` reference across 6 prompts × 20 tokens, also cross-checked against vllm-metal's MLX inline-cache path (three-way identical).

## Tests

`tests/test_granite_golden.py`:
- `test_golden_data_wellformed` (runs in CI under `-m "not slow"`): data-integrity guard on the golden tables (length, token validity, the two tables stay in sync).
- `TestGranitePagedGolden` (`slow`, opt-in): loads the 8B checkpoint and asserts the paged path matches the `mlx_lm` golden; also asserts layer 0 `self_attn` is replaced by the paged wrapper.

All tests pass; `scripts/lint.sh` (ruff / ruff format / mypy / shellcheck) is clean.

Marked experimental (🔵) in the matrix, consistent with other newly-introduced architectures.


LxYuan0420 (approved):
Approving after trimming this back to a simple support-matrix update. Let’s keep this model support doc high-level for now, since we will likely rewrite the matrix structure in the future rather than growing it into a per-checkpoint verification log.

hoobnn:
Thanks for the review and for trimming it down — makes sense to me. Appreciate the quick turnaround!

PR #404, opened 2026-05-30
Bump vLLM to 0.22.0

ricky-chaoju:
Bump vLLM to 0.22.0

Also update the optional `vllm-rs` install/docs path to use the Rust frontend vendored in upstream vLLM 0.22.0 instead of the archived external repo.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -95,7 +95,7 @@ install_vllm_rs() {
   echo "Installing vllm-rs from vLLM source: $vllm_src_dir/rust"
 
   # Run cargo from the vLLM repository root so rustup honors rust-toolchain.toml.
-  if ! ( cd "$vllm_src_dir" && cargo install --path rust/src/cmd --bin vllm-rs ); then
+  if ! ( cd "$vllm_src_dir" && cargo install --locked --path rust/src/cmd --bin vllm-rs ); then
     error "Failed to install vllm-rs."
     exit 1
   fi
--- a/install.sh
+++ b/install.sh
@@ -78,6 +78,8 @@ download_and_install_wheel() {
 }
 
 install_vllm_rs() {
+  local vllm_src_dir="$1"
+
   section "Installing vllm-rs (experimental Rust frontend)"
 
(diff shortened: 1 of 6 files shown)


install.sh
```diff
@@ -78,28 +78,24 @@ download_and_install_wheel() {
...
     exit 1
   fi
 
-  # Run cargo from inside the tree so rustup honors rust-toolchain.toml;
-  # `cargo install --git` would ignore it and use the system default toolchain.
-  if ! ( cd "$tmp_dir/vllm-frontend-rs" && cargo install --path src/cmd --bin vllm-rs ); then
+  echo "Installing vllm-rs from vLLM source: $vllm_src_dir/rust"
+
+  # Run cargo from the vLLM repository root so rustup honors rust-toolchain.toml.
+  if ! ( cd "$vllm_src_dir" && cargo install --path rust/src/cmd --bin vllm-rs ); then
```
LxYuan0420: Can we double-check whether this should pass ? The vLLM 0.22.0 tarball ships , and my understanding is that  ignores it unless  is set, which could let the installer resolve crate versions beyond the tested release.

LxYuan0420: Can we double-check whether this should pass `--locked`? The vLLM 0.22.0 tarball ships `rust/Cargo.lock`, and my understanding is that `cargo install --path` ignores it unless `--locked` is set, which could let the installer resolve crate versions beyond the tested release.

LxYuan0420:
Can you add one short manual smoke result before merge: `vllm serve Qwen/Qwen3-0.6B` plus one `/v1/completions` request on this bumped dependency stack? No need to block on the optional Rust frontend unless you already tested it.

LxYuan0420 (approved):
Approving with the inline `--locked` question and manual smoke request noted.

ricky-chaoju:
updated to use `cargo install --locked`.

Manual smoke passed with `vllm serve Qwen/Qwen3-0.6B` and one `/v1/completions` request returning 200. I also verified `./install.sh --with-vllm-rs` and a Rust frontend smoke separately.

PR #408, opened 2026-05-31
[Multimodal] Enable Qwen3-VL forward path

scyyh11:
## Summary

Flip `Qwen3VLMultimodalAdapter.forward_ready` to `True` so the Qwen3-VL multimodal path goes live: scheduled encoder inputs route through `_run_vision_encoders` and the paged mm forward runs end-to-end via `adapter.call_lm`, instead of failing fast at the encoder-dispatch gate.

This is **follow-up 3** from #350 (paged-core) — the user-visible gate flip plus the end-to-end parity test. The dependency blocker is cleared: #404 bumped vLLM to 0.22.0, which carries `llguidance >=1.7` (vllm [#42150](https://github.com/vllm-project/vllm/pull/42150)) and pins `mlx-vlm >=0.5.0`.

## What changes

- **`forward_ready: False → True`** on the adapter. 
- **`_reject_scheduled_encoder_inputs`: `NotImplementedError` → `RuntimeError`** describing the real gate condition (adapter missing / not `forward_ready`), now that the path is wired rather than "not connected yet".
- **Non-paged fail-fast**: when `VLLM_METAL_USE_PAGED_ATTENTION=0`, the gate now raises `NotImplementedError` for mm requests instead of running the encoder and falling through to the legacy `_prefill_single` path (which would feed raw image placeholder tokens to the LM without splicing image embeddings). Only the paged `_run_mm_paged_forward` splices.
- Drop the now-stale "encoder not connected" warning in `model_lifecycle`.
- Add **`tests/multimodal/test_qwen3_vl_parity.py`** (`slow`): end-to-end logits parity of `adapter.call_lm` vs `mlx_vlm.Model.__call__`; module-skips unless the model is in the HF cache.
- Sync encoder-dispatch gate tests (forward_ready=False → `RuntimeError`; dispatch now requires the paged backend) and add a non-paged mm fail-fast test.

## Scope

- **Paged-only** (RFC #319 hard rule 4). Non-paged prefill + decode is deferred; mm requests on the non-paged path now fail fast with a clear error.
- The #350 spec-decode reject in `_run_mm_paged_forward` **stays in place**. Correct per-query-token `segment_positions` is a separate change; mm + spec-decode is not triggerable today (the only spec-decode method, Gemma4 MTP, is text-only — no Qwen3-VL draft exists), so the reject is inert on the normal forward path.

## Validation

Re-validated this session under **vLLM 0.22.0 / mlx-vlm 0.5.0**:

**Logits parity** (`test_qwen3_vl_parity.py` — next-token argmax matches reference, `abs_max(diff) < 5e-2`):

| checkpoint | result |
| --- | --- |
| `Qwen3-VL-4B-Instruct-4bit` | ✅ pass |
| `Qwen3-VL-4B-Instruct-8bit` | ✅ pass |

**End-to-end greedy alignment vs `mlx_vlm.generate`** (`temperature=0`; re-run this session under vLLM 0.22.0 / mlx-vlm 0.5.0 — vllm-metal `LLM.generate` vs `mlx_vlm.stream_generate` on the same deterministic 128×128 image):

- **8-bit: 32/32 generated tokens match exactly** (vllm-metal capped at `max_tokens=32`; the mlx-vlm reference is token-identical over the shared length). The entire pipeline (vision encode → splice → M-RoPE → paged attention → call_lm → sampler) is numerically equivalent to `mlx_vlm.generate` up to bf16/quant noise.
- 4-bit: 8/33 prefix match, diverges at pos 8 — 4-bit quant logit precision flips a close-call argmax; the two continuations are the *same* content reordered ("abstract, pixelated, noisy" vs "pixelated, abstract, chaotic"), **not** a pipeline correctness issue.

Bisection confirmed M-RoPE positions are bit-exact vs `LanguageModel.get_rope_index`, per-feature vision output matches the batched reference, and the splice matches `Model.merge_input_ids_with_image_features`.

**Throughput** (`vllm bench serve --dataset-name random-mm`, re-run this session under vLLM 0.22.0 on the Metal server, `--ignore-eos`):

| run | total tok/s | output tok/s | TTFT med | TPOT med | ITL med |
| MM 32p, conc 4, 256×256, 128/128 | 401.6 | 155.8 | 292 ms | 19.6 ms | 19.5 ms |
| Text 32p, conc 4, 128/128 | 380.1 | 184.3 | 68 ms | 19.6 ms | 19.6 ms |
| MM 50p, conc 8, 384×384, 256/128 | 669.6 | 159.3 | 875 ms | 39.3 ms | 38.8 ms |

- Decode is **identical between MM and text** at conc 4 — TPOT 19.6 ms / ITL ~19.5 ms either way. The vision tower runs once in prefill; no per-token multimodal cost.
- MM prefill overhead is **+224 ms** median TTFT for one 256×256 image (292 vs 68 ms) at equal prompt length.
- At conc 8 / 384×384, median TTFT is 875 ms (TPOT ~39 ms under 8-way batched decode). Per-feature vision encoding is still serialized — batching scheduled encoder inputs into a single tower call is a documented `_run_vision_encoders` TODO (`adapter.encode_multimodal([all_features])` with `grid_thw.prod()`-based per-feature output slicing).

### DIFF
--- a/tests/multimodal/test_qwen3_vl_parity.py
+++ b/tests/multimodal/test_qwen3_vl_parity.py
@@ -1,0 +1,189 @@
+# SPDX-License-Identifier: Apache-2.0
+"""End-to-end parity: ``adapter.call_lm`` vs ``mlx_vlm.Model.__call__``.
+
+Marked ``slow`` — opt in with ``pytest -m slow`` (matches the existing
+real-model convention in ``tests/test_qwen35_smoke.py``).  Skips when
+the model is not pre-pulled into the HF cache; pre-pull locally with::
+
+    hf download mlx-community/Qwen3-VL-4B-Instruct-4bit
+
+Override via ``QWEN3_VL_PARITY_MODEL`` env var.
+"""
+
+from __future__ import annotations
+
+import os
+from typing import Any
+
+import mlx.core as mx
+import numpy as np
+import pytest
+import torch
+from PIL import Image
+from vllm.multimodal.inputs import MultiModalFieldConfig, MultiModalKwargsItem
+
+from vllm_metal.multimodal import (
+    MultiModalFeatureSpec,
+    PlaceholderRange,
+    merge_multimodal_embeddings,
+)
+from vllm_metal.multimodal.qwen3_vl import Qwen3VLMultimodalAdapter
+
+MODEL_ID = os.environ.get(
+    "QWEN3_VL_PARITY_MODEL", "mlx-community/Qwen3-VL-4B-Instruct-4bit"
+)
+
+
+def _model_in_cache(model_id: str) -> bool:
+    from huggingface_hub import scan_cache_dir
+    from huggingface_hub.errors import CacheNotFound
+
+    try:
+        info = scan_cache_dir()
+    except CacheNotFound:
+        return False
+    for repo in info.repos:
+        if repo.repo_id != model_id:
+            continue
+        for rev in repo.revisions:
+            if rev.size_on_disk > 100 * 1024 * 1024:
+                return True
+    return False
+
+
+if not _model_in_cache(MODEL_ID):
+    pytest.skip(
+        f"{MODEL_ID} not in HF cache; pre-pull with `hf download {MODEL_ID}`",
+        allow_module_level=True,
+    )
+
+
+@pytest.fixture(scope="module")
+def loaded():
+    from mlx_vlm import load
+
+    model, processor = load(MODEL_ID)
+    return model, processor
+
+
+def _build_image_inputs(model, processor):
+    """Encode 'describe the image' + one deterministic dummy image."""
+    from mlx_vlm.prompt_utils import apply_chat_template
+
+    rng = np.random.default_rng(0)
+    image = Image.fromarray((rng.random((128, 128, 3)) * 255).astype(np.uint8))
+    prompt = apply_chat_template(
+        processor, model.config, "describe the image", num_images=1
+    )
+    proc = processor(text=[prompt], images=[image], return_tensors="np")
+    input_ids = mx.array(np.asarray(proc["input_ids"]))
+    pixel_values = mx.array(np.asarray(proc["pixel_values"]))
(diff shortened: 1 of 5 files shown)


ricky-chaoju (commented):
I reran a comparable 50p / conc=8 / 384x384 / 256/128 random-mm benchmark locally and got P50 TTFT 2873 ms, much closer to the 3.5s TODO than the 875 ms table entry. Could you clarify the exact benchmark context for the 875 ms number, or update the PR body/table?

scyyh11:
@ricky-chaoju Thanks for the cross-check. I re-ran this cold but I can't reproduce 2873 ms.

**Exact reproduction**, vLLM 0.22.0 / mlx-vlm 0.5.0, `Qwen3-VL-4B-Instruct-4bit`:

```
# env: VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_USE_PAGED_ATTENTION=1 VLLM_METAL_MEMORY_FRACTION=auto
vllm serve mlx-community/Qwen3-VL-4B-Instruct-4bit --port 8000 \
  --max-model-len 4096 --limit-mm-per-prompt '{"image": 1}'

vllm bench serve --backend openai-chat --endpoint /v1/chat/completions \
  --model mlx-community/Qwen3-VL-4B-Instruct-4bit \
  --dataset-name random-mm --num-prompts 50 --max-concurrency 8 --request-rate inf \
  --random-input-len 256 --random-output-len 128 \
  --random-mm-bucket-config '{(384, 384, 1): 1.0}' --ignore-eos
```

→ **TTFT median 654 ms, P99 2638 ms**, TPOT 39.9 ms, 50/50 OK — same band as the 875 ms already in the table (run-to-run variance).

What device are you running the model on? I'm using M5 Max so I suspect it is device variance.

ricky-chaoju:
I reran the same benchmark shape locally on an Apple M2 Ultra with 192GB unified memory, so this is probably device variance.

ricky-chaoju (approved):

PR #409, opened 2026-06-01
[Bugfix] Lazy allocate hybrid GDN state slots

### LINKED ISSUE #400: Lazy GDN state allocation for hybrid Qwen3.5 models

ricky-chaoju:
Hybrid Qwen3.5 models eagerly allocate GDN recurrent state for `max_num_seqs` during paged-attention startup. On memory-constrained Apple Silicon, this can consume several GB before serving any request.

`HybridPagedAttentionBackend.initialize()` created `GDNPagedStateCache` with full `max_num_seqs` rows, and the paged cache planner treated that full GDN state as startup-reserved memory.
Fixes #400

### DIFF
--- a/vllm_metal/v1/cache_policy.py
+++ b/vllm_metal/v1/cache_policy.py
@@ -45,6 +45,8 @@ if TYPE_CHECKING:
 
 logger = init_logger(__name__)
 
+HYBRID_GDN_GROWTH_CUSHION_SLOTS = 2
+
 
 @dataclass(frozen=True, kw_only=True)
 class TurboQuantAttentionSpec(FullAttentionSpec):
@@ -757,13 +759,14 @@ class WorkerCachePlanner:
         max_num_seqs = runner.scheduler_config.max_num_seqs
         if max_num_seqs <= 0:
             return _HybridGDNReservation()
+        cushion_slots = min(max_num_seqs, HYBRID_GDN_GROWTH_CUSHION_SLOTS)
         return _HybridGDNReservation(
             bytes_per_slot=runner.linear_cache_bytes_per_slot(),
             # ``ensure_capacity`` grows by allocating a larger state pool and
-            # copying the old pool into it.  The worst exact-growth step is
-            # (max_num_seqs - 1) -> max_num_seqs, whose transient peak keeps
-            # both old and new GDN pools live.
-            reserved_slots=(2 * max_num_seqs) - 1,
+            # copying the old pool into it. Reserve a bounded growth cushion
+            # instead of the full scheduler cap so large max_num_seqs values
+            # still benefit from lazy GDN allocation.
+            reserved_slots=(2 * cushion_slots) - 1,
             max_num_seqs=max_num_seqs,
         )
 
--- a/vllm_metal/v1/cache_policy.py
+++ b/vllm_metal/v1/cache_policy.py
@@ -235,11 +235,7 @@ class _PagedAttentionPlan:
             "use a smaller or more quantized model",
         ]
         reservation = self.hybrid_gdn_reservation
-        if (
-            reservation.enabled
-            and reservation.reserved_slots == reservation.max_num_seqs
-            and reservation.max_num_seqs > 1
-        ):
+        if reservation.enabled and reservation.max_num_seqs > 1:
             seq_mitigation = (
                 "lower --max-num-seqs (for single-user serving, try --max-num-seqs 1)"
             )
@@ -253,18 +249,11 @@ class _PagedAttentionPlan:
         reservation = self.hybrid_gdn_reservation
         if not reservation.is_hybrid:
             return ""
-        if reservation.enabled:
-            return (
-                f"hybrid_gdn_state=lazy "
-                f"(first_slot_reserve={reservation.total_bytes / 1e9:.2f}GB, "
-                f"{reservation.bytes_per_slot / 1e6:.1f}MB/seq * "
-                f"reserved_slots={reservation.reserved_slots}/"
-                f"max_num_seqs={reservation.max_num_seqs})"
-            )
         return (
             "hybrid_gdn_state=lazy "
-            f"(first_slot_reserve=0.00GB, "
-            f"{reservation.bytes_per_slot / 1e6:.1f}MB/seq, "
+            f"(growth_peak_reserve={reservation.total_bytes / 1e9:.2f}GB, "
+            f"{reservation.bytes_per_slot / 1e6:.1f}MB/seq * "
+            f"peak_slots={reservation.reserved_slots}/"
             f"max_num_seqs={reservation.max_num_seqs})"
         )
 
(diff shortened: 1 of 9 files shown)


vllm_metal/v1/cache_policy.py
```diff
@@ -742,13 +761,15 @@ def _validate_paged_attention_plan(self, plan: _PagedAttentionPlan) -> None:
...
+        """Return lazy GDN headroom reserved outside the paged KV pool."""
         runner = self._worker.model_runner
         if not runner.is_hybrid:
             return _HybridGDNReservation()
+        max_num_seqs = runner.scheduler_config.max_num_seqs
         return _HybridGDNReservation(
             bytes_per_slot=runner.linear_cache_bytes_per_slot(),
-            max_num_seqs=runner.scheduler_config.max_num_seqs,
+            reserved_slots=1 if max_num_seqs > 0 else 0,
+            max_num_seqs=max_num_seqs,
```
LxYuan0420, line 770: Can the planner reserve the transient GDN growth peak, not just one steady-state slot? Runtime can grow past the first slot, and `ensure_capacity()` keeps old arrays live while materializing the larger arrays, so tight configs can pass startup but OOM on slot 2+. The budget should match that growth peak, with a constrained multi-slot growth test.

vllm_metal/v1/model_runner.py
```diff
@@ -931,28 +931,28 @@ def _start_paged_forward(
...
+            # ---- GDN slot mapping (hybrid models) ----
+            if self.is_hybrid:
+                ctx = get_context()
+                if ctx is not None:
+                    gdn_slots = []
+                    # Decode requests come first, then prefill
+                    for req_id, _ in decode_reqs:
+                        gdn_slots.append(self._gdn_alloc_slot(req_id))
+                    for pr in prefill_reqs:
+                        gdn_slots.append(self._gdn_alloc_slot(pr.req_id))
```
LxYuan0420: Can this assign GDN slots once per step instead of request-by-request? The current loop can grow/copy/eval once per new request and can leave partial `_gdn_req_to_slot` state if a later request fails. A single step-level assignment would make admission atomic and match the planner’s peak-memory accounting.

vllm_metal/v1/cache_policy.py
```diff
@@ -241,11 +251,20 @@ def format_mitigations(self) -> str:
...
+                f"max_num_seqs={reservation.max_num_seqs})"
+            )
         return (
-            f"hybrid_gdn_state={reservation.total_bytes / 1e9:.2f}GB "
-            f"({reservation.bytes_per_slot / 1e6:.1f}MB/seq * "
+            "hybrid_gdn_state=lazy "
+            f"(first_slot_reserve=0.00GB, "
+            f"{reservation.bytes_per_slot / 1e6:.1f}MB/seq, "
             f"max_num_seqs={reservation.max_num_seqs})"
         )
```
LxYuan0420, line 260: This branch looks stale now that hybrid planning reserves one slot whenever `max_num_seqs > 0`. `reservation.enabled` should already be true whenever `reservation.is_hybrid` is true, so can we collapse this formatting path?

LxYuan0420:
Thanks for tackling #400. The lazy-allocation direction is right, but I think the planner/runtime memory contract needs tightening before merge.

The planner reserves one GDN slot, while runtime can grow additional slots through a realloc/copy path whose peak is old + new state. Please make the planner budget match that growth peak, and make GDN slot assignment step-level/atomic instead of per-request mutation. There is also a small stale formatting branch in the cache-policy path.

LxYuan0420 (commented):
Can we also share the before/after result later?

ricky-chaoju:
Before/after with the same constrained setup:

Model: `Qwen3.5-0.8B`
`VLLM_METAL_MEMORY_FRACTION=0.03`
`max_num_seqs=2`
prompts: `["Hello", "Good morning"]`
`max_tokens=1`, greedy decoding

Before this PR:
`hybrid_gdn_state=0.04GB (19.5MB/seq * max_num_seqs=2)`
`kv_budget=2.01GB`
`num_blocks=300`
`Hybrid cache initialized ... 18 linear layers (2 slots)`
`GENERATED_BEFORE [',', '!']`

After this PR:
`hybrid_gdn_state=lazy (growth_peak_reserve=0.06GB, 19.5MB/seq * peak_slots=3/max_num_seqs=2)`
`kv_budget=1.98GB`
`num_blocks=296`
`Hybrid cache initialized ... 18 linear layers (0/2 GDN slots allocated)`
`GENERATED_AFTER [',', '!']`

 The same two-request run completes, but GDN state is no longer allocated at startup. The planner reserves the lazy growth peak instead of allocating the full state pool up front.

vllm_metal/v1/cache_policy.py
```diff
@@ -742,13 +750,21 @@ def _validate_paged_attention_plan(self, plan: _PagedAttentionPlan) -> None:
...
+        if max_num_seqs <= 0:
+            return _HybridGDNReservation()
         return _HybridGDNReservation(
             bytes_per_slot=runner.linear_cache_bytes_per_slot(),
-            max_num_seqs=runner.scheduler_config.max_num_seqs,
+            # ``ensure_capacity`` grows by allocating a larger state pool and
+            # copying the old pool into it.  The worst exact-growth step is
+            # (max_num_seqs - 1) -> max_num_seqs, whose transient peak keeps
+            # both old and new GDN pools live.
+            reserved_slots=(2 * max_num_seqs) - 1,
```
LxYuan0420: Can we revisit this reservation formula? The shared before/after result is helpful for `max_num_seqs=2`, but this scales as `2 * max_num_seqs - 1`; for the #400-style case (`max_num_seqs=256`, ~64.4MB/slot), that reserves ~32.9GB, which is more conservative than the old eager 256-slot reservation. That seems to defeat the main goal of lazy GDN allocation for large scheduler capacities. I think the planner should reserve a bounded startup/growth cushion, not the worst exact-growth peak across the full scheduler cap.

ricky-chaoju: Updated. This now reserves a bounded lazy-growth cushion instead of the worst peak across the full scheduler cap.

  The planner reserves `2 * min(max_num_seqs, 2) - 1` slots, so max_num_seqs=256 reserves 3 peak slots, not 511.

  Fresh before/after with Qwen3.5-0.8B, max_num_seqs=256, VLLM_METAL_MEMORY_FRACTION=0.05:

  Before:
  - hybrid_gdn_state=5.00GB (19.5MB/seq * max_num_seqs=256)
  - kv_budget=-0.64GB
  - startup failed

  After:
  - hybrid_gdn_state=lazy (growth_peak_reserve=0.06GB, 19.5MB/seq * peak_slots=3/max_num_seqs=256)
  - kv_budget=4.30GB
  - Hybrid cache initialized with 0/256 GDN slots allocated
  - generation completed: GENERATED_AFTER_256_F005 [',', '!']

LxYuan0420 (approved):
LGTM

PR #417, opened 2026-06-04
[Distributed] Support vLLM's Ray executor on Metal (single-node, TP=1)

WindChimeRan:
# Summary

Makes `MetalPlatform` a first-class citizen of vLLM's Ray distributed executor by advertising the Apple GPU as a custom Ray resource, instead of aborting at cluster init. No vLLM-core edits — everything is plugin-side. Validated single-node, TP=1, on the default Ray **V2** executor, with correct generation through Ray.

## Before / After

**Before** — any Ray-backed run aborts at engine-core init:

```console
$ vllm serve Qwen/Qwen3-0.6B --distributed-executor-backend ray --tensor-parallel-size 1
...
ValueError: current platform cpu does not support ray.
RuntimeError: Engine core initialization failed. ... Failed core proc(s): {'EngineCore': 1}
```

**After** — boots and serves (each node first advertises the `mlx` resource):

```console
$ IP=$(python -c "from vllm.utils.network_utils import get_ip; print(get_ip())")
$ ray start --head --node-ip-address="$IP" --resources='{"mlx": 1}'   # IP must match vLLM's get_ip()
...
Ray runtime started.

$ RAY_ADDRESS=auto vllm serve Qwen/Qwen3-0.6B --distributed-executor-backend ray --tensor-parallel-size 1
...
(RayWorkerProc) INFO [compat.py] vllm_metal: patched Ray V2 worker get_node_and_gpu_ids on RayWorkerProc
INFO:     Application startup complete.

$ curl -s localhost:8000/v1/completions -H 'Content-Type: application/json' \
    -d '{"model":"Qwen/Qwen3-0.6B","prompt":"The capital of France is","max_tokens":16,"temperature":0}'
{"choices":[{"text":" Paris. The capital of Italy is Rome. The capital of Spain is Madrid.", ...}]}
```

## What changed

- `MetalPlatform.ray_device_key = "mlx"` (+ `device_control_env_var = "VLLM_METAL_VISIBLE_DEVICES"`) → vLLM's Ray executor takes the generic `resources={mlx: n}` placement path instead of the CUDA `num_gpus` path.
- Override `RayWorkerProc.get_node_and_gpu_ids` (the Ray **V2** executor's worker actor; in `compat.py`, installed via a `worker_process_setup_hook`) to read the *assigned* `mlx` resource — Ray's `get_accelerator_ids()` never lists a custom resource, so the stock method `KeyError`s.
- `set_device` accepts a `torch.device` — vLLM's base `Platform.set_device` contract passes one; the old int-only signature was the latent mismatch.
- Register `VLLM_METAL_VISIBLE_DEVICES` in `envs.py` so `validate_environ()` does not warn.

## Scope / limitations

- **Ray V2 executor only** (the default, `VLLM_USE_RAY_V2_EXECUTOR_BACKEND=1`). The legacy `RayDistributedExecutor` (V1) is not supported.
- **Pipeline parallelism** is rejected at config time (no MLX cross-stage hand-off yet).

---

cc @IntuitIntelLLC


LxYuan0420 (commented):
Thanks for pushing this forward. I see this mainly as laying groundwork for Ray / future distributed support rather than something that improves the normal single-Mac serving path today.

For the currently validated scope, single-node `TP=1`, I don’t see Ray as practically better than `uni`; So I think we should be clear whether we want to support this direction longer-term. If yes, I’m supportive, but we should keep the scope tight and avoid accumulating more temporary compat patching without tests.

Before merge, can we add coverage for the Ray patch itself and fail fast for `tensor_parallel_size > 1`, since the PR only validates TP=1?

WindChimeRan:
- yeah we need to support multi-node via ray. TP / PP, maybe with RDMA #1 . Otherwise we'll block large models away. 
- guard tp==1, add more tests
- No end2end CI test yet. Will defer to real multi-node PRs.

LxYuan0420 (approved):

Issue #343, opened 2026-05-06
[Feature] Native MLX sampling to eliminate mlx_to_torch round-trip


EagerofLight:
## Motivation

Non-greedy sampling bridges MLX logits to PyTorch via `mlx_to_torch` on every decode step. This is a Metal-specific cost (on CUDA logits are already PyTorch tensors). The overhead comes from `mx.eval()` sync + memoryview + MPS device transfer. Noted as tech debt in #308. Same direction as abandoned PR #86.

### Scope

- **In**: temperature, top-k, top-p, min-p, penalties, categorical sampling, per-request seeded RNG. Paged path only.
- **Out**: non-paged path (#145 deprecated)

### Design

Extend the existing greedy fast path (`mx.argmax` in `sampling_batch.py:228`) to non-greedy. Whitelist routing: supported params → MLX path, everything else → PyTorch fallback automatically.

Algorithm: temperature → min-p → top-k (`mx.argpartition` O(n)) → top-p (pre-filter k=8192 + argsort subset) → `mx.random.categorical`

### Phases

| Phase | Scope |
|-------|-------|
| **PR1** | temperature + top-k + top-p + min-p + categorical. New `mlx_sampling.py`, whitelist fast path in `sample_from_logits()` |
| **PR2** | penalties (repetition/frequency/presence). Self-built to match vLLM full-history semantics |
| **PR3** | per-request seeded RNG (`mx.random.key`) |

### Improvements over PR #86

- Single entry point (`sample_from_logits`) not 3 places in model_runner
- Whitelist safety (unknown params fall back)

## Mini Benchmark for pre-filtering in top_p

Isolated `sample_from_logits()` call latency — measures the per-step sampling overhead only (not model forward pass). M4 Mac mini (16GB), 1000 iterations, temperature=0.8, top_p=0.9.

**Batch scaling (Qwen2.5-7B-Instruct-4bit, vocab=152,064):**

| batch | PyTorch bridge | MLX (full argsort) | MLX + pre-filter |
|:-----:|:-:|:-:|:-:|
| 1 | 2.42 ms | 1.04 ms (2.3x) | 0.80 ms (3.0x) |
| 4 | 3.94 ms | 3.46 ms (1.1x) | 1.94 ms (2.0x) |
| 8 | 6.71 ms | 7.91 ms (0.85x) | 4.03 ms (1.7x) |
| 16 | 13.50 ms | 17.48 ms (0.77x) | 8.98 ms (1.5x) |

[...]

PR #419, opened 2026-06-04
Add opt-in native MLX random sampling for no-logprobs decode

CAICAIIs:
Refs #343

## Summary

This PR adds a first, opt-in native MLX random sampling path for paged decode requests that do not request logprobs.

Supported in this first slice:

- temperature
- top_k
- top_p
- min_p
- categorical sampling

The existing greedy `mx.argmax` fast path is unchanged. `sample_from_logits()` remains the single sampling entry point.

## Scope

Native MLX random sampling is enabled only when `VLLM_METAL_NATIVE_RANDOM_SAMPLING=1`.

The new path conservatively falls back to the existing vLLM Torch sampler for:

- logprobs
- seeded generators / `SamplingParams.seed`
- frequency / presence / repetition penalties
- `allowed_token_ids`
- bad words
- active `min_tokens`
- active `logit_bias`
- unknown or unsupported logits processors
- prefill sampling

The env flag defaults to off because the current end-to-end benchmark does not show a stable throughput improvement.

## Correctness Notes

The MLX path applies filters in the same order as the existing sampler path:

1. temperature
2. min_p
3. top_k / top_p
4. categorical sampling

The top-k subset fast path preserves vLLM's boundary-tie semantics: if tokens are tied at the kth threshold, it falls back to the full filtering path so tied candidates are not dropped.

## Validation

Commands were run from the repository root with the project virtualenv.

```bash
.venv-vllm-metal/bin/python -m pytest tests/test_sampling.py -q
# 51 passed, 16 warnings

.venv-vllm-metal/bin/python -m pytest tests/test_sampling.py tests/test_v1_model_runner_generate.py -q
# 86 passed, 16 warnings

.venv-vllm-metal/bin/python -m pytest tests/test_paged_attention.py tests/test_grammar_bitmask.py -q
# 38 passed, 16 warnings

.venv-vllm-metal/bin/python -m pytest tests/test_register.py tests/test_config.py -q
# 21 passed, 2 warnings

.venv-vllm-metal/bin/python -m ruff check tests/test_sampling.py vllm_metal/v1/sampling_batch.py vllm_metal/envs.py
# All checks passed!

.venv-vllm-metal/bin/python -m ruff format --check tests/test_sampling.py vllm_metal/v1/sampling_batch.py vllm_metal/envs.py
# 3 files already formatted
```

## End-to-End Benchmark

This is not claimed as a performance win.

Benchmark setup:

- model: `Qwen/Qwen3-0.6B`
- revision: `c1899de289a04d12100db370d81485cdf75e47ca`
- path: paged attention
- env: `VLLM_METAL_USE_PAGED_ATTENTION=1`
- branch native env: `VLLM_METAL_NATIVE_RANDOM_SAMPLING=1`
- workload: random input 32 / output 128, 16 prompts, max concurrency 4
- sampling: `temperature=0.8`, `top_p=0.9`, `top_k=50`, `min_p=0.05`
- hardware: Apple M5, 16GB, macOS 26.4

3-run mean branch vs main:

| Metric | Result |
| --- | ---: |
| output throughput | -8.80% |
| total token throughput | -8.80% |
| mean TPOT | +16.87% |
| mean ITL | +34.95% |
| mean TTFT | +38.07% |

Given these results, the new path stays experimental and opt-in.

## Follow-up Work

Before this can be advertised as a performance improvement, the next steps are:

- profile the native path against the Torch bridge to identify whether the current cost comes from full-vocab sorting, extra MLX ops, or categorical sampling itself
- revisit the top-p pre-filter idea only if the candidate-set semantics can be kept exact
- evaluate larger-vocab and higher-concurrency decode-heavy workloads where sampler overhead is more visible
- add penalties and seeded RNG in separate PRs after their vLLM semantics are matched
- consider enabling the path by default only after a main-vs-branch end-to-end benchmark shows a stable win

### DIFF
--- a/tests/test_sampling.py
+++ b/tests/test_sampling.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for Metal sampling with vLLM Sampler integration."""
 
+from types import SimpleNamespace
+
 import mlx.core as mx
 import numpy as np
 import pytest
@@ -11,8 +13,9 @@ from vllm.utils.torch_utils import make_tensor_with_pad
 from vllm.v1.engine import EngineCoreOutput, EngineCoreRequest, FinishReason
 from vllm.v1.engine.output_processor import OutputProcessor
 from vllm.v1.outputs import LogprobsLists
-from vllm.v1.sample.logits_processor import LogitsProcessors
+from vllm.v1.sample.logits_processor import LogitsProcessors, build_logitsprocs
 from vllm.v1.sample.metadata import SamplingMetadata
+from vllm.v1.sample.ops.topk_topp_sampler import apply_top_k_top_p_pytorch
 from vllm.v1.sample.sampler import Sampler
 
 from tests.stub_runner import make_stub_runner
@@ -23,7 +26,15 @@ from vllm_metal.v1.model_runner import (
     _create_request_generator,
     _ExecutionBatch,
 )
-from vllm_metal.v1.sampling_batch import SamplingBatch, sample_from_logits
+from vllm_metal.v1.sampling_batch import (
+    SamplingBatch,
+    _mlx_apply_min_p,
+    _mlx_apply_top_k_top_p,
+    _mlx_top_k_subset_sample,
+    sample_decode_tokens,
+    sample_from_logits,
+    sample_prefill_tokens,
+)
 
 VOCAB_SIZE = 1024
 MAX_NUM_PROMPT_TOKENS = 64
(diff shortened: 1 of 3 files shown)


WindChimeRan (commented):
I run a sampling-only benchmark myself: 

A native sampler that's ~1.5× slower than the torch roundtrip it replaces (~6.6 vs ~4.4 ms/step) isn't acceptable even behind a flag

LxYuan0420:
I am not convinced we should introduce a user-visible env var for a path that is currently worse than the default. I would prefer to keep this out of `main` until we have data showing where native MLX sampling can actually win. Otherwise, we add another sampler path to maintain without a clear user benefit today.

CAICAIIs:
Thanks, that concern makes sense. Before pushing more changes, I did a local rework to make the path narrower and checked whether it has a real winning case.

In the local version, the native path stays default-off and is limited to no-logprobs paged decode for all-random batches. Logprobs, seeds/generators, penalties, allowed token filters, bad words, mixed greedy/random batches, custom processors, and prefill still fall back to the existing Torch sampler.

The local sampling-only repro is now faster than the Torch fallback:

- batch=4, vocab=151936, top_k=50, top_p=0.9, min_p=0.05: native 3.27 ms vs Torch fallback 5.08 ms
- batch=8, vocab=151936, top_p=0.9: native 4.25 ms vs Torch fallback 6.17 ms

I also ran a local main-vs-branch e2e A/B/A/B benchmark on a decode-heavy workload:

- model: Qwen/Qwen3-0.6B
- workload: random input 32, output 128, 16 prompts, concurrency 4
- sampling: temperature=0.8, top_p=0.9, top_k=50, min_p=0.05
- env: VLLM_METAL_USE_PAGED_ATTENTION=1, branch only VLLM_METAL_NATIVE_RANDOM_SAMPLING=1

| metric | main | local branch native | delta |
|---|---:|---:|---:|
| output tok/s | 163.249 ± 0.081 | 177.135 ± 0.237 | +8.51% |
| mean TPOT | 23.208 ± 0.069 ms | 21.441 ± 0.037 ms | -7.61% |
| p90 TPOT | 23.464 ± 0.061 ms | 21.579 ± 0.159 ms | -8.03% |

Would this narrower default-off direction, with the local e2e positive case above, be worth pushing for review? Or would you prefer to keep native random sampling out until there is broader workload coverage?

LxYuan0420:
Thanks for exploring this. I think we should close this PR for now. Sampling is correctness-sensitive, and a second sampler path behind an env var is not worth carrying without a much stronger end-to-end performance case.

CAICAIIs:
> Thanks for exploring this. I think we should close this PR for now. Sampling is correctness-sensitive, and a second sampler path behind an env var is not worth carrying without a much stronger end-to-end performance case.

got it, you are right

PR #422, opened 2026-06-04
[Multimodal] Add PaddleOCR-VL support

scyyh11:
## Summary

Add native PaddleOCR-VL (`model_type: paddleocr_vl`) support on the Metal paged multimodal path.

The main difference from Qwen3-VL is RoPE ownership: PaddleOCR-VL attention layers do not own a `rope` / `rotary_emb` module. Instead, the language model computes M-RoPE `(cos, sin)` embeddings at the model level and passes them into each `self_attn` call. This PR teaches the Metal attention wrapper to consume those caller-provided position embeddings and teaches the runner to provide explicit positions whenever this model needs them.

## What changes

- New `vllm_metal/multimodal/paddleocr_vl/` adapter:
  - resolves the PaddleOCR-VL vision tower, language model, patch merge size, and token embeddings at load time;
  - encodes image features with patch-count validation and dtype alignment to the patch embedding weights;
  - delegates M-RoPE position construction to `language_model.get_rope_index`;
  - forwards `inputs_embeds`, `cache`, and `position_ids` through `call_lm`.
- Metal attention now accepts PaddleOCR-VL's model-level `(cos, sin)` position embeddings, including the mlx-vlm convention where they arrive in the fourth positional argument slot.
- SDPA preparation applies PaddleOCR-VL M-RoPE with `rope_parameters["mrope_section"]`, bypassing the normal per-layer `apply_packed_rope` path for this architecture.
- Add `MultimodalRuntimeAdapter.requires_explicit_positions`. PaddleOCR-VL sets it to `True`, so text-only batches still route through the multimodal paged path and receive correct absolute prefill/decode positions. Qwen3-VL keeps the existing text path.
- Dispatch recognizes `model_type: paddleocr_vl` and `PaddleOCRVLForConditionalGeneration`.

## Scope

- **One image per request** for now. The pinned mlx-vlm 0.5.x `get_rope_index` collapses multiple vision-start indices; this is fixed upstream in [Blaizzy/mlx-vlm#1282](https://github.com/Blaizzy/mlx-vlm/pull/1282), but not in the current pin.
- **Spec-decode fails fast** on PaddleOCR-VL through the existing multimodal-path reject. That is preferable to the previous behavior, where text-only PaddleOCR-VL spec-decode could silently use position-0 RoPE. The `feat/multimodal-spec-decode-paged` branch has the per-query-token `segment_positions` split needed to lift the reject later.
- **Paged-only**, as in #408: multimodal requests on the non-paged path keep failing fast at the encoder-dispatch chokepoint.

## Validation

**Unit tests**: 348 passed across `tests/v1/mm`, `tests/multimodal`, attention, paged-attention, model-adapter, runner-generate, and Gemma4-MTP suites.

New coverage includes:

- adapter validation, encode, `call_lm`, position construction, multi-image guard, and `requires_explicit_positions`;
- `position_embeddings` SDPA path without a per-layer `rope` attribute;
- wrapper fallback forwarding of `(cos, sin)` when no paged context is active;
- text-only prefill/decode routing for explicit-position adapters;
- dispatch by `model_type` and architecture.

**Real model** (`PaddlePaddle/PaddleOCR-VL-1.6`, paged backend):

- Full-page OCR on a 1275×1650 arXiv page (~2610 image tokens) produced coherent markdown for title, authors, abstract, and two-column body in 7.0 s.
- Text-only greedy output is byte-identical to native `mlx_vlm generate --temperature 0` on the same snapshot. The explicit-position path avoids reusing position-0 RoPE during decode.

## Notes for reviewers

- mlx-vlm contracts were checked against the pinned 0.5.0 source: `apply_multimodal_rotary_pos_emb`, `Attention.rope_parameters["mrope_section"]`, `visual.embeddings.patch_embedding`, `get_rope_index`, and `LanguageModel.__call__`.
- The `(cos, sin)` detection is structural (`2-sequence of mx.array`). Current callers are unambiguous: Qwen3-VL passes one `(3, 1, L)` array in that slot; PaddleOCR-VL passes a two-array tuple.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ vLLM Metal is a plugin that enables vLLM to run on Apple Silicon Macs using MLX 
 
 ## Supported Models
 
-vllm-metal supports a growing set of text and multimodal models on Apple Silicon. See the full matrix in [docs/supported_models.md](docs/supported_models.md).
+vllm-metal supports a growing set of models on Apple Silicon. See the full matrix in [docs/supported_models.md](docs/supported_models.md).
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ vLLM Metal is a plugin that enables vLLM to run on Apple Silicon Macs using MLX 
 
 ## Supported Models
 
-vllm-metal supports a growing set of text-only language models on Apple Silicon. See the full matrix in [docs/supported_models.md](docs/supported_models.md).
+vllm-metal supports a growing set of text and multimodal models on Apple Silicon. See the full matrix in [docs/supported_models.md](docs/supported_models.md).
 
 ## Installation
 
--- a/docs/supported_models.md
+++ b/docs/supported_models.md
@@ -1,6 +1,8 @@
 # Supported Models
 
-vllm-metal currently focuses on text-only language models on Apple Silicon. Multi-modal (vision / audio input) models are not yet supported.
+vllm-metal supports text language models and a small set of native multimodal
+models on Apple Silicon. Multimodal support is currently vision-only and runs on
+the paged backend.
 
 ## Legend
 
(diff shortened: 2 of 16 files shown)


WindChimeRan (commented):
if a patch is specific to a model, then it should stay with its own file, instead of mess up with the hot path. e.g., attention/impls, model_runner.

Need to be clear on, which part is model specific patch, which one is model-agnostic improvement. 

One of the possible fix: Move the PaddleOCR M-RoPE application next to the Qwen3.5 one in varlen_rope_compat.py

WindChimeRan (commented):
need to update doc on supported models, then it's ready to go

WindChimeRan (approved):

PR #423, opened 2026-06-05
[CI] Remove metal & cpp JIT compile, release the binary

WindChimeRan:
## What & why

Ship the native paged-attention extension **prebuilt in the wheel** instead of compiling it on the user's machine at first run.

Today `get_ops()` JIT-builds `paged_ops.cpp` + nanobind with `clang++` into `~/.cache` on first use — fragile (needs Xcode CLT, matching MLX headers, and nanobind). After this PR, users run zero cpp compilers.

This will also pave the way for future simpler installation (one line pip install)

## Cold start: no first-run compilation

Today the native kernels compile on first use, and that first run is slow. Measured (`get_ops()` wall time, fresh process, M-series):

| startup kernel init                                   | main (JIT) | this PR |
|-------------------------------------------------------|-----------:|--------:|
| clang++ build of `paged_ops.cpp` + nanobind           |    ~3.8 s  |    —    |
| Metal shader source→AIR compile (9.8 MB v2 dominates) |    ~8.1 s  |    —    |
| **first run on a fresh machine**                      | **~12 s**  | **~0.35 s** |
| subsequent restarts (OS caches warm)                  |   ~0.05 s  | ~0.04 s |

A freshly-installed wheel reaches its first token **~11.5 s sooner**, with zero compilers or Metal toolchain invoked. macOS caches compiled kernels across processes, so on `main` this is a **one-time, per-machine** cost (it recurs on every kernel-source change or MLX bump); steady-state restart time is unchanged.

<sub>Method: measured in-tree by toggling `VLLM_METAL_BUILD_FROM_SOURCE` (same sources, only load strategy differs); cold shader compile isolated by salting the source to force a metal-compiler cache miss, leaving the shared system cache untouched (stable 8.0–8.2 s across runs).</sub>

## Changes

- **Prebuilt `.so`** — `build.py` emits `_paged_ops*.so` into the package at build time; `get_ops()` loads it and fails loud if it's missing (no silent runtime compile). maturin `include` bundles it; `mlx` pinned to `==0.31.2`.
- **Precompiled shaders** — the v2 / GDN / MLA Metal libraries are compiled to `.metallib` (`xcrun metal -O3 -fno-fast-math`, matching MLX's `setFastMathEnabled(false)`) and loaded by path — no first-request shader compile.
- **CI / release** — `release.sh` and the CI test job build the artifacts before packaging, behind a defensive Metal-toolchain check; the wheel is self-contained.
- **Staleness guard** — editing `paged_ops.cpp` or a `.metal` shader without rebuilding makes `get_ops()` warn (content-hash stamps); silent for wheel users.

## Contributors

No compiler needed to *use* it. To hack on the kernels, set `VLLM_METAL_BUILD_FROM_SOURCE=1` (see `docs/CONTRIBUTING.md`) — this compiles the `.cpp` and shaders from source at runtime, so edit + restart is enough.

## ABI note

The prebuilt `.so` links MLX's private headers and subclasses its `Primitive` types, and `libmlx.dylib` carries no SONAME version — so the wheel is ABI-safe only against the exact MLX it was built against. Hence the `mlx==0.31.2` pin and a rebuild on every MLX bump.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -147,6 +147,11 @@ main() {
   setup_dev_env
 
   if [ "$(uname)" == "Darwin" ]; then
+    # Build the native artifacts (.so + .metallib) into the package tree before
+    # install.sh's `uv pip install .`, so the wheel bundles them (maturin
+    # `include`) and CI exercises the prebuilt path users get, like release.sh.
+    ensure_metal_toolchain
+    build_native_artifacts
     installs
     # shellcheck source=/dev/null
     source .venv-vllm-metal/bin/activate
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -152,6 +152,21 @@ main() {
     # `include`) and CI exercises the prebuilt path users get, like release.sh.
     ensure_metal_toolchain
     build_native_artifacts
+
+    # Shift-left the release wheel guard. The pytest run below imports the source
+    # tree, which shadows the installed wheel, so a maturin `include` regression
+    # (artifacts silently dropped from the wheel) would pass CI here and only
+    # surface when release.sh runs on main. Build the wheel and assert it bundles
+    # the prebuilt artifacts now — the same check release.sh runs before publish.
+    section "Building wheel"
+    uv build
+    local wheels=(dist/*.whl)
+    if [ ! -f "${wheels[0]}" ]; then
+      error "No wheel found in dist/ after uv build."
+      exit 1
+    fi
+    verify_wheel_artifacts "${wheels[0]}"
+
     installs
     # shellcheck source=/dev/null
     source .venv-vllm-metal/bin/activate
--- a/vllm_metal/metal/build.py
+++ b/vllm_metal/metal/build.py
@@ -282,8 +282,14 @@ def stale_artifacts() -> list[Path]:
     the ``_paged_ops`` ``.so`` (vs ``paged_ops.cpp`` et al.) and each
     ``.metallib`` (vs its ``.metal`` shaders) — checked through the single
     :func:`is_stale` mechanism. Returns ``[]`` for wheel installs (no stamps)
-    without importing the build deps, and never raises: a staleness check must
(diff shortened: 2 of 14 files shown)


scripts/test.sh
```diff
@@ -147,6 +147,26 @@ main() {
   setup_dev_env
 
   if [ "$(uname)" == "Darwin" ]; then
+    # Build the native artifacts (.so + .metallib) into the package tree before
+    # install.sh's `uv pip install .`, so the wheel bundles them (maturin
+    # `include`) and CI exercises the prebuilt path users get, like release.sh.
+    ensure_metal_toolchain
+    build_native_artifacts
```
LxYuan0420, line 154: Since full wheel install is not ready yet, `install.sh` is still the main path. CI builds native artifacts here before install, but local `./install.sh` does not. Should the local install branch also build artifacts before `uv pip install .` so fresh clones do not install successfully and then fail when `get_ops()` expects prebuilt files?

vllm_metal/metal/build.py
```diff
@@ -139,17 +261,50 @@ def needs_rebuild() -> bool:
...
+    without importing the build deps, and never raises: a staleness check must
+    not break loading."""
+    artifacts = (_OUT, *(metallib_path(n) for n in METALLIB_NAMES))
+    if not any(_stamp_path(a).exists() for a in artifacts):
+        return []
+    try:
+        digests = {_OUT: _input_hash(_build_spec())}
+        for name in METALLIB_NAMES:
+            digests[metallib_path(name)] = _metallib_digest(_metallib_source(name))
+    except (OSError, ImportError, RuntimeError, KeyError):
```
LxYuan0420: Could we avoid catching `KeyError` here? A missing builder for a `METALLIB_NAMES` entry looks like a code bug, and silently returning `[]` would hide it during development.

LxYuan0420 (approved):

PR #424, opened 2026-06-05
Clean up dead private helper code

LxYuan0420:
This PR is:
- To remove a redundant multimodal mode branch where `auto` and `text-only-compat` intentionally share the same compatibility allowlist.
- To drop unused private helper arguments from the prefill and MLA materialized-prefill paths.
- To keep private helper signatures aligned with the data they actually consume after the attention refactor.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -10,7 +10,7 @@
 | `VLLM_METAL_USE_PAGED_ATTENTION` | `1` | Enable experimental paged KV cache |
 | `VLLM_METAL_KV_SHARING_FAST_PREFILL` | `1` | Enable Gemma4 YOCO fast prefill for eligible Gemma4 models on the paged KV path |
 | `VLLM_METAL_DEBUG` | `0` | Enable debug logging |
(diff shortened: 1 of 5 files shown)


WindChimeRan (commented):
https://docs.vllm.ai/projects/vllm-metal/en/latest/configuration/#environment-variables

the env vars doc needs to be cleaned up as well. 

(Also, I think we should delete the Gemma4 YOCO Fast Prefill env var and explanation. fast prefill is already the default. there's no reason to disable it.)

LxYuan0420:
Good catch on the multimodal mode docs. I pushed a small docs update so the external configuration docs now match the local docstring: `auto` and `text-only-compat` use the same compatibility allowlist.

For `VLLM_METAL_KV_SHARING_FAST_PREFILL`, I would keep that separate. It is still an active behavior flag with config handling and tests around disabling it, so deleting it would be a behavior/API cleanup rather than part of this dead private-code cleanup.

WindChimeRan (approved):

PR #426, opened 2026-06-06
Remove YOCO fast prefill env flag

LxYuan0420:
This PR is:

- To remove the user-visible YOCO fast prefill env/config flag now that Gemma4 YOCO fast prefill is the default eligible path.
- To enable Gemma4 YOCO fast prefill directly from paged-attention setup when the loaded model matches the supported Gemma4 text contract.
- To remove stale docs and benchmark references to disabling YOCO fast prefill.


WindChimeRan (approved):
Maybe we can remove this arg in the future as well. `use_paged_attention=config.use_paged_attention`

PR #427, opened 2026-06-06
[Distributed] Pipeline Paralell

WindChimeRan:
## TLDR

Verified on 2-node macbook, connected by thunderbolt. PP=2 Qwen3-0.6B

## Summary

Pipeline parallelism for Metal/MLX. Splits a model's transformer layers into stages (one worker per stage) and pipes the hidden state stage-to-stage. 

- Data plane = MLX mx.distributed ring (point-to-point send/recv); 
- Control plane = vLLM's Ray executor + gloo. PP with TP=1 only. 

The mlx_lm model files are untouched, and everything is gated on pipeline_parallel_size > 1, so the default single-stage path is byte-for-byte unchanged.

## Limitation

- **Peak memory = full model per node.** Each stage loads the whole model before dropping its non-owned layers, so Phase 0 splits compute, not load-time memory — it can't serve a model too large for one  Mac.
- **Co-located stages oversubscribe the KV budget.** Each stage applies `VLLM_METAL_MEMORY_FRACTION` to the whole device independently — neither knows the other exists — so two stages on one Mac claim roughly twice the fraction. Lower it when stacking. Separate Macs are unaffected.
- **Synchronous scheduling required.** Run with `--no-async-scheduling` (the engine fails loud otherwise).
- **TP=1 only.** PP+TP is rejected; tensor parallelism (`--tensor-parallel-size > 1`) is not implemented.
- **Model support.** YOCO / hybrid / MLA / pooling / VLM / non-paged / speculative decoding are rejected; other shapes (sliding-window, MoE) are untested.

## Next PR for PP

- **Peak memory = full model per node.** Each stage loads the whole model before dropping its non-owned layers, so Phase 0 splits compute, not load-time memory — it can't serve a model too large for one  Mac.


vllm_metal/platform.py
```diff
@@ -317,6 +350,39 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+        # validated. Reject loudly rather than run an untested combination.
+        if (
+            parallel_config.pipeline_parallel_size > 1
+            and vllm_config.speculative_config is not None
+        ):
+            raise NotImplementedError(
+                "Pipeline parallelism on Metal does not support speculative "
+                "decoding; remove --speculative-config or run with "
+                "pipeline_parallel_size=1."
+            )
```
LxYuan0420, line 372: Optional: should we also reject PP + LoRA here for now?

LoRA is not default-on, but if a user enables it with pipeline parallelism, non-last PP stages rebuild `RequestState` without preserving `lora_id`, and decode routing later reads `state.lora_id`. Since this PR does not claim PP + LoRA support, a fail-fast guard plus docs note may be safer.

LxYuan0420 (approved):

PR #429, opened 2026-06-07
Bump mlx-vlm and enable PaddleOCR-VL multi-image requests

scyyh11:
## Summary

Enable PaddleOCR-VL requests with multiple image features now that the pinned `mlx-vlm` version includes the multi-image M-RoPE fix, and keep Qwen3.5 paged-attention smoke compatible with the new `mlx-vlm` language forward.

- Bump Darwin/arm64 `mlx-vlm` from `>=0.5.0,<0.6.0` to `>=0.6.2,<0.7.0`.
- Remove the PaddleOCR-VL adapter guard that rejected more than one image per request.
- Keep the adapter-level offset-ordering contract for `image_grid_thw`, while avoiding a redundant sort and duplicate `image_grid_thw` extraction during validation.
- Expose the wrapped SDPA attention RoPE attributes (`rotary_emb` and `rope`) so `mlx-vlm` 0.6.2 Qwen3.5 can precompute position embeddings after paged-attention patching.
- Add unit coverage for multi-image feature delegation, multi-feature vision encoding, and wrapper RoPE attribute forwarding.

## Motivation

The initial PaddleOCR-VL adapter intentionally rejected multi-image requests because `mlx-vlm <0.6` mis-counted multiple vision blocks in `get_rope_index`, corrupting M-RoPE positions for prompts with more than one image. `mlx-vlm` v0.6.2 includes the upstream fix for that path, so the adapter can accept multiple image features.

The same `mlx-vlm` PR ports the fix to `glm4v`, `glm4v_moe`, `paddleocr_vl`, `qwen2_5_vl`, `qwen2_vl`, `qwen3_omni_moe`, and `qwen3_vl_moe`; its regression test also checks those seven plus `qwen3_vl` and `qwen3_5`. This vllm-metal PR only enables and tests PaddleOCR-VL, but the dependency bump makes the broader upstream fixes available.

The dependency bump also exposes a Qwen3.5 compatibility requirement: `mlx-vlm` 0.6.2 reads `layer.self_attn.rotary_emb` before calling the attention module, while vllm-metal replaces `self_attn` with `SDPAPagedAttentionWrapper` for paged attention. The wrapper now forwards the narrow RoPE attribute surface from the wrapped attention module instead of hiding it.

## Implementation Notes

- `get_mrope_input_positions()` now sorts features once by placeholder offset, validates them in that order, and reuses the collected `image_grid_thw` values for the `mlx-vlm` language-model call.
- `_validate_image_features_and_collect_grids()` validates modality, frame count, and placeholder length while returning the already-read grids. This avoids re-running `_grid_thw()` for the same features.
- Multi-image tests use representative placeholder offsets: offsets point at the first image placeholder token, not the preceding vision-start token.
- `SDPAPagedAttentionWrapper` exposes `rotary_emb` and `rope` properties that point at the wrapped attention module, preserving the RoPE contract expected by Qwen3.5 model-level forward code.

## Testing

- `.venv-vllm-metal/bin/python -m pytest -q tests/multimodal/test_paddleocr_vl.py`
- `.venv-vllm-metal/bin/python -m pytest -q tests/multimodal/test_paddleocr_vl.py tests/v1/mm/test_paged_forward_mm.py::TestDeepstackPerChunkSlice`
- `.venv-vllm-metal/bin/python -m pytest -q tests/test_paged_attention.py::TestSDPAPagedAttentionWrapper`
- `.venv-vllm-metal/bin/ruff check vllm_metal/attention/impls/sdpa_wrapper.py tests/test_paged_attention.py`
- `.venv-vllm-metal/bin/ruff format --check vllm_metal/attention/impls/sdpa_wrapper.py tests/test_paged_attention.py`
- `.venv-vllm-metal/bin/mypy vllm_metal/attention/impls/sdpa_wrapper.py`
- Manual PaddleOCR-VL smoke with all 10 pages from `2603.24373v1.pdf` using `limit_mm_per_prompt={"image": 10}`.
- Manual Qwen3.5 paged-attention smoke reached generation without the previous `rotary_emb` wrapper crash, and the CI smoke prompt `The capital of France is` matched the expected output.

### DIFF
--- a/tests/multimodal/test_paddleocr_vl.py
+++ b/tests/multimodal/test_paddleocr_vl.py
@@ -18,6 +18,9 @@ _SPATIAL_MERGE_SIZE = 2
 _IMAGE_GRID_THW_2X2 = (1, 4, 4)
 _IMAGE_TOKEN_COUNT_2X2 = 4
 _RAW_PATCH_COUNT_2X2 = 16
+_TEXT_TOKEN_ID = 10
+_VISION_START_TOKEN_ID = 20
+_IMAGE_PLACEHOLDER_TOKEN_ID = 30
 
 
 class _RecordingLanguageModel:
@@ -176,16 +179,26 @@ class TestPaddleOCRVLMultimodalAdapterValidation:
         with pytest.raises(error_type, match=match):
             _adapter().get_mrope_input_positions([1, 2, 3], [feature])
 
-    def test_multiple_images_per_request_rejected(self) -> None:
-        # Guard against mlx-vlm < 0.6 multi-image position corruption; see
-        # mlx-vlm#1282.  Remove together with the adapter guard on bump.
+    def test_multiple_images_per_request_delegated_in_offset_order(self) -> None:
+        language_model = _RecordingLanguageModel()
+        adapter = _adapter(language_model=language_model)
         features = [
-            _feature(offset=1, grid_thw=(1, 4, 4), length=4, raw_patches=16),
             _feature(offset=8, grid_thw=(1, 6, 4), length=6, raw_patches=24),
+            _feature(offset=2, grid_thw=(1, 4, 4), length=4, raw_patches=16),
         ]
+        input_tokens = (
+            [_TEXT_TOKEN_ID, _VISION_START_TOKEN_ID]
+            + [_IMAGE_PLACEHOLDER_TOKEN_ID] * _IMAGE_TOKEN_COUNT_2X2
+            + [_TEXT_TOKEN_ID, _VISION_START_TOKEN_ID]
+            + [_IMAGE_PLACEHOLDER_TOKEN_ID] * 6
+            + [_TEXT_TOKEN_ID, _TEXT_TOKEN_ID]
+        )
 
-        with pytest.raises(NotImplementedError, match="mlx-vlm#1282"):
-            _adapter().get_mrope_input_positions([1] * 12, features)
+        adapter.get_mrope_input_positions(input_tokens, features)
+
+        call = language_model.rope_calls[0]
+        assert call["input_ids"].tolist() == [input_tokens]
+        assert call["image_grid_thw"].tolist() == [[1, 4, 4], [1, 6, 4]]
 
 
 class TestPaddleOCRVLMultimodalAdapterPositions:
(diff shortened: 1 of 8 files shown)


tests/multimodal/test_paddleocr_vl.py
```diff
@@ -176,16 +179,26 @@ def test_invalid_image_features_raise(
         with pytest.raises(error_type, match=match):
             _adapter().get_mrope_input_positions([1, 2, 3], [feature])
 
-    def test_multiple_images_per_request_rejected(self) -> None:
-        # Guard against mlx-vlm < 0.6 multi-image position corruption; see
-        # mlx-vlm#1282.  Remove together with the adapter guard on bump.
+    def test_multiple_images_per_request_delegated_in_offset_order(self) -> None:
```
LxYuan0420, line 230: Can we exercise the real mlx-vlm 0.6.2 get_rope_index here? Since this PR removes the multi-image guard for an upstream M-RoPE bug, the test should prove the real two-image position_ids/delta path, not only that we pass grids in offset order.

vllm_metal/attention/impls/sdpa_wrapper.py
```diff
@@ -78,6 +78,16 @@ def __init__(
             self, "_mk_cache_idx", cache_idx if cache_idx is not None else layer_idx
         )
 
+    @property
+    def rotary_emb(self) -> Any:
```
LxYuan0420, line 82: This delegation makes sense for Qwen3-VL model-level rotary_emb access. One follow-up edge: if upstream takes the non-fused position_embeddings path, our precomputed M-RoPE helper is currently PaddleOCR-specific. Worth adding coverage or confirming supported Metal cannot hit that path.

scyyh11: In the supported Metal path Qwen3.5 normally keeps `MRoPERotaryEmbedding.fused_apply=True`, so the smoke test exercises the `rotary_emb.apply_rotary(...)` path rather than the precomputed `position_embeddings` fallback. I added coverage and support for the fallback anyway: `apply_precomputed_mrope()` now detects Qwen3.5-style interleaved `rotary_emb` and dispatches to `mlx_vlm.models.qwen3_5.language.apply_multimodal_rotary_pos_emb` instead of the PaddleOCR-specific sectioned helper, with a unit test covering that non-fused branch.

LxYuan0420 (approved):

Issue #415, opened 2026-06-04
Add native GGUF model loading and quantized Metal execution support on Apple Silicon.


ricky-chaoju:
**Problem:** vllm-metal should support GGUF without losing the main user benefit of GGUF: quantized, memory-efficient inference.

The first implementation step should not be a dense-only fallback, and it should no longer start with a custom raw Q8_0 Metal primitive. MLX already has a native GGUF repack path for Q8_0/Q4_0/Q4_1, so those qtypes should use MLX’s existing quantized representation and kernels. Custom raw-block Metal primitives should be reserved for GGUF qtypes that MLX does not support well.

This roadmap is being reset to account for both MLX’s GGUF path and the out-of-tree `vllm-gguf-plugin` UX before the next implementation PR.

### PR Stack

* [x] PR 1: MLX-native GGUF Q8_0/Q4_0 runtime boundary (#442). Q4_1 deferred — MLX's GGUF repack currently mis-decodes it (ml-explore/mlx#3664).
* [x] PR 2: GGUF-aware Linear and Embedding wrappers #444
* [x] PR 3: Upstream/plugin-aligned local GGUF loader #449 
* [x] PR 4: Opt-in lifecycle integration and docs #453
* [ ] Follow-ups: compat shims if still needed, qtypes not covered well by MLX, MoE raw-block execution, model matrix, and benchmarks

### Design

* Q8_0/Q4_0 use MLX's GGUF repack path and existing quantized ops. Q4_1 is deferred until a fixed MLX release is in our pinned range (ml-explore/mlx#3664).
* Supported MLX-native weights must stay quantized in the runtime path; no full dense weight copy.
* The MLX-native wrapper contract must define logical shape, packed weight shape, group size, bits, input dtype, and output dtype.
* Custom raw-block Metal primitives are reserved for qtypes MLX does not support well.
* Future raw-block Metal paths should keep raw GGUF block bytes in `qweight` and the GGUF quant type in `qweight_type`.
* Unsupported qtypes should fail clearly unless an explicit fallback is chosen.
* The UX should align with `vllm-gguf-plugin` where practical, including `repo:quant`-style references.
* Upstream `gguf` should own broad tensor-name mapping and dequantization behavior where possible.
* Model-specific MLX differences should live behind adapters/strategies, not inline branches in shared loader code.
* Loader, lifecycle, compat shims, and docs claims should be split from the first runtime-boundary PR.
* `gguf` should remain optional while experimental.
[...]

PR #434, opened 2026-06-09
[GGUF] Add experimental dense GGUF loading for local models

ricky-chaoju:
Adds an experimental local GGUF load path for vllm-metal generation models.

This PR supports local `.gguf` files and complete local GGUF shard groups by resolving a paired HF/MLX config/tokenizer directory, translating GGUF tensor names to MLX-LM weights, and loading supported GGUF tensor types through a dense fallback path.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ dependencies = [
     "transformers>=5.5.1",
     "accelerate>=0.26.0",
     "safetensors>=0.4.0",
+    "gguf>=0.17.0",
     # Native Metal extension JIT build
     "nanobind==2.10.2; platform_system == 'Darwin' and platform_machine == 'arm64'",
     # Core utilities
--- a/tests/test_gguf.py
+++ b/tests/test_gguf.py
@@ -113,18 +113,18 @@ def test_model_config_candidates_take_precedence(tmp_path: Path) -> None:
     assert reference.model_path == model_dir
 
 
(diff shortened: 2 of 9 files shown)


LxYuan0420 (commented):
Thanks for working on this. I’m not convinced this dense fallback is the right thing to merge yet. This path dequantizes GGUF weights into dense MLX arrays, so it does not preserve the main user benefit of GGUF: quantized, memory-efficient inference. Before we carry a new loader, dependency, and compat patches in `main`, can you clarify what concrete benefit this gives users over loading the equivalent HF/MLX safetensors directly?

I also have a few Implementation concerns:

- This does not follow an upstream-first approach closely enough. `gguf` already owns broad name mapping and dequantization behavior; local code should only cover narrow gaps.
- Model-specific behavior is scattered through shared loader functions: Gemma4 prefixes, Qwen norm shifts, architecture aliases, tensor reshapes. This should be isolated behind a strategy/adapter boundary if we proceed.
- The tests import many underscore-prefixed helpers directly, which suggests the implementation lacks clean public seams.
- The new compat monkeypatches need focused tests and clear removal conditions.
- `gguf` should be optional for an experimental local-GGUF path, not a default dependency.

Given the current design and maintenance surface, I think we should close this PR and continue the design discussion in #415?

ricky-chaoju:
Thanks, agreed. This PR started from a dense fallback loader path, but that does not preserve the main GGUF benefit.

I’ll close this and continue in #415 with a native-first stack.

PR #432, opened 2026-06-07
[Installation] Fix broken installation script

WindChimeRan:
Problem:

```bash
$ curl -fsSL https://raw.githubusercontent.com/vllm-project/vllm-metal/main/install.sh | bash
# ....
#  - vllm==0.17.1 (from file:...)
 + vllm==0.22.1+cpu (from file:///private/var/folders/s9/tdccvq7n731b5cvfkz9wng040000gn/T/tmp.MwdIMx5xJ6/vllm-0.22.1)
/Users/ran/workspace/vllm-metal-pp
Fetching latest release...

Error: No wheel file found in the latest release.
```

Solution: Pass the GitHub release JSON to Python via stdin instead of interpolating it into a source-string literal — release-notes content was corrupting json.loads and making install.sh report "No wheel file found" on releases that have one.

PR #438, opened 2026-06-11
[Installation] Install release wheel on local runs and parse release JSON via stdin

ChopkinsAtMethod:
## Summary
- Remove the local editable-install branch so `./install.sh` always installs the latest release wheel.
- Parse GitHub release JSON from stdin to avoid bash quoting/interpolation issues.
- Update contributor docs to match the new install flow.

## Verification
- `bash -n install.sh`
- `git diff --check origin/main...HEAD`
- Full uninstall/reinstall via `rm -rf .venv-vllm-metal && ./install.sh`
- Confirmed the installed wheel and artifacts are present in `.venv-vllm-metal`

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -8,15 +8,19 @@ Thanks for your interest in contributing! This plugin targets **Apple Silicon Ma
 git clone https://github.com/vllm-project/vllm-metal.git
 cd vllm-metal
 
-# Creates ./.venv-vllm-metal/, installs vLLM core + the plugin, and
-# prebuilds the native Metal kernels from your checkout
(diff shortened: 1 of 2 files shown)


WindChimeRan (commented):
Part A: This is the same way as #432 . But it's actually defensing on a unlikely case. Don't wanna be overdefensive this way. 

Part B: The contributor UX is worse.  The flow goes from two steps to three. 

So I tend to reject this PR.

LxYuan0420:
Closing it: aligning with @WindChimeRan's review.

Thanks for the contribution. If there's a specific failure case that #430 doesn't cover, happy to revisit with a reproducer.

PR #439, opened 2026-06-11
[GGUF] Add native GGUF Q8_0 primitive boundary

ricky-chaoju:
Add the first native GGUF boundary for vllm-metal: a raw GGUF tensor representation plus a Q8_0 Metal primitive.

This PR intentionally does not add the GGUF loader, model wrappers, lifecycle wiring, compatibility shims, non-Q8 kernels, MoE, or model-level enablement. Those depend on this primitive boundary and should be reviewed separately.

* Keep `qweight` as raw GGUF `uint8` block bytes.
* Keep `qweight_type` explicit.
* Define the Q8_0 layout contract: logical shape, raw block shape, row-major contiguous raw bytes, transpose convention, input dtype, and output dtype.
* Add a lazy GGUF Metal library init path so existing core Metal kernels still load as before.
* Add Q8_0 Metal kernels that quantize activations to Q8_1 internally and multiply against raw Q8_0 blocks.
* Add focused correctness tests through public seams only.

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -126,7 +126,7 @@ build_native_artifacts() {
 }
 
 # Fail unless the freshly built wheel actually bundles the prebuilt native
-# artifacts: the _paged_ops*.so extension and the three .metallib shader
+# artifacts: the _paged_ops*.so extension and all .metallib shader
 # libraries. maturin's `include` directive is what pulls these (gitignored)
 # files in; if that ever regresses, the wheel would install fine but fail at
 # first run with "Prebuilt native extension not found". The expected filenames
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -11,7 +11,7 @@ main() {
 
   setup_dev_env
 
-  # Build the prebuilt native extension (.so) and the three precompiled
+  # Build the prebuilt native extension (.so) and all precompiled
   # .metallib shader libraries into vllm_metal/metal/ before `uv build`, so the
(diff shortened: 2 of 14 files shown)


WindChimeRan (commented):
MLX core already repacks GGUF Q8_0/Q4_0/Q4_1 losslessly into its native quantized format at load (mlx/io/gguf.cpp:228), zero new kernels. 

And in an A/B on this branch it's ~3× faster at decode and ~10× more accurate than the raw-block kernel here.

 What does the raw Q8_0 primitive buy over that path, before we commit to the kernel + packaging surface?

LxYuan0420:
Ranran is right here. Q8_0/Q4_0/Q4_1 are already covered by MLX’s GGUF repack path and existing quantized kernels, so a custom raw Q8_0 Metal primitive is not the right first step.

My bad, I should have caught this earlier when I suggested starting with Q8_0.

Let’s close this PR and move the design discussion back to #415. Before the next implementation PR, we should first account for the existing work in MLX and vllm-gguf-plugin, then agree on the smallest vllm-metal-specific step that adds new value.

ricky-chaoju:
Thanks, agreed. I missed the MLX GGUF repack path and the vllm-gguf-plugin UX here.

Since Q8_0/Q4_0/Q4_1 already have a better MLX-native route, this raw Q8_0 primitive is not the right first step. I’ll close this PR and move the design discussion back to #415.

PR #442, opened 2026-06-13
[GGUF] Add MLX-native Q8_0/Q4_0 runtime boundary

ricky-chaoju:
The original roadmap had this PR covering Q8_0, Q4_0, and Q4_1. While implementing it, I found that MLX's GGUF repack path currently mis-decodes Q4_1 weights.

Specifically, `mx.load()` appears to silently corrupt Q4_1 tensors because it skips a 2-byte block header, while Q4_1 uses a 4-byte block header. I checked this against the `gguf` package's `dequantize` implementation: Q4_1 shows a max absolute error of about `5.0`, while Q8_0 and Q4_0 are around `0.002`.

I filed the upstream fix here: ml-explore/mlx#3664.

Because of that, this PR ships **Q8_0 + Q4_0 only**. Q4_1 is rejected at construction time with a clear error pointing to ml-explore/mlx#3664. Re-enabling Q4_1 should be done as a follow-up once a fixed MLX release is within our pinned range. We currently pin `mlx==0.31.2`, so that follow-up will also need the usual MLX bump and JIT-build check.

### DIFF
--- a/vllm_metal/gguf/mlx_native.py
+++ b/vllm_metal/gguf/mlx_native.py
@@ -1,18 +1,19 @@
 # SPDX-License-Identifier: Apache-2.0
-"""MLX-native quantized GGUF tensors and the primitives that compute with them.
+"""MLX-native quantized GGUF tensors that compute with their packed weights.
 
 MLX's GGUF loader (``mx.load``) repacks Q8_0 and Q4_0 weights into the affine,
 group-32 representation that ``mx.quantized_matmul`` already consumes: a
 ``uint32`` packed ``qweight`` alongside ``float16`` ``scales`` and ``biases``.
-``GGUFMLXQuantizedTensor`` wraps that triple in an explicit, validated contract,
-and :func:`qmatmul` / :func:`embedding_lookup` run on it directly so supported
-weights never get expanded into a dense copy.
+``GGUFMLXQuantizedTensor`` wraps that triple in an explicit, validated contract
+and exposes :meth:`~GGUFMLXQuantizedTensor.matmul` /
+:meth:`~GGUFMLXQuantizedTensor.embedding`, which run on the packed weights so
+supported weights never get expanded into a dense copy.
 
 Q4_1 is deliberately not supported here. MLX's repack mis-decodes Q4_1 blocks
 (it skips a 2-byte block header where Q4_1 uses 4), so ``mx.load`` returns
 corrupted Q4_1 weights. Fix tracked upstream at ml-explore/mlx#3664; Q4_1 can
 join once a fixed MLX release is in our supported range. K-quants and other
-qtypes MLX does not repack natively are left for the raw-block Metal path.
+qtypes MLX does not repack natively are out of scope for this path.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


vllm_metal/gguf/mlx_native.py
```diff
@@ -0,0 +1,252 @@
...
+MLX_NATIVE_GGUF_TYPES = frozenset(_BITS)
+
+# MLX's GGUF repack always groups these qtypes by 32 in the affine quant mode.
+_GROUP_SIZE = 32
+_QUANT_MODE = "affine"
+
+_WEIGHT_SUFFIX = ".weight"
+
+
+def _as_supported_qtype(value: GGMLQuantizationType | int) -> GGMLQuantizationType:
```
LxYuan0420: Please move this into `GGUFMLXQuantizedTensor`, for example as a private static method `_normalize_qtype()`.

Qtype normalization is part of the tensor contract, not a module-level utility.

vllm_metal/gguf/mlx_native.py
```diff
@@ -0,0 +1,252 @@
...
+    Construct directly when you already hold the affine arrays (e.g. a row slice
+    of another tensor), or via :meth:`from_mx_load` from an ``mx.load`` result.
+    """
+
+    qweight: mx.array
+    scales: mx.array
+    biases: mx.array
+    qweight_type: GGMLQuantizationType
+
+    def __post_init__(self) -> None:
```
LxYuan0420, line 75: Could we keep `__post_init__` short and move the existing validation checks into a private method like `_validate_contract()`?

The validation is useful because this object is built from `mx.load()` / GGUF file data, but the class would read cleaner as: normalize qtype, then validate contract.

vllm_metal/gguf/mlx_native.py
```diff
@@ -0,0 +1,252 @@
...
+    @property
+    def logical_shape(self) -> tuple[int, int]:
+        return (self.out_features, self.in_features)
+
+    @property
+    def packed_shape(self) -> tuple[int, ...]:
+        return tuple(self.qweight.shape)
+
+
+def qmatmul(x: mx.array, qt: GGUFMLXQuantizedTensor) -> mx.array:
```
LxYuan0420: I think `qmatmul()` / `embedding_lookup()` should be methods on the tensor, not public module functions.

The main reason is the future dispatch boundary. When K-quant raw-block tensors arrive, the wrapper layer should be able to depend on one small protocol:

```python
class GGUFTensor(Protocol):
    def matmul(self, x: mx.array) -> mx.array: ...
    def embedding(self, ids: mx.array, output_dtype: mx.Dtype) -> mx.array: ...
```

Then both `GGUFMLXQuantizedTensor` and a future `GGUFRawBlockTensor` can satisfy it, and wrappers can call `tensor.matmul(x)` without knowing the backend representation. Free functions make that polymorphism harder because they tend to grow type-dispatch branches.

vllm_metal/gguf/mlx_native.py
```diff
@@ -0,0 +1,252 @@
...
+        scales_name = f"{prefix}.scales"
+        biases_name = f"{prefix}.biases"
+        missing = [
+            n for n in (weight_name, scales_name, biases_name) if n not in arrays
+        ]
+        if missing:
+            raise ValueError(
+                f"{qweight_type.name} tensor {weight_name!r} is missing MLX repack "
+                f"arrays {missing}; MLX did not natively repack it (expected with "
+                f"this MLX version) — load it through the raw-block path instead"
```
LxYuan0420: This points users to a “raw-block path”, but that path does not exist yet in vllm-metal. Please keep the error fail-loud without referencing an unimplemented fallback.

For example: “MLX did not natively repack this tensor with the current MLX version; this qtype is not supported by the MLX-native GGUF path.”

LxYuan0420 (approved):
Great work finding the upstream MLX Q4_1 bug and laying the groundwork here. The main shape I would like us to land with is that `GGUFMLXQuantizedTensor` owns both the data contract and the compute methods:

```python
qt = GGUFMLXQuantizedTensor.from_mx_load(
    arrays,
    "model.layers.0.weight",
    qtype,
)

out = qt.matmul(x)
emb = qt.embedding(ids, output_dtype=mx.float16)
```

That gives us a clean future seam for MLX-native tensors and future raw-block/K-quant tensors to share the same small protocol. Then wrappers can call `tensor.matmul(x)` or `tensor.embedding(...)` without growing backend-specific dispatch branches

ricky-chaoju:
Thanks, all addressed in the latest commit. Kept the Protocol itself for PR 2 since the wrapper there is the first thing that uses it.

PR #443, opened 2026-06-14
Bump vLLM to 0.23.0

ricky-chaoju:
This keeps vllm-metal aligned with upstream vLLM 0.23.0.

The TurboQuant test is updated because vLLM 0.23 resolves KV cache specs through `KVCacheSpecRegistry` instead of the old `spec_manager_map`.


WindChimeRan (commented):
Can you check _register_turboquant_spec_manager() is now dead code on 0.23?

LxYuan0420 (approved):

PR #444, opened 2026-06-15
[GGUF] Add GGUF-aware Linear/Embedding wrappers

ricky-chaoju:
This PR adds the GGUF-aware Linear and Embedding modules and the small `GGUFTensor` protocol requested in the PR 1 review. This is the second PR of the #415 stack.

`GGUFTensor` (`tensor.py`) is the seam the wrappers depend on: `matmul`, `embedding`, `out_features`, and `in_features`. PR 1's `GGUFMLXQuantizedTensor` satisfies it structurally, unchanged here, and a future raw-block tensor can satisfy the same interface.

One deviation from the original sketch: I added `out_features` and `in_features` next to the two methods, since both wrappers' `_extra_repr` need the logical shape. Reading it through the protocol keeps the wrappers independent from the concrete tensor type. Happy to trim this back to just the two methods if preferred — that would be a one-line change.

`GGUFLinear` and `GGUFEmbedding` (`wrappers.py`) are thin `nn.Module` drop-ins that delegate all quant math to the tensor. The bias is dense, added after the matmul, and cast to the activation dtype. GGUF biases are f32, which would otherwise promote a bf16/f16 result. Shape and floating dtype are validated at construction.

`GGUFEmbedding.as_linear` covers the tied `lm_head` case. The packed quant arrays stay owned by the tensor, so they are intentionally not exposed as `Module.parameters()`; this prevents dtype sweeps from touching packed `uint32` data. The forward path runs through lazy eval.

Unsupported qtypes are rejected when the tensor is built, before any wrapper exists.

### DIFF
--- a/tests/test_gguf_wrappers.py
+++ b/tests/test_gguf_wrappers.py
@@ -1,0 +1,196 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the GGUF-aware Linear/Embedding wrappers.
+
+Wrappers are exercised around the REAL tensor (built from a gguf/mx.load fixture
+like the PR-1 tests), with parity checked against ``gguf.quants.dequantize`` — no
+SimpleNamespace/MagicMock tensor fakes. The wrapper modules are nn.Modules, so
+this suite needs MLX (it won't collect on a machine without a Metal device).
+"""
+
+from __future__ import annotations
+
+import mlx.core as mx
+import numpy as np
+import pytest
+from mlx.utils import tree_flatten
+
+gguf = pytest.importorskip("gguf")
+
+from vllm_metal.gguf.mlx_native import GGUFMLXQuantizedTensor  # noqa: E402
+from vllm_metal.gguf.wrappers import GGUFEmbedding, GGUFLinear  # noqa: E402
+
+GGMLQuantizationType = gguf.GGMLQuantizationType
+
+NATIVE_QTYPES = [GGMLQuantizationType.Q8_0, GGMLQuantizationType.Q4_0]
+
+
+def _write_quantized_gguf(path, weight: np.ndarray, qtype) -> dict[str, mx.array]:
+    raw = gguf.quants.quantize(weight, qtype)
+    writer = gguf.GGUFWriter(str(path), "llama")
+    writer.add_tensor("w.weight", raw, raw_shape=raw.shape, raw_dtype=qtype)
+    writer.write_header_to_file()
+    writer.write_kv_data_to_file()
+    writer.write_tensors_to_file()
+    writer.close()
+    return mx.load(str(path))
+
+
+def _make_tensor(tmp_path, qtype, shape=(64, 128)):
+    weight = np.random.default_rng(0).standard_normal(shape).astype(np.float32)
+    arrays = _write_quantized_gguf(tmp_path / f"{qtype.name}.gguf", weight, qtype)
+    qt = GGUFMLXQuantizedTensor.from_mx_load(arrays, "w.weight", qtype)
+    oracle = gguf.quants.dequantize(gguf.quants.quantize(weight, qtype), qtype).astype(
+        np.float32
+    )
+    return qt, oracle
+
+
+def test_quant_arrays_are_not_module_parameters(tmp_path):
+    qt, _ = _make_tensor(tmp_path, GGMLQuantizationType.Q8_0)
+    bias = mx.zeros((qt.out_features,), dtype=mx.float32)
+
+    # The packed quant arrays are owned by the tensor, not the Module, so a dtype
+    # cast / parameter sweep can't touch them.
+    assert dict(tree_flatten(GGUFLinear(qt).parameters())) == {}
+    emb = GGUFEmbedding(qt, output_dtype=mx.float16)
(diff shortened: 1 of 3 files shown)


vllm_metal/gguf/wrappers.py
```diff
@@ -0,0 +1,90 @@
...
+class GGUFLinear(nn.Module):
+    """A quantized linear layer: ``x @ W.T`` (plus an optional bias).
+
+    Mirrors ``mlx.nn.QuantizedLinear`` — the bias is a dense array added after
+    the matmul. It is cast to the activation dtype so the output follows ``x``'s
+    dtype (GGUF biases are stored as float32, which would otherwise promote a
+    float16/bfloat16 result).
+    """
+
+    def __init__(self, tensor: GGUFTensor, bias: mx.array | None = None) -> None:
```
LxYuan0420, line 34: `__init__` is carrying too much validation logic inline here. I want these wrapper constructors to stay lean. If we keep the bias validation, please move it into a small private helper so `__init__` just assigns validated state.

vllm_metal/gguf/wrappers.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class GGUFEmbedding(nn.Module):
+    """A quantized embedding table; ``as_linear`` reuses it for a tied lm_head.
+
+    Mirrors ``mlx.nn.QuantizedEmbedding``: ``__call__`` gathers and dequantizes
+    rows, ``as_linear`` runs the table as the output projection.
+    """
+
+    def __init__(self, tensor: GGUFTensor, output_dtype: mx.Dtype) -> None:
```
LxYuan0420, line 80: `output_dtype` needs validation here. `GGUFEmbedding(qt, output_dtype=mx.int32)` currently succeeds and silently truncates dequantized rows to integers. This is a public wrapper boundary, so it should fail fast on non-floating dtypes. Also keep `__init__` lean; put the check in a private helper if you keep it.

tests/test_gguf_wrappers.py
```diff
@@ -0,0 +1,297 @@
...
+    weight = np.random.default_rng(0).standard_normal(shape).astype(np.float32)
+    arrays = _write_quantized_gguf(tmp_path / f"{qtype.name}.gguf", weight, qtype)
+    qt = GGUFMLXQuantizedTensor.from_mx_load(arrays, "w.weight", qtype)
+    oracle = gguf.quants.dequantize(gguf.quants.quantize(weight, qtype), qtype).astype(
+        np.float32
+    )
+    return qt, oracle
+
+
+def test_tensor_satisfies_protocol(tmp_path):
```
LxYuan0420: I would drop this test. `runtime_checkable` Protocol `isinstance` only checks member presence, not signatures or return types, so this does not buy much as a contract test. The parity tests are the real proof.

tests/test_gguf_wrappers.py
```diff
@@ -0,0 +1,297 @@
...
+
+def test_linear_extra_repr_shows_dims(tmp_path):
+    qt, _ = _make_tensor(tmp_path, GGMLQuantizationType.Q8_0, shape=(64, 128))
+    repr_str = repr(GGUFLinear(qt))
+    assert "input_dims=128" in repr_str
+    assert "output_dims=64" in repr_str
+    assert "bias=False" in repr_str
+
+
+def test_embedding_extra_repr_shows_dims(tmp_path):
```
LxYuan0420, line 133: Tighten this assertion to the actual repr contract, e.g. assert `"64, 128"` is present, instead of checking `"64"` and `"128"` independently. That will catch ordering or format drift.

tests/test_gguf_wrappers.py
```diff
@@ -0,0 +1,297 @@
...
+    arrays = _write_quantized_gguf(
+        tmp_path / "q4_1.gguf", weight, GGMLQuantizationType.Q4_1
+    )
+    with pytest.raises(ValueError, match="ml-explore/mlx#3664"):
+        GGUFMLXQuantizedTensor.from_mx_load(
+            arrays, "w.weight", GGMLQuantizationType.Q4_1
+        )
+
+
+# --- Real generate smoke (opt-in: needs a local dense Qwen3.5 checkpoint) ------
```
LxYuan0420: Please remove this smoke test from the PR. At this point I only want the focused wrapper-contract coverage in this file. A local checkpoint-dependent generate smoke is fine for your own validation, but it should stay local for now rather than be committed into `tests/test_gguf_wrappers.py`.

tests/test_gguf_wrappers.py
```diff
@@ -0,0 +1,297 @@
...
+
+_QWEN35_DENSE = os.environ.get("VLLM_METAL_TEST_QWEN35_PATH")
+
+
+@pytest.mark.slow
+@pytest.mark.skipif(
+    not _QWEN35_DENSE, reason="set VLLM_METAL_TEST_QWEN35_PATH to a dense Qwen3.5"
+)
+def test_real_generate_matches_dense():
+    import mlx.nn as nn
```
LxYuan0420: This lazy import block can go away once the smoke test is removed. I do not want function-local imports in this unit file unless there is a real runtime or collection reason to keep them.

LxYuan0420 (approved):
Great work ; but please address those minor comments

PR #446, opened 2026-06-17
Close prefill/decode asymmetry in hybrid GDN state submission

LxYuan0420:
This PR is:
- To close the prefill/decode asymmetry in hybrid GDN state submission
- To add focused coverage for the hybrid submission path

Hybrid GDN decode updates mutate the conv/recurrent state-cache arrays as side effects of the forward, and those writes are not transitively forced by `mx.eval(logits)`. Previously the paged runner only submitted those state arrays to `mx.async_eval` when a batch contained prefill work. This change removes that prefill-only gate and always submits the authoritative GDN state arrays for hybrid paged runtimes.

### DIFF
--- a/tests/test_v1_model_runner_generate.py
+++ b/tests/test_v1_model_runner_generate.py
@@ -1182,26 +1182,6 @@ class TestV1MetalModelRunnerGDNSubmit:
         )
         return backend
 
-    def test_prefill_hybrid_submits_logits_and_updated_gdn_states(
-        self, monkeypatch
-    ) -> None:
-        # Arrange
-        submitted: list[tuple[object, ...]] = []
-        runner = make_stub_runner(_paged_attention_runtime=self._make_hybrid_backend())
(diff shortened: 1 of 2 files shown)


WindChimeRan (approved):
This PR LGTM. 

---


The GDN code in the model runner needs refactoring in future PRs. `_submit_paged_forward_outputs`: GDN knowledge leaking into the runner, maybe we can use polymorphic way to make it cleaner. Need to check upstream vllm solution on this as well.

LxYuan0420:
Agree, definitely an architecture pain point to me as well.

I want to keep this PR as the minimal correctness fix, then follow with a dedicated refactor that moves GDN-specific submission/materialization ownership out of `MetalModelRunner`. I checked upstream vLLM and the right direction is the same high-level pattern: keep the runner generic and delegate model-specific state transitions behind owned hooks/objects.

PR #448, opened 2026-06-17
Optimize PaddleOCR-VL single-segment attention

scyyh11:
## Summary

- Add a PaddleOCR-VL vision attention fast path for the common single-image segment case where `cu_seqlens == [0, seq_len]`.
- Skip construction and use of the dense block-diagonal vision attention mask for that case, and call MLX scaled dot product attention with `mask=None`.
- Preserve the upstream `mlx-vlm` attention implementation for all non-single-segment cases.
- Update the PaddleOCR-VL multimodal adapter test fixture so per-image vision encoding behavior remains explicit.

## Motivation

PaddleOCR-VL requests in vLLM Metal usually encode one page/image at a time before the language model consumes the resulting visual embeddings. In that path, the visual attention sequence has exactly one contiguous segment, so the dense attention mask contains one all-zero block and does not change the result. Avoiding that mask reduces vision tower work without changing behavior for multi-segment inputs.

## Implementation Notes

- The patch is installed from `PaddleOCRVLMultimodalAdapter.from_loaded_model`.
- It only applies when the loaded visual tower actually uses `mlx_vlm.models.paddleocr_vl.vision.Attention`.
- It is class-level and idempotent via `_vllm_metal_single_segment_fast_path`.
- Runtime guard:
  - fast path: `cu_seqlens.shape == (2,)`, `cu_seqlens[0] == 0`, and `cu_seqlens[1] == seq_len`
  - fallback: call the original `Attention.__call__`

## Benchmark

Local benchmark machine: Apple Silicon, MLX/Metal backend.

Baseline commit: `377d3322dee65b027e92eb02e1613b0f44d136c6`
Current commit: `d51ffa528e2556a9e5ac254e469268ca0c3a43b9`

### Isolated PaddleOCR-VL Vision Tower

This directly measures the changed path, after one warmup run, with 5 measured runs.

| input             | baseline mean | current mean | speedup | latency reduction |
| ----------------- | ------------- | ------------ | ------- | ----------------- |
| page 1, edge 768  | 1.3884s       | 1.3319s      | 1.04x   | 4.1%              |
| page 1, edge 1280 | 5.1287s       | 4.8111s      | 1.07x   | 6.2%              |

Median latency also improved:

| input             | baseline median | current median | speedup | latency reduction |
| ----------------- | --------------- | -------------- | ------- | ----------------- |
| page 1, edge 768  | 1.3946s         | 1.3070s        | 1.07x   | 6.3%              |
| page 1, edge 1280 | 5.1315s         | 4.7886s        | 1.07x   | 6.7%              |

### End-to-End vLLM Metal Sanity Check

HTTP E2E includes request handling, VLM preprocessing, generation, prefix cache, and MM cache effects, so it is noisier than the isolated vision benchmark.

| input             | baseline mean | current mean | change |
| ----------------- | ------------- | ------------ | ------ |
| 1 page, edge 768  | 1.970s        | 2.024s       | -2.7%  |
| 2 pages, edge 768 | 2.905s        | 2.940s       | -1.2%  |
| 4 pages, edge 768 | 4.035s        | 3.949s       | +2.1%  |
| 1 page, edge 1280 | 5.206s        | 4.960s       | +4.7%  |

The isolated benchmark is the primary signal for this PR because it directly targets the patched visual attention path.

### DIFF
--- a/tests/multimodal/test_paddleocr_vl.py
+++ b/tests/multimodal/test_paddleocr_vl.py
@@ -92,7 +92,12 @@ class _RecordingVisual:
                 "output_hidden_states": output_hidden_states,
             }
         )
-        return mx.ones((_IMAGE_TOKEN_COUNT_2X2, 8), dtype=mx.float32)
+        parts = []
+        for idx, grid in enumerate(image_grid_thw.tolist()):
+            t, h, w = (int(item) for item in grid)
+            tokens = t * (h // _SPATIAL_MERGE_SIZE) * (w // _SPATIAL_MERGE_SIZE)
+            parts.append(mx.full((tokens, 8), idx + 1, dtype=mx.float32))
+        return mx.concatenate(parts, axis=0)
 
 
 class _RealPaddleOCRVLGetRopeIndexLanguageModel:
--- a/vllm_metal/multimodal/paddleocr_vl/adapter.py
+++ b/vllm_metal/multimodal/paddleocr_vl/adapter.py
@@ -57,6 +57,7 @@ class PaddleOCRVLMultimodalAdapter:
         language_model = model.language_model
         spatial_merge_size = int(model.config.vision_config.spatial_merge_size)
         embed_tokens_fn = cls._resolve_embed_tokens(language_model)
+        cls._install_single_segment_attention_fast_path(visual)
         return cls(
             spatial_merge_size=spatial_merge_size,
             visual=visual,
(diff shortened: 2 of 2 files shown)


WindChimeRan (commented):
I think this PR should go to upstream mlx_vlm:

 PaddleOCR-VL is the only one build a dense mask, other vlms don't materialize the mask at all.

scyyh11:
> I think this PR should go to upstream mlx_vlm:
> 
> PaddleOCR-VL is the only one build a dense mask, other vlms don't materialize the mask at all.

Thanks for the feedback. I will close this one and create a PR to mlx-vlm.

PR #449, opened 2026-06-18
[GGUF] Add MLX-native local GGUF loader

ricky-chaoju:
This is the third PR of the #415 stack. It takes PR 1 (#442, the MLX-native Q8_0/Q4_0 repack) and PR 2 (#444, the `GGUFLinear`/`GGUFEmbedding` wrappers) and actually loads a model with them: read a local `.gguf`, map each tensor to its mlx-lm module, install the quantized wrappers, and load the plain tensors (norms, biases) the normal way. This is the part PR 1 and PR 2 deferred. Nothing before this did `mx.load(gguf)` all the way through to generate.

There are two new files in `vllm_metal/gguf/`. `loader.py` is the orchestration: `load_gguf_model()` does a reader-only preflight, then reads the file, sorts each tensor into quant/plain/bias, installs the wrappers, loads the rest, and checks every parameter actually ended up populated. `adapter.py` is the model-family side: the GGUF to mlx-lm name mapping (the `gguf` package's tensor map plus a small static override for the global tensors), arch resolution, and the scope guard.

I kept the scope narrow on purpose: dense decoder-only, Q8_0/Q4_0, and for now just the qwen2/qwen3 families. The arch check is an allowlist, so a hybrid/MoE/vision/K-quant file is rejected in preflight with a clear error instead of half-loading, and each rejection has a test. I only listed the archs I've actually verified; it grows as I test more.

For evidence, the main one is a real-file parity test: load `Qwen3-0.6B-Q8_0.gguf` through the loader, greedy-generate 30 tokens, and check they match the dense `Qwen3-0.6B` token for token (30/30 on my M-series Mac, mlx 0.31.2 / mlx_lm 0.31.3 / gguf 0.18.0). It's opt-in behind `VLLM_METAL_TEST_QWEN3_GGUF` and the dense dir. The offline suite (`pytest tests/test_gguf_loader.py -m "not slow"`, 20 tests) covers the synthetic-GGUF partition/install/tie-skip/bias/completeness paths and every rejection, plus the real Qwen3.5 and Q4_K_M files as fail-fast fixtures.

This PR is just the loader, nothing calls it in-tree yet. PR 4 wires it into `model_lifecycle.py` with a thin branch next to the AWQ loader, behind one opt-in flag, and adds a real `vllm serve` golden-token smoke. I split them so the loader and the lifecycle hook stay one change each. Two things for now: a local `.gguf` still needs its companion HF config/tokenizer dir (no remote `repo:quant` yet), and the bias / untied lm_head paths are covered by fixtures rather than a real model, since Qwen3-0.6B has neither.


vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+from typing import Any, cast
+
+import mlx.nn as nn
+from mlx.utils import tree_flatten
+
+# GGUF globals whose MLX target has no live parameter to invert from on a tied
+# model (``output`` -> ``lm_head``), so they must translate independently of the
+# live parameter tree. These three are the only load-bearing override entries for
+# the supported dense decoder set; everything else comes from the live-derived map.
+_STATIC_GLOBAL_OVERRIDE = {
```
LxYuan0420: Please move `_STATIC_GLOBAL_OVERRIDE` onto `GGUFModelAdapter` as a class-level constant. This is adapter-owned policy, not loose module state.

vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+    "token_embd.weight": "model.embed_tokens.weight",
+    "output_norm.weight": "model.norm.weight",
+    "output.weight": "lm_head.weight",
+}
+
+# GGUF tensors with no MLX-LM counterpart (precomputed buffers MLX recreates).
+_KNOWN_SKIP = frozenset({"rope_freqs.weight"})
```
LxYuan0420: Same for `_KNOWN_SKIP`: keep this on `GGUFModelAdapter` as a class-level constant. Skip policy belongs to the adapter owner.

vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+
+# Dense decoder architectures this loader is verified against. The scope guard is
+# an allowlist (default-deny): any other arch — linear-attention/SSM hybrids,
+# fused-QKV, MoE, vision-language — is rejected in preflight, so a new non-dense
+# arch can never be silently admitted. Grow this set as archs are tested.
+# Canonical ``gguf.MODEL_ARCH_NAMES`` spellings.
+SUPPORTED_DENSE_ARCHS = frozenset({"qwen2", "qwen3"})
```
LxYuan0420: `SUPPORTED_DENSE_ARCHS` should also be adapter-owned. Do not leave support policy split between module globals and a thin class.

vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+
+# Substrings flagging an out-of-scope tensor regardless of the declared arch:
+# fused QKV, SSM/Mamba, MoE experts, and vision/mmproj. Defense in depth behind
+# the arch allowlist — catches a file mislabeled with a supported arch. None of
+# these appear in a dense decoder's tensor names (token_embd/output/blk.*).
+OUT_OF_SCOPE_TENSOR_SUBSTRINGS = (
```
LxYuan0420: `OUT_OF_SCOPE_TENSOR_SUBSTRINGS` is part of the adapter boundary as well. Please move it onto `GGUFModelAdapter` so all GGUF name/scope policy lives in one owner.

vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+    "ssm_",
+    "_exps",
+    "mmproj",
+    "mm.",
+    "v.blk",
+    "v.patch",
+)
+
+
+def normalize_arch(name: str) -> str:
```
LxYuan0420: Please do not keep `normalize_arch()` as a public free function. Make this a private `@staticmethod` on `GGUFModelAdapter`.

vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+    """Lower-case and unify separators for architecture-name comparison."""
+    return name.strip().lower().replace("-", "_")
+
+
+def canonical_arch(model_type: str) -> str:
```
LxYuan0420: I do not think you need both `normalize_arch()` and `canonical_arch()` in this shape. Collapse this to one private class-owned normalization helper. Do this as a private `@staticmethod` or `@classmethod`, not as public module API.

vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+
+    The supported dense archs (``qwen2``/``qwen3``) share the same spelling
+    between HF ``model_type`` and the GGUF ``general.architecture`` KV, so a
+    normalize is sufficient; out-of-scope archs are rejected by the allowlist
+    before any spelling subtlety matters.
+    """
+    return normalize_arch(model_type)
+
+
+def resolve_arch_enum(gguf: Any, arch: str) -> Any | None:
```
LxYuan0420: Please move `resolve_arch_enum()` into `GGUFModelAdapter` as a private `@classmethod`. Enum resolution is adapter-owned behavior and should call the class-owned normalization logic.

vllm_metal/gguf/adapter.py
```diff
@@ -0,0 +1,152 @@
...
+    ``gguf.get_tensor_name_map`` requires the enum, not the string, so the caller
+    must resolve it here rather than passing the name through.
+    """
+    for arch_enum, name in getattr(gguf, "MODEL_ARCH_NAMES", {}).items():
+        if normalize_arch(str(name)) == arch:
+            return arch_enum
+    return None
+
+
+class GGUFModelAdapter:
```
LxYuan0420, line 27: This class should own the whole adapter boundary: constants, arch normalization, enum resolution, skip policy, and translation. Right now the ownership is split across too many free helpers.

Suggested shape:

```python
class GGUFModelAdapter:
    _STATIC_GLOBAL_OVERRIDE = {...}
    _KNOWN_SKIP = frozenset({...})
    SUPPORTED_DENSE_ARCHS = frozenset({"qwen2", "qwen3"})
    OUT_OF_SCOPE_TENSOR_SUBSTRINGS = (...)

    @staticmethod
    def _normalize_arch(name: str) -> str:
        return name.strip().lower().replace("-", "_")

    @classmethod
    def _resolve_arch_enum(cls, gguf: Any, arch: str) -> Any | None:
        for arch_enum, name in gguf.MODEL_ARCH_NAMES.items():
            if cls._normalize_arch(str(name)) == arch:
                return arch_enum
        return None

    @classmethod
    def from_model(...): ...
    def translate(self, gguf_name: str) -> str | None: ...
```

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+_WEIGHT_SUFFIX = ".weight"
+_BIAS_SUFFIX = ".bias"
+_NUM_LAYERS_KEYS = ("num_hidden_layers", "n_layers", "num_layers", "n_layer")
+
+
+class GGUFLoadError(RuntimeError):
+    """Raised when a local GGUF checkpoint cannot be loaded through the MLX path."""
+
+
+def load_gguf_model(
```
LxYuan0420: This repeats the same structural issue from the earlier GGUF PRs: the loader is implemented as a procedural helper bag instead of a real owner. Please introduce a loader class and keep `load_gguf_model()` as a thin convenience wrapper around it.

Suggested shape:

```python
class GGUFModelLoader:
    def __init__(
        self,
        gguf_path: str,
        *,
        config_dir: str,
        target_dtype: mx.Dtype,
        tokenizer_config: dict[str, Any] | None = None,
    ) -> None:
        ...

    def load(self) -> tuple[nn.Module, Any]:
        self._open_inputs()
        self._build_model_from_upstream_public_loader()
        self._preflight()
        self._partition_tensors()
        self._install_quant_modules()
        self._load_plain_weights()
        self._assert_complete()
        return self._model, self._tokenizer


def load_gguf_model(...):
    return GGUFModelLoader(...).load()
```

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+
+    Raises:
+        GGUFLoadError: On any unsupported file, arch, qtype, shape mismatch, or an
+            incompletely populated model (fail fast before/at load, never silent).
+    """
+    # The gguf extra is required to even import this module (mlx_native imports
+    # gguf at module top), so its absence surfaces as a clear ImportError there;
+    # PR 4's lifecycle hook catches that before the loader is reached.
+    import gguf
+    from mlx_lm.utils import _get_classes, load_config, load_tokenizer
```
LxYuan0420: Please stop depending on `mlx_lm.utils._get_classes`. That is a private upstream API. I verified the public `mlx_lm.utils.load_model(config_dir, strict=False)` path can already build the config-only skeleton. Align this loader to the public upstream contract instead of reconstructing the model-build flow manually.

vllm_metal/gguf/adapter.py
LxYuan0420: Forgot to mention: please list the most frequently used methods first, and the less frequently used methods last

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+            f"{config_arch!r}; the .gguf and config_dir describe different models."
+        )
+
+    _preflight(gguf, reader)
+
+    arch_enum = resolve_arch_enum(gguf, arch)
+    if arch_enum is None:  # pragma: no cover - allowlisted archs always resolve
+        raise GGUFLoadError(f"Unknown GGUF architecture {arch!r}.")
+
+    model_class, model_args_class = _get_classes(config=config)
```
LxYuan0420: This manual `_get_classes(...)` + direct model construction path is exactly what I want to avoid. Use the public upstream loader path and layer the GGUF-specific orchestration on top of it.

vllm_metal/gguf/adapter.py
LxYuan0420: Small clarification on the `@classmethod` ask: I want `resolve_arch_enum()` moved onto `GGUFModelAdapter` as a private `@classmethod`, not left as a free function. Enum resolution is adapter-owned behavior, and it should route through the adapter’s class-owned normalization logic so this policy lives in one place.

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+            f"({num_layers} layers); cannot map any weight."
+        )
+
+    tied = bool(config.get("tie_word_embeddings", False))
+    arrays = mx.load(str(gguf_file))
+    quant, plain, biases = _partition(
+        reader, arrays, model=model, adapter=adapter, tied=tied
+    )
+
+    if hasattr(model, "sanitize"):
```
LxYuan0420: `if hasattr(model, "sanitize")` is another symptom of manually rebuilding upstream load flow. Once this moves to the public `load_model(..., strict=False)` path, sanitize handling should be owned there rather than reconstructed here.

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+
+    tokenizer = load_tokenizer(
+        cfg_dir,
+        tokenizer_config_extra=tokenizer_config or {},
+        eos_token_ids=config.get("eos_token_id"),
+    )
+    return model, tokenizer
+
+
+def _num_hidden_layers(config: dict[str, Any]) -> int:
```
LxYuan0420: From here down, this should become loader methods, not module-level free functions. `_num_hidden_layers`, `_gguf_arch`, `_preflight`, `_partition`, `_install_quant_modules`, `_assert_complete`, `_validate_*` all belong on the loader owner.

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+
+def _gguf_arch(reader: Any) -> str:
+    """Read and canonicalize the GGUF ``general.architecture`` metadata string."""
+    field = reader.get_field("general.architecture")
+    if field is None:
+        raise GGUFLoadError("GGUF file has no general.architecture metadata.")
+    return canonical_arch(str(field.contents()))
+
+
+def _preflight(gguf: Any, reader: Any) -> None:
```
LxYuan0420: `_preflight()` is load-bearing behavior. Keep it, but make it a private method on the loader class. This should read as an owned workflow step, not as a loose module helper.

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+                f"{name!r}; only Q8_0/Q4_0 (and plain F32/F16/BF16) are supported."
+            )
+        if name.endswith(_BIAS_SUFFIX) and tensor.tensor_type not in plain_types:
+            raise GGUFLoadError(
+                f"Unsupported qtype {tensor.tensor_type.name} on bias {name!r}; "
+                "additive biases must be plain F32/F16/BF16."
+            )
+
+
+def _partition(
```
LxYuan0420: `_partition()` is another major workflow stage that should be class-owned. Right now the loader reads like a parser script instead of production loader code with clear ownership.

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+    model: nn.Module,
+    *,
+    quant: dict[str, GGUFMLXQuantizedTensor],
+    biases: dict[str, mx.array],
+    target_dtype: mx.Dtype,
+) -> list[tuple[str, GGUFMLXQuantizedTensor, nn.Module]]:
+    """Replace each quantized module with a GGUF wrapper on the live tree."""
+    installed: list[tuple[str, GGUFMLXQuantizedTensor, nn.Module]] = []
+    for module_path, qt in quant.items():
+        current = _get_module(model, module_path)
```
LxYuan0420: Please assert the supported target module types explicitly here. For this PR’s scope, mapped quant targets should be `nn.Linear` or `nn.Embedding`. Treating every non-Embedding target as `GGUFLinear` is too loose.

Suggested check:

```python
if isinstance(current, nn.Embedding):
    wrapper = GGUFEmbedding(qt, target_dtype)
elif isinstance(current, nn.Linear):
    ...
else:
    raise GGUFLoadError(
        f"Unsupported quantized target module {type(current).__name__!r} "
        f"at {module_path!r}"
    )
```

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+                    f"Model module {module_path!r} expects a bias but the GGUF "
+                    "has no matching .bias tensor."
+                )
+            wrapper = GGUFLinear(qt, bias=bias)
+        _set_module(model, module_path, wrapper)
+        installed.append((module_path, qt, wrapper))
+    return installed
+
+
+def _assert_complete(
```
LxYuan0420: `_assert_complete()` is useful behaviorally, but it should still be loader-owned. Please stop growing module-level private helpers for core workflow stages.

vllm_metal/gguf/loader.py
```diff
@@ -0,0 +1,398 @@
...
+) -> None:
+    expected = tuple(int(d) for d in _get_param(model, param_path).shape)
+    actual = tuple(int(d) for d in array.shape)
+    if actual != expected:
+        raise GGUFLoadError(
+            f"GGUF tensor {name!r} shape {actual} != model {param_path!r} {expected}."
+        )
+
+
+def _step(obj: Any, part: str) -> Any:
```
LxYuan0420: These path-walking helpers are a code smell in this shape. If the loader needs module-tree navigation, keep it as private instance methods on the loader object. Right now this is utility sprawl inside a boundary module.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the MLX-native local GGUF loader (vllm_metal.gguf.loader).
```
LxYuan0420: The scope of this file is too mixed. I only want focused loader-contract coverage here. Synthetic contract tests are good. Local real-file smoke/parity is a different concern and should not live in the same file.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+from vllm_metal.gguf.adapter import (  # noqa: E402
+    GGUFModelAdapter,
+    canonical_arch,
+    resolve_arch_enum,
+)
+from vllm_metal.gguf.loader import GGUFLoadError, load_gguf_model  # noqa: E402
+
+QT = gguf.GGMLQuantizationType
+
+# Dense greedy continuation of "The capital of France is" for dense Qwen3-0.6B
```
LxYuan0420: This is already a smell that the file is carrying the wrong scope. Golden tokens, prompt text, and local env-driven real-checkpoint inputs do not belong at the top of the focused loader test file.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+    315,
+    279,
+    7513,
+    9145,
+    13,
+    576,
+]
+_QWEN3_GOLDEN_PROMPT = "The capital of France is"
+
+_QWEN3_GGUF = os.environ.get("VLLM_METAL_TEST_QWEN3_GGUF")
```
LxYuan0420: Same for these env vars: this file is mixing local validation harness concerns with the repo-owned contract suite. Please move the real-file checks out of this file.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+
+@pytest.fixture
+def stub_tokenizer(monkeypatch):
+    """Stub mlx_lm tokenizer loading for synthetic dirs that carry no tokenizer.
+
+    The loader imports ``load_tokenizer`` from ``mlx_lm.utils`` at call time, so
+    patching the attribute is enough; synthetic model-load tests assert on model
+    state only and never tokenize.
+    """
+    import mlx_lm.utils as mlx_lm_utils
```
LxYuan0420: No lazy import here. If this file needs `mlx_lm.utils`, import it once at module scope. Function-local imports in this test file are not justified.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+
+    monkeypatch.setattr(mlx_lm_utils, "load_tokenizer", lambda *a, **k: object())
+
+
+# --------------------------------------------------------------------------- #
+# Adapter unit tests (pure; no model build)
+# --------------------------------------------------------------------------- #
+
+
+def test_translate_precedence_static_override_then_live_then_none():
```
LxYuan0420: I would drop this test. We do not need a test for basic Python dict override / lookup precedence here. This is not meaningful loader coverage.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+    assert (
+        adapter.translate("blk.0.attn_q.weight")
+        == "model.layers.0.self_attn.q_proj.weight"
+    )
+    # known-skip and unmapped both yield None
+    assert adapter.translate("rope_freqs.weight") is None
+    assert adapter.translate("blk.0.unknown.weight") is None
+
+
+def test_is_known_skip():
```
LxYuan0420: Same issue here. `is_known_skip()` is helper plumbing, not the level of behavior I want to spend dedicated test surface on.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+    assert GGUFModelAdapter.is_known_skip("rope_freqs.weight") is True
+    assert GGUFModelAdapter.is_known_skip("token_embd.weight") is False
+
+
+def test_canonical_arch_normalizes():
```
LxYuan0420: I do not want standalone tests for trivial string normalization helpers here. Once the adapter owns this privately, this test should disappear.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+    # Normalization only (lower-case, '-' -> '_', strip); the supported dense
+    # archs share spelling between HF model_type and GGUF general.architecture.
+    assert canonical_arch("Qwen3") == "qwen3"
+    assert canonical_arch("Qwen2-VL") == "qwen2_vl"
+    assert canonical_arch("  qwen2  ") == "qwen2"
+
+
+def test_resolve_arch_enum():
```
LxYuan0420: Same for `resolve_arch_enum()`. This is helper plumbing, not the core loader contract.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+
+def test_untied_qwen2_attaches_bias_and_installs_lm_head(tmp_path, stub_tokenizer):
+    gguf_path, cfg_dir = _build_dense_fixture(
+        tmp_path,
+        "qwen2",
+        config_overrides={"tie_word_embeddings": False},
+        has_qk_norm=False,
+        with_bias=True,
+    )
+    from vllm_metal.gguf.loader import _get_module
```
LxYuan0420: Please do not import private loader helpers into tests. If this behavior matters, assert it through the public loader contract or through a proper owned abstraction. Also: no function-local imports here.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+
+def test_completeness_fails_on_dropped_norm(tmp_path):
+    gguf_path, cfg_dir = _build_dense_fixture(
+        tmp_path, "qwen3", has_qk_norm=True, drop={"blk.1.ffn_norm.weight"}
+    )
+    with pytest.raises(GGUFLoadError, match="Incomplete GGUF load"):
+        load_gguf_model(gguf_path, config_dir=cfg_dir, target_dtype=mx.float32)
+
+
+def test_assert_complete_rejects_orphan_plain_key():
```
LxYuan0420: Same issue: this test reaches directly into `_assert_complete()`. I do not want coverage built around private internals unless that private internal is itself the owned abstraction. In this PR it is not.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+    ):
+        load_gguf_model(gguf_path, config_dir=cfg_dir, target_dtype=mx.float32)
+
+
+# --------------------------------------------------------------------------- #
+# Real-file fail-fast fixtures (opt-in)
+# --------------------------------------------------------------------------- #
+
+
+@pytest.mark.skipif(
```
LxYuan0420: Please remove these real-file tests from this PR’s unit file. A local checkpoint-dependent fail-fast smoke is fine for your own validation, but it should stay local for now or move later into a dedicated slow/integration file.

tests/test_gguf_loader.py
```diff
@@ -0,0 +1,615 @@
...
+            _Q4KM_GGUF, config_dir=_QWEN3_DENSE_DIR, target_dtype=mx.float32
+        )
+
+
+# --------------------------------------------------------------------------- #
+# Real-file headline parity (slow, opt-in)
+# --------------------------------------------------------------------------- #
+
+
+@pytest.mark.slow
```
LxYuan0420: Same for the real parity test. Do not commit this into the focused loader-contract file. Keep this file tight, deterministic, and repo-owned. Real checkpoint parity can move to a separate slow/integration path once the API shape is settled.

ricky-chaoju:
Thanks, all addressed. GGUFModelAdapter now owns its constants + normalization and translation, the loader is a GGUFModelLoader class built on the public load_model(..., strict=False) (no more _get_classes) with an explicit nn.Linear / nn.Embedding install check, and the tests are split: unit contract only, real-checkpoint parity moved to test_gguf_loader_real.py, trivial and private-internal ones dropped.

LxYuan0420:
I pushed a few cleanup commits onto this PR to simplify the GGUF loader structure and fix the omitted `tie_word_embeddings` case.

I kept the loader entrypoint OOP-style, tightened the workflow readability, and added focused regression coverage for the tie-handling fix.

LxYuan0420 (approved):

ricky-chaoju:
Thanks for taking this over and catching the tie flag, I'd only tested it with the flag set so I missed the omitted case. Looks good to me.

PR #451, opened 2026-06-21
[Distributed] Support data parallelism for dense models across Macs

ricky-chaoju:
This PR is:
- To relax `MetalPlatform.check_and_update_config` admission for dense data parallelism across Macs — Ray DP backend, one full replica per node, internal load balancer — the one shape validated end to end on two Macs.
- To fail fast on every other DP combination that relaxation newly reaches: MoE, `mp` backend, `data_parallel_size_local != 1` (including the external-DP sentinel `0`), external/hybrid LB, pipeline parallelism, speculative decoding, LoRA, multimodal, STT.
- To register the Apple-GPU worker hook at the Ray **job** level (`_register_dp_ray_worker_setup_hook`), after all rejects, so an unsupported config fails before any `ray.init` side effect.
- To add DP tests (admit, every reject, a real `vllm.config.ParallelConfig` contract test, hook idempotency / foreign-init) and a `docs/distributed.md` Data parallelism section.

vLLM already owns the dense-DP control plane (replica placement, DP coordinator, request load balancer) and dense DP needs no cross-device collective, so this is pure admission + reuse — no new control-plane code, no runner/attention/cache changes. The one Metal-specific piece: Ray honours a `worker_process_setup_hook` only from the job `runtime_env`, and the DP engine manager's `ray.init` does not forward `parallel_config.ray_runtime_env`, so the per-replica `RayWorkerProc` would `KeyError` on the custom `mlx` resource without the job-level registration. Marked `# SCAFFOLDING`, removable once upstream forwards `ray_runtime_env` on the DP path.

Smoke test (two Macs, Ray cluster up — see `docs/distributed.md`):
```
RAY_ADDRESS=auto VLLM_HOST_IP=<head-ip> VLLM_METAL_USE_PAGED_ATTENTION=1 \
VLLM_METAL_MEMORY_FRACTION=0.5 vllm serve mlx-community/Qwen3-8B-4bit \
    --max-model-len 8192 --data-parallel-size 2 --data-parallel-backend ray \
    --data-parallel-size-local 1 --data-parallel-address <head-ip>
```

Benchmark (2× AWS EC2 M4 Mac mini 24 GB, VPC Ethernet; Qwen3-8B-4bit, `--max-model-len 8192`, paged attention, `MEMORY_FRACTION=0.5`; `vllm bench serve`, sonnet, 128 output tokens, warmed — output tok/s). `--max-concurrency` caps total in-flight requests, which DP=2's load balancer splits across the two replicas, so the fair comparison matches per-replica load (1 Mac @ C vs DP=2 @ 2C, each replica sees C):

| per-replica concurrency | 1 Mac | DP=2 | speedup |
| --- | --- | --- | --- |
| 8 | 35.5 | 63.9 | 1.80× |
| 16 | 54.5 | 94.6 | 1.73× |
| 32 | 64.1 | 96.6 | 1.51× |

DP=2 saturates ~96 tok/s vs ~64 single (~1.5× peak). Sub-linear because the head Mac's replica shares the host with the API server + DP coordinator + load balancer (efficiency ~90% at light load → ~75% at saturation). DP scales throughput, not capacity.

### DIFF
--- a/vllm_metal/platform.py
+++ b/vllm_metal/platform.py
@@ -43,6 +43,15 @@ class MetalPlatform(Platform):
     # name that does not collide with CUDA_VISIBLE_DEVICES.
     device_control_env_var: str = "VLLM_METAL_VISIBLE_DEVICES"
 
+    # Dotted path of the worker setup hook that patches RayWorkerProc to read the
+    # custom "mlx" resource (see compat._patch_ray_distributed). Installed both via
+    # the executor ray_runtime_env (single-stage) and, for data parallelism, at the
+    # Ray job level (_register_dp_ray_worker_setup_hook).
+    _RAY_WORKER_SETUP_HOOK: str = "vllm_metal.compat._patch_ray_distributed"
+    # Set once this process has initialized Ray with the job-level worker hook, so
+    # the DP registration is idempotent across repeated check_and_update_config calls.
+    _dp_ray_hook_registered: bool = False
+
     @classmethod
     def get_device_name(cls, device_id: int = 0) -> str:
         """Get the name of the Metal device.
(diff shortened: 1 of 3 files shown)


vllm_metal/platform.py
```diff
@@ -229,6 +238,69 @@ def get_torch_device(cls, device_id: int = 0) -> torch.device:
...
+        # second in-process DP engine re-registers instead of skipping the hook.
+        cls._dp_ray_hook_registered = False
+        # Preserve a user-provided ray_runtime_env (working_dir / py_modules /
+        # env_vars the remote Macs may need) and add our worker hook: the DP manager
+        # reuses this job session without re-applying ray_runtime_env, so replacing
+        # it with a hook-only env would drop the user's settings from remote actors.
+        runtime_env: dict[str, object] = (
+            dict(ray_runtime_env) if isinstance(ray_runtime_env, dict) else {}
+        )
+        runtime_env["worker_process_setup_hook"] = cls._RAY_WORKER_SETUP_HOOK
```
LxYuan0420: Question on the DP hook contract: I have not verified this end to end on a live multi-Mac DP Ray setup, so I am asking this as a contract question rather than asserting a bug. In `_register_dp_ray_worker_setup_hook()` we seem to overwrite any existing `ray_runtime_env["worker_process_setup_hook"]` unconditionally, while the single-stage ray path rejects a foreign hook instead of silently replacing it. Is that difference intentional for the DP path?

LxYuan0420 (approved):

vllm_metal/platform.py
ricky-chaoju: Thanks. Both paths go through the same helper now, so DP rejects a foreign hook instead of overwriting it.

WindChimeRan:
Nice work! 

One small thing on the benchmark interpretation. I re-interpreted your results to below. Because I would expect 2x speedup for 2-node DP: 

| comparison         | per-replica load  | result      | speedup                            |
| ------------------ | ----------------- | ----------- | ---------------------------------- |
| 1Mac@16 vs DP=2@16 | 16 vs **8** each  | 43.0 → 64.5 | 1.50× *(unfair — DP under-loaded)* |
| 1Mac@16 vs DP=2@32 | 16 vs **16** each | 43.0 → 92.9 | **2.16× *(matched)***              |

ricky-chaoju:
@WindChimeRan Thanks! And agreed, matched per-replica load is the right comparison. Following that I re-ran with warmup (my original single-Mac numbers were cold, which pushed the ratio up). Matched per-replica:

| per-replica conc | 1 Mac | DP=2 (2C total) | speedup |
| --- | --- | --- | --- |
| 8 | 35.5 | 63.9 | 1.80x |
| 16 | 54.5 | 94.6 | 1.73x |
| 32 | 64.1 | 96.6 | 1.51x |

So on these two 24GB M4 minis it comes out ~1.5-1.8x rather than a clean 2x. DP=2 saturates around 96 vs 64 tok/s single; the gap to 2x is the head Mac sharing its replica with the API server + DP coordinator + LB (efficiency ~90% at light load, ~75% at saturation). Updated the PR with these. I'll re-test on bigger EC2 instances at some point to see if the head overhead shrinks toward the 2x you'd expect, though probably not for a while.

PR #452, opened 2026-06-21
Move hybrid GDN state ownership behind the paged runtime seam

LxYuan0420:
This PR is:
- To move hybrid GDN state ownership out of `MetalModelRunner`
- To give hybrid request-slot lifecycle a dedicated owner in `HybridGDNStateManager`
- To keep hybrid forward side effects and pending-state materialization behind the paged runtime seam
- To make paged GDN fail closed when `gdn_slot_mapping` is missing instead of silently falling back

Follow-up to PR 446. PR 446 fixed the immediate GDN state submission issue, but left `MetalModelRunner` owning hybrid-specific `_gdn_*` lifecycle directly: request-slot mapping, slot reuse, release/materialize ordering, and hybrid forward side effects. That made the shared runner responsible for a model-family-specific state system.

This PR moves that ownership behind the runtime seam:
- `MetalModelRunner` becomes a uniform orchestrator that talks to `PagedAttentionRuntime` through four small lifecycle hooks
- `HybridPagedAttentionRuntime` owns the hybrid integration
- `HybridGDNStateManager` owns the request-slot lifecycle
- the existing release / reuse / materialize semantics are preserved
- the previous silent fallback when `gdn_slot_mapping` is missing now raises


ricky-chaoju (commented):
LGTM. Thanks for moving the GDN slot lifecycle behind the runtime boundary.

ricky-chaoju (approved):

PR #453, opened 2026-06-23
[GGUF] Serve local .gguf models

ricky-chaoju:
This PR 3 (#449) shipped `GGUFModelLoader` but nothing called it; this wires it into `model_lifecycle.py` so a local `.gguf` serves:

vllm serve <model.gguf> --tokenizer <config-dir>

It routes on detection like the AWQ branch (vLLM already sets `quantization=gguf` for a `.gguf`), with no env flag. The loader gains the AWQ-style `for_model`/`cache_key` classmethods; a `.gguf` carries weights only, so the companion config/tokenizer dir comes from `--tokenizer`.

The real serve surfaced a bug, which is the `[Bugfix]` commit: the first forward crashed because `sdpa.py` derives head_dim from `k_proj.weight.shape[0]` and `GGUFLinear` hides its packed weight (no `.weight`). I made the derivation prefer an explicit `out_features` and fall back to `.weight.shape[0]`, and gave `GGUFLinear` an `out_features` property; it is behavior-neutral for `nn.Linear`/`nn.QuantizedLinear`, which both lack `out_features`. It is the only `.weight` introspection in the attention tree.

Evidence (M2 Ultra, macOS 26.4, mlx 0.31.2): `vllm serve Qwen3-0.6B-Q8_0.gguf --tokenizer Qwen3-0.6B`. Before the fix the first request is a 500 (the AttributeError above); after, greedy `/v1/completions` gives " Paris" and " 4", and the model stays quantized (`model_memory=0.67GB`). 94 unit tests, plus an opt-in slow paged-runner smoke (`LLM.generate` through the Metal runner, asserts "Paris", mirrors the AWQ e2e test) and a golden-token test against the dense mlx-lm reference. End-to-end greedy parity is Qwen3-0.6B Q8_0 only; qwen2 and Q4_0 are loader-shape and PR 1 tensor-parity tested. Scope stays narrow (local `.gguf`, dense qwen2/qwen3, Q8_0/Q4_0); K-quants, Q4_1, MoE, SSM, vision, and remote `repo:quant` fail fast.

### DIFF
--- a/tests/test_gguf_serve_golden.py
+++ b/tests/test_gguf_serve_golden.py
@@ -1,0 +1,132 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Opt-in greedy golden-token parity for the local GGUF serve path.
+
+Loads a real local ``.gguf`` and greedy-decodes a fixed prompt across two slow
+checks:
+
+- ``test_gguf_greedy_matches_committed_golden`` pins the first ``N`` token ids to
+  a committed golden sequence — a deterministic regression guard for the served
+  Q8_0 model.
+- ``test_gguf_greedy_matches_dense_reference_prefix`` asserts the leading tokens
+  match an ``mlx_lm`` dense reference of the same checkpoint (loaded at its
+  native dtype), proving the quantized model decodes the same high-confidence
+  prefix as the float reference. The quantized and dense decodes diverge after
+  that prefix (expected from quantization — for Qwen3-0.6B they agree through
+  index 11 and diverge at index 12), so only the agreeing prefix is asserted.
+
+These load the GGUF directly through ``GGUFModelLoader``; the full ``vllm serve``
+HTTP smoke through the Metal paged runner lives in ``tools/gguf_serve_smoke.py``.
+Slow and env-gated (skipped unless the model paths are set); kept OUT of the
+deterministic unit files (real checkpoints, not synthetic fixtures). Run with::
+
+    VLLM_METAL_MEMORY_FRACTION=0.5 \\
+    VLLM_METAL_TEST_GGUF_SERVE_PATH=<...Qwen3-0.6B-Q8_0.gguf> \\
+    VLLM_METAL_TEST_GGUF_TOKENIZER_PATH=<...Qwen3-0.6B dir> \\
+    VLLM_METAL_TEST_GGUF_DENSE_PATH=<...Qwen3-0.6B dir> \\
+    pytest tests/test_gguf_serve_golden.py -m slow
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+from typing import Any
+
+import mlx.core as mx
+import pytest
+
+pytest.importorskip("gguf")
+
+from mlx_lm import load as mlx_lm_load  # noqa: E402
+
+from vllm_metal.gguf.loader import GGUFModelLoader  # noqa: E402
+
+_PROMPT = "The capital of France is"
+_N = 16
+# Greedy decode of _PROMPT through Qwen3-0.6B-Q8_0.gguf (target_dtype=bf16).
+# The leading 12 ids match the dense mlx_lm reference; they diverge at index 12
+# (quantized vs dense), so the full sequence is pinned here as a GGUF-path
+# regression while the prefix is cross-checked against dense below.
+_GOLDEN_IDS = [
+    12095,
+    13,
+    576,
+    6722,
+    315,
+    9625,
+    374,
+    1083,
+    279,
+    6722,
+    315,
+    279,
+    5429,
+    315,
+    9625,
+    13,
+]
+# Compared against the dense reference; safely inside the 12-token agreement.
+_DENSE_AGREE_PREFIX = 8
+
+_GGUF_ENV = "VLLM_METAL_TEST_GGUF_SERVE_PATH"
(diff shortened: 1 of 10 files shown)


vllm_metal/v1/model_lifecycle.py
```diff
@@ -151,6 +151,18 @@ def _load_generation_model(self, model_name: str, is_vlm: bool) -> tuple[Any, An
         logger.info("Loading model: %s (VLM: %s)", model_name, is_vlm)
         start_time = time.time()
 
+        # GGUF checkpoints are owned end-to-end by GGUFModelLoader, dispatched on
```
LxYuan0420: Design note: `ModelLifecycle` is accumulating multiple concerns in one flow: load-family routing, shared cache behavior, runner field wiring, dimension resolution, and assistant setup. This PR should keep the new GGUF path intentionally small and owner-oriented.

For the GGUF addition, lifecycle should only route, cache, and delegate. `GGUFModelLoader` should own GGUF path/tokenizer/qtype/model-install policy. Please keep the new lifecycle code short, avoid extra defensive access around typed vLLM config fields, and avoid adding GGUF format validation here.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -151,6 +151,18 @@ def _load_generation_model(self, model_name: str, is_vlm: bool) -> tuple[Any, An
...
+        # the upstream-set quantization string (vLLM core sets quantization="gguf"
+        # for a local .gguf), mirroring the AWQ branch below which routes purely on
+        # detected config. The check is gguf-package-free, so a default install
+        # never imports the optional GGUF owner on this path; it is imported lazily
+        # only inside the branch, when a GGUF model is actually served.
+        if (
+            not is_vlm
+            and getattr(self._runner.model_config, "quantization", None) == "gguf"
```
LxYuan0420: This route is correct, but the comment is heavier than the code path needs.

Can we trim this to the actual contract and avoid defensive field probing? `ModelConfig.quantization` is a real vLLM field, and `EngineArgs.create_model_config()` normalizes local `.gguf` inputs to `quantization="gguf"` before lifecycle sees them.

Suggested:

```python
# EngineArgs normalizes local GGUF inputs to quantization="gguf";
# the GGUF owner handles format-specific validation and loading.
if not is_vlm and self._runner.model_config.quantization == "gguf":
    return self._load_gguf_generation_model(model_name, start_time)
```

If test stubs need it, they should set `quantization=None` explicitly rather than making production code use `getattr`.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -207,6 +219,61 @@ def _load_generation_model(self, model_name: str, is_vlm: bool) -> tuple[Any, An
         logger.info("Model loaded in %.2fs: %s", time.time() - start_time, model_name)
         return model, tokenizer
 
+    def _load_gguf_generation_model(
+        self, model_name: str, start_time: float
+    ) -> tuple[Any, Any]:
+        """Load a local GGUF checkpoint through ``GGUFModelLoader`` (orchestration).
```
LxYuan0420: The docstring/comment can be much smaller. Lifecycle should read as orchestration: import optional owner, derive dtype/cache key, delegate. GGUF path/tokenizer/qtype policy belongs in `GGUFModelLoader`.

Suggested docstring:

```python
"""Load a GGUF checkpoint through the optional GGUF owner."""
```

tests/test_gguf_serve_golden.py
```diff
@@ -0,0 +1,187 @@
...
+
+    assert gguf_ids == dense_ids, (
+        "Q8_0 GGUF greedy prefix must match the dense mlx_lm reference "
+        f"(gguf={gguf_ids}, dense={dense_ids})"
+    )
+    assert gguf_ids == _GOLDEN_IDS[:_DENSE_AGREE_PREFIX]
+
+
+@pytest.mark.slow
+def test_gguf_e2e_paged_runner_smoke() -> None:
```
LxYuan0420: Please move this live-model serve smoke out of pytest and into `tools/`.

This test depends on local checkpoint paths, starts the full vLLM engine, and is meant to prove the user-facing `vllm serve` workflow. That is better as a reproducible maintainer script, not a skipped slow unit test. A tool can start `vllm serve`, wait for `/health`, call `/v1/completions`, and check the expected continuation directly.

The loader/unit tests should stay in `tests/`; the live serve proof should live in `tools/gguf_serve_smoke.py` or an equivalent script.

vllm_metal/attention/impls/sdpa.py
```diff
@@ -59,6 +59,22 @@ def _has_packed_qkv_sdpa_contract(module: nn.Module) -> bool:
     )
 
 
+def _projection_out_features(proj: nn.Module) -> int:
```
LxYuan0420, line 62: This shared attention change is small and technically sound, but it is broader than just wiring the loader. Since it fixes the first-forward crash exposed by GGUF wrappers, I am okay with keeping it in this PR.

Please keep it called out explicitly in the PR description/commit story as a required runtime bugfix for serving GGUF, not incidental cleanup. That makes the bundling clear for future reviewers.

ricky-chaoju:
Thanks! All done.

LxYuan0420 (approved):
LGTM

PR #454, opened 2026-06-24
Close hybrid GDN runtime bypass paths

LxYuan0420:
This PR is:
- To route pooling and non-last PP paged forward submissions through the runtime side-effect seam
- To preserve caller-output ordering while appending runtime-owned GDN state arrays
- To materialize pending runtime state before non-last PP sample exits
- To cover the bypass paths with focused runner tests

Follow-up to #452. That PR moved hybrid GDN request-slot ownership behind the paged runtime seam, but a few runner paths could still bypass the runtime-owned forward side-effect contract.

The runner still decides when to submit and finish each paged forward step. The hybrid runtime still decides which GDN state arrays need to be included and materialized.

### DIFF
--- a/tests/test_v1_model_runner_generate.py
+++ b/tests/test_v1_model_runner_generate.py
@@ -1374,6 +1374,50 @@ class TestV1MetalModelRunnerGDNSubmit:
 
         assert submitted == [(logits,)]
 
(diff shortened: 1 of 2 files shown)


ricky-chaoju (approved):
LGTM

PR #457, opened 2026-06-29
[Refactor] Clarify model lifecycle load flow

LxYuan0420:
This PR is:

- To make `ModelLifecycle.load()` read as a clear load -> install -> resolve dims -> install extensions flow.
- To capture load-time runner state in immutable DTOs instead of re-reading runner config across phases.
- To keep model installation and runtime extension installation as separate lifecycle phases with explicit ordering.
- To pin that Gemma4 MTP extension loading sees resolved target model dimensions.


ricky-chaoju (approved):
LGTM. Thanks!

PR #458, opened 2026-06-29
Use pinned MLX APIs directly

LxYuan0420:
This PR is:
- To remove legacy MLX API fallbacks that existed when we supported older MLX versions with different API locations.
- To rely on the current pinned MLX contract instead: `mlx==0.31.2` provides `mx.synchronize()`, `mx.clear_cache()`, and `mx.set_wired_limit(...)`.
- To make MLX API drift fail clearly instead of silently degrading through compatibility shims or skipped cleanup paths.
- To update tests so they assert the supported API surface rather than simulating missing or incompatible MLX APIs.


WindChimeRan (approved):
I saw other tautology tests in CI. Will need to clean them up as well.

PR #459, opened 2026-06-30
[Model] Add EXAONE 4.0 (1.2B) support on Metal

DONGRYEOLLEE1:
EXAONE-4.0-1.2B is a plain GQA model (30 layers, all full_attention, 32/8 heads, head_dim=64, tied embeddings, llama3 RoPE) and runs on the standard paged GQA path. Bring-up required two fixes:

- compat: guard transformers' Exaone4Config against a div-by-zero in __post_init__ when sliding_window is None (it sets sliding_window_pattern=0, then derives layer_types via (i+1) % 0). Pre-fill layer_types as all full_attention so the model loads through AutoConfig / mlx_lm / vLLM.
- sdpa_wrapper: expose is_local on SDPAPagedAttentionWrapper. mlx_lm's exaone4 host forward reads self_attn.is_local to select the mask; the paged wrapper replaces self_attn and previously lacked the attribute.

The paged path is bit-for-bit identical to the mlx_lm reference on the golden prompt set. Adds tests/test_exaone4_golden.py and its generator tools/gen_exaone4_golden.py, plus a docs support row (experimental). Verified on M5 Pro with mlx-community/exaone-4.0-1.2b-4bit; only the 1.2B (no-sliding-window) variant is in scope.


LxYuan0420 (changes requested):
Superseded by the newer review.

The remaining ask is coverage: default CI should exercise the EXAONE config compat patch and `SDPAPagedAttentionWrapper.is_local` directly.

vllm_metal/compat.py
```diff
@@ -31,6 +31,49 @@ def apply_compat_patches() -> None:
...
+    if config_cls is None or getattr(config_cls, "_vllm_metal_exaone4_patched", False):
+        return
+
+    original_post_init = config_cls.__post_init__
+
+    def _patched_post_init(self: Any, **kwargs: Any) -> Any:
+        if self.layer_types is None and self.sliding_window is None:
+            num_layers = getattr(self, "num_hidden_layers", 0) or 0
+            self.layer_types = ["full_attention"] * num_layers
+        return original_post_init(self, **kwargs)
```
LxYuan0420, line 71: This patch is installed on the exact `Exaone4Config` class, whose constructor contract includes `num_hidden_layers`. If that changes upstream, we should fail loudly instead of silently producing `layer_types=[]`.

```suggestion
        if self.layer_types is None and self.sliding_window is None:
            self.layer_types = ["full_attention"] * self.num_hidden_layers
        return original_post_init(self, **kwargs)
```

Please add fast `tests/test_compat.py` coverage for no-sliding-window, sliding-window, explicit `layer_types`, and idempotent patching.

vllm_metal/attention/impls/sdpa_wrapper.py
```diff
@@ -88,6 +88,17 @@ def rope(self) -> Any:
...
+    @property
+    def is_local(self) -> Any:
+        """Expose the wrapped attention's local/global attention flag.
+
+        EXAONE 4.0's host model forward reads ``self_attn.is_local`` to select
+        the sliding-window vs global mask per layer. Attention modules without
+        the attribute (the common case) default to ``False`` — global
+        attention — so wrapping them is a no-op for this flag.
+        """
+        return getattr(self._inner, "is_local", False)
```
LxYuan0420, line 100: Please add fast `tests/test_paged_attention.py` coverage for this new contract: passthrough when inner exposes `is_local`, and default `False` when it does not.

tests/test_exaone4_golden.py
```diff
@@ -0,0 +1,198 @@
...
+        print("  result: NO MATCH")
+        print(f"  expected (mlx_lm): {mlx_expected}")
+        print(f"  expected (paged):  {paged_expected}")
+
+    assert mlx_match or paged_match, (
+        f"EXAONE4 output for {prompt!r} matched neither golden.\n"
+        f"Got:               {token_ids}\n"
+        f"Expected (mlx_lm): {mlx_expected}\n"
+        f"Expected (paged):  {paged_expected}"
+    )
```
LxYuan0420: Please delete this slow golden pytest file. It is skipped in normal CI, and the only default-CI test compares identical committed dicts, so it does not exercise compat patching, `is_local`, model loading, or paged attention.

Local EXAONE checkpoint proof belongs in `tools/`; CI should cover the changed seams with fast tests.

tools/gen_exaone4_golden.py
```diff
@@ -0,0 +1,112 @@
...
+
+    model_path = args[0]
+    if paged:
+        _print_dict("GOLDEN_PAGED", _paged_golden(model_path))
+    else:
+        _print_dict("GOLDEN_MLX_LM", _mlx_lm_golden(model_path))
+
+
+if __name__ == "__main__":
+    main()
```
LxYuan0420, line 81: With the pytest golden removed, please make this the local EXAONE smoke/compare tool: run mlx_lm and vllm-metal paged on the same prompts, print both token sequences, and exit non-zero if parity fails.

tools/gen_exaone4_golden.py
```diff
@@ -0,0 +1,112 @@
...
+        args = args[1:]
+    if len(args) != 1:
+        print("Usage: python tools/gen_exaone4_golden.py [--paged] <model-path>")
+        sys.exit(1)
+
+    model_path = args[0]
+    if paged:
+        _print_dict("GOLDEN_PAGED", _paged_golden(model_path))
+    else:
+        _print_dict("GOLDEN_MLX_LM", _mlx_lm_golden(model_path))
```
LxYuan0420: Since this becomes the manual validation tool, please use `argparse` instead of hand-rolled argv parsing so `--help`, unknown flags, and usage stay correct. `import sys` can then be removed.

LxYuan0420 (changes requested):
Requesting changes on coverage.

This PR changes two runtime seams that default CI does not exercise:

- `compat.py` patches `Exaone4Config.__post_init__`
- `SDPAPagedAttentionWrapper.is_local`

Please replace the slow golden pytest with a `tools/` smoke/compare script, and add fast seam tests for both runtime changes.

vllm_metal/compat.py
DONGRYEOLLEE1: Applied — using `self.num_hidden_layers` directly to fail loud. Added test_compat.py coverage for no-sliding-window / sliding-window / explicit layer_types / idempotent.

tests/test_exaone4_golden.py
DONGRYEOLLEE1: Deleted!

vllm_metal/attention/impls/sdpa_wrapper.py
DONGRYEOLLEE1: Added test_paged_attention.py coverage: passthrough when inner exposes `is_local`, default False otherwise.

tools/gen_exaone4_golden.py
DONGRYEOLLEE1: Done — reworked into an `argparse` smoke tool: runs mlx_lm and paged on the same prompts, prints both token sequences, and exits non-zero on divergence. Dropped `import sys` / `subprocess` / `json` (single process).

LxYuan0420 (approved):
LGTM

PR #461, opened 2026-06-30
Simplify model lifecycle loading

LxYuan0420:
This PR is:

- To remove the process-level lifecycle model cache, which added global state and test coupling without a clear runtime payoff for one-model-per-worker loading.
- To delete obsolete AWQ/GGUF cache-key APIs and cache-key-only tests that only existed for that cache.
- To make generation loading dispatch explicit across GGUF, MLX-VLM, AWQ, and MLX-LM paths.
- To keep `resolve_model_dims()` as a readable lifecycle phase with private helpers ordered by the runner state they install.

Model selection and runner wiring behavior are preserved: GGUF still routes through `GGUFModelLoader`, AWQ detection stays owned by `AWQQuantLoader`, VLM still uses `mlx-vlm`, and plain text still uses `mlx-lm` with the existing path adaptation.

### DIFF
--- a/tests/quant/test_awq_e2e_dense_matrix.py
+++ b/tests/quant/test_awq_e2e_dense_matrix.py
@@ -38,7 +38,6 @@ import torch
 from mlx.utils import tree_flatten
 
 from tests.stub_runner import make_stub_runner
-from vllm_metal.v1 import model_lifecycle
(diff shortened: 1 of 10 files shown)


ricky-chaoju (approved):
LGTM.

PR #462, opened 2026-06-30
Validate pipeline ring base port

zupengwang:
## Summary
- add a named integer parser for `VLLM_METAL_RING_BASE_PORT`
- reject privileged or out-of-range pipeline ring base ports before building the MLX host list
- cover non-integer, low-port, and overflow cases in the pure Python pipeline tests

## Rationale
`_ring_hosts()` derives one TCP port per pipeline rank from `VLLM_METAL_RING_BASE_PORT`. Invalid values currently pass through until the ring backend tries to use an unusable address. Validating the base port where the rank count is known keeps the failure local and actionable.

## Tests
- `uvx ruff check vllm_metal/envs.py vllm_metal/distributed/pipeline.py tests/test_pp_pipeline.py`
- `uvx ruff format --check vllm_metal/envs.py vllm_metal/distributed/pipeline.py tests/test_pp_pipeline.py`
- `python3.12 -m py_compile vllm_metal/envs.py vllm_metal/distributed/pipeline.py tests/test_pp_pipeline.py`
- `uv run --python 3.12 --with pytest --with mlx==0.31.2 --with vllm python -m pytest tests/test_pp_pipeline.py -q`

### DIFF
--- a/vllm_metal/envs.py
+++ b/vllm_metal/envs.py
@@ -18,17 +18,6 @@ import os
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
-
-def _get_int_env(name: str, default: int) -> int:
-    value = os.getenv(name)
-    if value is None:
-        return default
-    try:
-        return int(value)
-    except ValueError as exc:
-        raise ValueError(f"{name} must be an integer, got {value!r}") from exc
-
-
 if TYPE_CHECKING:
     VLLM_METAL_MEMORY_FRACTION: str = "auto"
     VLLM_METAL_USE_MLX: bool = True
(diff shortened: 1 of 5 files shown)


vllm_metal/envs.py
```diff
@@ -18,6 +18,17 @@
...
+def _get_int_env(name: str, default: int) -> int:
+    value = os.getenv(name)
+    if value is None:
+        return default
+    try:
+        return int(value)
+    except ValueError as exc:
+        raise ValueError(f"{name} must be an integer, got {value!r}") from exc
+
+
```
LxYuan0420: Can we revert the custom `_get_int_env()` helper and keep the existing upstream-style `int(os.getenv(...))` parser here? The meaningful validation belongs in `_ring_hosts()`, where the pipeline size is known.

LxYuan0420 (approved):

LxYuan0420 (approved):
Great work!

PR #463, opened 2026-06-30
Bump vLLM to 0.24.0

ricky-chaoju:
Bumps vLLM to 0.24.0 and aligns dependency bounds with the upstream 0.24.0 stack.

Also updates speculative decoding tests for scheduler-selected draft token counts, since vLLM 0.24 now passes the per-step speculative token count through scheduler output.

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -1365,12 +1365,7 @@ class MetalModelRunner:
             )
             batch.add_output(req_id, decode_token_ids[i], logprobs)
 
-        spec = self.vllm_config.speculative_config
-        num_speculative_tokens = getattr(
-            scheduler_output,
-            "num_spec_tokens_to_schedule",
-            spec.num_speculative_tokens if spec is not None else 0,
-        )
+        num_speculative_tokens = scheduler_output.num_spec_tokens_to_schedule
         draft_ctx = ProposeContext(
             target_hidden_states=target_hidden_states,
             decode_reqs=decode_reqs,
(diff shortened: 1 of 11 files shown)


WindChimeRan:
image

benchmark: No performance drops. (0.23 vs 0.24)

ricky-chaoju:
Thanks @WindChimeRan

LxYuan0420 (approved):
One ask: any other 0.24.0 runtime contracts we should honor here, or TODOs / compat shims this bump kills?

ricky-chaoju:
> One ask: any other 0.24.0 runtime contracts we should honor here, or TODOs / compat shims this bump kills?

I didn’t find another 0.24.0 runtime contract we need to handle in this PR. The existing compat/TODOs still look valid: FastAPI `<0.137` matches upstream 0.24.0, and the Gemma4 assistant config shim is still needed at our `transformers>=5.5.3` floor.

vllm_metal/v1/model_runner.py
```diff
@@ -1365,6 +1365,12 @@ def _sample_paged_batch(
             )
             batch.add_output(req_id, decode_token_ids[i], logprobs)
 
+        spec = self.vllm_config.speculative_config
+        num_speculative_tokens = getattr(
+            scheduler_output,
+            "num_spec_tokens_to_schedule",
+            spec.num_speculative_tokens if spec is not None else 0,
```
WindChimeRan: the fallback is dead code. the default of `spec.num_speculative_tokens` is now 0

vllm_metal/v1/ngram_proposer.py
WindChimeRan, line 105: Should we add the same <= 0 guard here? same with other spec decoding algorithm?

WindChimeRan (approved):

PR #464, opened 2026-07-01
Use upstream LoRA mapping and harden adapter slots

LxYuan0420:
This PR is:

- To reuse vLLM 0.24's upstream `LoRAMapping` instead of carrying a local duplicate.
- To remove dead LoRA metadata and keep active adapter state owned by the slot table.
- To fail fast when LoRA wrapping finds no target modules.
- To reject mappings that reference inactive or unknown LoRA IDs instead of silently routing them to the base model.

The recent vLLM 0.24 bump makes the upstream mapping type available, so this cleanup reduces local fork surface while tightening our owned MLX LoRA runtime contracts.


ricky-chaoju (approved):
LGTM

PR #466, opened 2026-07-03
[GGUF] Restore local GGUF serving under vLLM 0.24

ricky-chaoju:
vLLM 0.24 removed in-tree GGUF support and moved it to vllm-gguf-plugin (vllm#39583, vllm#39612). We picked up 0.24 in #463, so nothing sets `quantization="gguf"` anymore and the GGUF serving from #453 stopped working. CI didn't catch it because the GGUF serve tests need local checkpoints.

The plugin itself doesn't install on macOS (it builds a CUDA extension and imports triton), so this PR restores the same layer inside vllm-metal. A `vllm.general_plugins` entry point registers a marker quantization config and patches `create_model_config` the same way the plugin does. A local `.gguf` gets `quantization="gguf"`, the weights path goes into `model_weights`, and `model` points at the config source (hf_config_path, then tokenizer, then the parent dir). The lifecycle reads the `.gguf` from `model_config.model_weights`. Remote `repo:quant` is recognized but rejected for now. If the plugin becomes usable on macOS, `register()` already steps aside, so we can delete this and depend on it instead.

### DIFF
--- a/vllm_metal/gguf/vllm_integration.py
+++ b/vllm_metal/gguf/vllm_integration.py
@@ -1,0 +1,295 @@
+# SPDX-License-Identifier: Apache-2.0
+"""vLLM engine integration for GGUF references (post vLLM 0.24 removal).
+
+vLLM 0.24 migrated in-tree GGUF support to the CUDA/ROCm-only
+``vllm-project/vllm-gguf-plugin`` (RFC vllm#39583, PR vllm#39612), so nothing
+sets ``quantization="gguf"`` for a ``.gguf`` model anymore and vllm-metal's
+GGUF routing went dead. This module restores the engine-facing layer with the
+same semantics as the official plugin for the fields the Metal path consumes:
+a marker quantization config so ``quantization="gguf"`` validates, and two
+narrow wraps that detect a local GGUF file, carry it in
+``model_config.model_weights``, and point ``model`` at the config source.
+
+Mounted through the ``vllm.general_plugins`` entry point (the official
+plugin's own mechanism), which vLLM loads in ``EngineArgs.__post_init__`` and
+again inside the EngineCore/worker subprocesses, so the registration is
+spawn-safe without import-time patching.
+
+This module must stay importable WITHOUT the optional ``gguf`` package: a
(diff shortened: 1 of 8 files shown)


vllm_metal/v1/model_lifecycle.py
```diff
@@ -201,26 +200,30 @@ def _load_generation_model(
...
     ) -> tuple[Any, Any]:
         """Load a GGUF checkpoint through the optional GGUF owner."""
         from vllm_metal.gguf.loader import GGUFModelLoader
 
+        # The GGUF engine integration rewrites model_config.model to the config
+        # source and carries the .gguf path in model_config.model_weights.
         loader = GGUFModelLoader.for_model(
-            model_name,
+            model_config.model_weights,
             config_dir=model_config.tokenizer,
```
LxYuan0420: This should use the validated config source, not `model_config.tokenizer`. With `--tokenizer /tok --hf-config-path /cfg`, the integration sets `model_config.model = /cfg`, but this still passes `/tok` as `config_dir`. Please use `model_config.model` for GGUF model construction, or pass separate config/tokenizer dirs into the loader.

vllm_metal/gguf/vllm_integration.py
```diff
@@ -0,0 +1,295 @@
...
+                    )
+                # Resolve (and validate) the config source BEFORE touching any
+                # EngineArgs field, so a fail-fast never leaves the args
+                # half-rewritten.
+                config_source = cls._resolve_config_source(
+                    gguf_model,
+                    self.tokenizer if isinstance(self.tokenizer, str) else None,
+                    self.hf_config_path,
+                )
+                if self.quantization is None:
```
LxYuan0420: This should fail fast for `quantization` values other than `None` or `"gguf"`. `--model model.gguf --quantization awq` keeps `awq` but still rewrites `model_weights`, so lifecycle skips the GGUF branch and fails later on the wrong loader path.

LxYuan0420 (approved):
good work!

ricky-chaoju:
Thanks @LxYuan0420 !

PR #467, opened 2026-07-03
Clarify GGUF load source ownership

LxYuan0420:
This PR is:
- To follow up #466 with an internal cleanup of the local GGUF load path.
- To introduce `GGUFLoadSource` so lifecycle code no longer reads GGUF admission fields directly.
- To simplify `GGUFModelLoader` around explicit local weights, config, and tokenizer sources.
- To keep `--hf-config-path X --tokenizer Y` separation covered in lifecycle tests.

This does not expand GGUF support beyond the local-file path restored in #466. Remote `repo:quant` serving remains a future UX improvement.

### DIFF
--- a/tests/test_gguf_loader.py
+++ b/tests/test_gguf_loader.py
@@ -436,13 +436,9 @@ def test_rejects_missing_config(tmp_path):
         ).load()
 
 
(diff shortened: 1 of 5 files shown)


ricky-chaoju (approved):
LGTM. Thanks!

PR #469, opened 2026-07-03
[Kernel] Fuse paged K/V cache write via reshape_and_cache

KrxGu:
### What

The paged decode path writes K and V into the cache with two per-layer MLX scatters (`flat_k[slot_mapping] = k_3d; flat_v[slot_mapping] = v_3d` in `sdpa_forward`). `reshape_and_cache.metal` already exists under `kernels_v2/` but isn't part of the v2 library source and isn't exposed as an op, so it wasn't being used. This wires it up: it compiles into `paged_attention_v2_kern`, is exposed as a `reshape_and_cache` MLX primitive (same in-place aliasing as `tq_encode`), and `sdpa_forward` calls it in place of the two scatters.

### Why

One fused Metal dispatch instead of two host-driven scatters on the per-layer decode hot path, and two fewer Python-level ops per layer.

Qwen2.5-7B-4bit, M3 Ultra, ctx 256, eager, single-stream decode, 3 interleaved runs each in the same session:

  `baseline:  79.9 / 80.5 / 80.3 tok/s   (mean 80.2)`
  `this PR:   84.0 / 82.4 / 84.3 tok/s   (mean 83.6)`

About +4%, and the ranges do not overlap.

### Correctness

`reshape_and_cache` output is byte-identical to the scatter it replaces (`tests/test_reshape_and_cache.py`: parametrized over fp16/bf16/fp32 and a couple of head geometries, plus a negative-slot padding case). TurboQuant and fp8 paths are untouched; this covers only the non-quantized cache write.

### Note

`reshape_and_cache.metal` declared `use_fp8_scales` at `function_constant(10)`, which clashes by name and index with an existing constant once the kernel is concatenated into the v2 library. Renamed to `rac_use_fp8_scales` at 100.

### DIFF
--- a/tests/test_attention_sdpa.py
+++ b/tests/test_attention_sdpa.py
@@ -607,6 +607,16 @@ class TestSDPAForward:
         captured: dict[str, int] = {}
 
         class _FakeOps:
+            def reshape_and_cache(
+                self,
+                _key,
+                _value,
+                key_cache,
+                value_cache,
+                _slot_mapping,
+            ) -> tuple[mx.array, mx.array]:
+                return key_cache, value_cache
+
             def paged_attention_primitive(
                 self,
                 _query,
--- a/tests/test_reshape_and_cache.py
+++ b/tests/test_reshape_and_cache.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Parity test for the fused reshape_and_cache K/V paged scatter.
+
+The fused Metal op must produce byte-identical caches to the MLX scatter it
+replaces on the decode path (flat[slot_mapping] = kv).
+"""
+
+import mlx.core as mx
+import pytest
+
+from vllm_metal.metal import get_ops
+
+
+def _reference_scatter(key, value, num_blocks, block_size, slot_mapping):
+    h, d = key.shape[1], key.shape[2]
+    kc = mx.zeros((num_blocks, block_size, h, d), dtype=key.dtype)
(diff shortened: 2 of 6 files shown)


KrxGu:
CI failed on an unrelated engine-startup OOM (runner had 2.85/7 GiB free); the change is decode-path only and doesn't affect engine init.

can someone retrigger the CI. and review the pr.

ricky-chaoju:
Is the CI issue handled independently in #471?

LxYuan0420:
My bad: CI smoke-test changes don’t belong in this kernel PR. I reverted my tweak here. Let’s keep this PR focused on reshape_and_cache and handle the CI fix in #471.

KrxGu:
> My bad: CI smoke-test changes don’t belong in this kernel PR. I reverted my tweak here. Let’s keep this PR focused on reshape_and_cache and handle the CI fix in #471.

Sure, thanks for trying. looking forward to get this PR merged.

KrxGu:
@ricky-chaoju @LxYuan0420 is there anything that i am required to do/change? to get this pr merged.

ricky-chaoju:
@KrxGu Please bear with me. Once the issue is fixed, I’ll rerun CI so this PR can go through review again. : )

LxYuan0420 (approved):
Nice kernel work. Approving this first; we’re still investigating the CI runner/platform issue in PR #471. We’ll rerun CI and merge this PR once that CI issue is fixed

ricky-chaoju (approved):
I rebased onto main and fixed the _FakeOps stub to add reshape_and_cache, which the fused write path now calls. LGTM.

PR #470, opened 2026-07-04
[LoRA] Add QLoRA support for AWQ-quantized base models

narain1:
# Summary

This PR extends the Metal v1 LoRA runtime to support attaching LoRA adapters to AWQ/MLX-quantized base models (QLoRA). Previously, MLXLoRAModelManager only wrapped plain nn.Linear modules, so quantized models had zero LoRA-eligible targets. This adds a parallel wrapper for MLX's quantized-linear protocol so `--enable-lora` works on quantized checkpoints too.

## Scope

Supported in this PR:

LoRA adapters on AWQ/MLX-quantized Linear layers, mixed with plain nn.Linear layers in the same model
Adapter loading, activation, slot reuse, and capacity handling all previously-tested behavior now verified against quantized targets.
- float16 / bfloat16 adapter dtypes on top of a quantized base


ricky-chaoju:
Hi @narain1, could you please rebase onto main to include the CI fix from #471?

narain1:
Sure, let me do that

vllm_metal/v1/lora/layers.py
```diff
@@ -18,6 +18,28 @@ def can_wrap(module: Any) -> bool:
     )
 
 
+def can_wrap_qlora(module: Any) -> bool:
+    return (
+        not isinstance(module, nn.Linear)
+        and hasattr(module, "bits")
+        and hasattr(module, "group_size")
+        and callable(module)
+    )
```
LxYuan0420, line 37: This is too broad: it matches `nn.QuantizedEmbedding`, but the wrapper calls `add_lora_linear`. Please restrict this to the quantized linear protocol, require `scales`, explicitly reject quantized embeddings, and add a regression for that case.

vllm_metal/v1/lora/layers.py
```diff
@@ -111,3 +133,88 @@ def __call__(self, x: mx.array) -> mx.array:
...
+class MLXQuantizedLinearWithLoRA(nn.Module):
+    def __init__(
+        self,
+        base_layer: Any,
+        max_loras: int,
+        max_lora_rank: int,
+        dtype: mx.Dtype,
+    ):
+        super().__init__()
+        self._base = base_layer
```
LxYuan0420: Do not store the wrapped module as `_base`; MLX skips private module attributes in `parameters()`, so the quantized base weights disappear. Use `self.base_layer`, same as `MLXLinearWithLoRA`, and test that wrapped quantized params remain visible.

vllm_metal/v1/lora/layers.py
```diff
@@ -111,3 +133,88 @@ def __call__(self, x: mx.array) -> mx.array:
...
+        y_flat = y.reshape(-1, y.shape[-1]) if y.ndim > 2 else y
+        out = self.punica_wrapper.add_lora_linear(
+            y_flat.astype(lora_dtype),
+            x_flat.astype(lora_dtype),
+            self.lora_a_stacked,
+            self.lora_b_stacked,
+            scale=1.0,
+        )
+        out = out.astype(y.dtype)
+        return out if out.shape == shape else out.reshape(shape)
```
LxYuan0420, line 248: This wrapper must remain substitutable for the projection it replaces. At minimum match `MLXLinearWithLoRA`: expose `base_layer`, `weight`, and `bias` where the base has them; runtime code may inspect `.out_features` or `.weight.shape[0]`. Please add a focused test for that surface.

vllm_metal/v1/lora/layers.py
```diff
@@ -111,3 +133,88 @@ def __call__(self, x: mx.array) -> mx.array:
...
+    def prepare_lora_weights(
+        self, slot: int, lora_a: mx.array, lora_b: mx.array
+    ) -> tuple[mx.array, mx.array]:
+        if lora_a.ndim != 2:
+            raise ValueError(
+                f"lora_a must be 2-D (rank, in_features); got shape {lora_a.shape}"
+            )
+        rank, in_ = int(lora_a.shape[0]), int(lora_a.shape[1])
+        out = int(lora_b.shape[0])
+        b_rank = int(lora_b.shape[1]) if lora_b.ndim == 2 else None
```
LxYuan0420: Validate `lora_b.ndim == 2` before reading `lora_b.shape`, matching the plain LoRA wrapper. A bad B tensor should raise a clear LoRA weight error, not a broadcast error later.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+
+    class _FakeQuantLinear(nn.Module):
+        bits = 4
+        group_size = 128
+        scales = mx.zeros((64, 2))  # (out, groups)
+
+        def __call__(self, x: mx.array) -> mx.array:
+            return x
+
+    assert can_wrap_qlora(_FakeQuantLinear())
```
LxYuan0420: Please delete or rewrite this duck-typed happy-path test. It encodes the current overly broad predicate; the important contract is that only quantized linear-like leaves are accepted and quantized embeddings are rejected.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+    b_stacked = np.array(w.lora_b_stacked)
+
+    # Slot 0 must still be zero.
+    assert not a_stacked[0].any()
+    assert not b_stacked[0].any()
+    # Slot 1: first 2 rows of A and first 2 cols of B are ones.
+    np.testing.assert_array_equal(a_stacked[1, :2, :], np.ones((2, 128)))
+    np.testing.assert_array_equal(a_stacked[1, 2:, :], np.zeros((2, 128)))
+    np.testing.assert_array_equal(b_stacked[1, :, :2], np.ones((64, 2)))
+    np.testing.assert_array_equal(b_stacked[1, :, 2:], np.zeros((64, 2)))
```
LxYuan0420: Please delete this slot-padding test. It re-tests generic LoRA slot layout rather than QLoRA-specific behavior; keep the dim inference and bad-shape tests instead.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+    w = MLXQuantizedLinearWithLoRA(ql, max_loras=1, max_lora_rank=2, dtype=mx.float32)
+
+    w.set_lora(
+        slot=0,
+        lora_a=mx.array(np.ones((2, 128), dtype=np.float32)),
+        lora_b=mx.array(np.ones((64, 2), dtype=np.float32)),
+    )
+    w.reset_lora(0)
+    assert not np.array(w.lora_a_stacked[0]).any()
+    assert not np.array(w.lora_b_stacked[0]).any()
```
LxYuan0420: Please delete this reset-slot test. Resetting stacked LoRA buffers is generic wrapper behavior already covered by the plain LoRA path, not a QLoRA-specific contract.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+    base_out = np.array(model(x).astype(mx.float32))
+    punica._no_lora = False
+
+    # With adapter active
+    out = np.array(model(x).astype(mx.float32))
+
+    # The lm_head is identity, so output = q_proj_out (+ delta).
+    delta = (b_np @ a_np @ x_np.T).T  # (1, 128)
+    expected = base_out + delta
+    np.testing.assert_allclose(out, expected, rtol=1e-3, atol=1e-3)
```
LxYuan0420, line 500: This is useful wrapper coverage, but still synthetic. Please add a short PR smoke note with a small AWQ base + real LoRA adapter showing no-LoRA works and adapter output changes.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+        max_num_batched_tokens=2,
+        dtype=mx.float32,
+    )
+    a = _make_adapter(7, module_name="q_proj", in_dim=128, out_dim=64)
+    manager.add_adapter(a)
+    # pin acknowledges a registered adapter; unknown ids return False.
+    assert manager.pin_adapter(7) is True
+    assert manager.pin_adapter(8) is False
+    assert manager.remove_adapter(7) is True
+    assert 7 not in manager.list_adapters()
```
LxYuan0420: Please delete this block. Slot reuse, capacity, and pin/remove are generic LoRA manager behavior, not QLoRA-specific behavior; they should stay covered by the existing LoRA manager tests.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+            "language_model.q_proj": _w("language_model.q_proj", 128, 64),
+            "vision_model.q_proj": _w("vision_model.q_proj", 128, 64),
+            "out_proj": _w("out_proj", 64, 32),
+        },
+    )
+    manager.add_adapter(ambiguous)
+    with pytest.raises(ValueError, match="ambiguous suffix matches"):
+        manager.activate_adapter(42)
+    assert all(sid is None for sid in manager.lora_index_to_id)
+    assert 42 not in manager.lora_index_to_id
```
LxYuan0420: Please delete this block. Zero-module match and ambiguous suffix matching are generic adapter-resolution behavior; QLoRA does not change that contract.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+
+    # Without load_inplace, duplicate is a no-op.
+    assert manager.add_adapter(_stub_lora_request(7)) is False
+
+    # With load_inplace, weights swap.
+    state[7] = v2
+    assert manager.add_adapter(_stub_lora_request(7, load_inplace=True)) is True
+    slot = manager._mm.lora_index_to_id.index(7)
+    b_col = np.array(manager._mm.modules["q_proj"].lora_b_stacked[slot])[:, 0]
+    assert np.allclose(b_col, 9.0, atol=1e-5)
```
LxYuan0420: Please delete this worker-manager block. Worker code is unchanged; one manager-level test proving a loaded adapter activates on quantized `q_proj` is enough.

tests/test_qlora.py
```diff
@@ -0,0 +1,914 @@
...
+    regular_dir = _write_peft_adapter(tmp_path / "regular", use_rslora=False)
+    rslora_dir = _write_peft_adapter(tmp_path / "rslora", use_rslora=True)
+
+    regular = peft_loader_mod.load_peft_adapter(regular_dir, lora_id=1)
+    rslora = peft_loader_mod.load_peft_adapter(rslora_dir, lora_id=2)
+
+    key = "layers.0.self_attn.q_proj"
+    # lora_alpha=4, r=2 → regular: 4/2=2.0, rslora: 4/sqrt(2)≈2.828
+    assert regular.weights[key].scaling == pytest.approx(2.0)
+    assert rslora.weights[key].scaling == pytest.approx(4.0 / np.sqrt(2.0), rel=1e-4)
```
LxYuan0420: Please delete the pure PEFT loader/scaling checks here. PEFT parsing and RSLoRA scaling are not QLoRA-specific; keep only the round-trip that proves a loaded adapter activates on quantized `q_proj`.

vllm_metal/v1/lora/layers.py
narain1: should we consider having a `setattr` for both these classes? because when i try to do 

> w.bias = mx.zeros((64,), dtype=mx.float32)

```
plain wrapper bias set FAILED: AttributeError property 'bias' of 'MLXLinearWithLoRA' object has no deleter
```

i get this error, this is because of nn.Module's setattr tries to intercept with mx.array's delete operation.

with something like this

```python
def __setattr__(self, name: str, value: Any) -> None:
    if name in ("weight", "bias") :
        type(self).__dict__[name].fset(self, value)
        return
    super().__setattr__(name, value)
```

LxYuan0420: I do not think we need assignment support here. The clean fix is just to mirror the plain wrapper surface: expose `base_layer`, `weight`, and `bias` where present, then add a focused test that runtime introspection still works after wrapping.

LxYuan0420 (approved):

PR #471, opened 2026-07-04
Restore Metal CI after transformers 5.13 breakage

ricky-chaoju:
CI serve smokes went red on 2026-07-03 evening with no repo change: main was green at 13:42 and red from 18:47, on the identical macos-15 image. The failures looked like memory problems on vLLM's CPU worker, but that was a symptom. vLLM was falling back to its stock CPU platform because MLX stopped seeing Metal inside the full vLLM process, while a bare mlx import on the same runner still saw it.

Bisecting the dependency delta between the last green and the first red run left three candidates, and only transformers 5.13.0 (released 16:05 that day, between the two runs) fits the window. An A/B on the same runner proved it: with the shipped 5.13.0 the platform resolves to CpuPlatform, and after downgrading only transformers to 5.12.1 it resolves to MetalPlatform. A bare "import torch, transformers" does not reproduce it, so the trigger needs the fuller vllm import context; I have not pinned down the exact mechanism yet.

So this PR does two things. It caps transformers below 5.13 in pyproject, the same temporary-cap shape as the existing fastapi pin, which restores the real Metal smokes and pytest on the stock macos-15 runners. And it adds a preflight assertion in scripts/test.sh that fails fast with the resolved platform name when vLLM does not resolve MetalPlatform, so the next regression like this surfaces as one clear line instead of confusing CPU-worker errors. Verified on a fork run with the cap: the full pipeline resolves MetalPlatform, both smokes pass their goldens, and 1361 tests pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,12 @@ dependencies = [
     "mlx-vlm>=0.6.2,<0.7.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
     # Model loading and weights
     # Keep aligned with vLLM 0.24.0's Transformers lower bound.
-    "transformers>=5.5.3",
+    # Temporary cap: transformers 5.13.0 (released 2026-07-03) breaks MLX Metal
+    # availability inside the vLLM process on GitHub's macOS runner VMs, so
+    # platform resolution falls back to CPU. Verified A/B on the same runner:
+    # 5.13.0 -> CpuPlatform, 5.12.1 -> MetalPlatform.
+    # TODO: remove once the interaction is understood/fixed upstream.
+    "transformers>=5.5.3,<5.13",
     "accelerate>=0.26.0",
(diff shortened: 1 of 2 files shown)


LxYuan0420 (approved):
Approving to unblock CI. Let's merge and monitor. 

Thanks Ricky!

ricky-chaoju:
Please wait a moment, I'm still confirming the issue.

ricky-chaoju:
@LxYuan0420 CI red was transformers 5.13.0 (5.13.0 → CpuPlatform, 5.12.1 → MetalPlatform on the same runner); this caps it below 5.13 like the fastapi pin plus a one-line Metal-platform preflight, fork-verified green.

LxYuan0420:
Great work and thanks Ricky. Good lesson for future version bumps: check platform resolution first, otherwise we end up debugging CPUWorker symptoms that are not the real problem.

PR #472, opened 2026-07-04
Hybrid + TurboQuant: size scheduler-visible KV pages with the packed layout

### LINKED ISSUE #468: KV cache memory-sufficiency check ignores TurboQuant compression — blocks a --max-model-len that would otherwise easily fit
## Title
KV cache memory-sufficiency check ignores TurboQuant compression — blocks a `--max-model-len` that would otherwise easily fit

## Environment
- `vllm-metal` version: 0.3.0.dev20260701212152
- `vllm` version: 0.24.0
- Hardware: Mac Studio, Apple M4 Max, 36 GB unified memory
- macOS: 26.5.1 (build 25F80)
- Model: `mlx-community/Qwen3.6-35B-A3B-4bit-DWQ` (native max context 262,144 tokens)

## Summary
TurboQuant KV cache compression works correctly and is genuinely applied — logs confirm a real 3.37x reduction in KV cache memory footprint. However, the startup "is there enough KV cache memory" validation check appears to compute its requirement using the **uncompressed (fp16) KV cache size**, regardless of whether TurboQuant is enabled. This causes the server to refuse to start at a `--max-model-len` that TurboQuant's actual (compressed) memory usage would easily support.

## Reproduction

```bash
vllm serve mlx-community/Qwen3.6-35B-A3B-4bit-DWQ \
  --port 8001 \
  --max-model-len 262144 \
  --enable-auto-tool-choice \
  --tool-call-parser qwen3_xml \
  --additional-config '{"turboquant": true, "k_quant": "q4_0", "v_quant": "q4_0"}'
```

Log output (abridged, in order):
```
INFO [kv_cache.py:227] TurboQuant KV cache (packed): 1547.3 MB (K: 4b->128d, V: 4b->128d, vs 5212.1 MB fp16, 3.37x compression)
INFO [cache_policy.py:610] Paged attention enabled: 40 layers patched, 241 blocks allocated (block_size=1056, mla=False, turboquant=True, k_quant=q4_0)
ERROR [core.py:1231] ValueError: To serve at least one request with the model's max seq len (262144), (5.08 GiB KV cache is needed, which is larger than the available KV cache memory (4.85 GiB). Based on the available memory, the estimated maximum model length is 251328. ...
```

The engine itself reports the real, compressed size (1,547.3 MB ≈ 1.51 GiB) — a 3.37x reduction from the fp16 baseline (5,212.1 MB ≈ 5.09 GiB). But the error message that then aborts startup still cites the **uncompressed 5.08 GiB figure** as "needed", and rejects based on that number rather than the ~1.51 GiB TurboQuant would actually require.

## Evidence this isn't a real memory shortage

Running the same model+config at a smaller, working `--max-model-len` (65536) with TurboQuant enabled and inspecting the actual allocated cache:

```
INFO [cache_policy.py:580] Paged attention memory breakdown: metal_limit=30.15GB, fraction=0.9, usable_metal=27.14GB, model_memory=20.66GB, overhead=1.07GB, kv_budget_before_hybrid=5.41GB, kv_budget=5.21GB, per_block_bytes=21626880, num_blocks=241, max_tokens_cached=254496
INFO [kv_cache.py:227] TurboQuant KV cache (packed): 1547.3 MB (K: 4b->128d, V: 4b->128d, vs 5212.1 MB fp16, 3.37x compression)
INFO [kv_cache_utils.py:2146] GPU KV cache size: 239,305 tokens
INFO [platform.py:667] Metal memory: 38.7GB total, 8.6GB available
```

With the *same* 5.21 GB KV budget, TurboQuant fits **239,305 tokens** of cache (vs. roughly the 65,536-token ceiling fp16 hits in that same budget) — nearly the model's full native 262,144-token context — and still leaves **8.6 GB free** after full engine init. This strongly suggests the full native context would fit fine with TurboQuant if the startup check used the compressed size instead of the fp16 size.

## Expected behavior
The "not enough KV cache memory" check should compute its requirement using the actual configured KV cache dtype/compression (TurboQuant `k_quant`/`v_quant`), matching what `kv_cache.py`'s own packed-size calculation already correctly reports, rather than always assuming fp16.

## Additional context
No speed regression observed from TurboQuant at a context length that does start successfully (65536) — baseline pp512/tg128 throughput was statistically identical to the non-quantized run (~61 tok/s generation, ~500ms median TTFT).

This is being filed separately from a related-but-distinct issue we also found in this session (a permanent EngineCore freeze on single requests above ~58k-63k input tokens) since the two appear to be independent bugs in different code paths.

abcgco:
Fixes #468

The hybrid runtime compresses SDPA layers when TurboQuant is enabled (`_build_hybrid_backend` forwards `turboquant=config.turboquant`), but `_use_turboquant()` excluded hybrid models from every scheduler-visible sizing path. Root-cause analysis with a model-free repro is in [#468 (comment)](https://github.com/vllm-project/vllm-metal/issues/468#issuecomment-4883093460) — the spec page the engine checks is exactly the compression ratio (3.37x for q4_0/q4_0) larger than the page the runtime allocates, so `check_enough_kv_cache_memory` rejects contexts that fit and the planner budgets ~3.4x fewer blocks than the packed cache supports.

What this changes:

- `ModelCachePolicy._use_turboquant()` no longer excludes hybrids: SDPA layers of hybrid models get `TurboQuantAttentionSpec` (packed `real_page_size_bytes`), and `get_cache_block_size_bytes()` budgets the paged pool with the packed page.
- The TurboQuant branch of `estimate_one_sequence_kv_bytes()` now adds the GDN linear state (`linear_cache_bytes_per_slot()`) for hybrids, mirroring the fp16 branch.
- New `MetalPlatform._realign_hybrid_block_size_for_turboquant()`, called after upstream's `update_block_size_for_backend`: upstream `_align_hybrid_block_size` sizes the attention block with the standard dtype formula, so with packed pages the attention page would no longer cover the mamba page and upstream `unify_kv_cache_spec_page_size` would fail (the packed page does not evenly divide the fp16-padded mamba page). The realignment mirrors upstream's math with the packed per-token page: it grows `block_size` accordingly and pads `mamba_page_size_padded` to exactly the packed attention page, so page-size unification sees equal pages by construction.

Testing:

- `pytest tests/test_turboquant_hybrid_sizing.py` — 6 new regression tests (specs / per-block bytes / one-sequence estimate / alignment); all 6 fail before the fix, pass after.
- `pytest -m "not slow" tests/` — 1198 passed, 17 skipped. The 104 failing tests are the Metal-kernel suites failing at import ("Prebuilt native extension not found") in this dev env without prebuilt artifacts — verified identical on `main`.
- `ruff check`, `ruff format --check`, `mypy vllm_metal` — clean.

Worth discussing as a follow-up: vLLM 0.24 has native TurboQuant spec machinery (`TQFullAttentionSpec`, `cache_dtype="turboquant_*"` alignment in `Platform._align_hybrid_block_size`); migrating the plugin's TurboQuant onto it would remove the mirrored alignment added here.

### DIFF
--- a/vllm_metal/platform.py
+++ b/vllm_metal/platform.py
@@ -757,7 +757,6 @@ class MetalPlatform(Platform):
         # (``_align_hybrid_block_size``) handles hybrid alignment. The kernel
         # layer (``_pick_kernel_block_size``) validates the final
         # ``block_size`` at request time.
-        #
         cache_config = vllm_config.cache_config
         user_block_size = (
             cache_config.block_size if cache_config.user_specified_block_size else None
@@ -783,20 +782,7 @@ class MetalPlatform(Platform):
         user_block_size: int | None,
         user_mamba_block_size: int | None,
     ) -> None:
-        """Redo hybrid block-size alignment with TurboQuant page sizes.
-
-        Upstream ``Platform._align_hybrid_block_size`` sizes the attention
-        block with the standard dtype page formula, but TurboQuant packs
-        SDPA pages ~3-4x smaller (vllm-metal#468). Keeping the fp16-based
-        block would make every scheduler-visible page disagree with the
-        runtime layout: ``check_enough_kv_cache_memory`` rejects contexts
-        that fit, and the paged-attention planner budgets ~3-4x fewer
-        blocks than the TurboQuant cache actually needs.
-
-        Mirrors the upstream alignment (same mamba-state math, same
-        kernel-alignment source) with ``attn_page_size_1_token`` computed
-        from the packed TurboQuant layout.
-        """
+        """Redo hybrid alignment with TurboQuant's packed SDPA page size."""
         from vllm.model_executor.models import ModelRegistry
         from vllm.utils.math_utils import cdiv
         from vllm.v1.attention.backend import MultipleOf
@@ -808,10 +794,6 @@ class MetalPlatform(Platform):
         model_config = vllm_config.model_config
         cache_config = vllm_config.cache_config
 
-        # The resolved Metal config is authoritative here: ``check_and_update_config``
-        # populates the singleton on the driver, and ``MetalWorker.__init__`` re-applies
-        # TurboQuant from ``additional_config`` in each worker process before this hook
-        # runs, so the singleton already reflects TurboQuant in both cases.
         turboquant = metal_config.turboquant
         k_quant = metal_config.k_quant
         v_quant = metal_config.v_quant
@@ -824,7 +806,6 @@ class MetalPlatform(Platform):
         ):
             return
 
-        # Same mamba-state computation as upstream _align_hybrid_block_size.
         model_cls, _ = ModelRegistry.resolve_model_cls(
             model_config.architecture,
             model_config=model_config,
@@ -837,8 +818,6 @@ class MetalPlatform(Platform):
         if mamba_page_size == 0:
             return
 
-        # _turboquant_page_size_bytes is linear in block_size, so the
-        # one-token value scales exactly to any block size.
         tq_page_size_1_token = _turboquant_page_size_bytes(
             block_size=1,
             num_kv_heads=model_config.get_num_kv_heads(vllm_config.parallel_config),
@@ -859,8 +838,6 @@ class MetalPlatform(Platform):
         )
 
         if cache_config.mamba_cache_mode == "all":
-            # Mirror upstream: with prefix caching, align to the mamba chunk
-            # size for kernel performance.
             from math import lcm
 
             base_chunk_size = (
--- a/vllm_metal/platform.py
+++ b/vllm_metal/platform.py
@@ -766,12 +766,14 @@ class MetalPlatform(Platform):
             if cache_config.user_specified_mamba_block_size
             else None
         )
+        hash_block_size = cache_config.hash_block_size
         super().update_block_size_for_backend(vllm_config)
 
         cls._realign_hybrid_block_size_for_turboquant(
             vllm_config,
             user_block_size=user_block_size,
             user_mamba_block_size=user_mamba_block_size,
+            hash_block_size=hash_block_size,
         )
 
     @classmethod
@@ -781,6 +783,7 @@ class MetalPlatform(Platform):
         *,
         user_block_size: int | None,
         user_mamba_block_size: int | None,
+        hash_block_size: int | None,
     ) -> None:
         """Redo hybrid alignment with TurboQuant's packed SDPA page size."""
         metal_config = get_config()
@@ -798,6 +801,8 @@ class MetalPlatform(Platform):
             or not model_config.is_hybrid
         ):
             return
+
+        from math import lcm
 
         from vllm.model_executor.models import ModelRegistry
         from vllm.utils.math_utils import cdiv
@@ -832,14 +837,13 @@ class MetalPlatform(Platform):
             s.base if isinstance(s, MultipleOf) else s
             for s in backend_cls.get_supported_kernel_block_sizes()
         )
-        kernel_block_alignment_size = max(
+        kernel_block_alignment_size = lcm(
             backend_block_alignment_size,
-            user_block_size or backend_block_alignment_size,
+            user_block_size or 1,
+            hash_block_size or 1,
         )
 
         if cache_config.mamba_cache_mode == "all":
-            from math import lcm
-
             base_chunk_size = (
                 user_mamba_block_size or model_config.get_mamba_chunk_size()
             )
--- a/vllm_metal/platform.py
+++ b/vllm_metal/platform.py
@@ -397,15 +397,17 @@ class MetalPlatform(Platform):
         # Disable features not supported on Metal
         parallel_config.disable_custom_all_reduce = True
 
-        if (
-            model_config is not None
-            and model_config.is_hybrid
-            and vllm_config.cache_config.enable_prefix_caching
-        ):
-            raise NotImplementedError(
-                "Prefix caching is not supported for hybrid GDN models on Metal "
-                "because GDN recurrent state cannot be restored from KV blocks."
-            )
+        if model_config is not None and model_config.is_hybrid:
+            cache_config = vllm_config.cache_config
+            if (
+                cache_config.enable_prefix_caching
+                or cache_config.mamba_cache_mode != "none"
+            ):
+                raise NotImplementedError(
+                    "Prefix caching and Mamba cache modes are not supported for "
+                    "hybrid GDN models on Metal because GDN recurrent state cannot "
+                    "be restored from KV blocks."
+                )
 
         # Pipeline parallelism is supported on Metal/MLX: each stage runs in its
         # own worker process and the inter-stage activations cross the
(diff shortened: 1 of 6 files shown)


abcgco:
Pushed a follow-up commit (f445c68) after an adversarial self-review of the first one:

- **`is_kv_cache_spec_uniform` compatibility**: vLLM probes uniformity via `values()[0].merge(all_values)` and treats `AssertionError` as "not uniform". For hybrids whose *first* layer is SDPA, `TurboQuantAttentionSpec.merge` answered the legitimately mixed TurboQuant+Mamba dict with `TypeError`, which escaped the probe's `except AssertionError` and aborted engine init. It now raises `AssertionError`, matching `KVCacheSpec.merge`'s assert-based contract. Regression test covers both dict orders (linear-first and SDPA-first).
- **Cross-process gating**: the realignment gate read only the process-local Metal config singleton, which executors can consult in worker processes before it is populated (PP>1 / mp executor). It now gates on `vllm_config.additional_config` — which travels with the pickled config — with the singleton as fallback.

8 tests in `tests/test_turboquant_hybrid_sizing.py`, all failing before their respective fixes; `ruff` + `mypy` clean.

vllm_metal/platform.py
```diff
@@ -749,6 +749,139 @@ def update_block_size_for_backend(cls, vllm_config: "VllmConfig") -> None:
         # ``block_size`` at request time.
         super().update_block_size_for_backend(vllm_config)
 
+        cls._realign_hybrid_block_size_for_turboquant(vllm_config)
```
LxYuan0420: This runs too late to know what the user actually asked for. `super()` has already rewritten both `block_size` and `mamba_block_size`, while `user_specified_mamba_block_size` still remains true, so this pass can treat an upstream-mutated value as the user value. Please capture the original user-specified block settings before `super()` and add end-to-end coverage through `update_block_size_for_backend` with a real `CacheConfig` (`--block-size 64` / hash block 64, and `mamba_cache_mode="all"` with user Mamba block 128). The current isolated `SimpleNamespace` test masks this bug.

vllm_metal/v1/cache_policy.py
```diff
@@ -77,21 +77,27 @@ def real_page_size_bytes(self) -> int:
...
             if not isinstance(spec, TurboQuantAttentionSpec):
-                raise TypeError(
+                raise AssertionError(
                     "All attention layers in the same KV cache group must be "
                     "TurboQuantAttentionSpec."
                 )
             turbo_specs.append(spec)
         if not turbo_specs:
-            raise ValueError("TurboQuantAttentionSpec.merge() requires specs")
+            raise AssertionError("TurboQuantAttentionSpec.merge() requires specs")
```
LxYuan0420: I don’t think these should all become `AssertionError`. That exception is only needed for vLLM’s mixed-type uniformity probe when a non-`TurboQuantAttentionSpec` appears. Empty input and mixed k/v TurboQuant settings are bad configs, so they should stay `ValueError` and fail loudly instead of being swallowed by the fallback path.

vllm_metal/platform.py
```diff
@@ -749,6 +749,139 @@ def update_block_size_for_backend(cls, vllm_config: "VllmConfig") -> None:
...
+        metal_config = get_config()
+        model_config = vllm_config.model_config
+        cache_config = vllm_config.cache_config
+
+        # Gate on additional_config rather than only the process-local
+        # singleton: executors invoke this hook in worker processes too,
+        # where the singleton may not have been populated yet, while
+        # ``additional_config`` travels with the pickled ``vllm_config``.
+        # Defaults mirror ``check_and_update_config``.
+        add = getattr(vllm_config, "additional_config", None) or {}
```
LxYuan0420: This fallback is doing a second parse of TurboQuant settings for a state real executors should not reach. By this hook, `additional_config` should already be resolved into the Metal config. Let’s consume the resolved config here and drop the fresh-singleton fallback and test.

vllm_metal/v1/cache_policy.py
```diff
@@ -410,13 +416,16 @@ def estimate_one_sequence_kv_bytes(
...
-            return num_kv_layers * _turboquant_page_size_bytes(
+            tq_bytes = num_kv_layers * _turboquant_page_size_bytes(
                 block_size=aligned_tokens,
                 num_kv_heads=self._runner.num_kv_heads,
                 head_dim=self._runner.head_dim,
                 k_quant=config.k_quant,
                 v_quant=config.v_quant,
             )
+            if self._runner.is_hybrid:
+                return tq_bytes + self.linear_cache_bytes_per_slot()
```
LxYuan0420: This looks like test-only behavior. TurboQuant requires paged attention, and paged attention does not use this one-sequence estimator for capacity. Please remove this branch and its direct-call test, then cover the public paged sizing path instead.

abcgco:
Applied all four in `f4cd7c3`.

1. **Ordering:** `update_block_size_for_backend` now snapshots the user's `mamba_block_size` before `super()` and threads it into the realign, so the `mamba_cache_mode="all"` branch uses the user's value rather than the one upstream `_align_hybrid_block_size` overwrites (with `user_specified_mamba_block_size` still True). Swapped the isolated stub for a real-`CacheConfig` regression test (`--block-size 64`, `mamba_cache_mode="all"`, user mamba block 128) that fails if the snapshot moves after `super()`.
2. **AssertionError:** kept only the non-`TurboQuantAttentionSpec` mismatch as `AssertionError` (the uniformity probe's "not uniform" signal); empty-input and mixed-quant are back to `ValueError` so genuine misconfigs fail loudly instead of being swallowed by the probe.
3. **additional_config fallback:** dropped — the resolved singleton is authoritative here. `check_and_update_config` populates it on the driver, and `MetalWorker.__init__` re-applies TurboQuant from `additional_config` in each worker before this hook runs, so gating on the singleton is correct on both the uni and mp/PP paths. Removed the gating test too.
4. **is_hybrid estimator branch:** removed, with its direct-call test. TurboQuant forces paged attention, whose capacity is reported through the paged planner (`num_blocks × per-block bytes`), so `estimate_one_sequence_kv_bytes` is never reached for hybrid+TQ — the branch was test-only.

Affected + adjacent unit tests pass locally (`test_turboquant_hybrid_sizing`, `test_v1_worker`, `test_platform*`, `test_per_layer_kv_cache`); ruff clean.

One thing to surface, not for this PR: upstream `_align_hybrid_block_size` already has a `cache_dtype.startswith("turboquant_")` branch, so this override can look like a duplicate. It isn't reachable for us — vllm-metal signals TurboQuant through the Metal config rather than `cache_config.cache_dtype`, and the packed page here is asymmetric (`q8_0` K vs `q3_0` V, plus MLX scale-group overhead) versus upstream's symmetric `TQFullAttentionSpec`. Happy to look at routing through `cache_dtype` in a follow-up if you'd rather upstream own the hybrid alignment.

LxYuan0420 (approved):
Approved after the cleanup pass. The branch now preserves user block/hash constraints, rejects unsupported hybrid state-cache modes, and passes the focused tests plus real TurboQuant serve smoke.

arnaudlb:
Thank you for the fix. I will test and confirm

Issue #465, opened 2026-07-02
[RFC] Redundant KV-cache reads during speculative-decode verification


WindChimeRan:
## Problem

Speculative-decode verification checks K drafted tokens against the target model in one forward pass over a K+1-token window. On Metal, that window currently gets split into K+1 separate single-token attention calls, so each one re-reads the full KV cache independently — the context gets streamed K+1 times instead of once. Normal one-token decoding is unaffected (one token, one read, already optimal); this is a decoder-kernel limitation specific to the multi-token verification shape, not a spec-decoding bug.

Measured in isolation: a K=3-5 verification step's attention costs ~2-2.7x a single decode step's, purely from this redundancy.

## Prior attempt

I built one kernel that tiles a verification window's queries together per KV-head so each block is read once and shared across the window instead of once per token, replacing both existing shape-routed kernels. It fixed the redundancy in isolation, but the measured end-to-end serving speedup couldn't be distinguished from run-to-run benchmark noise, so I shelved it rather than ship the added complexity for an unproven win.

:(  I spent 500$ claude code fable-5 on discussion & design & experiment monitor

## What a fix needs to clear

1. No new (third) attention kernel.
2. Bitwise-correct under greedy speculative decoding.
3. A real end-to-end serving speedup, clearly outside run-to-run noise.
4. No regression to plain (non-speculative) decoding.

Ideas or a different angle on the dispatch/kernel design welcome.

ricky-chaoju:
Did some digging. The K+1 split is actually made in Python, not the kernel: `prepare_unified` emits one `cu_seqlens` span per verification token, so the dispatcher sees pure decode and re-reads KV per row. Both kernels already handle multi-token segments.

I tried the fix as a window mode in the existing decode kernel behind a function constant (`0` = today's codegen untouched): a threadgroup takes 2 rows of a window and shares the K/V loads, bigger windows become several 2-row threadgroups, split-KV still applies. 2 rows per threadgroup is the register cliff; 4+ collapses to ~2x per row. Only pure-verify batches route there, prefill stays tiled. Output is bitwise identical to the current path (checked fp16/bf16/fp32/SWA/TurboQuant), and SD goldens pass with 100% acceptance.

[...]

PR #473, opened 2026-07-04
Add isolation benchmark for speculative-verification attention (RFC #465)

abcgco:
Companion tool for the measurements posted in #465 ([analysis](https://github.com/vllm-project/vllm-metal/issues/465#issuecomment-4883407900), [e2e follow-up](https://github.com/vllm-project/vllm-metal/issues/465#issuecomment-4883604329)), so anyone iterating on the verification-window shape can reproduce them or evaluate a candidate kernel with one command.

What it measures:

- the K+1-token verification window under both segmentations: the current per-token split (`prepare_unified` emits one `cu_seqlens` span per token) vs one fused (K+1)-token segment, which loses split-KV partitioning at the `pure_decode` gate in `paged_ops.cpp`;
- both variants parity-checked against `ref_paged_attn` (flagged loudly on failure);
- the single-token decode floor, i.e. the lower bound for an ideal fused-window kernel that reads each KV block once.

Timings are medians over 5 runs with min/max spread printed alongside — sub-millisecond cells at short prefixes flip ordering between runs (observed ±25% on an M4 Pro), and the spread column makes that visible instead of letting a single-shot number mislead.

Sample output on an M4 Pro (16-core GPU):

```
 prefix  K+1      split µs (min..max)      fused µs (min..max) fused/split   ref |Δ|
   8192    4     1218.9 (1012.5..1846.2)     2530.2 (2125.2..4128.1)       2.08x    0.0001
  32768    4     2358.5 (2235.0..2489.4)     8007.7 (7898.0..8361.6)       3.40x    0.0000
  32768    6     2907.2 (2675.7..2999.8)     8123.3 (8072.2..8334.7)       2.79x    0.0000

 prefix   1-token decode µs (min..max)   (ideal fused-window floor)
  32768           1298.6 (1203.0..1586.1)
```

Testing: ran end to end on M4 Pro (all parity checks pass, table above); `ruff check` and `ruff format --check` clean. No production code touched — the tool only adds a file under `tools/`.

### DIFF
--- a/tools/spec_verify_window_bench.py
+++ b/tools/spec_verify_window_bench.py
@@ -1,0 +1,173 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Isolation benchmark for speculative-verification attention (RFC #465).
+
+A speculative-decode verification step runs attention for a K+1-token query
+window over a long cached prefix. ``prepare_unified`` currently emits that
+window as K+1 single-token ``cu_seqlens`` segments, so the kernel re-reads
+the prefix once per token; the obvious alternative is one (K+1)-token
+segment, which however routes past the ``pure_decode`` gate in
+``paged_ops.cpp`` and loses split-KV partitioning. This tool measures both
+segmentations against each other and against the single-token decode floor,
+so kernel work on the verification shape can be judged with numbers:
+
+  A. split (current) vs fused (one segment) per prefix length and window,
+     with both variants parity-checked against ``ref_paged_attn``.
+  B. Single-token decode on the same prefixes — the lower bound for an
+     ideal fused-window kernel that reads each KV block once.
+
+Timings are medians over REPEATS runs with min/max spread printed
+alongside: sub-millisecond cells at short prefixes flip ordering between
+runs (observed +-25% on an M4 Pro), so single-shot readings below ~4k
+prefix are not meaningful.
+
+Run with:
+    uv run python tools/spec_verify_window_bench.py
+"""
+
+from __future__ import annotations
+
+import time
+
+import mlx.core as mx
+import numpy as np
+
+from tools.attention_bench_utils import ref_paged_attn
+from vllm_metal.metal import get_ops
+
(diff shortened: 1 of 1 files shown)


LxYuan0420 (commented):
It seems like this PR measures current split decode vs the naive one-segment reroute,
which we already know regresses because it drops split-KV partitioning. Are we sure this moves #465 forward, rather than adding a baseline harness for a path we do not plan to ship?

I would like a second opinion from @ricky-chaoju and @WindChimeRan before we move this forward.

abcgco:
Fair question. The fused variant is indeed the non-shipping path — its value on main is as the measured cliff behind the `pure_decode` gate: the prepare_unified-only reroute looks strictly better on paper and is 2.5-4x worse in practice, and this is the one-command demonstration of that (plus `ref_paged_attn` parity for both segmentations and the 1-token floor any window kernel gets judged against). That temptation will outlive this thread — it was #465's first proposed fix.

That said, once the window kernel lands, the A/B that matters on main becomes split vs the real window mode, and ricky's `spec_window_bench.py` already does that better (drives the actual kernel, covers concurrency). So two clean options, no attachment on my side:

1. park this until `feat/spec-verify-window-kernel` merges, then I rebase it so the second variant drives the shipped window mode, keeping the floor + reference-parity + median-with-spread pieces; or
2. close this, and I fold those pieces (1-token floor, `ref_paged_attn` parity, median/spread reporting) into ricky's bench as a follow-up on his branch.

Happy either way — whichever @ricky-chaoju and @WindChimeRan prefer.

WindChimeRan:
Prefer not to merge this PR. But It's good to keep it here (open or close, not merged), for discussion and repro. 

Currently I'm still not very clear on how to solve #465 or maybe it's just a non-issue.

LxYuan0420:
Closing this for now since it doesn’t move #465 forward from here. We can reopen it once there is a clearer next step.

PR #474, opened 2026-07-05
Route the Metal (MLX) KV cache through LMCache

lokic233:
## Summary

Enables **LMCache KV caching on the Apple-Silicon Metal/MLX backend**, end to end in the serving loop, with **bit-identical output** and a **measured TTFT win**. Companion PR on the LMCache side adds the host-memory (CPU) connector.

> Opening as a **draft** for design feedback — the LMCache-side companion PR is required for this to run.

## Why this is non-trivial

vllm-metal keeps its paged KV cache as **MLX arrays in unified memory** (`MetalPagedKVCache`) and does **not** drive vLLM's standard v1 `KVConnector` hooks, so LMCache's stock connector (which binds vLLM's torch paged tensors) has nothing to attach to. Three facts make the integration work:

1. **Unified memory makes MLX↔torch zero-copy AND bidirectional** (`tensor_bridge`) — writes to the bridged torch tensor land back in the MLX cache.
2. **LMCache's CPU path is fully functional without CUDA** — `lmcache.c_ops` falls back to the vectorized `python_ops_fallback.multi_layer_kv_transfer`.
3. **The correct control point is the scheduler-side KVConnector** — the vLLM core scheduler folds externally-cached tokens into `num_computed_tokens`, so the worker computes only the suffix. (A worker-only prefix skip desynchronizes the scheduler and corrupts output — we tried it, it's wrong, and this PR does it the right way.)

## What's in this PR

| File | Purpose |
|---|---|
| `vllm_metal/lmcache_kv_connector.py` | `MetalLMCacheKVConnector` — full vLLM v1 KVConnector (scheduler + worker roles) |
| `vllm_metal/lmcache_integration.py` | `MetalLMCacheConnector` — bridges the live `MetalPagedKVCache` to an LMCache engine, touching only the request's own blocks (`mx.take`/scatter) |
| `vllm_metal/v1/model_runner.py`, `worker.py` | bind the MLX cache, drive `start_load_kv` (before forward) / `wait_for_save` (after), `ensure_kv_transfer_initialized` |
| `vllm_metal/bootstrap.py` | `bootstrap_metal_platform()` — fixes a silent platform-cache race where importing vLLM cached `CpuPlatform` before the Metal plugin resolved (this had been making "Metal" runs silently execute on CPU torch) |
| `vllm_metal/compat.py` | tolerate `AutoTokenizer.register()` string keys so `mlx_lm` imports under transformers ≥5.x |
| `docs/`, `examples/` | dev-env recipe, benchmark results + figures, runnable roundtrip / correctness / TTFT harnesses |

## Results (Qwen3-0.6B, M4 Pro, MLX bf16)

**Correctness:** output **bit-identical** across cold / metal-native APC / LMCache.

**Latency (TTFT) — Metal-native APC vs LMCache vs cold recompute:**

| Prompt (tokens) | Cold | Metal-native APC | LMCache | APC speedup | LMCache speedup |
|---|---|---|---|---|---|
| 638 | 224 ms | 46 ms | 78 ms | 5× | 3× |
| 1284 | 446 ms | 64 ms | 144 ms | 7× | 3× |
| 2559 | 976 ms | 88 ms | 212 ms | 11× | 5× |
| 5124 | 2707 ms | 149 ms | 338 ms | 18× | **8×** |

Decode throughput is ~equal across all three (caching is a prefill win).

## Honest positioning

Metal-native APC is **faster than LMCache when both apply** (in-process, in-memory, un-evicted). LMCache's unique value is **persistence, cross-process/node sharing, and offload** — the LMCache column above is measured with APC reset, i.e. exactly the case where APC has nothing and the alternative is a full cold recompute. They're complementary: APC = fast L1 (in-GPU); LMCache = L2 (persistent/shared/offloaded).

## Notes for reviewers

- Requires the companion LMCache change (host-memory CPU connector). Link to follow.
- The prebuilt `_paged_ops` native artifacts must match the current vLLM-metal sources — rebuild native artifacts when running against latest `main`.
- Enable via `KVTransferConfig(kv_connector="MetalLMCacheKVConnector", kv_connector_module_path="vllm_metal.lmcache_kv_connector", kv_role="kv_both")` + `VLLM_METAL_ENABLE_LMCACHE=1`.

🤖 Authored by Navi on behalf of @lokic233.


lokic233:
Companion LMCache PR (required): LMCache/LMCache#4003

WindChimeRan (commented):
Could you please explain the motivation of the design? 

The comparison with APC is not very meaningful to me. 

This PR & design seems to be, offloading KV cache to CPU, which is in our case, already UMA? Why should we do that?

LxYuan0420:
Closing this: it puts LMCache-specific lifecycle and policy into vllm-metal before the upstream contract is established. feel free to reopen after upstream part is ready

PR #475, opened 2026-07-05
[Distributed] Derive PP wire descriptor without probing the embedding

ricky-chaoju:
Our pipeline-parallel forward wrapper (`PipelinedModel`) works out the shape and dtype of the hidden state that crosses between stages by calling `model.model.embed_tokens` on every stage. That only holds today because Phase-0 pipeline parallelism (#427) loads the full model on each Mac and then slices, so every stage happens to own the embedding. It is the wrong thing to read for where this is heading. The next PR streams each stage's own layers off disk, and a non-first stage will not hold `embed_tokens` at all, so the probe would break.

This PR derives the wire descriptor from what every stage genuinely owns. The width comes from `model.args.hidden_size` on the built model, and the dtype comes from the first floating-point parameter of the stage's first owned layer. A packed quantized weight is `uint32` and gets skipped, while the norm weights and quant scales carry the real compute dtype. I also turned the probe's bare `AttributeError` into source-aware fail-fasts, and made the split fail loud when there are more stages than layers instead of tripping an `IndexError` later when the descriptor reads `layers[0]`.

There is no intended change to the currently validated Qwen3 load-then-slice path. I re-ran the 2-stage ring parity harness on Qwen3-0.6B and it is still bit-exact.

### DIFF
--- a/tests/test_pp_pipeline.py
+++ b/tests/test_pp_pipeline.py
@@ -198,14 +198,11 @@ class TestPipelinedModel:
     def test_singleton_runs_full_model_without_recv(self) -> None:
         # size-1 group: is_first and is_last are both true, so the wrapper runs
         # the full model (no recv) and passes input_embeddings=None — the model
-        # embeds the tokens itself.
+        # embeds the tokens itself. The wire descriptor is never read for a
+        # singleton, so the stub needs nothing beyond __call__.
         class _Stub:
             def __init__(self) -> None:
-                self.args = SimpleNamespace(hidden_size=32)
                 self.input_embeddings: object = "unset"
-                self.model = SimpleNamespace(
-                    embed_tokens=nn.Embedding(4, 32), layers=[_StubLayer(32)]
-                )
 
             def __call__(self, input_ids, *, cache=None, input_embeddings=None):
                 self.input_embeddings = input_embeddings
@@ -235,28 +232,3 @@ class TestPipelinedModel:
         assert wrapper._wire_hidden == 32
         assert wrapper._wire_dtype == layer.q_proj.scales.dtype
         assert wrapper._wire_dtype != layer.input_layernorm.weight.dtype
-
-    def test_wire_descriptor_rejects_non_floating_stage(self) -> None:
-        # A stage whose first layer exposes no floating compute parameter cannot
-        # declare a valid wire dtype — fail loud rather than deadlock the ring.
-        class _IntLayer(nn.Module):
-            def __init__(self) -> None:
-                super().__init__()
-                self.w = mx.zeros((4, 4), dtype=mx.int32)
-
-        model = SimpleNamespace(
-            args=SimpleNamespace(hidden_size=4),
-            model=SimpleNamespace(layers=[_IntLayer()]),
-        )
-        with pytest.raises(TypeError, match="no floating-point parameter to derive"):
-            PipelinedModel(model, _pp(1, 2))
-
-    def test_wire_descriptor_rejects_missing_hidden_size(self) -> None:
-        # No model.args.hidden_size -> source-aware fail-fast, not a bare
-        # AttributeError.
-        model = SimpleNamespace(
-            args=SimpleNamespace(),
-            model=SimpleNamespace(layers=[_StubLayer(32)]),
-        )
-        with pytest.raises(TypeError, match="model.args.hidden_size as a positive"):
-            PipelinedModel(model, _pp(1, 2))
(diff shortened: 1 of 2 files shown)


vllm_metal/distributed/pipeline.py
```diff
@@ -233,17 +241,52 @@ class PipelinedModel:
     forward evals.
     """
 
+    # Floating-point dtypes a stage's activation ("wire") may cross the ring with;
+    # a packed quantized weight is uint32 and never the wire dtype.
+    _FLOATING_DTYPES = (mx.float16, mx.bfloat16, mx.float32)
```
LxYuan0420: I do not think we need our own floating dtype tuple here. It is easy for this list to drift if MLX adds or starts exposing another floating dtype, and the intent is really “any floating parameter,” not “one of these three names.” `mx.issubdtype(param.dtype, mx.floating)` expresses that directly and stays aligned with MLX.

vllm_metal/distributed/pipeline.py
```diff
@@ -233,17 +241,52 @@ class PipelinedModel:
...
+        """The activation width, from ``model.args.hidden_size``. Fail loud with a
+        source-aware message rather than a bare ``AttributeError`` if a model lacks
+        a positive integer hidden size."""
+        hidden = getattr(getattr(model, "args", None), "hidden_size", None)
+        if not isinstance(hidden, int) or hidden <= 0:
+            raise TypeError(
+                "PP wire descriptor needs model.args.hidden_size as a positive "
+                f"integer, got {hidden!r}."
+            )
+        return hidden
```
LxYuan0420: This helper feels unnecessary. `model.args.hidden_size` is the mlx_lm model contract here, so I would inline it and delete the synthetic missing-hidden-size test.

vllm_metal/distributed/pipeline.py
```diff
@@ -233,17 +241,52 @@ class PipelinedModel:
...
+        # Annotate the tree_flatten result: it is typed list|dict, and iterating the
+        # dict branch would yield str keys — pin the (path, array) list for mypy.
+        flat: list[tuple[str, Any]] = tree_flatten(first_layer.parameters())
+        for _path, param in flat:
+            if param.dtype in cls._FLOATING_DTYPES:
+                return param.dtype
+        raise TypeError(
+            "PP stage's first layer exposes no floating-point parameter to derive "
+            "the wire dtype from."
+        )
```
LxYuan0420: This can be simpler: `next(param.dtype for _, param in tree_flatten(first_layer.parameters()) if mx.issubdtype(param.dtype, mx.floating))`. The no-floating branch feels like defensive coverage for a state a real transformer layer should not reach.

tests/test_pp_pipeline.py
```diff
@@ -191,16 +216,47 @@ def __call__(self, input_ids, *, cache=None, input_embeddings=None):
...
+            def __init__(self) -> None:
+                super().__init__()
+                self.w = mx.zeros((4, 4), dtype=mx.int32)
+
+        model = SimpleNamespace(
+            args=SimpleNamespace(hidden_size=4),
+            model=SimpleNamespace(layers=[_IntLayer()]),
+        )
+        with pytest.raises(TypeError, match="no floating-point parameter to derive"):
+            PipelinedModel(model, _pp(1, 2))
```
LxYuan0420: I would delete this test. It only covers the defensive no-floating branch with an impossible transformer layer.

tests/test_pp_pipeline.py
```diff
@@ -191,16 +216,47 @@ def __call__(self, input_ids, *, cache=None, input_embeddings=None):
...
+
+    def test_wire_descriptor_rejects_missing_hidden_size(self) -> None:
+        # No model.args.hidden_size -> source-aware fail-fast, not a bare
+        # AttributeError.
+        model = SimpleNamespace(
+            args=SimpleNamespace(),
+            model=SimpleNamespace(layers=[_StubLayer(32)]),
+        )
+        with pytest.raises(TypeError, match="model.args.hidden_size as a positive"):
+            PipelinedModel(model, _pp(1, 2))
```
LxYuan0420: This can go with the helper. Missing `model.args.hidden_size` is a stale internal contract bug, not user input to validate here.

tests/test_pp_pipeline.py
```diff
@@ -191,16 +216,47 @@ def __call__(self, input_ids, *, cache=None, input_embeddings=None):
         assert out == "logits"
         assert stub.input_embeddings is None
```
LxYuan0420, line 214: This singleton test now needs a heavier scaffold because `PipelinedModel.__init__` computes an unused wire descriptor for `pp.size == 1`. Optional, but worth considering.

ricky-chaoju:
All done. Thanks~

LxYuan0420 (approved):

PR #478, opened 2026-07-05
Fix Whisper segment seek offsets after chunking

### LINKED ISSUE #477: Whisper segment seek offsets reset for each long-audio chunk
## Description

Whisper transcription converts chunk-relative timestamp tokens into global `start` and `end` values by adding `time_offset`, but calculates `seek` from the local timestamp alone.

As a result, verbose timestamp output for every chunk after the first contains a `seek` position relative to that chunk rather than to the original recording. This makes `seek` inconsistent with `start` and `end`, and later chunks can report seek positions near zero again.

## Reproduction

The segment extraction path demonstrates the inconsistency directly:

```python
segments = transcriber._extract_segments(
    [timestamp_0s, text_token, timestamp_3s],
    time_offset=10.0,
)

segment = segments[0]
```

Current values:

```python
segment.seek == 0
segment.start == 10.0
segment.end == 13.0
```

## Expected behavior

`seek` should identify the position in the original audio, consistent with Whisper segment semantics and the global `start`/`end` fields:

```python
segment.seek == 1000
segment.start == 10.0
segment.end == 13.0
```

`SEEK_MULTIPLIER` is 100, so `(local timestamp + chunk offset) * 100` produces the global seek frame.

## Impact

Consumers of verbose transcription output can receive repeated or non-monotonic seek positions for valid recordings that use the default long-audio chunking path. This can break segment navigation or downstream alignment even though the displayed start/end timestamps look correct.

## Why this fix is correct

- Whisper timestamp tokens are local to each decoded chunk.
- `time_offset` is the chunk start in the original recording.
- `start` and `end` already add `time_offset`; `seek` must use the same coordinate system.
- `SEEK_MULTIPLIER=100` converts seconds to Whisper seek units of 10 ms.
- Therefore the correct calculation is `int((seg_start + time_offset) * SEEK_MULTIPLIER)`.

The implementation changes only the two segment-construction paths (paired timestamps and an unfinished final segment). It does not change audio chunking, token decoding, or public APIs. Regression tests cover both paths with a nonzero chunk offset.

## Proposed fix

Include `time_offset` when calculating `seek` in both paired-timestamp and unfinished-segment construction paths, with regression coverage for both.

## Reference implementation

[OpenAI Whisper initializes `seek` from the global clip position and writes that value directly into each segment](https://github.com/openai/whisper/blob/04f449b8a437f1bbd3dba5c9f826aca972e7709a/whisper/transcribe.py#L226-L253). It also derives the chunk `time_offset` from that same global seek position. vLLM Metal decodes chunks separately, so adding `time_offset` to the chunk-local timestamp restores the same original-audio coordinate system before converting seconds to 10 ms seek units.

harivilasp:
## Summary

Make Whisper segment `seek` positions global for long-audio transcription, consistent with the existing global `start` and `end` timestamps.

## Details

- Include each chunk `time_offset` when calculating `TranscriptionSegment.seek`.
- Correct both paired-timestamp and unfinished-segment construction paths.
- Extend segment extraction tests to cover nonzero chunk offsets in both paths.

Before this change, a segment spanning local `0s` to `3s` in a chunk starting at `10s` reported `seek=0`, `start=10`, and `end=13`. It now reports `seek=1000`, `start=10`, and `end=13`.

## Test Plan

- Isolated segment-extraction harness on Python 3.12: verified `(seek, start, end) == (1000, 10.0, 13.0)`.
- `uvx ruff format --check vllm_metal/stt/whisper/transcriber.py tests/test_transcribe.py`
- `uvx ruff check vllm_metal/stt/whisper/transcriber.py tests/test_transcribe.py`
- `python3.12 -m py_compile vllm_metal/stt/whisper/transcriber.py tests/test_transcribe.py`

The complete pytest suite was not run locally because the documented macOS setup requires compiling vLLM from source and that build did not complete during local verification. CI should exercise the regression in the supported integration environment.

## Related Issue

Resolves #477

## AI disclosure

Code and tests were generated with Codex GPT-5. The commit includes `Generated-by: Codex GPT-5` and the required DCO sign-off.

## Reference implementation

[OpenAI Whisper initializes `seek` from the global clip position and writes that value directly into each segment](https://github.com/openai/whisper/blob/04f449b8a437f1bbd3dba5c9f826aca972e7709a/whisper/transcribe.py#L226-L253). It also derives the chunk `time_offset` from that same global seek position. vLLM Metal decodes chunks separately, so adding `time_offset` to the chunk-local timestamp restores the same original-audio coordinate system before converting seconds to 10 ms seek units.

### DIFF
--- a/tests/test_transcribe.py
+++ b/tests/test_transcribe.py
@@ -101,8 +101,22 @@ class TestExtractSegments:
         segments = transcriber._extract_segments(tokens, time_offset=10.0)
 
         assert len(segments) == 1
+        assert segments[0].seek == 1000
         assert segments[0].start == pytest.approx(10.0)
         assert segments[0].end == pytest.approx(13.0)
+
+    def test_time_offset_applies_to_unfinished_segment_seek(
+        self, transcriber: WhisperTranscriber
+    ) -> None:
+        start_tok = transcriber._get_token_id("<|0.00|>")
+
+        segments = transcriber._extract_segments(
+            [start_tok, 2425],
+            time_offset=10.0,
+        )
+
+        assert len(segments) == 1
+        assert segments[0].seek == 1000
 
     def test_empty_tokens(self, transcriber: WhisperTranscriber) -> None:
         assert transcriber._extract_segments([]) == []
--- a/vllm_metal/stt/whisper/transcriber.py
+++ b/vllm_metal/stt/whisper/transcriber.py
@@ -335,7 +335,7 @@ class WhisperTranscriber:
                         segments.append(
                             TranscriptionSegment(
                                 id=seg_id,
-                                seek=int(seg_start * SEEK_MULTIPLIER),
+                                seek=int((seg_start + time_offset) * SEEK_MULTIPLIER),
                                 start=round(seg_start + time_offset, 2),
                                 end=round(ts + time_offset, 2),
                                 text=seg_text,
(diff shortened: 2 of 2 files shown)


LxYuan0420 (approved):
Thanks for fixing it!

PR #480, opened 2026-07-07
[GGUF] Add llama GGUF loading with q/k RoPE un-permutation

ricky-chaoju:
This adds the llama architecture to the local GGUF loader (#415; qwen2/qwen3 shipped in #466). A llama `.gguf` was rejected at the arch allowlist before; now it loads and serves.

The load-bearing part is a real correctness bug I only caught with golden parity. llama.cpp stores `q_proj` / `k_proj` in its own RoPE-permuted row layout, while mlx builds the HF layout, so a llama `.gguf` loads cleanly but produces broken attention. Per-op forward cosine against the mlx_lm dense reference was `mlp.gate_proj` 0.99998 and `embed` 0.99997, but `q_proj` 0.014 and `k_proj` 0.042 (orthogonal). qwen is unaffected because llama.cpp does not permute its q/k. The inverse is a pure reorder of the output rows, and applying it restored q and k to cosine 1.0000, including GQA.

I kept the ownership where the merged code already puts it. The row reorder is a `permute_rows` method on `GGUFMLXQuantizedTensor`; it owns its packed `qweight` / `scales` / `biases`, so the three gather together and the weight stays quantized with no dense copy. Which arch's q/k need it, and the precomputed index, are adapter-owned: a `permute_index` table built in `from_model` next to the translation map, read off the built q/k projections so an omitted `head_dim` can't bite. The loader consults the adapter and applies it in both the quantized and plain branches, and to the q/k bias, with no arch conditional. Adding llama also makes Mixtral/MoE-llama files (`architecture=llama`) reachable, so `ffn_gate_inp` joins the out-of-scope preflight next to `_exps`.

### DIFF
--- a/tests/test_gguf_loader.py
+++ b/tests/test_gguf_loader.py
@@ -251,6 +251,257 @@ def test_untied_qwen2_attaches_bias_and_installs_lm_head(tmp_path):
     assert out.shape[-1] == 256
 
 
+def _write_minimal_tokenizer(config_dir: str, vocab_size: int) -> None:
+    """Write a tiny WordLevel fast tokenizer into a synthetic config dir.
+
+    transformers builds a llama tokenizer through ``LlamaTokenizer``, which
+    needs real tokenizer files (unlike qwen, which constructs from config
+    alone), so the llama fixtures must ship a tokenizer to reach ``load()``.
+    """
+
+    def _special(i: int, content: str) -> dict:
+        return {
+            "id": i,
+            "content": content,
+            "special": True,
+            "single_word": False,
+            "lstrip": False,
+            "rstrip": False,
+            "normalized": False,
+        }
+
+    vocab = {"<unk>": 0, "<s>": 1, "</s>": 2}
+    vocab.update({f"t{i}": i for i in range(3, vocab_size)})
+    tokenizer = {
+        "version": "1.0",
+        "truncation": None,
+        "padding": None,
+        "added_tokens": [_special(0, "<unk>"), _special(1, "<s>"), _special(2, "</s>")],
+        "normalizer": None,
+        "pre_tokenizer": {"type": "Whitespace"},
+        "post_processor": None,
+        "decoder": None,
+        "model": {"type": "WordLevel", "vocab": vocab, "unk_token": "<unk>"},
+    }
+    dir_path = Path(config_dir)
+    (dir_path / "tokenizer.json").write_text(json.dumps(tokenizer))
+    (dir_path / "tokenizer_config.json").write_text(
+        json.dumps({"tokenizer_class": "PreTrainedTokenizerFast"})
+    )
+
+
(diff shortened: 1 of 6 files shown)


vllm_metal/gguf/loader.py
```diff
@@ -199,18 +199,27 @@ def _partition(
                     f"{module_path!r} is absent from the model."
                 ) from exc
 
+            # llama.cpp permutes q/k rows for its RoPE layout regardless of qtype;
+            # the adapter returns the row index to undo it (None for every other
+            # tensor / arch), applied identically to the quantized or plain weight.
+            permute = adapter.rope_permute_index(name)
+
             if suffix == _BIAS_SUFFIX:
```
LxYuan0420, line 209: We should validate the raw GGUF tensor before applying the llama q/k row permutation. GGUF is a file boundary, and `array[permute]` can hide malformed row counts by producing an expected-shaped result before `_validate_plain_shape` runs. Please switch the order to: read source tensor → validate source shape → then permute into model layout.

```python
if suffix == _BIAS_SUFFIX:
    bias = arrays[name]
    self._validate_plain_shape(model, translated, bias, name)
    biases[module_path] = bias[permute] if permute is not None else bias
elif tensor.tensor_type in MLX_NATIVE_GGUF_TYPES:
    qt = GGUFMLXQuantizedTensor.from_mx_load(arrays, name, tensor.tensor_type)
    self._validate_quant_shape(module_path, current, qt, name)
    quant[module_path] = qt.permute_rows(permute) if permute is not None else qt
else:
    weight = arrays[name]
    self._validate_plain_shape(model, translated, weight, name)
    plain[translated] = weight[permute] if permute is not None else weight
```

tests/test_gguf_loader.py
```diff
@@ -251,6 +251,257 @@ def test_untied_qwen2_attaches_bias_and_installs_lm_head(tmp_path):
...
+        "model": {"type": "WordLevel", "vocab": vocab, "unk_token": "<unk>"},
+    }
+    dir_path = Path(config_dir)
+    (dir_path / "tokenizer.json").write_text(json.dumps(tokenizer))
+    (dir_path / "tokenizer_config.json").write_text(
+        json.dumps({"tokenizer_class": "PreTrainedTokenizerFast"})
+    )
+
+
+@pytest.mark.parametrize("quant_type", [QT.Q8_0, QT.Q4_0])
```
LxYuan0420, line 294: This proves wrappers install, but not the main quantized llama q/k behavior. Please add one fast AAA test that loads a synthetic Q8_0/Q4_0 llama GGUF, then asserts the installed q/k `GGUFLinear` tensors match `raw_qt.permute_rows(_rope_inv_index(...))` and do not match the unpermuted raw tensor.

LxYuan0420 (approved):

PR #447, opened 2026-06-17
[Spec] DraftModel Speculative decoding

WindChimeRan:
## Correctness
- use same model for both draft and verification, got 100% acceptance rate on 1 request, 94% acceptance rate on batch request

## Limitation
- no async scheduling
- Greedy requests only
- Draft KV cache is a second, un-budgeted store
- SWA draft model may have bugs. (there are some ongoing bug fixing in the upstream vLLM. But I think user would rarely use gemma-4-e2b as draft model. )
- **[Next PR]: Verification cost on memory bandwidth: this is a pre-existing kernel limitation, now single forward on K+1 verification requires K KV cache reading.** 

## Benchmark

**Setup.** Apple M5 Pro (64 GB), Qwen3-8B target + Qwen3-0.6B draft, paged path, greedy (`temperature=0`), `--ignore-eos`, output 128, `VLLM_METAL_MEMORY_FRACTION=0.55`, `--max-model-len 2048`.

 Dataset: a **198-prompt balanced sample** (~22 each across the 9subsets) of [`RedHatAI/speculator_benchmarks`](https://huggingface.co/datasets/RedHatAI/speculator_benchmarks)
(HumanEval, qa, rag, summarization, tool_call, translation, math_reasoning, question, writing) — natural prompts, fed via vLLM's `--dataset-name spec_bench`.

> Baseline runs with `--no-async-scheduling` to match SD (which forces sync), so the comparison is at equal scheduling.

<details open>
<summary><b>Single-stream (concurrency 1, 30 prompts)</b></summary>

| Config | Out tok/s | tput× | Mean TPOT (ms) | TPOT× | Draft acceptance |
|---|---:|---:|---:|---:|---:|
| Baseline (no SD) | 16.7 | — | 58.3 | — | — |
| SD K=3 | 22.2 | **1.33×** | 42.8 | **1.36×** | 64% (len 3.0) |
| SD K=5 | 23.7 | **1.42×** | 39.4 | **1.48×** | 52% (len 3.6) |

</details>

<details open>
<summary><b>Batch (concurrency 32, 100 prompts)</b></summary>

| Config | Out tok/s | tput× | Mean TPOT (ms) | TPOT× | Draft acceptance |
|---|---:|---:|---:|---:|---:|
| Baseline (no SD) | 170.7 | — | 143.3 | — | — |
| SD K=3 | 190.3 | **1.11×** | 135.0 | **1.06×** | 64% (len 3.1) |
[...]

PR #481, opened 2026-07-07
[Spec Decode] DFlash

WindChimeRan:
## Benchmark (single-stream, Qwen3-8B, natural-data spec bench, greedy)

`vllm bench serve`, spec_bench dataset, 30 prompts × 128 output tokens, seed 0, concurrency 1, sync scheduling for all rows, prefix cache hit rate 0%. Apple M5 Pro, bf16.

| config | out tok/s | accept len | speedup |
|---|---|---|---|
| baseline (sync) | 16.51 | — | 1.00x |
| **DFlash K=7** | **19.70** | **2.62** | **1.19x** |
| DFlash K=5 / K=3 | 12.96 / 12.65 | ~1.6 | 0.78x |
| draft_model Qwen3-0.6B K=3 (reference) | 21.53 | 2.95 | 1.30x |

## Limitation

same with #447

### DIFF
--- a/docs/speculative_decoding.md
+++ b/docs/speculative_decoding.md
@@ -1,15 +1,15 @@
 # Speculative Decoding
 
-vllm-metal supports three speculative decoding methods on the paged attention
+vllm-metal supports four speculative decoding methods on the paged attention
 path. All require synchronous scheduling and greedy sampling.
 
-| | MTP | Draft Model | N-gram |
-|---|---|---|---|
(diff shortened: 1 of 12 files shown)


anthonyalayo:
I am very curious how the performance of qwen will look like with both this and #584 together! I am unaware of alternative inference servers that have both implemented.

WindChimeRan:
@anthonyalayo  This implementation is too slow ... haven't figured out the root cause. I'd expect dflash far faster than draft_model.

Issue #482, opened 2026-07-07
[Performance]: draft-model spec decode is net-negative: per-request full-prompt re-ingest into the draft KV + ~2x steady-state proposer overhead


abcgco:
**TL;DR** — two stacked costs in the draft-model proposer:

1. Every new request re-prefills its entire prompt into the draft KV on its first propose (4.6 s at 8k) — the draft allocator has no content-based reuse, so even an identical resubmitted prompt pays it again.
2. Steady-state propose costs 24.7 ms/token vs 14.6 ms for a complete nospec engine step (~1.7x, even at 100% acceptance).

Measurements, mechanism, and fix directions below.

---

### Summary

Following up on @KrxGu's note in #465 that the draft-model path runs net-negative (0.6B draft ~3.3x slower than nospec on an M3 Ultra with an 8B target), flagged there as a separate issue — this is that issue, with the mechanism pinned down on an M4 Pro (24 GB). Notably, the "isn't on the MLX fast path" guess turns out not to be it: a dispatch probe shows every draft forward on the expected kernels. The loss is proposer-side, and it is two stacked problems.

### Setup

`draft == target == Qwen/Qwen3-0.6B` (greedy → 100% acceptance). An equal-size draft caps the *ideal* speedup at ~break-even by construction — chosen deliberately so that any measured loss is pure proposer overhead rather than acceptance math (that a realistically-sized 0.6B→8B pair also loses 3.3x suggests the same overheads dominate there too). Warm-decode wall/token over 128 generated tokens, prefix caching on, `num_speculative_tokens=3`, `VLLM_ENABLE_V1_MULTIPROCESSING=0`, one engine per process. M4 Pro 24 GB (16-core GPU), macOS 15.6, vllm-metal @ b19b8be (vLLM v0.24.0).

| prefix | nospec | ngram | draft-model |
|---|---|---|---|
| 2k | 10.7 ms/tok | 5.9 | 22.8 (**2.1x slower**) |
| 8k | 14.6 | 8.2 | 62.4 (**4.3x slower**) |

ngram beating nospec shows the verify side is healthy; the loss is in the proposer.

### Problem 1 — every new request re-prefills the entire prompt into the draft KV

[...]

PR #483, opened 2026-07-08
[Spec Decode] Move hidden-state policy into Gemma4 MTP proposer

samithaj:
## What

This PR narrows the spec-decode cleanup to the standalone Gemma4 MTP change.

- Move target hidden-state policy into `Gemma4MTPProposer`.
- Keep the existing Gemma4 MTP install path as the only MTP proposer path in this PR.
- Add positive install-path coverage for Gemma4 MTP.

The native-MTP registry, build context, and generic runner dispatch from the earlier version have been removed.

## Testing

Validated locally after rebasing on current `main`:

- `ruff check vllm_metal/v1/model_runner.py vllm_metal/v1/proposer.py vllm_metal/v1/spec_decode.py tests/test_gemma4_mtp_proposer.py tests/test_spec_decode_metadata.py tests/test_v1_model_runner_generate.py`
- `ruff format --check vllm_metal/v1/model_runner.py vllm_metal/v1/proposer.py vllm_metal/v1/spec_decode.py tests/test_gemma4_mtp_proposer.py tests/test_spec_decode_metadata.py tests/test_v1_model_runner_generate.py`
- `pytest -q tests/test_gemma4_mtp_proposer.py tests/test_spec_decode_metadata.py tests/test_v1_model_runner_generate.py`

Result: `70 passed`.


vllm_metal/v1/model_runner.py
```diff
@@ -722,6 +727,20 @@ def install_drafter(self, *, num_blocks: int, block_size: int) -> None:
                 vllm_config=self.vllm_config,
                 controller=self._spec_decode_controller,
             )
+        elif spec.method == "mtp":
+            # Route mtp drafts through the native-head registry. It stays empty
+            # until the GLM-4.7-Flash head is wired to a proposer in a later PR,
+            # so find_native_mtp_head returns None here and we fail loud, naming
+            # the draft model_type + the registered native heads.
+            if find_native_mtp_head(spec) is not None:
```
LxYuan0420: I would not add a registered-native-head branch whose only behavior is “proposer not wired yet.” If the registry is empty in this PR, let all non-Gemma4 `mtp` configs fail through `unsupported_mtp_message(spec)`. If the registry lands now, the positive path should be real and delegate to the head/spec to build a `MetalProposer`, so `model_runner` stays thin.

vllm_metal/v1/mtp_heads/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+
+from __future__ import annotations
+
+from collections.abc import Callable, Mapping
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass(frozen=True, slots=True)
+class MTPHeadSpec:
```
LxYuan0420: This should be a class-owned runtime contract, not just metadata plus a loader. Existing speculative paths are class-owned, so I would make the native head/spec own proposer construction, or introduce a small `NativeMTPHead` protocol. Otherwise #485 has to keep adding free functions and runner-side dispatch around an incomplete DTO.

tests/test_v1_model_runner_generate.py
```diff
@@ -1632,3 +1632,33 @@ def test_is_mla_false_for_standard_mha(self) -> None:
...
+            ),
+        )
+
+    def test_unregistered_mtp_draft_raises_naming_model_type(self) -> None:
+        runner = make_stub_runner(tokenizer=object())
+        runner.vllm_config = SimpleNamespace(
+            speculative_config=self._make_mtp_spec(model_type="eagle"),
+        )
+
+        with pytest.raises(NotImplementedError) as excinfo:
```
LxYuan0420: This covers the unregistered-MTP error, but the load-bearing invariant is the positive Gemma4 dispatch: `install_drafter()` with a Gemma4 MTP config installs `Gemma4MTPProposer`. The new hidden-state policy depends on that install-time binding, and existing tests mostly set `_drafter` manually.

tests/test_gemma4_mtp_proposer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_needed_with_decode_segments(self) -> None:
+        assert _proposer().needs_target_hidden_states(
+            [_decode_segment()], has_final_prefill=False
+        )
+
+    def test_needed_with_final_prefill_only(self) -> None:
+        assert _proposer().needs_target_hidden_states([], has_final_prefill=True)
+
+    def test_not_needed_for_intermediate_prefill_only(self) -> None:
```
LxYuan0420: These two negative tests exercise the same branch: no decode segments and no final prefill. Please keep one; the intermediate-prefill meaning can be captured in the test name.

tests/test_mtp_head_registry.py
```diff
@@ -0,0 +1,123 @@
...
+def _mtp_config(*, model_type: str) -> SimpleNamespace:
+    return SimpleNamespace(
+        method="mtp",
+        draft_model_config=SimpleNamespace(
+            hf_config=SimpleNamespace(model_type=model_type),
+        ),
+    )
+
+
+def test_returns_none_for_missing_config() -> None:
```
LxYuan0420: The `returns_none` cases are mostly the same behavior with different synthetic missing fields. Please parametrize them or keep only the meaningful boundary cases. The important coverage is registered match plus unsupported-message clarity.

vllm_metal/v1/mtp_heads/__init__.py
```diff
@@ -0,0 +1,105 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Registry of native MTP head implementations, keyed by draft ``model_type``.
```
LxYuan0420: Please trim the staging narration. One module note explaining “Gemma4 stays on its dedicated path; native heads register here” is enough. Repeating “empty until GLM / later PR / not wired yet” makes this read like PR scaffolding rather than a stable subsystem contract.

LxYuan0420 (changes requested):
Before this becomes the base for #484/#485, please either keep this PR to the hidden-state policy move, or make the native-MTP contract class-owned and runnable enough that a registered head can build/install a `MetalProposer` while keeping `model_runner` thin.

samithaj:
Thanks for the thorough pass — agreed on all counts, and going with your second option: class-owned and runnable.

Restructure plan across the stack:

- **This PR**: keep the hidden-state policy move; replace the `MTPHeadSpec` DTO with a `NativeMTPHead` protocol that owns validation and proposer construction (`build_proposer(...) -> MetalProposer`); `install_drafter`'s positive path delegates to the registered head and stays thin, with non-registered `mtp` configs still failing through `unsupported_mtp_message`. Adding the positive Gemma4 dispatch test (agreed that's the load-bearing one), deduplicating the negative proposer tests, parametrizing the registry `returns_none` cases, and rewriting the module docs as a contract rather than staging narration.
- **#484**: light rebase — loader/model internals are unchanged by this.
- **#485**: `Glm4MoeLiteMTPHead` implements the protocol and owns its runtime guards (prefix-caching rejection, `num_speculative_tokens` cap), so the runner-side free functions and dispatch you flagged go away; the proposer/slab internals are untouched.

Force-pushes on all three within a day, with a summary comment on each.

samithaj:
Force-push summary:

Reworked per review into two clean commits. The `MTPHeadSpec` DTO is gone; `vllm_metal/v1/mtp_heads` now defines a class-owned `NativeMTPHead` protocol whose `build_proposer(*, speculative_config, controller, vllm_config, target_config, dtype)` owns both runtime validation and proposer construction, so `install_drafter`'s mtp branch is a thin delegation — look the head up by draft `model_type`, install `head.build_proposer(...)`, or fail loud via `unsupported_mtp_message`. The registry stays empty in this PR, so all non-Gemma4 mtp drafts fail through. The package import stays light (no mlx); a head imports its heavy implementation lazily inside `build_proposer`.

Tests add the positive Gemma4 dispatch (an assistant config installs `Gemma4MTPProposer` through `install_drafter`) and a `NativeMTPHead` test double proving the mtp path builds and installs whatever a registered head returns; the registry negatives are parametrized, the duplicate Gemma4 no-sample test is collapsed, and the module docs read as a subsystem contract.

Tests: `pytest -m "not slow" tests/` — 1337 passed, 17 skipped. `ruff check` + `ruff format --check` clean.

vllm_metal/v1/mtp_heads/__init__.py
```diff
@@ -0,0 +1,123 @@
+# SPDX-License-Identifier: Apache-2.0
```
LxYuan0420: Please move this implementation out of `mtp_heads/__init__.py` into a real module, e.g. `mtp_heads/registry.py`. The package root currently owns the protocol, registry state, config lookup, and unsupported-message formatting. That is too much ownership for `__init__.py`, and #484/#485 will make this package grow quickly. Keep `__init__.py` empty or re-export-only, with runner importing the registry seam from an explicit module.

vllm_metal/v1/mtp_heads/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def build_proposer(
+        self,
+        *,
+        speculative_config: Any,
+        controller: Any,
+        vllm_config: Any,
+        target_config: Any,
+        dtype: Any,
+    ) -> MetalProposer:
```
LxYuan0420: Please replace this wide kwargs protocol with a small immutable build context, e.g. `NativeMTPBuildContext(speculative_config, controller, vllm_config, target_config, dtype)`, then `build_proposer(context) -> MetalProposer`. Five `Any` kwargs make the interface unstable and force future native heads to depend on runner internals instead of one narrow contract.

vllm_metal/v1/model_runner.py
```diff
@@ -722,6 +726,21 @@ def install_drafter(self, *, num_blocks: int, block_size: int) -> None:
...
+            head = find_native_mtp_head(spec)
+            if head is None:
+                raise NotImplementedError(unsupported_mtp_message(spec))
+            self._drafter = head.build_proposer(
+                speculative_config=spec,
+                controller=self._spec_decode_controller,
+                vllm_config=self.vllm_config,
+                target_config=self.model_args,
+                dtype=self.kv_cache_dtype,
+            )
```
LxYuan0420: Please keep this call site thinner once the native-MTP build context exists: build the context and call `head.build_proposer(context)`. Right now `model_runner` still shapes every dependency one kwarg at a time, which makes the runner part of the native-head construction contract.

tests/test_v1_model_runner_generate.py
```diff
@@ -1632,3 +1632,97 @@ def test_is_mla_false_for_standard_mha(self) -> None:
...
+
+    model_type = "recording_mtp"
+
+    def __init__(self) -> None:
+        self.calls: list[dict[str, object]] = []
+        self.proposer = object()
+
+    def build_proposer(self, **kwargs: object) -> object:
+        self.calls.append(kwargs)
+        return self.proposer
```
LxYuan0420: Please make this fake return a tiny proposer-shaped test double that implements the `MetalProposer` surface. Returning a plain `object()` proves the runner installs any object, not that a registered head builds a valid proposer. This test should protect the contract we are introducing, not bypass it.

tests/test_mtp_head_registry.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+
+class _FakeHead:
+    """Minimal :class:`NativeMTPHead` for registry-lookup tests."""
+
+    model_type = "dummy_mtp"
+
+    def build_proposer(self, **_: Any) -> object:
+        return object()
```
LxYuan0420: Please make `_FakeHead.build_proposer()` return a minimal proposer-shaped object, or keep this test strictly about lookup and avoid exercising `build_proposer`. Returning `object()` weakens the native-head contract in the test layer.

tests/test_mtp_head_registry.py
```diff
@@ -0,0 +1,121 @@
...
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    head = _FakeHead()
+    monkeypatch.setattr(mtp_heads, "MTP_HEAD_REGISTRY", {"dummy_mtp": head})
+
+    assert find_native_mtp_head(_mtp_config(model_type="dummy_mtp")) is head
+
+
+def test_fake_head_satisfies_native_mtp_head_protocol() -> None:
+    assert isinstance(_FakeHead(), NativeMTPHead)
```
LxYuan0420: Please delete this test. `@runtime_checkable` Protocol only checks attribute presence, not the `build_proposer` signature or return type, so this mostly tests Python Protocol mechanics. The meaningful proof is registered-head dispatch with a proposer-shaped return.

tests/test_mtp_head_registry.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def test_registered_mtp_head_types_is_empty_by_default() -> None:
+    assert registered_mtp_head_types() == []
```
LxYuan0420: Please delete this staging-specific test. It will flip once a native head registers. The unregistered-error test already covers the current empty-registry behavior without making 'empty by default' a durable contract.

tests/test_mtp_head_registry.py
```diff
@@ -0,0 +1,121 @@
...
+
+    assert registered_mtp_head_types() == ["dummy_mtp"]
+
+
+def test_unsupported_mtp_message_names_model_type_and_registered_heads() -> None:
+    message = unsupported_mtp_message(_mtp_config(model_type="eagle"))
+
+    assert "'eagle'" in message
+    # Registry is empty, so the registered-heads list renders empty.
+    assert "registered native heads: []" in message
```
LxYuan0420: Please stop asserting the exact rendered list format. The contract is that the unsupported error names the draft model type and points users toward supported heads. Exact text like `registered native heads: []` makes formatting part of the test contract.

tests/test_mtp_head_registry.py
```diff
@@ -0,0 +1,121 @@
...
+def test_unsupported_mtp_message_lists_registered_heads(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.setattr(
+        mtp_heads, "MTP_HEAD_REGISTRY", {"glm4_moe_lite_mtp": _FakeHead()}
+    )
+
+    message = unsupported_mtp_message(_mtp_config(model_type="eagle"))
+
+    assert "registered native heads: ['glm4_moe_lite_mtp']" in message
```
LxYuan0420: Please assert that the registered model type appears, not the exact Python list rendering. Message formatting should be free to change without breaking a behavior test.

tests/test_v1_model_runner_generate.py
```diff
@@ -1632,3 +1632,97 @@ def test_is_mla_false_for_standard_mha(self) -> None:
...
         )
         assert runner.is_mla is False
+
+
+class _RecordingMTPHead:
+    """A native MTP head double that records its ``build_proposer`` call.
+
+    Proves the mtp dispatch delegates: whatever this head returns is installed
+    as ``_drafter``, with the runner's config threaded through unchanged.
+    """
```
LxYuan0420: Please trim this explanatory docstring/comment. The test name and assertions already show that a registered head builds and installs a proposer. Comments that narrate the staging story will age poorly once #485 lands.

tests/test_gemma4_mtp_proposer.py
```diff
@@ -0,0 +1,43 @@
...
+from types import SimpleNamespace
+
+from vllm_metal.v1.proposer import Gemma4MTPProposer
+from vllm_metal.v1.spec_decode import PagedDecodeSegment
+
+
+def _proposer() -> Gemma4MTPProposer:
+    # The runner is unused by needs_target_hidden_states; the proposer is only
+    # installed for Gemma4 MTP configs, so the policy needs no config lookup.
+    return Gemma4MTPProposer(runner=SimpleNamespace())
```
LxYuan0420, line 13: Please remove or shorten this helper comment. The runner is intentionally unused for this method-level test, and `SimpleNamespace()` already makes that clear. Keep comments for non-obvious invariants, not mocked dependencies.

samithaj:
All applied, in one push.

The contract now lives in `vllm_metal/v1/mtp_heads/registry.py`. The package `__init__.py` just re-exports the registry symbols and imports the small bundled head modules so they register. The five-kwargs protocol is gone: `build_proposer` now takes a frozen `NativeMTPBuildContext`, which is also the extension point for anything future heads need. `install_drafter` builds the context and installs whatever the head returns, nothing more. In #485 the GLM head moved to its own small module, `mtp_heads/glm4.py`, and registers itself — `registry.py` never names a head, and importing the package still pulls in no mlx.

On the test comments: the fake head now returns a proposer-shaped object, the protocol-conformance and registry-empty tests are deleted, the error-message tests check meaning (draft model type, supported heads) instead of exact wording, and the narration comments are trimmed. We also went through #484 and #485 with the same standard — removed the staging/PR-reference sentences from docstrings, cut explanatory comments down to their invariants, and fixed the one remaining exact-wording assertion there too.

samithaj:
Force-push summary:

The contract moved to `mtp_heads/registry.py`, and `__init__.py` re-exports it. `build_proposer` takes a frozen `NativeMTPBuildContext` instead of five keyword arguments. `model_runner` imports the registry module directly, builds the context, and installs what the head returns. On the tests: the fake head returns a proposer-shaped object, the protocol-conformance and registry-empty tests are deleted, message assertions check meaning instead of exact wording, and narration comments are trimmed.

Tests: `pytest -m "not slow" tests/` — 1335 passed, 17 skipped (2 fewer because of the deleted tests). `ruff check` + `ruff format --check` clean.

LxYuan0420:
I pushed a few cleanup commits to align the seam: class-owned registry, thin runner dispatch, no package-root implementation dump, and tighter contract tests.

#483 is technically in good shape now, but I’ll hold approval until #485 is reviewable enough to prove this seam with a real native head. If #485 does not land cleanly, we should keep this PR to the hidden-state policy move and defer the registry.

samithaj:
Thanks for taking this over. I rebased #484/#485 on top: the loader no longer re-checks the config after `NativeMTPHeadRegistry.find` has matched, and the GLM head registers from `mtp_heads/glm4.py`.

LxYuan0420:
I narrowed #483 to the standalone Gemma4 change: hidden-state policy now lives in Gemma4MTPProposer, with install-path coverage. The native-MTP registry and wiring have been removed.

As discussed in #482, #484/#485 will remain as reference work and are no longer planned merge candidates.

LxYuan0420 (approved):
Merging it as a standalone maintainability improvement for the existing Gemma4 MTP path. 

TLDR: It moves the needs_target_hidden_states decision into Gemma4MTPProposer while the runner continues to collect hidden states through the generic proposer contract. 


The broader native-MTP discussion continues in #482. #484/#485 remain reference work.

PR #487, opened 2026-07-09
Gate native greedy sampling on logits-processor activity, not presence

KrxGu:
Found this while profiling why single-stream greedy decode trails mlx-lm on the same model and kernels.

can_use_native_greedy() bails to the torch sampler whenever any non-argmax-invariant logits processor is registered. The problem is that build_logitsprocs() always registers the MinTokens and LogitBias builtins, so that condition is true for essentially every request. The result is that greedy decode never actually takes the native MLX argmax path. Every step it copies the full-vocab logits out to torch and samples there, even when nothing is set that could change the argmax.

Those builtins are no-ops unless a request sets min_tokens or logit_bias, so the fix is to gate on whether they're active rather than merely present. The check now skips the bridge only when no request sets logit_bias or min_tokens, and it still forces the torch path for any non-argmax-invariant processor it doesn't recognize, so it stays correct if new ones are added. Output is identical either way, since the skipped builtins wouldn't have touched the logits for these requests.

image

Smaller models gain more, since the fixed bridge cost is a bigger share of the step.

The existing test built a batch with an empty processor set, which is why this never showed up. Added a test that registers the real builtins: it fails on current main, passes with the change, and also checks that active logit_bias/min_tokens and unrecognized processors still fall back to the torch sampler.

### DIFF
--- a/vllm_metal/v1/sampling_batch.py
+++ b/vllm_metal/v1/sampling_batch.py
@@ -14,6 +14,10 @@ from vllm.sampling_params import SamplingParams
 from vllm.utils.torch_utils import make_tensor_with_pad
 from vllm.v1.outputs import LogprobsLists
 from vllm.v1.sample.logits_processor import LogitsProcessors
+from vllm.v1.sample.logits_processor.builtin import (
+    LogitBiasLogitsProcessor,
+    MinTokensLogitsProcessor,
+)
 from vllm.v1.sample.metadata import SamplingMetadata
 from vllm.v1.sample.sampler import Sampler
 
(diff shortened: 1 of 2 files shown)


vllm_metal/v1/sampling_batch.py
```diff
@@ -166,7 +166,22 @@ def merge_logprobs_rows(
...
+        # populated at this point, so params are the source of truth.)
+        _known_non_argmax = (
+            "MinTokensLogitsProcessor",
+            "LogitBiasLogitsProcessor",
+            "MinPLogitsProcessor",
+        )
+        if any(
+            type(_proc).__name__ not in _known_non_argmax
+            for _proc in self.logitsprocs.non_argmax_invariant
+        ):
```
LxYuan0420: The fast path should only whitelist the exact upstream builtins whose dormant behavior we know. `MinPLogitsProcessor` is argmax-invariant upstream, so please remove it, and avoid matching by class name. Unknown/custom processors should fall back.

```python
from vllm.v1.sample.logits_processor.builtin import (
    LogitBiasLogitsProcessor,
    MinTokensLogitsProcessor,
)

if any(
    type(proc) not in (MinTokensLogitsProcessor, LogitBiasLogitsProcessor)
    for proc in self.logitsprocs.non_argmax_invariant
):
    return False
```

KrxGu: Updated. Removed MinPLogitsProcessor since its is_argmax_invariant is True upstream, so it never shows up in non_argmax_invariant anyway. Switched to identity checks against the imported MinTokensLogitsProcessor / LogitBiasLogitsProcessor; anything else falls back to the torch path.

KrxGu:
@LxYuan0420 i have addressed all your reviews, CI is all green.
LGTM!!

LxYuan0420 (approved):

PR #489, opened 2026-07-09
Release invalidated GDN runtime state

### LINKED ISSUE #488: [Bug][Hybrid] Preempted-then-resumed requests on hybrid GDN models continue from stale recurrent state
## Summary

On hybrid GDN models (Qwen3.5 / Qwen3.6 / Qwen3-Next family), a request that the
scheduler preempts under block pressure (recompute mode) and later resumes
re-prefills from position 0 — but its GDN recurrent-state slot is neither
released nor reset. The re-prefill kernel loads the slot's existing
mid-sequence recurrent state as its initial state, so every linear-attention
layer of the resumed request computes from corrupted state from the resume
point onward. Full-attention (SDPA) layers are unaffected because the
re-prefill overwrites their paged KV rows. Silent — no assert fires; the
request completes "successfully" with a wrong continuation.

Found while designing hybrid spec-decode verification (the rollback design
needs exactly this "re-seed a slot from a known state" primitive).

## Mechanism (main @ 0e10f10)

- Per-request GDN state lives in stable slots, separate from paged KV:
  `HybridGDNStateManager` (`vllm_metal/attention/state/gdn.py`) maps
  `req_id -> slot` into `GDNPagedStateCache`
  (`vllm_metal/attention/caches/gdn_cache.py:81-89`).
- Slots are freed ONLY for finished requests:
  `model_runner.py:2169-2197` (`_cleanup_finished_requests`) →
  `HybridPagedAttentionRuntime.release_requests`
  (`attention/runtime/hybrid.py:240-241`) →
  `HybridGDNStateManager.release_requests` (`state/gdn.py:110-123`).
- The resume path neither releases nor resets:
  `model_runner.py:1934-1943` (`_update_cached_request_blocks`) detects
  `req_id in cached_reqs.resumed_req_ids`, replaces `state.block_ids`, sets
  `state.generated_tokens = 0`, and pops `_paged_request_seq_lens[req_id]` —
  triggering a full re-prefill from position 0. The GDN slot is untouched.
- On the re-prefill step, `assign_step_slots` (`state/gdn.py:57-104`)
  short-circuits on the existing `req_id` mapping (lines 66-69) and returns
  the OLD slot with its mid-sequence state. Zeroing (`reset_slot`,
  `gdn_cache.py:164-175`) only happens for slots reused by a *new* request.
- The prefill kernel then seeds the scan from that stale state:
  `gdn_recurrent_prefill.metal:29-33` loads `state_in` (the slot's current
  recurrent state) as the initial state.

### End-to-end (Qwen/Qwen3.5-0.8B, greedy, measured)

Setup: paged attention on, in-process engine
(`VLLM_METAL_USE_PAGED_ATTENTION=1`, `VLLM_ENABLE_V1_MULTIPROCESSING=0`),
paged-KV block pool clamped to 12 blocks, 5 concurrent prompts,
`max_tokens=700`, `ignore_eos=True`, `max_model_len=1600`. Preemption and
resume confirmed via `SchedulerOutput.preempted_req_ids` /
`CachedRequestData.resumed_req_ids`: request 1 preempted and resumed at total
length 545, request 3 at total length 381 — identically on both code
versions, so the two runs are step-aligned.

Fixed-vs-unfixed pressure runs (same seed-free greedy config):

| request | preempted/resumed | first fixed-vs-unfixed diff (output pos) |
|---|---|---|
| 0, 2, 4 | no | none — bit-identical (700/700) |
| 1 | resumed @ total 545 | **535 = exactly the resume point** |
| 3 | resumed @ total 381 | **375 = exactly the resume point** |

Ground truth: feeding each run's identical pre-resume token prefix back as a
solo prompt and greedily continuing on a clean engine:

| request | clean-vs-FIXED first diff | clean-vs-UNFIXED first diff |
|---|---|---|
| 1 | 18 (batch-shape tie-wobble) | **0 — first post-resume token wrong** |
| 3 | 36 (batch-shape tie-wobble) | **0 — first post-resume token wrong** |

The unfixed continuation is wrong from the first token computed after resume;
the fixed path reproduces the clean continuation until ordinary batched-decode
tie-wobble. The corrupted continuation stays locally fluent (GDN's decay
gating attenuates the stale contribution over the re-prefill), which makes the
bug silent in practice — wrong content, no crash.

### Unit-level

Driving the real `HybridGDNStateManager` + `GDNPagedStateCache` through the
runner's preempt+resume bookkeeping: after resume the request still maps to
its original slot and the slot still holds the full mid-sequence state
(`max|state| = 5.0` where a clean re-prefill requires `0.0`).

## Fix

Zero the request's GDN slot when the runner detects resume-for-recompute,
keeping the `req_id -> slot` mapping (the request is still live):
`HybridGDNStateManager.reset_requests()`, surfaced through the runtime
protocol with a no-op default for pure paged-KV runtimes, called from
`_update_cached_request_blocks` for `resumed_req_ids`. Regression tests
assert the slot is zeroed at re-prefill time and the mapping is preserved.

**PR available.**

samithaj:
## What

Fixes #488 by changing hybrid GDN lifecycle handling from resume-time slot reset to runtime-state invalidation.

When the scheduler finishes, preempts, or resumes a hybrid GDN request, the request's recurrent runtime state is no longer valid. Finished requests also drop runner/MM metadata; preempted and resumed requests keep runner metadata, but release their GDN slot so the next assignment starts from clean runtime state.

This PR also rejects hybrid prefix caching at the platform config boundary. GDN recurrent state cannot currently be restored from KV blocks, so nonzero prefix-cache resume is unsupported until explicit GDN state snapshots exist.

## Testing

Validated locally after rebasing on current `main`:

- `ruff format --check` on touched files
- `ruff check` on touched files
- `pytest tests/test_platform.py tests/attention/test_hybrid_gdn_state_manager.py tests/test_paged_prefix_caching.py tests/test_v1_model_runner_generate.py tests/v1/mm/test_model_runner_multimodal_cache.py -q`

Result: `144 passed`.


vllm_metal/v1/model_runner.py
```diff
@@ -1941,6 +1943,14 @@ def _update_cached_request_blocks(
...
             state.generated_tokens = 0
             self._paged_request_seq_lens.pop(req_id, None)
+            resumed_req_ids.add(req_id)
+
+        # A resumed request re-prefills from position 0 but keeps any
+        # per-request recurrent state the runtime holds for it (it never
+        # finished, so release_requests never ran); reset that state before
+        # this step's forward is built.
+        if resumed_req_ids:
+            runtime.reset_requests(resumed_req_ids)
```
LxYuan0420: Please release runtime-owned GDN state when a request is preempted/resumed, instead of keeping the same slot and resetting it on resume. A waiting request should not keep a bounded GDN slot reserved; otherwise another request can need a third slot before the preempted one resumes. Keep `RequestState`, release only the GDN mapping/state for `preempted_req_ids | resumed_req_ids`, and let `assign_step_slots()` reuse and zero the slot. Then `reset_requests()` can be deleted.

tests/test_paged_prefix_caching.py
```diff
@@ -347,6 +349,69 @@ def test_resumed_request_replaces_blocks_and_resets_prefill_state(self):
...
+            all_token_ids={},
+            num_computed_tokens=[4],
+            num_output_tokens=[0],
+        )
+        runner._update_cached_request_blocks(cached_reqs)
+
+        mx.eval(cache.conv_states[0], cache.recurrent_states[0])
+        assert manager.request_slots == {"req-1": slot}
+        assert np.all(np.array(cache.conv_states[0][slot]) == 0)
+        assert np.all(np.array(cache.recurrent_states[0][slot]) == 0)
```
LxYuan0420: Please make this test use `num_computed_tokens=[0]`. With `[4]`, the runner would skip tokens 0-3 and forward only from `start_pos=4`, but the test also zeroes the GDN state for those skipped tokens. That is not a valid resume shape unless we implement state restore.

LxYuan0420 (approved):

PR #491, opened 2026-07-10
Generic KVConnector worker host contract on Metal (connector-agnostic)

lokic233:
## Summary

Implements the **generic KVConnector worker host contract** on the Metal
model runner/worker, so that **any** configured KV-transfer connector can run
on Apple Silicon through the standard `KVTransferConfig` mechanism. This
**replaces** the earlier LMCache-specific approach (`vllm_metal/lmcache_mp.py`)
— that file is gone.

The Metal runner/worker do not inherit vLLM's `GPUModelRunner` /
`KVConnectorModelRunnerMixin`, so the worker-side KVConnector lifecycle that GPU
workers get for free never ran on Metal. This adds it, using **only** the
generic public vLLM API (`get_kv_transfer_group` / `has_kv_transfer_group`), and
is a **no-op unless a `kv_transfer_config` is set**.

### Connector-agnostic by construction
This patch contains **no** knowledge of any specific connector: no LMCache
imports, no library-specific env vars, no connector-specific metadata types,
no connector-specific block tracking, no layout logic, no `lmcache_mp.py`, and
no `if connector is X` special-casing. `git diff` is **2 files / +275 lines**
with **zero** `LMCache` references.

## What it adds

**`worker.py`**
- `initialize_from_config`: call the generic
  `ensure_kv_transfer_initialized(vllm_config, kv_cache_config)` before KV-cache
  init, mirroring `GPUWorker`. **This is the hook the Metal worker was missing** —
  without it there is no worker-side KV transfer group, so `register_kv_caches`
  (and the entire worker lifecycle) never fires.
- `get_kv_connector_handshake_metadata`: mirror `GPUWorker` so the engine core
  can collect per-worker handshake metadata.

**`model_runner.py`**
- expose the live paged KV cache to the configured connector via
  `register_kv_caches`, keyed by the engine's ordered layer names;
- drive the standard `bind_connector_metadata → start_load_kv → (forward) →
  wait_for_save → get_finished` lifecycle and return a `KVConnectorOutput`;
- poll the connector on zero-token scheduler steps so async loads for requests
  in `WAITING_FOR_REMOTE_KVS` can complete;
- narrow the MLX↔view sync to the blocks touched each step.

## Honest note on the exposed buffers

The per-layer buffers exposed to the connector are `torch.stack([K, V])` CPU
views over the live MLX key/value caches. **`torch.stack` allocates, so the
stacked view is a COPY, not an alias** — hence the MLX→view sync before a store
and view→MLX writeback after a load. (The individual per-tensor MLX↔torch views
*do* alias unified memory; only the stack breaks it.) A future connector
enhancement that accepts **separate K and V tensors** would let the per-K/per-V
CPU views alias the MLX cache with zero copies and remove the sync/writeback.

## Why this belongs in vllm-metal (and not in the connector)

`ensure_kv_transfer_initialized` and the worker-side lifecycle are the generic
vLLM host contract that `GPUWorker`/`GPUModelRunner` already implement in-tree.
Metal is out-of-tree and doesn't inherit them, so they must be provided here —
for the benefit of **every** connector, not any specific one. Ideally this
(or an equivalent mixin) is upstreamed so out-of-tree runners get it for free.

## Companion

- LMCache/LMCache#4003 — the LMCache-side explicit KV-layout support
  (canonical key `lmcache.kv_layout`). All LMCache-specific behavior (layout
  resolution, MP protocol, tracking) lives there.

Configuration (standard `KVTransferConfig`, no `VLLM_METAL_ENABLE_LMCACHE_MP`):
```python
KVTransferConfig(
    kv_connector="LMCacheMPConnector",
    kv_connector_module_path="lmcache.integration.vllm.lmcache_mp_connector",
    kv_role="kv_both",
    kv_connector_extra_config={
        "lmcache.mp.host": "tcp://localhost", "lmcache.mp.port": "5555",
        "lmcache.mp.mp_transfer_mode": "engine_driven",
        "lmcache.kv_layout": "NHD",
    },
)
```

🤖 Authored by Navi on behalf of @lokic233.

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -36,12 +36,16 @@ from vllm.v1.kv_cache_interface import KVCacheConfig, KVCacheSpec
 from vllm.v1.outputs import (
     EMPTY_MODEL_RUNNER_OUTPUT,
     DraftTokenIds,
+    KVConnectorOutput,
     LogprobsLists,
     ModelRunnerOutput,
 )
 from vllm.v1.sample.logits_processor import build_logitsprocs
 from vllm.v1.sample.metadata import SamplingMetadata
 from vllm.v1.sample.sampler import Sampler
+from vllm.v1.worker.kv_connector_model_runner_mixin import (
+    KVConnectorModelRunnerMixin,
+)
 
 from vllm_metal.attention.context import (
     OffsetCache,
@@ -260,7 +264,7 @@ class _PagedForwardState(NamedTuple):
     pooling_hidden_states: mx.array | None = None
 
 
-class MetalModelRunner:
+class MetalModelRunner(KVConnectorModelRunnerMixin):
     """Model runner for MLX-based inference on Metal.
 
     Implements the vLLM v1 model runner interface for Apple Silicon.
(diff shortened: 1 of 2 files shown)


lokic233:
Superseded by #497.

Per review direction, the generic KVConnector host contract has been moved into **vLLM core** (vllm-project/vllm#48389) instead of hand-rolled here. vllm-metal now merely **consumes** that device-neutral contract: `MetalWorker`/`MetalModelRunner` inherit the vLLM mixins, and the only Metal-specific code is exposing MLX buffers to vLLM as torch tensors — no KV-cache/connector/LMCache awareness, no `lmcache_mp.py`.

Closing in favor of #497 (validated bit-exact on Apple M4 with the stock `LMCacheMPConnector`).

— via Navi on behalf of dengcchi

PR #495, opened 2026-07-10
Optimize paged-attention slot mapping preparation

trevorprater:
## Summary

- Optimize `prepare_unified()` slot mapping construction by appending contiguous physical KV-block spans as ranges.
- Add a dedicated fast path for batched single-token decode requests.
- Preserve non-contiguous block behavior by appending one range per physical block rather than calculating every token slot individually.
- Add reference-oracle coverage for non-contiguous blocks, non-power-of-two block sizes, speculative decode, mixed prefill/decode, prefix hits, and zero-length prefill.

## Why

Paged-attention metadata preparation rebuilt each slot with Python division, indexing, multiplication, and list append operations. Prefill and prefix-cache workloads usually cover long logical token spans, and physically contiguous KV blocks make those slots a single arithmetic range. Constructing that range directly removes most of the per-token Python overhead while preserving the exact metadata contract.

## Performance

These are CPU-side metadata microbenchmark results on Apple Silicon, comparing the original implementation and this change in the same process with alternating measurement order. They are not end-to-end serving throughput claims.

| Workload | Original | Optimized | Speedup |
| --- | ---: | ---: | ---: |
| 1 x 2048 contiguous prefill | 82.91 us | 16.01 us | 5.18x |
| 4 x 2048 contiguous prefill | 332.40 us | 71.24 us | 4.67x |
| Mixed decode/prefill, contiguous | 174.95 us | 44.28 us | 3.95x |
| 4 x 2048 non-contiguous prefill | 333.15 us | 144.21 us | 2.31x |
| Mixed decode/prefill, non-contiguous | 173.92 us | 79.45 us | 2.19x |
| 128 single-token decode requests | 17.90 us | 14.09 us | 1.27x |

The aggregate geometric-mean ratio across the full benchmark suite was **3.109x**.

## Validation

- `ruff check .`
- `python -m compileall -q vllm_metal tests`
- `VLLM_METAL_BUILD_FROM_SOURCE=1 python -m pytest -m 'not slow' -q` (`1342 passed, 17 skipped, 35 deselected`)
- Reference oracle against the original implementation across fixed edge cases and 120 seeded randomized workloads

### DIFF
--- a/tests/test_paged_attention.py
+++ b/tests/test_paged_attention.py
@@ -18,6 +18,81 @@ from vllm_metal.attention.context import (
 )
 from vllm_metal.attention.impls.sdpa_wrapper import SDPAPagedAttentionWrapper
 
+DecodeRequest = tuple[list[int], int] | tuple[list[int], int, int]
+PrefillRequest = tuple[list[int], int, int]
+PrepareMetadata = tuple[
+    list[int],
+    list[int],
+    list[list[int]],
+    list[int],
+    list[int],
+    int,
+]
+
+
+def _reference_prepare_unified(
+    decode_requests: list[DecodeRequest],
+    prefill_requests: list[PrefillRequest],
+    block_size: int,
+) -> PrepareMetadata:
+    slot_mapping: list[int] = []
+    cu_seqlens: list[int] = [0]
+    block_tables: list[list[int]] = []
+    context_lens: list[int] = []
+    offsets: list[int] = []
+
+    for decode_request in decode_requests:
+        if len(decode_request) == 2:
+            block_ids, seq_len = decode_request
+            num_tokens = 1
+        else:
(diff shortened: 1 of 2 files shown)


trevorprater:
I profiled this change in real paged-attention generation after opening the PR. Across prefill-heavy, balanced, decode-heavy, and shared-prefix Qwen3-0.6B workloads, the original function represented approximately 0.0096% of end-to-end runtime, and the estimated overall improvement was only 0.0025%. Replaying the exact production metadata traces also reduced the function-level improvement from the synthetic benchmark's 3.109x to 1.36x.

Given the added implementation complexity and negligible practical impact, I am closing this rather than asking maintainers to review it. The optimization itself is correct, but the production path is not a meaningful hotspot. Thanks for your time.

PR #496, opened 2026-07-11
Fix packed decode RoPE with vector offsets

trevorprater:
This PR is:
- To batch the current single-token packed decode RoPE path.
- To pass vector offsets into native `nn.RoPE`, avoiding MLX scalar-offset corruption for batched one-token rows.
- To keep custom RoPE, `B > 1`, and multi-token/prefill segments on the existing per-segment path.

## Context

MLX can produce wrong results for batched `[B, H, 1, D]` RoPE when using a scalar offset. The paged decode path already packs multiple one-token segments as `[1, H, N, D]`, so this PR transposes that into `[N, H, 1, D]` and calls native `nn.RoPE` once for Q and once for K with vector offsets.

This keeps the real decode-path win without adding broader RoPE batching semantics that current callers do not own.

### DIFF
--- a/tests/test_paged_attention.py
+++ b/tests/test_paged_attention.py
@@ -226,21 +226,21 @@ class TestPrepare:
 class TestPackedRoPE:
     """Tests for per-request RoPE position reset in packed prefill."""
 
+    class RecordingRoPE(nn.RoPE):
+        def __init__(self) -> None:
+            super().__init__(dims=8, traditional=False, base=10_000.0)
+            self.calls = []
+
+        def __call__(self, x, offset=0):
+            self.calls.append((x.shape, offset))
+            return super().__call__(x, offset=offset)
+
     def test_native_rope_batches_single_token_segments(self):
         from vllm_metal.attention.impls.varlen_rope_compat import (
             apply_packed_rope,
         )
 
-        class RecordingRoPE(nn.RoPE):
-            def __init__(self) -> None:
-                super().__init__(dims=8, traditional=False, base=10_000.0)
-                self.calls = []
-
-            def __call__(self, x, offset=0):
-                self.calls.append((x.shape, offset))
-                return super().__call__(x, offset=offset)
-
-        rope = RecordingRoPE()
+        rope = self.RecordingRoPE()
         reference_rope = nn.RoPE(dims=8, traditional=False, base=10_000.0)
         module = SimpleNamespace(rope=rope)
         q = mx.arange(1 * 3 * 4 * 8, dtype=mx.float32).reshape(1, 3, 4, 8) / 100
@@ -280,16 +280,7 @@ class TestPackedRoPE:
             apply_packed_rope,
         )
 
-        class RecordingRoPE(nn.RoPE):
-            def __init__(self) -> None:
-                super().__init__(dims=8, traditional=False, base=10_000.0)
-                self.offsets = []
-
-            def __call__(self, x, offset=0):
-                self.offsets.append(offset)
-                return super().__call__(x, offset=offset)
-
-        rope = RecordingRoPE()
+        rope = self.RecordingRoPE()
         module = SimpleNamespace(rope=rope)
         cu_seqlens = [0, 1, 2, 3, 4]
         q = mx.zeros((1, 3, 4, 8), dtype=mx.float32)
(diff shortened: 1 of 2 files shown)


tests/test_paged_attention.py
```diff
@@ -223,6 +227,173 @@ def test_prepare_unified_mixed(self):
...
+        module = SimpleNamespace(rope=nn.RoPE(dims=8, traditional=False, base=10_000.0))
+        q = mx.random.normal((batch, 3, total_len, 8))
+        k = mx.random.normal((batch, 2, total_len, 8))
+
+        expected_q = self._reference_rope(module.rope, q, cu_seqlens, offsets)
+        expected_k = self._reference_rope(module.rope, k, cu_seqlens, offsets)
+        q_out, k_out = apply_packed_rope(module, q, k, cu_seqlens, offsets=offsets)
+        mx.eval(expected_q, expected_k, q_out, k_out)
+
+        assert mx.allclose(q_out, expected_q, rtol=1e-5, atol=1e-5).item()
```
LxYuan0420, line 281: This still passes if batching never happens. Please spy on the RoPE calls and assert the actual fast-path contract: two calls total for Q/K, batched input shapes, and vector offsets of length `segment_count`. Numeric equivalence alone does not protect the performance change.

vllm_metal/attention/impls/varlen_rope_compat.py
```diff
@@ -52,6 +54,74 @@ def _apply_mrope_segment_with_positions(
     return apply_multimodal_rotary_pos_emb(q_seg, k_seg, cos, sin)
 
 
+def _supports_batched_offsets(rope_fn: object) -> bool:
+    """Return whether a RoPE callable supports vectorized MLX offsets."""
+    return isinstance(rope_fn, nn.RoPE) or (
```
LxYuan0420: Let’s not support `partial(mx.fast.rope)` until a real model path produces it. The current supported path uses `nn.RoPE` or custom RoPE objects, so this branch and its test are synthetic.

vllm_metal/attention/impls/varlen_rope_compat.py
```diff
@@ -52,6 +54,74 @@ def _apply_mrope_segment_with_positions(
...
+                mx.reshape(
+                    x,
+                    (batch, heads, segment_count, segment_len, head_dim),
+                ),
+                (0, 2, 1, 3, 4),
+            ),
+            (batch * segment_count, heads, segment_len, head_dim),
+        )
+
+    if batch == 1 or isinstance(batch_offsets, int):
```
LxYuan0420: This defines a `B > 1` paging contract we do not own today. Production packed forwards use one flattened sequence row. Please remove the multi-batch branch and `batch=2` tests from this PR.

vllm_metal/attention/impls/varlen_rope_compat.py
```diff
@@ -170,9 +240,50 @@ def apply_packed_rope(
...
+    if (
+        rope_fn is not None
+        and positions is None
+        and segment_count > 1
+        and cu_seqlens[0] == 0
+        and cu_seqlens[-1] == queries.shape[2]
+        and (not apply_keys or cu_seqlens[-1] == keys.shape[2])
+        and _supports_batched_offsets(rope_fn)
+    ):
+        segment_len = cu_seqlens[1]
```
LxYuan0420: The measured win is single-token decode. This also reroutes equal-length prefill, which is broader than the evidence. Please keep the fast path to `segment_len == 1` unless prefill correctness and perf are covered too.

vllm_metal/attention/impls/varlen_rope_compat.py
```diff
@@ -170,9 +240,50 @@ def apply_packed_rope(
     rope_fn = getattr(attn_module, "rope", None)
     rotary_emb = getattr(attn_module, "rotary_emb", None) if rope_fn is None else None
 
+    segment_count = len(cu_seqlens) - 1
+    if (
+        rope_fn is not None
+        and positions is None
+        and segment_count > 1
+        and cu_seqlens[0] == 0
```
LxYuan0420: These are `prepare_unified` internal invariants. Silent fallback hides bad metadata as a missed optimization. Trust the contract here, or validate once at the context owner, not again in the hot RoPE path.

LxYuan0420 (changes requested):
I think there is a real win here, but the PR currently carries extra paths that are not part of the measured production case.

LxYuan0420 (approved):
I pushed a small cleanup pass to keep the useful fix while narrowing the scope.

The branch now only batches the production single-token decode RoPE path. I also removed the speculative `partial(mx.fast.rope)`, `B > 1`, and equal-length multi-token prefill paths because they are not owned by current callers and were only covered by synthetic tests.

vllm_metal/attention/impls/varlen_rope_compat.py
```diff
@@ -170,9 +186,34 @@ def apply_packed_rope(
...
+        and positions is None
+        and segment_count > 1
+        and queries.shape[0] == 1
+        and queries.shape[2] == segment_count
+        and (not apply_keys or keys.shape[2] == segment_count)
+        and all(cu_seqlens[i] == i for i in range(segment_count + 1))
+        and isinstance(rope_fn, nn.RoPE)
+    ):
+        # MLX 0.31.2 corrupts rows after the first for [B, H, 1, D] RoPE
+        # with a scalar offset (MLX #3494). Vector offsets select the correct
```
WindChimeRan, line 215: nit: this issue has been fixed in the upstream? (we need to pin higher version of mlx)

vllm_metal/attention/impls/varlen_rope_compat.py
```diff
@@ -170,9 +186,34 @@ def apply_packed_rope(
...
+    segment_count = len(cu_seqlens) - 1
+    if (
+        rope_fn is not None
+        and positions is None
+        and segment_count > 1
+        and queries.shape[0] == 1
+        and queries.shape[2] == segment_count
+        and (not apply_keys or keys.shape[2] == segment_count)
+        and all(cu_seqlens[i] == i for i in range(segment_count + 1))
+        and isinstance(rope_fn, nn.RoPE)
```
WindChimeRan: suggestion (feel free to reject it): 

```python
from mlx_lm.models.rope_utils import Llama3RoPE, ProportionalRoPE, SuScaledRoPE, YarnRoPE

_VECTOR_OFFSET_ROPE = (nn.RoPE, Llama3RoPE, YarnRoPE, SuScaledRoPE, ProportionalRoPE)
...
and isinstance(rope_fn, _VECTOR_OFFSET_ROPE)
```

WindChimeRan (approved):
Nice job on the performance optimization! Thanks

vllm_metal/attention/impls/varlen_rope_compat.py
LxYuan0420: ya. upstream has fixed it. I will add a short removal cond doc and add a follow up PR.

PR #498, opened 2026-07-12
[GGUF] Add mistral GGUF loading via llama arch alias

ricky-chaoju:
This adds mistral to the local GGUF loader (#415; qwen2/qwen3 shipped in #466, llama in #480). Classic Mistral GGUFs declare general.architecture="llama" (llama.cpp converts MistralForCausalLM under the llama arch), and mlx-lm itself builds model_type="mistral" through its llama module, so the only thing rejecting the pair was the adapter's gguf-arch-vs-config cross-check. This PR adds a "mistral" to "llama" alias as an adapter class constant folded into the existing normalization helper, and the mismatch error now names the raw config model_type alongside its canonical form. The llama q/k RoPE un-permutation from #480 applies through the alias unchanged, and the real v0.3 config omits head_dim, so the built-object read from #480 gets exercised by a real checkpoint. The docs scope sentence now lists all four archs, which also catches up the llama entry #480 never added.

For evidence, per-op forward cosine of Mistral-7B-Instruct-v0.3 Q8_0 against its dense bf16 mlx-lm reference gives q_proj/k_proj 0.99997 (v/gate 0.99998), greedy decode agrees with the dense reference for all 16 golden tokens, and tools/gguf_serve_smoke.py passes with VLLM_METAL_MEMORY_FRACTION=0.5, the completion matching the golden text verbatim. The slow goldens run with VLLM_METAL_TEST_GGUF_MISTRAL_SERVE_PATH / _TOKENIZER_PATH / _DENSE_PATH set, via `pytest -m slow tests/test_gguf_serve_golden.py -k mistral`. Unit tests cover the untied, head_dim-omitted mistral fixture (q/k asserted un-permuted over qweight, scales, and biases), and the resolve_arch contract pins mistral3/mistral4 staying rejected. Mixtral (MoE) stays rejected by the tensor preflight; sliding-window semantics follow mlx-lm (bare mistral configs have no layer_types, so every layer runs full attention; v0.3 sets it null). Local `.gguf` only, verified on Mistral-7B-Instruct-v0.3 Q8_0.

### DIFF
--- a/tests/test_gguf_loader.py
+++ b/tests/test_gguf_loader.py
@@ -343,11 +343,12 @@ def test_quantized_llama_qk_are_row_unpermuted(tmp_path, quant_type):
     assert not bool(mx.array_equal(got_k.qweight, raw_k.qweight).item())
 
 
-def test_loads_untied_mistral_via_llama_alias(tmp_path):
+def test_loads_untied_mistral_via_llama_arch_mapping(tmp_path):
     # Real Mistral GGUFs declare general.architecture="llama" while the config
-    # says model_type="mistral"; the adapter alias admits the pair and the llama
-    # RoPE q/k un-permutation applies through it. Mirrors the real v0.3 shape:
-    # untied (lm_head present) and head_dim OMITTED (derived hidden // heads).
+    # says model_type="mistral"; the adapter mapping admits the pair and the
+    # llama RoPE q/k un-permutation applies through it. Mirrors the real v0.3
+    # shape: untied (lm_head present) and head_dim OMITTED (hidden // heads).
+    # Arrange
     d = _dims(_tiny_config("mistral"))
     gguf_path, cfg_dir = _build_dense_fixture(
         tmp_path,
@@ -368,10 +369,14 @@ def test_loads_untied_mistral_via_llama_alias(tmp_path):
     exp_q = raw_q.permute_rows(_rope_inv_index(d["qd"], cfg["num_attention_heads"]))
     exp_k = raw_k.permute_rows(_rope_inv_index(d["kvd"], cfg["num_key_value_heads"]))
 
+    # Act
     model, _ = GGUFModelLoader(
         gguf_path, config_dir=cfg_dir, target_dtype=mx.float32
     ).load()
+    out = model(mx.array([[1, 2, 3]]))
+    mx.eval(out)
 
+    # Assert
     assert isinstance(model.lm_head, GGUFLinear)  # untied head installed
     got_q = model.model.layers[0].self_attn.q_proj.tensor
     got_k = model.model.layers[0].self_attn.k_proj.tensor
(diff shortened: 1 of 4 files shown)


vllm_metal/gguf/adapter.py
```diff
@@ -207,9 +216,10 @@ def _rope_permute_index(out_features: int, n_head: int) -> mx.array:
             .reshape(out_features)
         )
 
-    @staticmethod
-    def _normalize_arch(name: str) -> str:
-        return name.strip().lower().replace("-", "_")
+    @classmethod
+    def _normalize_arch(cls, name: str) -> str:
```
LxYuan0420: Please split this into two steps: one helper only normalizes a GGUF arch string (`strip/lower/replace`), and another maps config `model_type` to GGUF arch. Also rename `_ARCH_ALIASES` to `_CONFIG_MODEL_TYPE_TO_GGUF_ARCH`. We should accept `gguf_arch="llama"` + `model_type="mistral"`, but not a GGUF that declares raw `mistral` as `llama`. The mismatch error should show the original GGUF arch, original config type, and mapped config arch.

tests/test_gguf_loader.py
```diff
@@ -342,6 +343,76 @@ def test_quantized_llama_qk_are_row_unpermuted(tmp_path, quant_type):
...
+
+    assert isinstance(model.lm_head, GGUFLinear)  # untied head installed
+    got_q = model.model.layers[0].self_attn.q_proj.tensor
+    got_k = model.model.layers[0].self_attn.k_proj.tensor
+    for got, exp, raw in ((got_q, exp_q, raw_q), (got_k, exp_k, raw_k)):
+        assert bool(mx.array_equal(got.qweight, exp.qweight).item())
+        assert bool(mx.array_equal(got.scales, exp.scales).item())
+        assert bool(mx.array_equal(got.biases, exp.biases).item())
+        assert not bool(mx.array_equal(got.qweight, raw.qweight).item())
+    out = model(mx.array([[1, 2, 3]]))
```
LxYuan0420: Please keep this test in AAA order. Arrange all fixture data and expected q/k tensors first, act by loading and running the forward pass, then keep assertions together. Right now the test asserts q/k before the forward pass, which makes the behavior harder to scan.

tests/test_gguf_loader.py
```diff
@@ -342,6 +343,76 @@ def test_quantized_llama_qk_are_row_unpermuted(tmp_path, quant_type):
...
+    mx.eval(out)
+    assert out.shape == (1, 3, 256)
+
+
+def test_resolve_arch_mistral_alias_contract():
```
LxYuan0420: Please split this mixed contract test. Keep one success test for `gguf_arch="llama"` + `model_type="mistral"`, then add parameterized rejection tests for raw `mistral`/`mistral3`/`mistral4` and mismatch errors. Success, unsupported arch, and mismatch are different behaviors.

tests/test_gguf_serve_golden.py
```diff
@@ -199,3 +199,73 @@ def test_llama_gguf_greedy_matches_dense_reference_prefix() -> None:
...
+    1782,
+]
+_MISTRAL_DENSE_AGREE_PREFIX = 16
+
+_MISTRAL_GGUF_ENV = "VLLM_METAL_TEST_GGUF_MISTRAL_SERVE_PATH"
+_MISTRAL_TOK_ENV = "VLLM_METAL_TEST_GGUF_MISTRAL_TOKENIZER_PATH"
+_MISTRAL_DENSE_ENV = "VLLM_METAL_TEST_GGUF_MISTRAL_DENSE_PATH"
+
+
+@pytest.mark.slow
```
LxYuan0420: The Qwen, Llama, and Mistral golden tests now repeat the same flow. Please parameterize the model cases instead: env names, golden ids, dense prefix, target dtype, and model label. Then the same two slow tests can run for every GGUF family without copy/paste.

LxYuan0420 (approved):
Approving it first since there are just minor edits needed.

PR #497, opened 2026-07-12
Consume vLLM's device-neutral KVConnector host contract (replaces #491)

lokic233:
## Summary

Make the Metal worker/runner KV-transfer-capable by **consuming vLLM's device-neutral KVConnector host-contract mixins** — not by hand-rolling the connector lifecycle. Metal gains **no** knowledge of any KV-cache library, layout, connector metadata, or block tracking. It only (a) inherits the generic vLLM mixins and (b) exposes its MLX paged KV buffers to vLLM as CPU torch tensors. Everything is a no-op without a `kv_transfer_config`.

> **Replaces #491.** That PR hand-rolled the KVConnector lifecycle inside vllm-metal. This one moves the generic contract into vLLM core and has Metal merely consume it, so Metal stays backend-only.

## Depends on

**vllm-project/vllm#48389** — "[V1] Add device-neutral KVConnector host-contract mixins" (adds `KVConnectorWorkerMixin` and the `register_kv_caches_with_connector` / `sync_kv_caches_before_save` / `writeback_kv_caches_after_load` methods this PR consumes). This PR should land after that.

## Changes

**`worker.py`** (+9):
- `MetalWorker` inherits `KVConnectorWorkerMixin`; `initialize_from_config` calls the generic `maybe_initialize_kv_transfer` before `initialize_kv_cache`.
- `get_kv_connector_handshake_metadata` is **inherited from the mixin** — no Metal code.

**`model_runner.py`** (+204):
- `MetalModelRunner` inherits `KVConnectorModelRunnerMixin`.
- `initialize_kv_cache` exposes the MLX key/value caches as per-layer CPU torch views (`mlx_to_torch`) and registers them via the generic `register_kv_caches_with_connector`.
- `execute_model` drives the generic lifecycle: `kv_connector_no_forward` on empty steps; otherwise `bind_connector_metadata` + `start_load_kv`, then loaded KV is written back into MLX. `sample_tokens` finalizes (`wait_for_save` + `get_finished`) once the paged forward completes.
- Overrides the two vLLM sync hooks. The connector-facing per-layer buffer is `torch.stack([K, V])` — and `torch.stack` **allocates** (a copy, not an MLX alias) — so Metal refreshes it from live MLX before a store and writes loaded KV back after a load, narrowed to the blocks touched each step (derived from generic scheduler data, not connector metadata).

The **only** Metal-specific code is the MLX↔torch bridging. There are **zero** references to any KV-cache library.

[...]

PR #499, opened 2026-07-13
Expose live split K/V to vLLM's generic KVConnector contract (replaces #497)

lokic233:
> **Draft** — depends on the generic KVConnector step lifecycle in vllm-project/vllm#48464. Will leave draft after rebasing onto the merged vLLM API.

## Summary
Make the Metal worker/runner KV-transfer-capable by exposing Metal's native MLX paged KV cache to whatever generic vLLM `KVConnector` is configured, and driving vLLM's generic connector step lifecycle around the Metal forward.

## Architecture / ownership
- **Configuration** selects the concrete connector (e.g. an `LMCacheMPConnector`); this code never names or imports it.
- **vLLM** owns the generic registration/lifecycle contract and constructs `KVConnectorOutput`.
- **vllm-metal (this PR)** exports stable live split K/V aliases exactly once and drives the generic lifecycle.
- **The connector** validates and interprets the split/fused representation and transfers the live buffers.

## What this PR does
1. **Live split K/V, no staging.** Exposes each layer as an independent `(key_view, value_view)` pair of zero-copy torch aliases over the live MLX arrays — **not** a fused `torch.stack([k, v])`. Connector writes on RETRIEVE land directly in the live MLX cache; reads on STORE observe the live KV. No fused staging buffer, no MLX→staging store copy, no staging→MLX writeback.
2. **Register once; pointer stability is an explicit invariant.** The in-place `reshape_and_cache` kernel keeps K/V buffer addresses stable, so the bridge registers the live aliases **exactly once** and records their data pointers + shapes. Per step it only (a) `mx.eval`-materializes pending MLX writes (memory-ordering barrier) and (b) verifies the live pointers still match the registered ones. `register_kv_caches` is not a replace/update operation, so if a pointer moved the bridge **fails fast** with a clear error rather than silently re-registering or continuing with stale pointers. There is **no dynamic re-registration**.
3. **Lifecycle + abort.** `begin` in `execute_model`, `finish` in `sample_tokens`; both wrappers abort the in-flight step (clear connector metadata exactly once via vLLM's generic `abort_kv_connector_step`) and re-raise if the body fails. No broad-exception retry of `start_load_kv`.
4. **Version dependency via capability detection.** When a KV connector is configured, the bridge checks the generic mixin for the required begin/finish/abort lifecycle and raises a clear, connector-name-agnostic capability error on an incompatible vLLM. Importing vllm-metal and running normal inference **without** a `kv_transfer_config` is unaffected on older supported vLLM (proven: normal generation works with no connector configured).
5. **Strict layer-identity validation.** A mismatch between the vLLM `KVCacheConfig` layer names and the Metal cache raises `ValueError`.

## Accurate scope
- vllm-metal is **connector-name agnostic**; it exposes a generic split-KV registration. A configured connector **must** support that split registration form — this PR does **not** claim every `KVConnector` works on Metal.
- This integration **requires** the vLLM generic begin/finish/abort KVConnector step lifecycle (vllm-project/vllm#48464). It does **not** claim independence from that API.
- The Metal KV storage is registered **once** and required to remain stable; **dynamic re-registration is not supported**.
- No cache-sized staging allocation or KV copy is used.

## Tests
`tests/test_kv_connector_bridge.py` (no model/server): register-once (no second registration across steps), pointer-movement fail-fast, pointer check does not allocate, layer-count mismatch raises, capability error when the lifecycle is absent, and no-connector no-op.

## Cross-repo dependency
Depends only on vllm-project/vllm#48464 (the generic lifecycle + registration type).

### DIFF
--- a/tests/test_kv_connector_bridge.py
+++ b/tests/test_kv_connector_bridge.py
@@ -1,0 +1,175 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Unit tests for MetalKVConnectorBridge: register-once, pointer-stability
+fail-fast, and capability detection. No model/server required; uses a fake
+runner + fake MLX KV arrays and a fake vLLM KV-transfer group.
+"""
+import types
+
+import mlx.core as mx
+import pytest
+
+
+def _install_fake_kv_transfer(monkeypatch, connector):
+    """Point vllm.distributed.kv_transfer.{has,get}_kv_transfer_group at a fake."""
+    import vllm.distributed.kv_transfer as kvt
+
+    monkeypatch.setattr(kvt, "has_kv_transfer_group", lambda: connector is not None)
+    monkeypatch.setattr(kvt, "get_kv_transfer_group", lambda: connector)
+    # the bridge imports these names into its own module namespace
+    import vllm_metal.v1.kv_connector as bridge_mod
+
+    monkeypatch.setattr(bridge_mod, "has_kv_transfer_group",
+                        lambda: connector is not None)
+    monkeypatch.setattr(bridge_mod, "get_kv_transfer_group", lambda: connector)
+
+
+class _FakeConnector:
+    def __init__(self):
+        self.register_count = 0
+        self.registered = None
+
+    def register_kv_caches(self, kv_caches):
+        self.register_count += 1
+        self.registered = kv_caches
+
+
+class _FakeKVCache:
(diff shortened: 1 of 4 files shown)


LxYuan0420:
Closing this draft until it is ready and demonstrates meaningful improvement.

PR #500, opened 2026-07-13
[Spec Decode] Cache draft KV so shared prefixes skip re-ingestion (#482)

KrxGu:
The draft model re-prefills the entire prompt into its KV on every new request. The allocator keys everything by request id and prunes on finish, so nothing carries over, not even an identical resubmitted prompt. On an 8k prefix that first propose is ~4.59s versus ~95ms once warm (#482).

This gives the draft allocator the same prefix caching the target already has. Full prompt blocks are hashed (chained, so a match means the whole prefix matches), cached by content to a physical draft block, and refcounted so shared blocks survive while any request holds them and idle ones are evicted under pressure. A new request reuses the longest matching leading run and only ingests the uncached suffix.

Output is byte-identical, since a hash match means identical tokens and therefore identical KV. Checked on M3 Ultra with an 8B target and 0.6B draft: a repeated 4k prefix reuses 3200 tokens of draft KV and produces the exact same output, and the saving grows with prefix length.

Scope is direction 1 only. Fresh-prompt latency (direction 3) and the steady-state proposer overhead (Problem 2) are separate follow-ons. Tests cover reuse, the exact-block-multiple edge case, refcount sharing, and eviction.

### DIFF
--- a/tests/test_draft_prefix_cache.py
+++ b/tests/test_draft_prefix_cache.py
@@ -22,20 +22,20 @@ def test_hash_prefix_blocks_full_only_deterministic_and_chained():
     p = _proposer(block_size=16)
     toks = list(range(40))  # 2 full blocks (32) + 8 trailing partial
 
-    hashes = p._hash_prefix_blocks(toks)
+    hashes = p._hash_prefix_blocks("r1", toks)
     assert len(hashes) == 2  # only full blocks are hashed
-    assert p._hash_prefix_blocks(toks) == hashes  # deterministic
+    assert p._hash_prefix_blocks("r1", toks) == hashes  # cache hit, idempotent
 
     # Chaining: same first block but a different second block -> same h[0],
     # different h[1] (h[1] depends on h[0] as parent, plus its own tokens).
     toks_b = list(range(16)) + list(range(100, 124))
-    hb = p._hash_prefix_blocks(toks_b)
+    hb = p._hash_prefix_blocks("r2", toks_b)
     assert hb[0] == hashes[0]
     assert hb[1] != hashes[1]
 
     # A different first block changes every downstream hash.
     toks_c = list(range(50, 90))
-    hc = p._hash_prefix_blocks(toks_c)
+    hc = p._hash_prefix_blocks("r3", toks_c)
     assert hc[0] != hashes[0]
 
 
@@ -82,14 +82,81 @@ def test_shared_prefix_block_not_freed_while_referenced():
     reused = p2.block_ids[:2]
     p._prune_finished({"r2": object()})  # r1 finishes, r2 still live
     for blk in reused:
-        assert blk not in p._free_blocks  # shared blocks survive
+        # r1's release must only drop the refcount, not free the block: r2
+        # still holds a reference. (A block registered in the prefix cache
+        # is never appended to a raw free list regardless of refcount, so
+        # asserting free-list membership here would pass even if the
+        # refcount bookkeeping were broken -- assert on the count itself.)
+        assert p._pool.blocks[blk].ref_cnt > 0
 
 
 def test_eviction_reuses_idle_cached_block_under_pressure():
-    p = _proposer(block_size=16, num_blocks=4)  # tiny pool
+    # 5 physical blocks -> 4 usable (BlockPool reserves id 0 as a null
+    # placeholder), a tiny pool that exactly fits one requester at a time.
+    p = _proposer(block_size=16, num_blocks=5)
     _warm(p, "a", list(range(48)))  # 3 full blocks cached + 1 partial
     p._prune_finished({})  # a done; 3 idle cached blocks
-    assert len(p._cached) == 3
+    assert len(p._pool.cached_block_hash_to_block) == 3
     p2 = p._make_plan("b", _state(list(range(100, 148))), 4)  # distinct content
     assert p2 is not None  # allocation succeeded by evicting idle cached blocks
-    assert len(p._cached) < 3
+    assert len(p._pool.cached_block_hash_to_block) < 3
+
+
+# -- Adversarial cases from abcgco's PR #500 review (independently verified on
+# an M4 Pro: pass on the prefix-cache branch, fail on its pre-cache base). --
+
+
+def test_partial_prefix_match_reuses_only_leading_run():
+    """A prompt diverging mid-way must reuse exactly the matching leading
+    run of blocks and ingest everything after the divergence."""
+    p = _proposer(block_size=16, num_blocks=64)
+    _warm(p, "r1", list(range(40)))  # blocks 0,1 cached
+    p._prune_finished({})
+    tokens_b = list(range(16)) + list(range(100, 140))  # shares block 0 only
+    p2 = p._make_plan("r2", _state(tokens_b), 4)
+    assert p2 is not None
+    assert p2.draft_seq_len == 16  # block 0 reused, block 1 diverged
+    assert p2.ingest_tokens == tokens_b[16:]
+
+
+def test_multi_turn_history_reuses_generated_blocks():
+    """Turn 2 resubmits turn 1's prompt + generated tokens as its prefix.
+    Full blocks containing generated content are registered too, so the
+    whole turn-1 history is reused, not just the original prompt."""
+    p = _proposer(block_size=16, num_blocks=64)
+    history = list(range(40)) + list(range(200, 224))  # 64 = 4 full blocks
+    _warm(p, "r1", history)
+    p._prune_finished({})
+    turn2 = history + list(range(300, 310))
+    p2 = p._make_plan("r2", _state(turn2), 4)
+    assert p2 is not None
+    assert p2.draft_seq_len == 64
+    assert p2.ingest_tokens == turn2[64:]
+
+
+def test_reused_blocks_never_come_from_free_pool():
+    """A reused block must not simultaneously sit in the pool's free queue
+    (would be handed out again and overwritten under another request)."""
+    p = _proposer(block_size=16, num_blocks=64)
+    _warm(p, "r1", list(range(40)))
+    p._prune_finished({})
+    p2 = p._make_plan("r2", _state(list(range(40))), 4)
+    assert p2 is not None
+    free_ids = {b.block_id for b in p._pool.free_block_queue.get_all_free_blocks()}
+    for blk in p2.block_ids[:2]:
+        assert blk not in free_ids
+
+
+def test_same_content_different_history_not_reused():
+    """A block whose tokens reappear under a different preceding block must
+    not be reused: its KV was computed attending to a different history, and
+    the chained hash (parent-dependent) is what prevents the false hit."""
+    p = _proposer(block_size=16, num_blocks=64)
+    _warm(p, "r1", list(range(40)))  # r1 blocks: [0..16), [16..32)
+    p._prune_finished({})
+    # Same second-block tokens (16..32) behind a different first block.
+    shifted = list(range(100, 116)) + list(range(16, 32)) + list(range(200, 208))
+    p2 = p._make_plan("r2", _state(shifted), 4)
+    assert p2 is not None
+    assert p2.draft_seq_len == 0
+    assert p2.ingest_tokens == shifted
--- a/tests/test_draft_prefix_cache.py
+++ b/tests/test_draft_prefix_cache.py
@@ -1,0 +1,95 @@
+"""Tests for draft-KV prefix caching in DraftModelProposer (#482 direction 1)."""
+
+from types import SimpleNamespace
+
+from vllm_metal.v1.draft_model_proposer import DraftModelProposer
+
+
+def _proposer(block_size: int = 16, num_blocks: int = 64) -> DraftModelProposer:
+    # _make_plan / _hash_prefix_blocks / allocator logic never touch the model,
+    # controller, or extract_logits, so stubs are fine for these tests.
+    return DraftModelProposer(
+        model=None,
+        block_size=block_size,
+        num_blocks=num_blocks,
+        num_layers=1,
+        controller=None,
+        extract_logits=None,
+    )
+
+
+def test_hash_prefix_blocks_full_only_deterministic_and_chained():
+    p = _proposer(block_size=16)
+    toks = list(range(40))  # 2 full blocks (32) + 8 trailing partial
+
+    hashes = p._hash_prefix_blocks(toks)
+    assert len(hashes) == 2  # only full blocks are hashed
+    assert p._hash_prefix_blocks(toks) == hashes  # deterministic
+
+    # Chaining: same first block but a different second block -> same h[0],
+    # different h[1] (h[1] depends on h[0] as parent, plus its own tokens).
+    toks_b = list(range(16)) + list(range(100, 124))
+    hb = p._hash_prefix_blocks(toks_b)
+    assert hb[0] == hashes[0]
+    assert hb[1] != hashes[1]
+
+    # A different first block changes every downstream hash.
+    toks_c = list(range(50, 90))
+    hc = p._hash_prefix_blocks(toks_c)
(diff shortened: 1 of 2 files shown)


abcgco:
Validated on an M4 Pro (24 GB): the fix does what it says. Branch (74ed6d8) vs its own base (0e10f10), same environment (mlx 0.31.2 / mlx-lm 0.31.3), greedy, K=3, draft==target==Qwen3-0.6B, 96 output tokens, identical resubmitted prompt under a fresh request id:

| prefix | metric (resubmit) | base | #500 |
|---|---|---|---|
| 8k | first propose | 3985 ms | **489 ms** |
| 8k | first plan | ingest 8193 / reuse 0 | ingest **1** / reuse **8192** |
| 8k | e2e ms/tok | 79.8 | **43.7** |
| 2k | first propose | 903 ms | **37 ms** |
| 2k | first plan | ingest 2049 / reuse 0 | ingest **1** / reuse **2048** |
| 2k | e2e ms/tok | 31.0 | **20.3** |

Each timing cell is a single engine boot; the structural columns (ingest / reuse counts) are deterministic. The timing deltas (8-100x) dwarf the boot-to-boot variance we've measured (the cold first-propose itself ranges ~4.0-5.7 s across this session's boots). All e2e here is conc=1 resubmission — concurrent shared-prefix reuse is exercised by the PR's refcount unit tests, not end-to-end on this box.

- **Token identity**: outputs are token-identical between base and branch at both prefixes — full greedy token-id sequences compared across all 12 generations (6 engine boots), not just hashes.
- **Cleanest pair** (fresh engine, prompt submitted twice, nothing else): the second-ever submission goes 4543 ms first propose / 85.3 ms/tok on base to **41 ms / 34.6 ms/tok** on the branch.
- **True cold start**: with nothing cached the first plan is ingest 8193 / reuse 0 on both trees — no false reuse, and the cold-path cost is unchanged within boot-to-boot noise (188 base vs 160 branch ms/tok). Direction 3 stays a follow-on, as scoped.
- **First-reuse propose**: 35-59 ms in 4 of 5 samples; one run showed 489 ms at 8k — still 8x under base, but worth knowing it varies.
- **Steady state untouched**: propose median is 97-98 ms on the branch vs 96-100 on base at 8k — the per-propose hashing doesn't show up, and Problem 2 is intact as scoped.
- **Tests**: the PR's 5 pass here. The regression shape from the thread (no full-prompt draft prefill once warm) is exactly what the first-plan row asserts end-to-end, and `test_reuses_cached_prefix_after_finish` covers it at plan level. Four extra adversarial cases pass on the branch — and all four fail on the base, as regression tests should (below; they reuse your file's helpers): **same block content behind a different history is not reused** — the chained-hash property that keeps reuse position-safe; mid-block divergence reuses exactly the matching leading run; multi-turn history (prompt + generated tokens) is reused, which quietly makes the multi-turn case work; a reused block never sits in the free pool.
- Benchmark-design note for anyone reproducing: a `max_tokens=1` warm generate already seeds the draft cache (propose runs for it too), so cold timings need a fresh engine.

One logistical note: the branch base (0e10f10) is now 8 commits behind main — including the #487/#492 decode-path changes and the #496 kernel fix — so the numbers above isolate this change against its own base; worth a rebase before merge.

<details><summary>4 proposed test additions (pass on 74ed6d8, fail on 0e10f10; drop into tests/test_draft_prefix_cache.py — they reuse its helpers)</summary>

```python
def test_partial_prefix_match_reuses_only_leading_run():
    """A prompt diverging mid-way must reuse exactly the matching leading
    run of blocks and ingest everything after the divergence."""
    p = _proposer(block_size=16, num_blocks=64)
    _warm(p, "r1", list(range(40)))  # blocks 0,1 cached
    p._prune_finished({})
    tokens_b = list(range(16)) + list(range(100, 140))  # shares block 0 only
    p2 = p._make_plan("r2", _state(tokens_b), 4)
    assert p2 is not None
    assert p2.draft_seq_len == 16  # block 0 reused, block 1 diverged
    assert p2.ingest_tokens == tokens_b[16:]


def test_multi_turn_history_reuses_generated_blocks():
    """Turn 2 resubmits turn 1's prompt + generated tokens as its prefix.
    Full blocks containing generated content are registered too, so the
    whole turn-1 history is reused, not just the original prompt."""
    p = _proposer(block_size=16, num_blocks=64)
    history = list(range(40)) + list(range(200, 224))  # 64 = 4 full blocks
    _warm(p, "r1", history)
    p._prune_finished({})
    turn2 = history + list(range(300, 310))
    p2 = p._make_plan("r2", _state(turn2), 4)
    assert p2 is not None
    assert p2.draft_seq_len == 64
    assert p2.ingest_tokens == turn2[64:]


def test_reused_blocks_never_come_from_free_pool():
    """A reused block must not simultaneously sit in the free pool (would be
    handed out again and overwritten under another request)."""
    p = _proposer(block_size=16, num_blocks=64)
    _warm(p, "r1", list(range(40)))
    p._prune_finished({})
    p2 = p._make_plan("r2", _state(list(range(40))), 4)
    assert p2 is not None
    for blk in p2.block_ids[:2]:
        assert blk not in p._free_blocks


def test_same_content_different_history_not_reused():
    """A block whose tokens reappear under a different preceding block must
    not be reused: its KV was computed attending to a different history, and
    the chained hash (parent-dependent) is what prevents the false hit."""
    p = _proposer(block_size=16, num_blocks=64)
    _warm(p, "r1", list(range(40)))  # r1 blocks: [0..16), [16..32)
    p._prune_finished({})
    # Same second-block tokens (16..32) behind a different first block.
    shifted = list(range(100, 116)) + list(range(16, 32)) + list(range(200, 208))
    p2 = p._make_plan("r2", _state(shifted), 4)
    assert p2 is not None
    assert p2.draft_seq_len == 0
    assert p2.ingest_tokens == shifted
```
</details>

<details><summary>Validation harness (bench_pr500.py)</summary>

```python
"""PR #500 validation: before/after draft-KV prefix caching (issue #482, direction 1).

Two timed generates of the SAME prompt in one process:
  gen1_cold      -- fresh request, cold draft cache
  gen2_resubmit  -- identical prompt, new request id
On the base every new request pays the full-prompt draft re-ingest inside its
first propose; with #500 gen2 must reuse the cached prefix blocks and ingest
only the uncached suffix. Greedy, token ids recorded for identity checks.

The warm 1-token generate at the start heats the TARGET prefix cache only:
max_tokens=1 finishes on the prefill step, so propose() never runs and the
draft cache stays cold for gen1 on both trees.

  VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_MEMORY_FRACTION=0.4 \
    python bench_pr500.py <prefix_tokens>
"""

import hashlib
import json
import os
import sys
import time

from vllm_metal.v1 import draft_model_proposer as dmp

PREFIX = int(sys.argv[1]) if len(sys.argv) > 1 else 8192
K = 3
GEN = 96
MODEL = "Qwen/Qwen3-0.6B"

DURS: list[float] = []
PLANS: list[tuple[int, int]] = []  # (draft_seq_len, n_ingest) per plan

_orig_propose = dmp.DraftModelProposer.propose
_orig_make_plan = dmp.DraftModelProposer._make_plan


def _timed_propose(self, ctx):
    t0 = time.perf_counter()
    r = _orig_propose(self, ctx)  # self-evals its drafts before returning
    DURS.append(time.perf_counter() - t0)
    return r


def _logged_make_plan(self, req_id, state, num_speculative_tokens):
    plan = _orig_make_plan(self, req_id, state, num_speculative_tokens)
    if plan is not None:
        PLANS.append((plan.draft_seq_len, len(plan.ingest_tokens)))
    return plan


dmp.DraftModelProposer.propose = _timed_propose
dmp.DraftModelProposer._make_plan = _logged_make_plan

llm = LLM(
    model=MODEL,
    max_model_len=PREFIX + 256,
    max_num_seqs=1,
    enable_prefix_caching=True,
    async_scheduling=False,
    speculative_config={
        "method": "draft_model",
        "model": MODEL,
        "num_speculative_tokens": K,
    },
)
tok = llm.get_tokenizer()
UNIT = (
    "The engineer examined the trace carefully, noting where the latency "
    "spiked and which subsystem held the lock the longest before yielding. "
)
prompt = UNIT
while len(tok.encode(prompt)) < PREFIX:
    prompt += UNIT
prompt = tok.decode(tok.encode(prompt)[:PREFIX])

if not os.environ.get("SKIP_WARM"):
    llm.generate([prompt], SamplingParams(temperature=0, max_tokens=1))  # warm target

results = []
sp = SamplingParams(temperature=0, max_tokens=GEN)
for label in ("gen1_cold", "gen2_resubmit"):
    DURS.clear()
    PLANS.clear()
    t0 = time.perf_counter()
    out = llm.generate([prompt], sp)
    dt = time.perf_counter() - t0
    ids = list(out[0].outputs[0].token_ids)
    ms = sorted(d * 1000 for d in DURS)
    first_plan = PLANS[0] if PLANS else (None, None)
    results.append(
        {
            "label": label,
            "prefix": PREFIX,
            "gen": len(ids),
            "wall_s": round(dt, 3),
            "tpot_ms": round(dt / len(ids) * 1000, 2),
            "propose_calls": len(ms),
            "propose_first_ms": round(DURS[0] * 1000, 1) if DURS else None,
            "propose_median_ms": round(ms[len(ms) // 2], 1) if ms else None,
            "first_plan_reused_tokens": first_plan[0],
            "first_plan_ingest_tokens": first_plan[1],
            "token_sha": hashlib.sha256(str(ids).encode()).hexdigest()[:16],
            "token_ids": ids,
        }
    )
    sys.stderr.write(f"=== {label} done ===\n")
    sys.stderr.flush()

print("PR500_RESULT " + json.dumps(results))
```
</details>

KrxGu:
@WindChimeRan @LxYuan0420 can we get this pr merged? i have few other prs waiting in pipeline.

WindChimeRan:
@KrxGu  Thanks. Will review now!

WindChimeRan (commented):
Verified locally on
* correctness: output identical to non-speculative decoding. 
* effectiveness: Output throughput on a shared-prefix workload at conc=8 (TPOT −15.1%, TTFT −34.3%). Neutral on the #447 dataset.

The issue is true, the design is sound and the result is solid. 

Looking at the implementation details now.

vllm_metal/v1/draft_model_proposer.py
```diff
@@ -260,23 +326,36 @@ def _make_plan(
...
+            blocks.append(blk)
+            self._refcount[blk] = self._refcount.get(blk, 0) + 1
         return blocks
 
+    def _alloc_block(self, req_id: str, needed: int) -> int:
+        """A free physical draft block: from the pool, else by evicting an
+        idle (refcount-0) cached block, else fail loudly."""
+        if self._free_blocks:
+            return self._free_blocks.pop()
+        for h, blk in list(self._cached.items()):
```
WindChimeRan: `for h, blk in reversed(self._cached.items())`

Eviction destroys the chain root first

vllm_metal/v1/draft_model_proposer.py
```diff
@@ -117,6 +120,14 @@ def __init__(
...
         self._free_blocks: list[int] = list(range(num_blocks))
         self._req_blocks: dict[str, list[int]] = {}
+        # Content-based prefix cache so shared/repeated draft prefixes reuse
+        # draft KV blocks instead of re-prefilling the prompt (#482 direction 1).
+        self._caching_hash_fn = sha256_cbor
+        init_none_hash(self._caching_hash_fn)
+        self._cached: dict[Any, int] = {}
+        self._refcount: dict[int, int] = {}
+        self._block_hash: dict[int, Any] = {}
+        self._prefix_reuse_tokens = 0
```
WindChimeRan: _prefix_reuse_tokens dead variable.

tests/test_draft_prefix_cache.py
```diff
@@ -0,0 +1,95 @@
...
+def test_shared_prefix_block_not_freed_while_referenced():
+    p = _proposer(block_size=16, num_blocks=64)
+    tokens = list(range(40))
+    _warm(p, "r1", tokens)  # r1 caches 2 blocks, still live
+    p2 = p._make_plan("r2", _state(tokens), 4)  # r2 reuses them
+    assert p2.draft_seq_len == 32
+    reused = p2.block_ids[:2]
+    p._prune_finished({"r2": object()})  # r1 finishes, r2 still live
+    for blk in reused:
+        assert blk not in p._free_blocks  # shared blocks survive
```
WindChimeRan: it never failed.

vllm_metal/v1/draft_model_proposer.py
```diff
@@ -239,11 +267,49 @@ def _collect_draft_plans(
...
+            hashes.append(h)
+            parent = h
+        return hashes
+
     def _make_plan(
         self, req_id: str, state: RequestState, num_speculative_tokens: int
     ) -> _DraftPlan | None:
         committed_len = len(state.token_ids)
         draft_seq_len = self._draft_seq_lens.get(req_id, 0)
+        prefix_hashes = self._hash_prefix_blocks(state.token_ids[:committed_len])
```
WindChimeRan: `_hash_prefix_blocks `: The whole prompt is re-hashed on every step

WindChimeRan (commented):
Suggest to start with "can we reuse vLLM's BlockPool instead of hand-rolling?" 

Try to simplify the code and reuse or import vllm code instead of re-implementation.

KrxGu:
Rebased onto main (was 20 commits behind) and swapped the hand-rolled allocator for vLLM's own BlockPool — no reason to reimplement hash caching/refcounting/eviction that already exists. Also made the hashing/registration incremental instead of rescanning every step, fixed the weak test, killed the dead counter, and pulled in the 4 regression tests from this thread.

vllm_metal/v1/draft_model_proposer.py
KrxGu: Went with BlockPool instead of the local reversed() fix. Its free queue already evicts a chain's tail before its root when you free blocks tail-first — same thing single_type_kv_cache_manager.py does internally. Figured better to fix it the way vLLM already does it than patch around it.

vllm_metal/v1/draft_model_proposer.py
KrxGu: Gone — it was write-only, nothing ever read it.

tests/test_draft_prefix_cache.py
KrxGu: Yeah, looked into it — any block sitting in the prefix cache never lands in _free_blocks regardless of refcount, so that assert couldn't fail even if refcounting was broken. Switched it to check ref_cnt directly on the block, which actually depends on the refcount logic being right.

vllm_metal/v1/draft_model_proposer.py
KrxGu: Fixed — hashing now picks up where it left off per request instead of rehashing from token 0 every step (committed tokens never change once written, so that's safe). _register_prefix had the same issue, fixed that too.

KrxGu:
@WindChimeRan @LxYuan0420 CI is all green, LGTM!!

KrxGu:
@WindChimeRan @LxYuan0420 can we get this merged?

KrxGu:
Hey @WindChimeRan @LxYuan0420 is there any problem with this PR to get merged? Happy to iterate further.

LxYuan0420:
Thanks for updating this to use vLLM’s `BlockPool`; that was the right allocator direction. I still do not think this PR is mergeable because the draft cache is still owned in the proposer/worker path, while the real cache policy lives in the scheduler/request layer.

The proposer does not see `cache_salt`, per-request cache-read rules, or the full allocation limit. So it can reuse KV when it should not, and it can still exhaust draft blocks near the request limit.

I am going to close this PR. The right direction is a scheduler-managed draft cache, or a typed scheduler-to-proposer contract that carries the same cache policy and limits.

PR #501, opened 2026-07-13
Bump MLX to 0.32.0

LxYuan0420:
This PR is:
- To bump the exact MLX runtime pin from `0.31.2` to `0.32.0`.
- To compile the native Metal bridge against the current MLX command-encoder API instead of keeping old compatibility shims
- To trim stale RoPE workaround comments/tests now that the MLX scalar-offset fix used by #496 is in the pinned wheel

note:

PR #496 added the packed decode RoPE batching path and used vector offsets to avoid the MLX 0.31.2 scalar-offset corruption. MLX 0.32.0 includes the upstream fix for that bug, so this branch moves the exact pin forward and removes the local code that only existed to support the old pinned wheel.


ricky-chaoju (approved):
Thanks @LxYuan0420 !

PR #502, opened 2026-07-13
[Spec Decode] Route small draft ingests down the decode path

abcgco:
Problem 2 in #482: steady-state propose costs ~95 ms/step at an 8k prefix while a complete nospec engine step of the same model is ~14 ms. Decomposing the propose graph (CPU fences around its single `mx.eval`, then sweeps over K and context) puts ~70 ms of that in the ingest forward alone: ~24 ms context-independent plus ~5.6 µs per context token (35.6 ms at 2k → 69.8 ms at 8k), while each single-token draft step costs ~10.7 ms at 8k on the split-KV path.

The cause is routing, not volume: the proposer submits its steady K+1-token ingest as a prefill segment, so the forward lands on the tiled prefill kernel, which reads the entire context regardless of query size. The target's verify rows have the same shape (a few fresh tokens over a long cache) and ride the decode path as expanded single-query segments. A dispatch probe confirms both sides: on main, steady ingests print the tiled-prefill dispatch; with this change they dispatch as N single-query decode segments — identical to verify rows — and no small ingest hits the tiled path.

The change: ingests of ≤ 16 tokens are submitted as decode specs, one row per token; larger ingests (the first-propose catch-up, where the tiled kernel is the right choice) keep the prefill path.

Measured on an M4 Pro 24 GB (greedy, K=3, prefix caching on, medians over in-run steady proposes):

| setup | propose/step before | after |
|---|---|---|
| draft==target==Qwen3-0.6B, 8k | 94–95 ms (3 boots) | **53.2 / 53.9 ms** (2 boots) |
| draft==target==Qwen3-0.6B, 2k | 48.6 ms | **38.7 ms** |
| Qwen3-0.6B draft + Qwen3-8B-4bit target, 8k prose | 93.7 ms | **49.3 ms** |

Output token ids are unchanged in every configuration (sha-256 over the full greedy id sequence matches the pre-change runs: both prefixes on the 0.6B setup, and the spec-path output on the 8B pair).

Also prototyped direction 2 from the issue (skip re-ingesting accepted tokens whose KV the draft steps already wrote): on top of this change it saves only ~3 ms — rows inside one decode dispatch largely amortize — so it is deliberately left out of scope.

Tests cover the routing decision and the prepared-context shape on both sides of the threshold, plus the mixed-batch case (one oversized plan keeps the whole forward on the prefill path, since the prepared context is global per forward).

### DIFF
--- a/vllm_metal/v1/draft_model_proposer.py
+++ b/vllm_metal/v1/draft_model_proposer.py
@@ -88,6 +88,13 @@ class _DraftPlan:
     ingest_tokens: list[int]
 
 
+# Ingests at or below this size are submitted as expanded decode rows instead
+# of a prefill segment (see _ingest_and_draft_first). Covers the steady-state
+# K+1-token ingest for any practical num_speculative_tokens while keeping
+# full-prompt catch-up ingests on the tiled prefill kernel.
+_DECODE_INGEST_MAX_TOKENS = 16
+
+
 class DraftModelProposer:
     """:class:`vllm_metal.v1.proposer.MetalProposer` backed by a separate model."""
 
@@ -280,10 +287,6 @@ class DraftModelProposer:
     def _ingest_and_draft_first(
         self, plans: list[_DraftPlan], offset_caches: list[OffsetCache]
     ) -> mx.array:
-        prefill_specs = [
-            (plan.block_ids, len(plan.ingest_tokens), plan.draft_seq_len)
-            for plan in plans
-        ]
         packed: list[int] = []
         last_rows: list[int] = []
         for plan in plans:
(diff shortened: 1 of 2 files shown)


abcgco:
Merged main in after #534 landed — no conflicts, and the interaction is clean: this PR's small-ingest routing submits the expanded decode form (one single-query segment per token), which #534 leaves untouched — it's the reference side of the new window-parity tests; the window mode itself applies to the target's verify batches. Routing + window-parity + paged-attention + YOCO suites: **158 passed, 1 skipped** on the 16-core M4 Pro at `b54c2c0`.

Adopting window mode for the ingest itself (sharing K/V block loads across the K+1 ingest rows, same as verify) looks like a natural follow-up once this lands — happy to take it.

vllm_metal/v1/draft_model_proposer.py
```diff
@@ -88,6 +88,13 @@ class _DraftPlan:
     ingest_tokens: list[int]
 
 
+# Ingests at or below this size are submitted as expanded decode rows instead
+# of a prefill segment (see _ingest_and_draft_first). Covers the steady-state
+# K+1-token ingest for any practical num_speculative_tokens while keeping
+# full-prompt catch-up ingests on the tiled prefill kernel.
+_DECODE_INGEST_MAX_TOKENS = 16
```
LxYuan0420, line 95: Worth doing more benchmark tests before deciding this magic number

LxYuan0420 (approved):
LGTM; I think this is a focused fix for #482  problem 2

PR #504, opened 2026-07-14
[Distributed] Route PP dummy forwards through the stage shape

ricky-chaoju:
Our profiling does not behave like serving under pipeline parallelism: profile_run and warm_up run the full top-level model call on every rank, so a middle stage materializes the embedding and head it never uses and computes a full-vocab logits buffer it never produces. On a Qwen3-8B-4bit middle stage (rank 1 of 3) that is 2.014 GB peak against 1.309 GB for the stage shape.

This PR adds PipelinedModel.dummy_forward, which runs the stage locally with a zeros hidden state built from the wire descriptor instead of the ring recv, and routes the runner's dummy forward through it. The stage body is shared with call, so the wire-dtype fail-fast now fires at startup instead of on the first request. This mirrors upstream vLLM's _dummy_run with make_empty_intermediate_tensors on non-first ranks.

For evidence, python tools/pp_lazy_rss.py mlx-community/Qwen3-8B-4bit dummy-full|dummy-stage 3 1 reproduces 2.014 vs 1.309 GB, the 2-stage ring parity check stays bit-exact, and the full not-slow suite passes. At 70B the saving is shape arithmetic, not measured; the 2-node validation pass is the follow-up. Generic mlx_lm text path under pipeline parallelism only.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -47,6 +47,7 @@ def make_stub_runner(
         "_draft_token_ids": None,
         "_execute_model_state": None,
         "pp": None,
+        "_pp_model": None,
         "_model_adapter": DefaultModelAdapter(),
         "_spec_decode_controller": SpeculativeDecodeController(),
         "kv_heads_per_layer": None,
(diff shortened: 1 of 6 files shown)


LxYuan0420 (approved):
Thanks for working on this part!

PR #506, opened 2026-07-14
Bump vLLM to 0.25.1

ricky-chaoju:
This bumps vLLM from 0.24.0 to 0.25.1. I went to 0.25.1 rather than 0.25.0 because 0.25.0 pulls in torchcodec (vllm#46609), whose import raises `RuntimeError` when there is no system FFmpeg, which is the normal state of a Mac. vLLM 0.25.0 guards that import with `except ImportError` only, so it escapes and `import vllm.multimodal` dies. `vllm/v1/serial_utils.py` imports it, so the whole v1 serve path goes with it. vllm#47888 widens the guard and landed in 0.25.1, so the bump needs no workaround.

Three upstream contracts moved and the rest of the diff absorbs them. vLLM 0.25 pads a newly admitted decode request up to the uniform speculative width and hands back `[-1] * num_spec_tokens` with no invalid-token accounting (`pad_spec_decode`, `v1/core/sched/scheduler.py`), so `SpeculativeDecodeController.active_spec_decode_tokens()` drops that exact signature and all three consumers read through it. `Platform.update_block_size_for_backend` gained a heterogeneous-KV phase gated on `cache_config.kv_cache_dtype_skip_layers`, which upstream fills in for any `--kv-cache-dtype turboquant_*`, so `check_and_update_config` rejects it up front instead of letting two owners resize the same block pool from inside a spawned worker. And `SpeculativeConfig.use_heterogeneous_vocab` made `verify_equal_vocab_size_if_draft_model()` conditional, so a draft model with a different vocabulary is newly reachable and `check_and_update_config` rejects it. A padded request clipped by `long_prefill_token_threshold` breaks both the padding signature and upstream's own token accounting, so that combination is rejected there as well.

No TODOs or compat shims come out with this bump, and the fastapi cap stays. Its TODO said to drop it once we pin a release containing vllm#45629, and that fix is already an ancestor of 0.24.0, so the condition was met before this bump started. vLLM 0.25.1 still caps `fastapi[standard] < 0.137.0` itself though, for a different and still-live reason, so I rewrote the comment to name the real gate rather than dropping the bound. The Ray scaffolding in `platform.py` and the GGUF plugin shim are not vLLM-gated and are untouched.

Benchmark with `vllm bench serve` on the sonnet dataset, Qwen3-0.6B, 100 prompts, `--temperature 0 --ignore-eos --seed 0`.

| Metric | vLLM 0.24.0 | vLLM 0.25.1 | Delta |
| --- | ---: | ---: | ---: |
| Output tok/s | 1614.61 | 1647.95 | +2.1% |
| Total tok/s | 7454.76 | 7608.72 | +2.1% |
| Mean TTFT (ms) | 1867.66 | 1691.52 | -9.4% |
| P99 TTFT (ms) | 1873.26 | 1696.24 | -9.4% |
| Mean TPOT (ms) | 49.77 | 49.69 | -0.1% |
| P99 TPOT (ms) | 49.77 | 49.69 | -0.1% |
| Duration (s) | 9.29 | 9.10 | -2.0% |


WindChimeRan:
~~could you please check if it’s compatible with #508 ? (it’s around the latest main)~~

I'll review it tonight.

vllm_metal/v1/structured_output.py
```diff
@@ -85,7 +85,9 @@ def apply_paged(
         structured_req_ids = set(grammar_output.structured_output_request_ids)
         spec_req_ids = {
             req_id
-            for req_id, tokens in scheduler_output.scheduled_spec_decode_tokens.items()
+            for req_id, tokens in SpeculativeDecodeController.active_spec_decode_tokens(
```
LxYuan0420, line 88: This passes the wrong row count into `_build_constrained_rows`. Filtered spec tokens are right for Metal logits rows, but grammar bitmask rows still follow raw `scheduler_output.scheduled_spec_decode_tokens`.

Suggested shape:

```python
raw_spec_tokens = scheduler_output.scheduled_spec_decode_tokens
grammar_row_counts = {
    req_id: 1 + len(raw_spec_tokens.get(req_id, ()))
    for req_id in grammar_output.structured_output_request_ids
}

constrained = self._build_constrained_rows(
    grammar_output.structured_output_request_ids,
    row_targets,
    grammar_bitmask_row_count=grammar_bitmask.shape[0],
    grammar_row_counts=grammar_row_counts,
)
```

Then delete `consume_row_spans`; the helper can always advance by raw grammar span and only pair the logits rows that exist. Add a mixed active+padded structured test.

WindChimeRan (commented):
Tested on #508  wheel. Also tested spec decoding using the setting `Qwen3-8B target + Qwen3-0.6B draft, K=3, greedy, spec_bench prompts`

e2e works on both non-spec and spec path

LxYuan0420 (approved):

Issue #436, opened 2026-06-09
[Roadmap] Precompiled wheels: zero-compile install for vllm-metal


WindChimeRan:
Every reported install failure (#413, #435) is a compile step running on the user's machine. This roadmap removes them all.

- [x] **Step 1 — Ship vllm-metal binaries prebuilt** (#423). No more Metal/C++ JIT at first run.
- [x] **non-blocking** — fix vllm-cpu macos for longer sequence. https://github.com/vllm-project/vllm/pull/45649
- [x] **Step 2 — Upstream macOS arm64 vllm wheel** ([vllm#37216](https://github.com/vllm-project/vllm/pull/37216)). Removes the vllm-from-sdist build (the failure in #435). Status: build recipe verified on current main; blocked on a macOS Buildkite runner for the release pipeline.
- [ ] **Step 3 — One-command install**: a fresh Mac without Xcode CLT can run `pip install vllm-metal`. From https://pypi.org/project/vllm-metal/

ericcurtin:
Thanks for opening this, the main blocker has been this "Upstream macOS arm64 vllm wheel", without that, we can't avoid the build steps

ericcurtin:
@mgoin anything blocking this https://github.com/vllm-project/vllm/pull/37216 ?

reneleonhardt:
37216 has been closed, the new PR for macOS arm64 wheels is https://github.com/vllm-project/vllm/pull/46660

WindChimeRan:
close because #508 merged. This is the best we can do for now. Pypi publish requires more upstream work in both vllm and mlx_lm. So pypi is not planned.

reneleonhardt:
Great to hear! 🚀

Are vllm and mlx_lm efforts being coordinated, are there corresponding issues?

PR #508, opened 2026-07-14
Prebuilt vllm wheel

WindChimeRan:
pre-built vllm-cpu-macos wheel! now local toolchain is not needed. 

@mgoin Thanks for making it happen!!

We are very close to #436 !

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,9 +5,10 @@
 - macOS on Apple Silicon
 - Native arm64 Python 3.12. Rosetta/x86_64 Python is not supported.
 
-> **No compiler required.** vLLM core and the vllm-metal plugin both install as
-> prebuilt wheels, so nothing is compiled on your machine. Contributors editing
-> the Metal kernels need a toolchain; see [Contributing](CONTRIBUTING.md).
+> **No compiler required.** The install script below fetches vLLM core and the
+> vllm-metal plugin as prebuilt wheels, so nothing is compiled on your machine.
+> Installing from a source checkout instead builds the native Metal kernels
+> locally and needs a toolchain; see [Contributing](CONTRIBUTING.md).
 
 `uv` is bootstrapped automatically.
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -4,15 +4,13 @@
 
 - macOS on Apple Silicon
 - Native arm64 Python 3.12. Rosetta/x86_64 Python is not supported.
-- Xcode Command Line Tools (`xcode-select --install`). The installer compiles vLLM core from source via `clang++`.
 
-> **Metal kernels are prebuilt.** Release wheels ship the native paged-attention
-> extension (`_paged_ops*.so`) and its precompiled Metal shader libraries
-> (`*.metallib`), so `pip install` requires no compiler or Metal toolchain to
-> *run* the kernels — nothing is compiled on first request. The Command Line
-> Tools above are still needed because the installer builds vLLM core from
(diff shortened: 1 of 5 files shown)


install.sh
```diff
@@ -1,5 +1,15 @@
...
 
+# Prebuilt macOS arm64 vLLM core wheel. Installing this is what keeps the
+# installer compiler-free — vLLM used to be built here from its sdist.
+#
+# Pinned by full URL because there is nothing to resolve against: wheels.vllm.ai
+# publishes macOS wheels only under a commit's flat /<sha>/<file> path, never in
+# its PEP 503 indexes, so --extra-index-url cannot see them.
+#
+# To bump: paste the macOS wheel URL of a newer vllm main commit.
+VLLM_WHEEL_URL="https://wheels.vllm.ai/56557d98703c30190ddb70de23ed7d0297474564/vllm-0.23.1rc1.dev958%2Bg56557d987.cpu-cp312-cp312-macosx_11_0_arm64.whl"
```
LxYuan0420: I like the move to a prebuilt vLLM wheel. One thing I think we should tighten: the commit-shaped URL pins where we fetch from, but not the exact bytes we install. Could we add a SHA-256 or provenance check before `uv pip install`?

LxYuan0420: One concern: this wheel reports `vllm==0.23.1rc1.dev958+g56557d987.cpu`, while #506 moves us to the vLLM 0.25.1 contract. Can we rebase after #506 and point this at a 0.25.1-compatible wheel, or clearly prove this wheel has the scheduler behavior #506 relies on?

docs/installation.md
```diff
@@ -4,15 +4,10 @@
...
-
-> **Metal kernels are prebuilt.** Release wheels ship the native paged-attention
-> extension (`_paged_ops*.so`) and its precompiled Metal shader libraries
-> (`*.metallib`), so `pip install` requires no compiler or Metal toolchain to
-> *run* the kernels — nothing is compiled on first request. The Command Line
-> Tools above are still needed because the installer builds vLLM core from
-> source. Building the Metal kernels yourself is only for contributors editing
-> them; see [Contributing](CONTRIBUTING.md).
+
+> **No compiler required.** vLLM core and the vllm-metal plugin both install as
```
LxYuan0420: Maybe scope this wording to the release/curl install path. Local checkout install still does editable/native-artifact build steps, so contributors may still need the toolchain.

install.sh
WindChimeRan: already tested. the contract is on for both versions. (0.23.1rc1.dev is actually ahead of 0.25.1)

And yes. we need to land #506 first.

LxYuan0420: I did a quick local check and confirmed this wheel has the scheduler contract we need. One concern: it is not simply ahead of `v0.25.1`; GitHub compare shows it diverged with `ahead_by: 96`, `behind_by: 19`.

I think this can work, but I would prefer not to make it the default install path until we have a stable macOS wheel matching the vLLM version we support. “Ahead” is not the same as “compatible with the stable release.”

WindChimeRan: SHA-256 would be overkill. Will check if we can get a stable version wheel.

WindChimeRan: Let's pin it to 0.26.0! Tracking PR here https://github.com/vllm-project/vllm/pull/48771

WindChimeRan:
@LxYuan0420 vllm 0.26.0 wheels uploaded!

LxYuan0420 (approved):
Thanks for working on this!

PR #509, opened 2026-07-15
Fix stale AWQ slow-test model configs

### LINKED ISSUE #507: AWQ e2e slow tests broken
Found this while validating the 0.25.1 bump, but it reproduces on main with vLLM 0.24.0 too, so it is not bump fallout.

pytest tests/quant/ -m slow fails with:

    tests/quant/test_awq_e2e_qwen25_15b.py:61: in test_awq_load_aligns_non_quant_dtypes_to_runner_target
    vllm_metal/v1/model_lifecycle.py:193: in _load_generation_model
    vllm_metal/gguf/source.py:20: in from_model_config
    E   AttributeError: 'types.SimpleNamespace' object has no attribute 'quantization'

#467 added the GGUF probe to _load_generation_model, and GGUFLoadSource.from_model_config reads model_config.quantization directly. The AWQ e2e stubs predate that (#461) and never set the field, so every lifecycle load in those tests now dies before reaching the checkpoint. Same for the three parametrized cases in test_awq_e2e_dense_matrix.py. Nobody noticed because CI doesn't run slow tests.

The production read is fine, that's how lifecycle is supposed to read it. The fix is two lines: the _runner_model_config helpers in test_awq_e2e_qwen25_15b.py and test_awq_e2e_dense_matrix.py set quantization=None, same as the stub in test_model_lifecycle.py:88 already does.

PinJhih:
The GGUF load-source probe added in #467 reads model_config.quantization, but the AWQ lifecycle stubs predate that field.
Set quantization=None so the dtype-alignment tests reach the AWQ checkpoints again.

Tests: pytest tests/quant/ -m slow -v --tb=short
(5 passed, 39 deselected)

Fixes #507

### DIFF
--- a/tests/quant/test_awq_e2e_dense_matrix.py
+++ b/tests/quant/test_awq_e2e_dense_matrix.py
@@ -74,6 +74,7 @@ def _runner_model_config(repo: str, *, dtype):
         is_multimodal_model=False,
         trust_remote_code=False,
         dtype=dtype,
+        quantization=None,
     )
 
 
--- a/tests/quant/test_awq_e2e_qwen25_15b.py
+++ b/tests/quant/test_awq_e2e_qwen25_15b.py
@@ -46,6 +46,7 @@ def _runner_model_config(*, dtype):
         is_multimodal_model=False,
         trust_remote_code=False,
         dtype=dtype,
+        quantization=None,
     )
 
 


ricky-chaoju (approved):
LGTM. Thanks for the quick fix!

PR #510, opened 2026-07-15
Fail fast on model warm-up errors

LxYuan0420:
This PR is:
- To stop swallowing exceptions from `MetalModelRunner.warm_up()`
- To fail startup at the original warm-up error when dummy forward/eval breaks
- To add a focused regression test for warm-up error propagation

Note: Warm-up is an internal startup validation path, not an optional fallback. If it fails, continuing only hides the real issue until a later request path


ricky-chaoju (approved):
LGTM. Thanks!

Issue #505, opened 2026-07-14
M1 Ultra 128GB RAM fails to load https://huggingface.co/google/gemma-4-31B-it


ericcurtin:
If anybody has the hardware to debug this, it would be appreciated:

https://github.com/docker/model-runner/issues/1009

LxYuan0420:
@ricky-chaoju, if you still have access to your M2 Ultra, could you try reproducing this with google/gemma-4-31B-it and report back with the full startup traceback?

ricky-chaoju:
I'll take a look later.

ricky-chaoju:
Reproduced on my M2 Ultra 192GB on current main with VLLM_METAL_MEMORY_FRACTION=0.5 vllm serve <path-to-gemma-4-31B-it>. The EngineCore dies at startup with:

ValueError: To serve at least one request with the model's max seq len (262144), (220.0 GiB KV cache is needed, which is larger than the available KV cache memory (20.13 GiB). Based on the available memory, the estimated maximum model length is 23984.

max_model_len derives to 262144 from the config, and we report all 60 layers as full attention, so the 50 sliding layers (window 1024) get charged at full length too. That is 220 GiB for one max-length request, which no 128GB Mac can fit at any memory fraction. --max-model-len 16384 serves fine as a workaround.

I'll submit a PR later that defaults an unset --max-model-len to vLLM's auto-fit, so the engine boots at the largest context that fits instead of failing.

LxYuan0420:
Thanks Ricky. I do not think making unset `--max-model-len` behave like auto is the right long-term fix. The real fix is our memory estimate: we are charging Gemma 4 sliding-window layers like full-attention layers, so startup thinks the KV cache needs far more memory than it should.

For now, Docker/users can pass `--max-model-len auto` or a smaller value like `16384`. That lets vLLM pick a context length that fits the machine. Could you try the `auto` path?

ricky-chaoju:
Tried auto, it works: boots at the largest length that fits and greedy output matches mlx_lm. Agreed on dropping the default change.

One caveat before we point users at auto: the fitted length lands exactly on pool capacity, but the block pool reserves one block as the null placeholder, so a request at exactly that length hangs forever. Same code upstream; I'll file an issue there. The one-block fix is the second commit in #511.

[...]

PR #511, opened 2026-07-15
Default an unset --max-model-len to vLLM auto-fit

ricky-chaoju:
gemma-4-31B-it fails engine init on any 128GB Mac (#505, docker/model-runner#1009). The config derives `max_model_len=262144` and the packed KV layout prices all 60 layers at full length, so the admission check asks for 220 GiB before anything can serve.

This PR defaults an omitted `--max-model-len` to vLLM's own auto-fit on the paged path, by setting `original_max_model_len` to `-1`, the same value `--max-model-len auto` writes (vllm-project/vllm#29431). The engine then boots at the largest context that fits and the worker warns with the old and new lengths. Explicit values are untouched. Spec decode keeps the old behavior because our worker does not refresh drafter lengths after a reduction.

The second commit reserves BlockPool's null block in the auto-fit estimate. Without it, the fitted length lands exactly on pool capacity and a request at that length waits forever; I hit this live while validating. The bug exists upstream too; I'll file it there and the SCAFFOLDING patch goes away once it's fixed.

For evidence, the main one is the before/after serve: main dies with the 220 GiB error, this branch boots at 23968 and greedy chat answers normally. Greedy token ids match the mlx_lm reference, an explicit 16384 stays 16384, and the 31B goldens pass. Ran on my M2 Ultra 192GB at `VLLM_METAL_MEMORY_FRACTION=0.5`. Full 262144 still needs sliding-window KV groups, which I'll write up as a design issue.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -42,3 +42,11 @@ model pairing, and memory considerations.
 | `auto` | `1` | Yes | Paged KV path (default); defaults to 0.9 internally |
 | `0.7` | `1` | Yes | Paged KV path with explicit memory budget |
 | `0.7` | `0` | No | Explicit fraction without paged KV is invalid |
+
+On the paged KV path, an unset `--max-model-len` defaults to vLLM's auto-fit:
+if the model's full context does not fit the KV budget, the engine starts at
+the largest length that fits and logs the reduction (upstream CUDA fails
+instead). Pass `--max-model-len` to pin the context length; explicit values
+are never adjusted. Speculative-decoding runs keep the derived default and
+must pin `--max-model-len` themselves. Auto-fit maximizes single-sequence
+context, so pinning a smaller value restores batching headroom.
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -70,6 +70,7 @@ class TestMetalPlatform:
                 disable_custom_all_reduce=False,
             ),
             model_config=None,
+            speculative_config=None,
         )
         with pytest.raises(
             NotImplementedError, match="alone or combined with pipeline"
@@ -97,6 +98,7 @@ class TestMetalPlatform:
                     model="test-model",
                     disable_cascade_attn=False,
                     tokenizer=None,
+                    original_max_model_len=None,
                     max_model_len=32768,
                     multimodal_config=None,
                     hf_config=SimpleNamespace(model_type="qwen3"),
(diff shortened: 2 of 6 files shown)


LxYuan0420 (commented):
I suggest we close this one and start fresh with a narrower fix at the right layer: the KV memory calc

ricky-chaoju:
The narrow fix at the KV calc is actually already in here — it's the second commit. The auto-fit estimate doesn't account for the block pool's reserved null block, so auto fits to a length the scheduler can never serve, and a request at exactly that length hangs forever.

Happy to strip this PR down to just that.

LxYuan0420:
Thanks Ricky. I think we are talking about two different calculations.

The #505 fix I mean is Gemma 4 KV accounting/layout: 50 sliding-window layers and 10 full-attention layers are currently priced as if all 60 layers need full `262144` context.

The null-block auto-fit issue may be real, but it is a separate upstream auto-fit bug. Please file that upstream first; if we need a local workaround, we should do it in a separate PR with a clear removal condition.

ricky-chaoju:
Closing this one. The null-block fix is up upstream as vllm-project/vllm#48724 with a plain CPU-backend before/after repro. I'll send the one-block reservation as a separate PR here with #48724 as the removal condition, so `--max-model-len auto` is safe on Metal in the meantime.

For the Gemma 4 fix, the shape I have in mind: emit `SlidingWindowSpec` for the 50 sliding layers so vLLM prices them at the window instead of full context (220 GiB -> ~27 GiB at 262144). That switches on the hybrid KV manager with per-group block tables, so the runner has to pick each layer's group table instead of reading `block_ids[0]`, and the packed pool needs a per-group layout so sliding blocks can actually be reclaimed.
@LxYuan0420 WDYT?

LxYuan0420:
> Closing this one. The null-block fix is up upstream as [vllm-project/vllm#48724](https://github.com/vllm-project/vllm/pull/48724) with a plain CPU-backend before/after repro. I'll send the one-block reservation as a separate PR here with #48724 as the removal condition, so `--max-model-len auto` is safe on Metal in the meantime.
> 
> For the Gemma 4 fix, the shape I have in mind: emit `SlidingWindowSpec` for the 50 sliding layers so vLLM prices them at the window instead of full context (220 GiB -> ~27 GiB at 262144). That switches on the hybrid KV manager with per-group block tables, so the runner has to pick each layer's group table instead of reading `block_ids[0]`, and the packed pool needs a per-group layout so sliding blocks can actually be reclaimed. @LxYuan0420 WDYT?

Sounds reasonable. Let's try it in a focused PR and test it properly. probably something with a long prompt that exercise the sliding window path

PR #513, opened 2026-07-15
Reserve the null block in the vLLM auto-fit estimate

ricky-chaoju:
--max-model-len auto fits the context to the full KV pool, but vLLM's BlockPool keeps one block as the null placeholder, so a request at exactly the fitted length can never get its last block and waits forever. I hit this while validating auto for #505: at fraction 0.5 the engine fitted 262144 down to 19216 (= 1201 blocks x 16), and a request at that length sat in Waiting at 0% KV cache usage indefinitely.

The upstream fix is up as vllm-project/vllm#48724. Until it merges and reaches our pinned vLLM, this carries a small compat patch that shrinks the auto-fit search budget by one pool block. It installs with the other compat patches and is re-ensured from update_block_size_for_backend, because plugin activation can run while vLLM is still mid-import and silently skip the patch. SCAFFOLDING-marked with #48724 as the removal condition.

With the patch, the same serve fits to 23968 and a completion at exactly that total length finishes instead of hanging. The tests drive vLLM's real get_kv_cache_configs and pin the fitted-length-equals-usable-blocks contract plus the one-block-pool fail-fast. 1466 unit tests and mypy clean. Only the auto path changes; explicit --max-model-len values are untouched.

### DIFF
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -11,7 +11,10 @@ import torch
 from vllm.config import ParallelConfig
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
 from vllm.v1.attention.selector import AttentionSelectorConfig
+from vllm.v1.core.kv_cache_utils import get_kv_cache_configs, max_memory_usage_bytes
+from vllm.v1.kv_cache_interface import FullAttentionSpec, KVCacheSpec
 
+from vllm_metal.compat import ensure_vllm_auto_fit_null_block_patch
 from vllm_metal.config import reset_config
 from vllm_metal.platform import MetalPlatform
 from vllm_metal.v1.cache_policy import WorkerCachePlanner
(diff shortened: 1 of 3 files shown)


LxYuan0420 (commented):
Thanks Ricky. The upstream fix looks valid, but im hesitant to add another private vLLM patch here. Lets wait for upstream feedback before deciding whether we need a temporary local workaround.

ricky-chaoju:
I think we should keep this one. Even if #48724 merges fast, we only get it at the next vLLM bump, and we just did 0.25.1 yesterday. That's weeks of the hang sitting on the path we're pointing #505 users at. I'll delete the patch myself in the bump PR that picks up the fix, and if upstream changes shape it's a one-commit revert here. WDYT?

LxYuan0420 (commented):
ok with keeping this patch as short term patch but keep the removal condition documented. Another thing I want to clarify before merge: what is the tradeoff between patching _estimate_max_model_len_from_groups here versus patching the same _auto_fit_max_model_len layer as upstream?

ricky-chaoju:
Both layers can be wrapped, I just went with the estimator because the wrapper stays dumb: it gets the groups and the budget together, subtracts one block, and calls the original. Wrapping `_auto_fit` means redoing its per-worker pairing inside the wrapper, which breaks quietly if upstream ever changes that loop. Either way we double-reserve once a bump picks up #48724, until the patch is deleted.

The estimator also feeds the "estimated maximum model length" suggestion in the admission error, so with the wrap that suggestion becomes a value you can actually pin without hanging. If you want the patch strictly auto-only, I can wrap `_auto_fit` instead.

LxYuan0420 (approved):

PR #514, opened 2026-07-15
Wire up --gpu-memory-utilization on the Metal paged KV path

### LINKED ISSUE #512: --gpu-memory-utilization has no effect on Metal (paged KV path)
### What

On Apple Silicon with paged attention, `--gpu-memory-utilization` has no effect.
In auto mode (`VLLM_METAL_MEMORY_FRACTION=auto`, the default),
`WorkerCachePlanner._memory_fraction` (`vllm_metal/v1/cache_policy.py`) returns a
hardcoded `0.9` and never reads `cache_config.gpu_memory_utilization`:

```python
def _memory_fraction(self) -> float:
    if self._worker.metal_config.is_auto_memory:
        ...
        return PAGED_ATTENTION_DEFAULT_MEMORY_FRACTION   # 0.9, ignores the flag
    return self._worker.metal_config.memory_fraction
```

### Why it's confusing

`--gpu-memory-utilization` is the standard vLLM flag for the KV-cache budget. On
Metal it is dropped without a warning in two cases:

1. When the env var is unset (the default) and the CLI flag is set, the flag is
   ignored and `0.9` is used.
2. When both are set, the env var overrides the CLI flag, which reverses the
   usual CLI-over-env precedence.

Both resolve to `0.9` by default, so the flag looks like it works until you set
a different value and nothing changes.

### Repro

Model-independent: the flag never reaches `_memory_fraction`, which runs for
every model on the paged KV path. Serving any model with an explicit fraction:

```bash
vllm serve <any-model> --gpu-memory-utilization 0.5
```

The startup log reports the fixed default regardless of the flag value:

```
Paged attention: VLLM_METAL_MEMORY_FRACTION=auto, defaulting to 0.90 for paged path
```

`VLLM_METAL_MEMORY_FRACTION=0.5` changes it; `--gpu-memory-utilization 0.5` does not.

### Impact

The `0.9` default reserves ~90% of usable unified memory for the KV cache. On a
memory-constrained Mac this pushes the rest of the system into swap. Lowering
`--gpu-memory-utilization` does nothing; the knob that works is
`VLLM_METAL_MEMORY_FRACTION`, which is not obvious from the flag's behavior.

### Suggestion

A warning is probably enough. When `cache_config.gpu_memory_utilization` is set
to a non-default value but ignored (either case above), log that it has no
effect on Metal and point to `VLLM_METAL_MEMORY_FRACTION`.

### Environment

- Apple Silicon (Metal / paged attention).
- Code pointer: `vllm_metal/v1/cache_policy.py::WorkerCachePlanner._memory_fraction`.
  Present on current `main` (`75241ab`); `gpu_memory_utilization` does not appear
  anywhere under `vllm_metal/`.

mhdimo:
## Problem

On Apple Silicon (paged attention), `--gpu-memory-utilization` had no effect: `WorkerCachePlanner._memory_fraction()` returned a hardcoded `0.9` in auto mode and never read `cache_config.gpu_memory_utilization` (#512). The flag was silently dropped, so the standard vLLM knob for the KV-cache budget did nothing on Metal — only the non-obvious `VLLM_METAL_MEMORY_FRACTION` env var worked.

## Fix

`_memory_fraction()` now resolves the fraction with explicit precedence:

1. `VLLM_METAL_MEMORY_FRACTION` set explicitly (not `auto`) → wins as the Metal-specific knob; warns if `--gpu-memory-utilization` is also set to a conflicting value.
2. `auto` (the default) → honors `cache_config.gpu_memory_utilization`, so the standard vLLM flag controls the Metal KV-cache budget.
3. Falls back to `PAGED_ATTENTION_DEFAULT_MEMORY_FRACTION` only if `gpu_memory_utilization` is unavailable.

Startup logging is now source-explicit (`using gpu_memory_utilization=…` / `using VLLM_METAL_MEMORY_FRACTION=…`) instead of the misleading `defaulting to 0.90`.

## Behavior change

Plain `auto` runs (no flag, no env var) move `0.9 → 0.92` (vLLM's `gpu_memory_utilization` default). The flag now provides the lowering lever requested in the issue.

## Verification

- `tests/test_v1_worker.py::test_memory_fraction_precedence` — 5 parametrized cases covering auto + `--gpu-memory-utilization 0.5` and explicit `VLLM_METAL_MEMORY_FRACTION` precedence.
- Live serve: `vllm serve --gpu-memory-utilization 0.5` → `fraction=0.5`, `kv_budget=11.59 GB`, `num_blocks=6314` (≈half of the old 0.9 path), confirming the flag drives the real allocation.
- `ruff`, `mypy`, `shellcheck` clean; `pytest -m "not slow" tests/` green.

Closes #512.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,7 +4,7 @@
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `VLLM_METAL_MEMORY_FRACTION` | `auto` | `auto` allocates just enough memory plus a minimal KV cache, or `0.?` for fraction of memory |
+| `VLLM_METAL_MEMORY_FRACTION` | `auto` | Metal memory budget mode; see [Paged KV vs MLX KV Memory Settings](#paged-kv-vs-mlx-kv-memory-settings) |
 | `VLLM_METAL_USE_MLX` | `1` | Use MLX for compute (1=yes, 0=no) |
 | `VLLM_MLX_DEVICE` | `gpu` | MLX device (`gpu` or `cpu`) |
 | `VLLM_METAL_USE_PAGED_ATTENTION` | `1` | Enable experimental paged KV cache |
@@ -34,11 +34,11 @@ model pairing, and memory considerations.
 
 - MLX path (`VLLM_METAL_USE_PAGED_ATTENTION=0`): `VLLM_METAL_MEMORY_FRACTION` must be `auto`.
 - Paged KV path (`VLLM_METAL_USE_PAGED_ATTENTION=1`): `VLLM_METAL_MEMORY_FRACTION` can be `auto` or a numeric fraction in `(0, 1]`.
-- For paged KV with `VLLM_METAL_MEMORY_FRACTION=auto`, vllm-metal uses a default fraction of `0.9`.
+- For paged KV with `VLLM_METAL_MEMORY_FRACTION=auto`, vllm-metal uses vLLM's `--gpu-memory-utilization` value.
 
 | `VLLM_METAL_MEMORY_FRACTION` | `VLLM_METAL_USE_PAGED_ATTENTION` | Valid? | Notes |
 |--|--|--|--|
 | `auto` | `0` | Yes | MLX path |
-| `auto` | `1` | Yes | Paged KV path (default); defaults to 0.9 internally |
+| `auto` | `1` | Yes | Paged KV path (default); uses `--gpu-memory-utilization` |
 | `0.7` | `1` | Yes | Paged KV path with explicit memory budget |
 | `0.7` | `0` | No | Explicit fraction without paged KV is invalid |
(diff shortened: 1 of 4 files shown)


LxYuan0420 (approved):
Thanks for fixing this. I think the bug came from historical drift: Metal kept its old paged-KV memory path while vLLM’s --gpu-memory-utilization became the standard knob.

I cleaned it up so the precedence is smoother and explicit: numeric VLLM_METAL_MEMORY_FRACTION wins; otherwise auto uses vLLM’s --gpu-memory-utilization. With that, we no longer need the old PAGED_ATTENTION_DEFAULT_MEMORY_FRACTION.

mhdimo:
Thank you!!

javijiujitsu:
Your work on speculative decoding in vLLM is very close to what we’re building at [Particle](https://particle.ai/).

PR #519, opened 2026-07-16
Validate Metal boolean env vars

### LINKED ISSUE #517: Boolean Metal env vars silently ignore common true values
## Problem

Several `VLLM_METAL_*` boolean environment variables currently check only whether the raw string equals `"1"`. Common shell/config spellings such as `true`, `yes`, or `on` are therefore treated as false.

For example, setting `VLLM_METAL_USE_PAGED_ATTENTION=true` disables paged attention even though the user clearly intended to enable it. Invalid values such as `enabled` are also silently treated as false.

## Expected behavior

Boolean Metal env vars should accept common boolean spellings such as `1`/`0`, `true`/`false`, `yes`/`no`, and `on`/`off`, and should reject unrecognized values with a clear error.

## Impact

This can make configuration typos hard to diagnose, especially for paged attention and debug settings where the process continues with a different backend or logging behavior than requested.

harivilasp:
## Summary

- centralize parsing for Metal boolean environment variables
- accept common boolean spellings: `1`/`0`, `true`/`false`, `yes`/`no`, and `on`/`off`
- reject unrecognized boolean values with a clear `ValueError` instead of silently treating them as false
- document accepted boolean spellings in the configuration guide

Fixes #517.

## Why

The previous parser compared each boolean env var directly with `"1"`. That meant values like `VLLM_METAL_USE_PAGED_ATTENTION=true` disabled paged attention, and typos such as `enabled` also silently selected the false path.

## Tests

- `.venv-vllm-metal/bin/python -m pytest tests/test_config.py tests/test_macos_defaults.py`
- `ruff check vllm_metal/envs.py tests/test_config.py`
- `.venv-vllm-metal/bin/python -m py_compile vllm_metal/envs.py tests/test_config.py`
- `git diff --check`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -5,7 +5,7 @@
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `VLLM_METAL_MEMORY_FRACTION` | `auto` | `auto` allocates just enough memory plus a minimal KV cache, or `0.?` for fraction of memory |
-| `VLLM_METAL_USE_MLX` | `1` | Use MLX for compute (1=yes, 0=no) |
+| `VLLM_METAL_USE_MLX` | `1` | Use MLX for compute |
 | `VLLM_MLX_DEVICE` | `gpu` | MLX device (`gpu` or `cpu`) |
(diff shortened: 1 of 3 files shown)


harivilasp:
Closing per author request.

PR #520, opened 2026-07-16
Clear MLX cache after VLM warm-up

### LINKED ISSUE #515: [Performance] Reduce Qwen3-VL first-batch TTFT by clearing MLX cache after VLM warm-up
## Summary

Qwen3-VL multimodal serving on Apple Silicon appears sensitive to MLX buffer-cache state left after model warm-up. In a small paired A/B benchmark, clearing MLX cache once after a successful multimodal model warm-up reduced first-batch `random-mm` TTFT for Qwen3-VL.

This issue proposes adding a guarded cache release for VLMs only:

- run the normal model warm-up;
- only continue if warm-up succeeded;
- only apply to native multimodal adapters;
- only call `mx.clear_cache()` when `mx.get_cache_memory()` reports nonzero cache memory;
- log the released cache size.

This is a TTFT / prefill issue. It should not be treated as a decode throughput optimization.

## Motivation

I first benchmarked a Qwen3-VL vision-encoder batching candidate. That did not produce a stable TTFT improvement. Diagnostic timing suggested that, on this MLX/Metal stack, batched vision tower calls were only about 3% faster than serial steady-state calls for the tested image sizes. First-call cache / warm-up variance dominated the end-to-end result.

A narrower hypothesis performed better: release MLX cache after VLM warm-up so the first real multimodal requests start from a less cache-constrained state.

## Environment

Sanitized local test environment:

| Item | Value |
| --- | --- |
| Hardware | Apple Silicon, Apple M5, 16 GiB unified memory |
| macOS | 26.4 |
| Python | 3.12.12 |
| vLLM | 0.24.0+cpu |
| vllm-metal | 0.3.0 |
| MLX | 0.32.0 |
| mlx-vlm | 0.6.4 |
| Model | `mlx-community/Qwen3-VL-4B-Instruct-4bit` |
| Model snapshot | `2fd8dacbdb8f1e54b8c005f081ec5bf79c56376b` |
| Backend | vLLM OpenAI-compatible chat endpoint |

No private paths, usernames, or local cache directories are required to reproduce the benchmark.

## Benchmark Setup

Server shape:

```bash
VLLM_ENABLE_V1_MULTIPROCESSING=0 \
VLLM_METAL_MEMORY_FRACTION=0.70 \
GLOO_SOCKET_IFNAME=lo0 \
VLLM_METAL_BUILD_FROM_SOURCE=1 \
vllm serve mlx-community/Qwen3-VL-4B-Instruct-4bit \
  --host 127.0.0.1 \
  --port 8000 \
  --max-model-len 1024 \
  --max-num-seqs 4 \
  --max-num-batched-tokens 1024 \
  --generation-config vllm \
  --limit-mm-per-prompt '{"image": 1}'
```

Benchmark shape:

```bash
vllm bench serve \
  --backend openai-chat \
  --base-url http://127.0.0.1:8000 \
  --endpoint /v1/chat/completions \
  --model mlx-community/Qwen3-VL-4B-Instruct-4bit \
  --dataset-name random-mm \
  --num-prompts 4 \
  --max-concurrency 4 \
  --request-rate inf \
  --random-input-len 32 \
  --random-output-len 1 \
  --random-mm-bucket-config '{(320, 320, 1): 1.0}' \
  --ignore-eos \
  --temperature 0 \
  --metric-percentiles 50,90,99 \
  --percentile-metrics ttft,tpot,itl,e2el \
  --save-result \
  --save-detailed
```

I used ABBA ordering and restarted the server for every round:

1. baseline A1
2. candidate B1
3. candidate B2
4. baseline A2
5. baseline A3
6. candidate B3

I also added an 8 second delay after `/health` before running the benchmark client. Earlier runs showed that `/health` could respond before the server was reliably ready for chat-completion requests.

## Candidate

The lower-risk candidate is intentionally narrow:

```python
if warmed_up and self._multimodal_adapter is not None:
    cache_bytes = mx.get_cache_memory()
    if cache_bytes == 0:
        return
    mx.clear_cache()
    logger.info(
        "MLX cache cleared after multimodal model warm-up (%d bytes)",
        cache_bytes,
    )
```

This avoids changing text-only models, failed warm-ups, and empty-cache cases.

### Guarded cache release, `320x320`, max concurrency 4

| Run | Variant | Completed / Failed | Mean TTFT ms | P90 TTFT ms | Req/s | Total tok/s | Cache released |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| A1 | baseline | 4 / 0 | 5444.45 | 5445.72 | 0.734 | 104.99 | - |
| B1 | guarded cache release | 4 / 0 | 3220.82 | 3221.89 | 1.241 | 177.47 | 929205476 bytes |
| B2 | guarded cache release | 4 / 0 | 2952.95 | 2954.30 | 1.353 | 193.49 | 618048892 bytes |
| A2 | baseline | 4 / 0 | 3461.25 | 3463.10 | 1.154 | 165.06 | - |
| A3 | baseline | 4 / 0 | 3209.34 | 3211.29 | 1.245 | 177.98 | - |
| B3 | guarded cache release | 4 / 0 | 2048.40 | 2050.43 | 1.949 | 278.65 | 562477436 bytes |

Grouped:

| Metric | Baseline mean | Baseline std | Candidate mean | Candidate std | Change |
| Mean TTFT ms | 4038.35 | 1224.22 | 2740.72 | 614.35 | -32.13% |
| Request throughput req/s | 1.044 | 0.286 | 1.514 | 0.381 | +45.02% |
| Total token throughput tok/s | 149.34 | 40.85 | 216.54 | 54.55 | +45.00% |

Pairwise TTFT direction:

- B1 vs A1: `-40.84%`
- B2 vs A2: `-14.69%`
- B3 vs A3: `-36.17%`

The candidate was faster in 3 of 3 paired comparisons. The second pair narrowly missed a 15% single-pair threshold, so I would describe the result as a bounded TTFT improvement, not a universal speedup.

### Earlier unguarded prototype

The first prototype cleared cache unconditionally after VLM warm-up. It also looked positive:

| Workload | Baseline mean TTFT | Candidate mean TTFT | Change | Direction |
| --- | ---: | ---: | ---: | --- |
| `320x320`, max concurrency 2 | 1358.72 ms | 1060.45 ms | -21.95% | 2 / 3 faster |
| `320x320`, max concurrency 4 | 2989.97 ms | 2088.07 ms | -30.16% | 3 / 3 faster |

The guarded version is preferable because it keeps the same Qwen3-VL performance path while reducing side effects.

## Non-goals / risk boundaries

- This does not claim decode throughput improvement. The benchmark used `--random-output-len 1`.
- This does not claim a general vllm-metal speedup across all models.
- The benchmark is small: 4 prompts, synthetic `random-mm`, 320x320 images, low concurrency.
- The result may be strongest on memory-constrained Apple Silicon systems.
- More real-world workloads may behave differently, especially with larger images, mixed image sizes, longer prompts, or longer decode.

## Related observation: encoder batching was not enough

The original vision-encoder batching candidate did not pass the same performance gate:

- requested `128x128`, max concurrency 2: candidate mean TTFT was worse;
- forced `320x320`, max concurrency 4, original patch limit: grouped mean TTFT changed by only `-0.44%`;
- experimental patch limit to allow a 3-image tower call: grouped mean TTFT was `+1.62%` worse;
- steady-state timing suggested the batched Qwen3-VL vision tower call saved only about 3% for the tested `320x320` and `448x448` images.

This is why I think the cache-release path is a better issue / PR candidate than presenting encoder batching as the main performance win.

## Proposed acceptance criteria

- Add a guarded VLM-only MLX cache release after successful model warm-up.
- Keep text-only and STT paths unchanged.
- Log the released cache size when the clear happens.
- Add unit tests for:
  - multimodal adapter with nonzero cache clears once;
  - text-only model does not clear;
  - failed warm-up does not clear;
  - zero cache does not clear.
- Include paired A/B benchmark results in the PR, with the TTFT-only claim and the workload boundary above.

harivilasp:
## Summary

- clear MLX buffer-cache memory after a successful multimodal model warm-up when the cache is non-empty
- keep text-only warm-up behavior unchanged
- skip the clear when warm-up fails or when MLX reports zero cached bytes
- log the number of cache bytes released

Fixes #515.

## Why

Qwen3-VL first-batch multimodal prefill can start with substantial MLX buffer-cache memory left by warm-up. Releasing that cache after successful VLM warm-up gives the first real multimodal batch a cleaner memory state while avoiding text-only and failed-warm-up paths.

## Tests

- `PYTHONPATH=/Users/panjwaniha/experiment/vllm-docs-bf16-env .venv-vllm-metal/bin/python -m pytest tests/test_v1_model_runner_generate.py -q`
- `PYTHONPATH=/Users/panjwaniha/experiment/vllm-docs-bf16-env .venv-vllm-metal/bin/python -m pytest tests/test_v1_worker.py -q`
- `ruff check vllm_metal/v1/model_runner.py tests/test_v1_model_runner_generate.py`
- `.venv-vllm-metal/bin/python -m py_compile vllm_metal/v1/model_runner.py tests/test_v1_model_runner_generate.py`
- `git diff --check`

### DIFF
--- a/tests/test_v1_model_runner_generate.py
+++ b/tests/test_v1_model_runner_generate.py
@@ -68,6 +68,30 @@ class ForwardOutputRuntimeStub:
         return None
 
 
+class FakeWarmupMX:
+    int32 = object()
+
+    def __init__(self, cache_bytes: int) -> None:
+        self.cache_bytes = cache_bytes
+        self.clear_calls = 0
+        self.eval_calls = 0
+
+    def array(self, values, *, dtype):
+        assert values == [[1, 2, 3]]
+        assert dtype is self.int32
+        return values
+
+    def eval(self, *outputs) -> None:
+        assert outputs
+        self.eval_calls += 1
+
+    def get_cache_memory(self) -> int:
+        return self.cache_bytes
+
+    def clear_cache(self) -> None:
+        self.clear_calls += 1
+
+
 def test_gemma4_mtp_config_installs_gemma4_proposer() -> None:
     runner = make_stub_runner(tokenizer=object())
     runner.vllm_config = SimpleNamespace(
@@ -97,6 +121,65 @@ class TestV1MetalModelRunnerGenerate:
 
         with pytest.raises(RuntimeError, match="dummy forward failed"):
             runner.warm_up()
+
+    def test_warm_up_clears_nonzero_mlx_cache_for_multimodal_model(
+        self, monkeypatch
+    ) -> None:
+        fake_mx = FakeWarmupMX(cache_bytes=123)
+        monkeypatch.setattr(mr, "mx", fake_mx)
+
+        runner = self._make_runner()
+        runner._multimodal_adapter = object()
+        runner._dummy_forward_outputs = Mock(return_value=[object()])
+
+        runner.warm_up()
+
+        assert fake_mx.eval_calls == 1
+        assert fake_mx.clear_calls == 1
+
+    def test_warm_up_keeps_text_only_mlx_cache(self, monkeypatch) -> None:
+        fake_mx = FakeWarmupMX(cache_bytes=123)
+        monkeypatch.setattr(mr, "mx", fake_mx)
+
+        runner = self._make_runner()
+        runner._dummy_forward_outputs = Mock(return_value=[object()])
+
+        runner.warm_up()
+
+        assert fake_mx.eval_calls == 1
+        assert fake_mx.clear_calls == 0
+
+    def test_warm_up_skips_multimodal_clear_when_mlx_cache_empty(
+        self, monkeypatch
+    ) -> None:
+        fake_mx = FakeWarmupMX(cache_bytes=0)
+        monkeypatch.setattr(mr, "mx", fake_mx)
+
+        runner = self._make_runner()
+        runner._multimodal_adapter = object()
+        runner._dummy_forward_outputs = Mock(return_value=[object()])
+
+        runner.warm_up()
+
+        assert fake_mx.eval_calls == 1
+        assert fake_mx.clear_calls == 0
+
+    def test_warm_up_does_not_clear_cache_after_dummy_forward_failure(
+        self, monkeypatch
+    ) -> None:
+        fake_mx = FakeWarmupMX(cache_bytes=123)
+        monkeypatch.setattr(mr, "mx", fake_mx)
+
+        runner = self._make_runner()
+        runner._multimodal_adapter = object()
+        runner._dummy_forward_outputs = Mock(
+            side_effect=RuntimeError("dummy forward failed")
+        )
+
+        with pytest.raises(RuntimeError, match="dummy forward failed"):
+            runner.warm_up()
+
+        assert fake_mx.clear_calls == 0
 
     def test_accumulates_streamed_segments(self, monkeypatch) -> None:
         captured: dict[str, object] = {}
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -758,10 +758,32 @@ class MetalModelRunner:
 
         dummy_tokens = mx.array([[1, 2, 3]], dtype=mx.int32)
         mx.eval(*self._dummy_forward_outputs(dummy_tokens))
-        logger.info("Model warm-up complete")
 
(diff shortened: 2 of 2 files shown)


harivilasp:
Closing per author request.

PR #521, opened 2026-07-16
Validate boolean VLLM_METAL_* env vars

### LINKED ISSUE #517: Boolean Metal env vars silently ignore common true values

varad-more:
## Summary

- Centralize boolean env-var parsing in `vllm_metal/envs.py` behind a `_bool_env` helper.
- Accept `1`/`0`, `true`/`false`, `yes`/`no`, `on`/`off` — case-insensitive, surrounding whitespace ignored.
- Unset or empty (e.g. a docker-compose `VAR=${UNSET}` expansion) keeps the variable's default.
- Any other value now fails loudly instead of silently reading as false:
  `ValueError: Invalid VLLM_METAL_DEBUG='enabled'. Must be one of: 1/0, true/false, yes/no, on/off (case-insensitive).`
- Document accepted spellings in `docs/configuration.md`; add `tests/test_envs.py` (~180 parametrized cases across all six boolean vars).

Fixes #517.

## Why

`VLLM_METAL_USE_PAGED_ATTENTION=true` currently *disables* paged attention: every boolean var is compared with `== "1"`. All six parse sites live in the lazy registry in `vllm_metal/envs.py`, and every consumer reads the already-parsed bool, so the helper fixes the whole plugin in one place. Same shape as the merged ring-port validation (#462).

## Behavior changes (intended)

| Input | Before | After |
|---|---|---|
| `true` / `yes` / `on` (any case) | `False` | `True` |
| `false` / `no` / `off` (any case) | `False` | `False` (unchanged) |
| unset | default | default (unchanged) |
| set but empty / whitespace | `False`, even for default-true vars | default |
| anything else (`enabled`, `2`, …) | silently `False` | `ValueError` naming the var, raised at first read |

Notes:
- The error surfaces where the var is first read — engine startup for most vars; the first absorbed-MLA eligibility check for `VLLM_METAL_MLA_KERNEL`.
- vLLM's `compile_factors` calls registered getters inside `try/except`, so an invalid value additionally logs the "Skipping environment variable … while hashing compile factors" warning; the hard error still fires at real use.

Deliberately untouched: `VLLM_USE_MODELSCOPE` (vLLM's own `True`/`False` convention) and `MTL_CAPTURE_ENABLED` (Apple's `=1` convention). Possible follow-ups: a `docs/configuration.md` table row for `VLLM_METAL_BUILD_FROM_SOURCE`; a friendlier error for a non-integer `VLLM_METAL_RING_BASE_PORT`.

## Testing

- `ruff check .`, `ruff format --check .` (210 files clean), `mypy vllm_metal` — all green.
- `pytest tests/test_envs.py tests/test_config.py` — 199 passed.
- Full `pytest -m "not slow" tests/` locally: the only failures (104, all in native Metal-kernel tests) are byte-identical with and without this change on the same machine — pre-existing local-environment findings (no fresh native build), zero new failures.
- Behavior smokes: `VLLM_METAL_USE_MLX=yes` → `True`; `VLLM_METAL_USE_PAGED_ATTENTION=ON` → `True`; `VLLM_METAL_USE_MLX=` (empty) → `True` (default); `VLLM_METAL_DEBUG=enabled` → the `ValueError` above.

Picks up where #518/#519 (self-closed by their author, never reviewed) left off, as a fresh implementation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -17,6 +17,15 @@
 | `VLLM_METAL_VISIBLE_DEVICES` | — | Set automatically by the Ray executor per worker (the device-control var); not user-configurable. See [Distributed](distributed.md). |
 | `VLLM_METAL_RING_BASE_PORT` | `32323` | Base TCP port for the MLX ring data plane under pipeline parallelism; stage *r* binds `base + r` (so the default is `32323`/`32324` for two stages). Set the **same** value on every node to move the ring off a busy port — e.g. when an `mlx.launch` job, a restart still in `TIME_WAIT`, or another PP job holds the default. See [Distributed](distributed.md#pipeline-parallelism). |
 
+### Boolean values
+
+Boolean `VLLM_METAL_*` variables (those with a `0`/`1` default above, plus
+`VLLM_METAL_BUILD_FROM_SOURCE`) accept `1`/`0`, `true`/`false`, `yes`/`no`,
+and `on`/`off`, case-insensitively. Leaving a variable unset or empty keeps
+its default. Any other value is rejected at first read with a `ValueError`
+naming the variable. (`VLLM_USE_MODELSCOPE` is vLLM's own variable and keeps
+vLLM's `True`/`False` convention.)
+
 ## Multimodal Serve Modes
 
 - `auto`: use the text-only compatibility path for checkpoints on the compatibility allowlist, such as Gemma4 and Qwen3.5/Qwen3.6 FP8 conditional-generation wrappers.
--- a/tests/test_envs.py
+++ b/tests/test_envs.py
@@ -1,0 +1,95 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for boolean environment variable parsing in ``vllm_metal.envs``."""
+
+import pytest
+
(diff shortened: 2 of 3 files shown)


LxYuan0420:
Thanks for working on this. The current `0`/`1` format is simple and already documented, so we do not think adding more formats is worth the extra code and maintenance. We will close this PR for now.

PR #524, opened 2026-07-18
Accept only numerical-tie draft rejections in spec e2e

ricky-chaoju:
#501 bumped MLX to 0.32 and it broke one line in the draft-model spec-decode e2e. With draft == target == Qwen3 0.6B, greedy, every draft should be accepted. On 0.31.2 that holds; on 0.32 "One plus one equals" drops to 6/9.

Not our code. The same checkout gives 9/9 on MLX 0.31.2 and 6/9 on 0.32.0. I logged the rejected draft: the target's top two logits are bit identical (17.125 vs 17.125), a hard tie, and the M=1 draft and M=K+1 verify forwards pick different sides. Output still matches the golden. FP tie at equal logits, not a bug.

### DIFF
--- a/tools/test_spec_decode_draft_model_e2e.py
+++ b/tools/test_spec_decode_draft_model_e2e.py
@@ -2,8 +2,9 @@
 """End-to-end gate for draft-model speculative decoding on Metal.
 
 With draft == target == ``Qwen/Qwen3-0.6B`` under greedy decoding, every draft
-token equals the target's own argmax, so all drafts are accepted and
-speculative decoding must reproduce plain greedy decoding token-for-token. The
+token equals the target's own argmax, so drafts are accepted and speculative
+decoding reproduces plain greedy decoding token-for-token, except where the
+draft and the target winner are a numerical tie the two forwards can split. The
 greedy goldens are the same ones ``test_paged_deterministic`` pins.
 
 Process isolation: build exactly ONE engine per process (target + draft model +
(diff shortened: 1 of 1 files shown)


LxYuan0420 (commented):
Can we make this test prompt-agnostic and tie-aware instead of hardcoding `"One plus one equals": 3`? That is more sustainable across future prompts and MLX releases.

A draft/target mismatch should still fail by default. For each mismatch, inspect the target verifier logits at that position and accept it only when the draft token is exactly tied with, or within a small named ULP window of, the target winner. Keep failing when there is a meaningful score gap.

Conceptually:

```python
if draft_token != target_token:
    assert is_numerical_tie(
        target_logits,
        draft_token,
        target_token,
    )
```

tools/test_spec_decode_draft_model_e2e.py
```diff
@@ -115,28 +148,47 @@ class TestDraftModelSpecDecode:
...
                 logits, decode_reqs, decode_segments, logitsprocs=logitsprocs
             )
+            # verify stops at the first per-segment mismatch, so accepted =
+            # len(output) - 1 locates the one rejected draft (if any).
             for segment, output_ids in zip(decode_segments, result, strict=True):
-                counts["drafted"] += len(segment.draft_token_ids)
-                counts["accepted"] += len(output_ids) - 1
+                drafts = segment.draft_token_ids
+                stats["drafted"] += len(drafts)
+                accepted = len(output_ids) - 1
```
LxYuan0420, line 173: Please keep an `accepted` counter and assert it is non-zero. `drafted > 0` only proves the proposer ran; a broken draft cache could still have every proposal corrected as a permitted near-tie.

tools/test_spec_decode_draft_model_e2e.py
```diff
@@ -59,6 +64,34 @@
...
+    """Whether the draft and the target winner are within MAX_TIE_ULPS bfloat16
+    ULPs — a tie the M=1 draft and M=K+1 verify forwards may split, not a real
+    mismatch."""
+    winner = float(logits_row[target_token])
+    other = float(logits_row[draft_token])
+    gap = winner - other
+    if gap <= 0.0:
+        return True
+    mag = max(abs(winner), abs(other))
+    if mag == 0.0:
```
LxYuan0420, line 88: probably unreachable; the last line comparison operator ( <= ) is enough

LxYuan0420 (approved):

LxYuan0420 (approved):

PR #529, opened 2026-07-20
Describe the KV layout that was actually allocated for MLA and YOCO

### LINKED ISSUE #528: KV capacity under-reported for MLA (0.50x) and YOCO (0.43x) layouts: get_kv_cache_spec describes a layout that isn't allocated
### Your current environment

- Apple M4, 32 GiB, macOS 15.7.4 — where this was found
- Apple M1 Max, 64 GiB, macOS 15.7.1 — where the measurements below were taken
- vllm-metal `main` (`12ee169`), vLLM `0.25.1`, MLX `0.32.0`, Python 3.12

### Describe the bug

For MLA and YOCO models the block count vLLM plans against is a fixed fraction of the pool vllm-metal actually allocated. The Metal buffers exist; the scheduler simply never addresses the remainder.

Measured on real weights, `--max-model-len 4096 --gpu-memory-utilization 0.5`:

| model | layout | pool | engine sees | ratio | max concurrency @4,096 |
|---|---|---:|---:|---:|---:|
| `mlx-community/MiniCPM3-4B-4bit` | MLA, 62L, latent 288 | 38,498 | 19,249 | **0.500** | 75.19x |
| `mlx-community/gemma-4-e2b-it-4bit` | YOCO, 35L, 15 owning | 73,582 | 31,535 | **0.4286** | 123.18x |

With the spec corrected (see below) both round-trip to 1.000, and concurrency rises to 150.02x and 287.43x respectively — the YOCO figure being exactly 35/15 of the original.

The ratios are shape-invariant: `1/2` for MLA, `num_owning_layers / num_layers` for YOCO. Gemma4 E4B (42L, 24 owning) computes to 0.571 by the same rule.

This is the complement of #526/#527: that guard rejects `config.num_blocks > pool`, and its message says *"otherwise this is a capacity-accounting bug, please report it."* The under-report direction passes silently.

### Root cause is the spec, not the byte budget

The byte budget is correct in both cases — it describes what was really allocated. `available_memory = pool_blocks * get_cache_block_size_bytes()` (`cache_policy.py:666-667`), and the engine converts back via `available // page_size // group_size` (`kv_cache_utils.py:967-987`). The disagreement is that `get_kv_cache_spec` describes a layout vllm-metal does not allocate.

**MLA.** `cache_policy.py:302-309` emits a plain `FullAttentionSpec` for every attention layer, including MLA. `AttentionSpec.real_page_size_bytes` (`vllm/v1/kv_cache_interface.py:196-202`) hardcodes a leading `2 *` for K and V. But MLA stores a **single** latent tensor: `MLAPagedLatentCache` allocates one `mx.zeros((num_blocks, block_size, latent_dim))` per layer (`attention/caches/mla_cache.py:37`), with `latent_dim = kv_lora_rank + qk_rope_head_dim` already carrying the rope part, and `model_lifecycle.py:349-353` sets `num_kv_heads=1, head_size=latent_dim`.

So `_kv_factor()` returning 1 for MLA (`cache_policy.py:566-567`) is right, and it contradicts the spec emitted a few hundred lines earlier in the same file. vLLM's divisor ends up 2x the truth.

Upstream already has the matching class: `MLAAttentionSpec` (`vllm/v1/kv_cache_interface.py:363`) omits the `2 *`. vllm-metal never instantiates it — though `TurboQuantAttentionSpec`'s own docstring (`cache_policy.py:65-66`) already cites it as the pattern being followed.

**YOCO.** `get_kv_cache_spec` loops `range(self._runner.num_layers)` and emits one spec per layer, while `_mha_cache_layout` (`cache_policy.py:569-579`) and `model_lifecycle.py:360-362` allocate only `num_cache_layers` unique caches. The sharing is invisible to vLLM, so `group_size` is the full layer count against a smaller pool. The mapping is installed unconditionally at load (`model_lifecycle.py:143,356-361`) via `build_yoco_cache_mapping`, which keys only on `num_kv_shared_layers` — no flag, no opt-in.

Uniform attention round-trips because `_kv_factor()==2` matches the spec's `2 *`; per-layer shapes still round-trip once they collapse to `UniformTypeKVCacheSpecs`.

### Impact beyond lost concurrency

The pool is allocated before it is reported — `determine_available_memory` calls `setup_paged_attention` → `backend.initialize(plan.num_blocks)` (`cache_policy.py:610`), *then* reports. So this is allocated-but-unreachable memory.

It also breaks length fitting. With `max_model_len=8192` against a 1000-block MLA pool, startup does not merely under-report — it fails:

```
ValueError: ... 1.07 GiB KV cache is needed, which is larger than the available
KV cache memory (1.05 GiB) ... estimated maximum model length is 8000
```

Same specs, same byte budget; changing only the spec class yields `num_blocks=1000`.

### Scope — what is not affected

Hybrid GDN round-trips exactly. Serving `Qwen/Qwen3.5-0.8B` (24 layers, `full_attention_interval=4` → 18 linear / 6 full):

```
[cache_policy.py:610] Paged attention enabled: 24 layers patched, 308 blocks allocated (block_size=544)
[cache_policy.py:315] KV cache config received: 308 blocks
```

Grouping is `[6 Mamba][6 Mamba][6 Mamba][6 FullAttention]`, so `group_size == num_sdpa_layers` and the budgets agree. Real hybrids always derive a 1:(k−1) ratio from `full_attention_interval`, so this holds generally.

Hybrid GDN with `n_attn < n_mamba` can show a similar ratio for synthetic layer counts, but that is a different thing and I am not proposing a fix for it: vLLM assumes one unified pool serving both groups, while vllm-metal's paged pool holds only SDPA pages, with GDN recurrent state reserved lazily *outside* it (`cache_policy.py:709-710, 743-744`). No shipped hybrid layout actually diverges.

### Suggested fix

Fix the spec, not the byte budget. Inflating the budget would encode a known-false byte count into `estimate_max_model_len` and the log line at `cache_policy.py:668-674`.

1. **MLA** — emit `MLAAttentionSpec` when `self._runner.is_mla`, with the same field values passed today. `num_kv_heads` is already 1 and `head_size` already the latent dim, so no dimension change is needed.
2. **YOCO** — emit **no spec at all** for layers that share a cache, i.e. iterate only the leading `num_cache_layers`. This mirrors upstream, which expresses sharing by omission: `GPUModelRunner.get_kv_cache_spec` skips such layers "*so that KV cache management logic will act as this layer does not exist*" (`vllm/v1/worker/gpu_model_runner.py:7574-7585`).

`initialize_kv_cache` consumes only `kv_cache_config.num_blocks` and ignores `kv_cache_groups`/`layer_names` (`cache_policy.py:313-334`), and MLX routes layers via its own `cache_idx_map`, so omitting layer names changes no addressing.

I checked this isn't deliberate conservatism: no comment or docstring rationalizes it, `git log -L` on the relevant lines shows only refactors (#402, #472, #277, #273), and #513 pins a "fitted-length-equals-usable-blocks" contract with one block of sanctioned slack — not 43–50%.

**#529 implements both fixes** with regression tests, and carries the measured before/after above.

<sub>Note: `docs/supported_models.md:84` lists the Gemma 4 reference checkpoint as `mlx-community/gemma-4-E2B-it`, which returns 401. The published MLX conversions are lowercase — `mlx-community/gemma-4-e2b-it-4bit` is what these numbers come from.</sub>

hyunnnchoi:
## Purpose

Fixes #528.

`determine_available_memory` reports the paged pool to the engine as a **byte budget**, and vLLM converts it back to a block count by dividing by the page size its own specs imply (`kv_cache_utils.py:967-987`). For uniform attention both sides agree and the count round-trips exactly. For MLA and YOCO, `get_kv_cache_spec` describes a layout the plugin does not allocate, so the engine plans against a fraction of the pool while the Metal buffers sit allocated and unreachable.

**MLA.** `MLAPagedLatentCache` allocates a single latent tensor per layer (`attention/caches/mla_cache.py:37`) — which is why `_kv_factor()` bills it at 1 (`cache_policy.py:566-567`). But `get_kv_cache_spec` emitted a plain `FullAttentionSpec`, whose `real_page_size_bytes` hardcodes the 2x K/V factor (`vllm/v1/kv_cache_interface.py:196-202`). vLLM's divisor came out twice the truth and halved the block count.

This emits `MLAAttentionSpec` instead — upstream's class for exactly this layout, which omits the `2 *` (`kv_cache_interface.py:363`). It is registered with `FullAttentionManager` and `uniform_type_base_spec=FullAttentionSpec` (`single_type_kv_cache_manager.py:1539`) and is included in the block-table / prefix-caching type checks (`:274,307`), so grouping and addressing are unchanged. The MLA-specific group-splitting path is gated on `SlidingWindowMLASpec`, which this never emits. Fields are left at their defaults — `cache_dtype_str`, `alignment`, `compress_ratio` are DeepSeek-V3.2/V4 FlashMLA concerns and are inert here.

`TurboQuantAttentionSpec`'s own docstring already cites `MLAAttentionSpec` as the pattern it mirrors (`cache_policy.py:65-66`); this applies the same pattern to MLA itself.

**YOCO.** Only the leading `num_cache_layers` own a cache and the rest reuse it (`_mha_cache_layout`, `cache_policy.py:569-579`; `model_lifecycle.py:360-362`), but a spec was emitted for every layer, so `group_size` counted layers that hold nothing.

This emits specs only for the owning layers, mirroring upstream, which expresses sharing by omission — `GPUModelRunner.get_kv_cache_spec` skips such layers "*so that KV cache management logic will act as this layer does not exist*" (`vllm/v1/worker/gpu_model_runner.py:7574-7585`). `initialize_kv_cache` consumes only `kv_cache_config.num_blocks` and ignores `kv_cache_groups`/`layer_names` (`cache_policy.py:313-334`), and MLX routes layers via its own `cache_idx_map`, so dropping the shared layer names changes no addressing. `build_yoco_cache_mapping` assigns owners first, so the sharers are exactly the trailing layers and `_cache_layer_shapes` already slices `[:num_cache_layers]`.

Deliberately **not** changing the byte budget: it is physically correct, and inflating it would encode a known-false figure into `estimate_max_model_len` and the breakdown log at `cache_policy.py:668-674`.

## Test Plan

New `tests/test_kv_cache_spec_capacity.py` asserts the spec describes the pool that was allocated, by replicating vLLM's sizing (`available // page_size // group_size`) against a fixed 1000-block pool:

- `test_mla_spec_matches_single_latent_cache` — MLA emits `MLAAttentionSpec` and round-trips
- `test_yoco_shared_layers_get_no_spec` — only owning layers appear
- `test_plain_attention_spec_is_unchanged` — the common path is untouched

```bash
pytest tests/test_kv_cache_spec_capacity.py -q
pytest tests/test_v1_worker.py tests/test_per_layer_kv_cache.py tests/test_kv_cache_override_guard.py -q
```

## Test Result

Engine `num_blocks` against a 1000-block pool:

| layout | before | after |
|---|---:|---:|
| MLA, 61L latent 576 | 500 (0.500) | **1000** |
| YOCO — Gemma4 E2B (35L, 15 owning) | 428 (0.428) | **1000** |
| YOCO — Gemma4 E4B (42L, 24 owning) | 571 (0.571) | **1000** |
| uniform full attention, 36L | 1000 | 1000 (unchanged) |

Without the fix the new tests fail as expected (`assert 35 == 15`). With it: `3 passed`. Regression set: `28 passed`. `ruff check` and `ruff format --check` clean on both files.

I also verified the hybrid path is unaffected by actually serving `Qwen/Qwen3.5-0.8B` — 308 blocks allocated, 308 received by the engine, exact round-trip.

**Validated on real weights** (M1 Max 64 GB, macOS 15.7.1): `mlx-community/MiniCPM3-4B-4bit` goes 0.500 → 1.000 and `mlx-community/gemma-4-e2b-it-4bit` goes 0.4286 → 1.000, with max concurrency at 4,096 tokens rising 75.19x → 150.02x and 123.18x → 287.43x. Outputs are identical before and after, and 24 concurrent requests completed cleanly against each. Measured numbers are in the comment below.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -59,6 +59,7 @@ def make_stub_runner(
         "_logitsprocs": None,
         "_structured_output_applier": MetalStructuredOutputApplier(),
         "_lora": MetalLoRARuntime(),
+        "_yoco_cache_mapping": None,
         "model_args": _model_args,
     }
 
--- a/tests/test_kv_cache_spec_capacity.py
+++ b/tests/test_kv_cache_spec_capacity.py
@@ -1,0 +1,78 @@
+# SPDX-License-Identifier: Apache-2.0
+"""The scheduler-visible spec must describe the pool that was allocated."""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import mlx.core as mx
+import torch
+from vllm.v1.core.kv_cache_utils import get_uniform_page_size
+from vllm.v1.kv_cache_interface import FullAttentionSpec, MLAAttentionSpec
+
+from tests.stub_runner import make_stub_runner
+
+BLOCK_SIZE = 16
+DTYPE = torch.bfloat16
+POOL_BLOCKS = 1000
+
+
+def _engine_blocks(specs: dict, metal_per_block: int) -> int:
+    """Replicate vLLM's sizing: available_memory // page_size // group_size."""
+    available = POOL_BLOCKS * metal_per_block
+    return available // get_uniform_page_size(list(specs.values())) // len(specs)
+
+
+def _policy(*, num_layers, is_mla=False, yoco=None, num_kv_heads=4, head_dim=256):
+    # ``is_mla``/``is_hybrid`` are derived from the model config, not settable.
+    model_args = {"kv_lora_rank": head_dim} if is_mla else {}
+    runner = make_stub_runner(
+        model_args=model_args,
+        num_layers=num_layers,
+        _yoco_cache_mapping=yoco,
+        num_kv_heads=num_kv_heads,
+        head_dim=head_dim,
+        kv_cache_dtype=mx.bfloat16,
+        cache_config=SimpleNamespace(block_size=BLOCK_SIZE),
+        kv_heads_per_layer=None,
+        head_dim_per_layer=None,
+    )
+    return runner._cache_policy
+
+
+def test_mla_spec_matches_single_latent_cache() -> None:
+    """MLA stores one latent tensor, so its spec must not bill K and V."""
+    layers, latent = 61, 576
+    policy = _policy(num_layers=layers, is_mla=True, head_dim=latent, num_kv_heads=1)
+    specs = policy.get_kv_cache_spec()
+
+    assert all(isinstance(s, MLAAttentionSpec) for s in specs.values())
+
+    # _kv_factor() == 1 for MLA
+    metal_per_block = 1 * BLOCK_SIZE * DTYPE.itemsize * layers * 1 * latent
+    assert _engine_blocks(specs, metal_per_block) == POOL_BLOCKS
+
+
+def test_yoco_shared_layers_get_no_spec() -> None:
+    """Only the layers that own a cache may appear in the spec."""
+    layers, owned = 35, 15
+    policy = _policy(num_layers=layers, yoco=(owned, list(range(owned))))
+    specs = policy.get_kv_cache_spec()
+
+    assert len(specs) == owned
+
+    metal_per_block = 2 * BLOCK_SIZE * DTYPE.itemsize * owned * 4 * 256
+    assert _engine_blocks(specs, metal_per_block) == POOL_BLOCKS
+
+
+def test_plain_attention_spec_is_unchanged() -> None:
+    """Non-MLA, non-YOCO models keep emitting one FullAttentionSpec per layer."""
+    layers = 36
+    policy = _policy(num_layers=layers)
+    specs = policy.get_kv_cache_spec()
+
+    assert len(specs) == layers
+    assert all(type(s) is FullAttentionSpec for s in specs.values())
+
+    metal_per_block = 2 * BLOCK_SIZE * DTYPE.itemsize * layers * 4 * 256
+    assert _engine_blocks(specs, metal_per_block) == POOL_BLOCKS
--- a/vllm_metal/v1/cache_policy.py
+++ b/vllm_metal/v1/cache_policy.py
@@ -10,7 +10,12 @@ from typing import TYPE_CHECKING, Literal
 import mlx.core as mx
 import torch
 from vllm.logger import init_logger
-from vllm.v1.kv_cache_interface import FullAttentionSpec, KVCacheConfig, KVCacheSpec
+from vllm.v1.kv_cache_interface import (
+    FullAttentionSpec,
+    KVCacheConfig,
+    KVCacheSpec,
+    MLAAttentionSpec,
+)
 
 from vllm_metal.attention.caches.turboquant import (
     BLOCK_SIZE as TQ_BLOCK_SIZE,
@@ -273,8 +278,17 @@ class ModelCachePolicy:
         use_turboquant = self._use_turboquant(config)
 
         kv_heads, head_dims = self._cache_layer_shapes(self._runner.num_layers)
+        # Under YOCO KV sharing only the leading ``num_cache_layers`` layers own
+        # a cache; the trailing ones reuse it (see ``_mha_cache_layout``, whose
+        # mapping assigns the owners first).  Emit no spec for the sharers, so
+        # the engine sizes against the layers that were actually allocated --
+        # the same way upstream expresses sharing by omission in
+        # ``GPUModelRunner.get_kv_cache_spec``.
+        num_spec_layers = self._runner.num_layers
+        if self._runner._yoco_cache_mapping is not None:
+            num_spec_layers, _ = self._runner._yoco_cache_mapping
         specs: dict[str, KVCacheSpec] = {}
-        for layer_idx in range(self._runner.num_layers):
+        for layer_idx in range(num_spec_layers):
             if (
                 self._runner.is_hybrid
                 and layer_idx not in self._runner.sdpa_layer_indices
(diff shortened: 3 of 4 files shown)


hyunnnchoi:
## Validated on real weights — both MLA and YOCO

M1 Max 64 GB / macOS 15.7.1, `--max-model-len 4096 --gpu-memory-utilization 0.5`.

**MLA** — `mlx-community/MiniCPM3-4B-4bit` (62 layers, latent 288)

| | before | after |
|---|---:|---:|
| `blocks allocated` | 38,498 | 38,406 |
| `KV cache config received` | 19,249 | 38,406 |
| ratio | 0.500 | 1.000 |
| `Maximum concurrency` @ 4,096 | 75.19x | 150.02x |

**YOCO** — `mlx-community/gemma-4-e2b-it-4bit` (35 layers, 20 shared → 15 owning)

| | before | after |
|---|---:|---:|
| `blocks allocated` | 73,582 | 73,582 |
| `KV cache config received` | 31,535 | 73,582 |
| ratio | 0.4286 | 1.000 |
| `Maximum concurrency` @ 4,096 | 123.18x | 287.43x |

Both match the predicted ratios of 0.500 and 0.428 → 1.000. YOCO concurrency rises 2.333x, exactly 35/15.

The MLA pool differs 0.24% between runs because it is sized from free Metal memory at profile time; the ratio is computed within each run. The two YOCO runs happened to allocate identically.

Outputs are identical before and after in both cases, and 24 concurrent requests completed cleanly against each patched build, so the spec changes do not disturb addressing.

Measured by installing the release wheel, then applying only this PR's `cache_policy.py` change on top, so the two runs differ by exactly this diff.

LxYuan0420 (approved):
LGTM!

PR #533, opened 2026-07-21
[Spec Decode] Skip n-gram matching for requests that keep missing

KrxGu:
The match kernel scans a request's whole committed history every decode step whether or not it finds anything, plus a full history copy into the scratch buffer, paid regardless of outcome. A request with nothing exploitable in it, plain prose mostly, pays that cost every step for its whole lifetime with nothing to show for it. That's why n-gram spec decode looked like a dead end when it was first tried here on a prose workload and came back at basically zero gain.

Turns out the technique wasn't the problem, the workload was. On a task that actually reproduces spans from its own context, editing a file or annotating code, it's a real 1.45-1.51x on decode throughput, byte-identical output verified across multiple runs on an 8B target. The difference between doing nothing and 1.5x is just whether the generation echoes back what it already saw.

Rather than expecting a user to know which bucket their traffic falls into before flipping method=ngram, this gives the proposer a memory: after enough consecutive misses for a request, stop scanning it for a while, and retry periodically in case the content turns repetitive partway through, a response that starts as prose and later quotes earlier context, say. A hit resets the count immediately. That's what makes it safe to leave n-gram spec decode on for mixed traffic instead of a per-workload opt-in: the requests that don't benefit stop paying for it on their own, without touching the ones that do.

Everything lives inside the existing NgramProposer wrapper. Nothing upstream or in the verify path changes, and the throttle can only ever skip a proposal, never invent or drop an accepted token, so correctness is untouched by construction. Checked with a token-hash comparison across throttled and unthrottled runs on the repetitive workload, output matches exactly. Tests cover the miss-then-throttle transition, a hit resetting the streak mid-run, cooldown expiry letting a request retry, and bookkeeping getting cleared when a request finishes.

The headline numbers it highlights: 1.45-1.51x decode throughput on repetition-heavy generation, byte-identical correctness, and the structural point, it converts n-gram spec-decode from a niche opt-in into something safe to leave on for all traffic.

### DIFF
--- a/vllm_metal/v1/ngram_proposer.py
+++ b/vllm_metal/v1/ngram_proposer.py
@@ -22,6 +22,12 @@ prose, for instance) stops paying the match kernel's per-step scan cost after
 a few misses in a row, with periodic retries in case the content turns
 repetitive later. See ``_record_miss``/``_on_cooldown``.
 
+A lookup match is not itself evidence the request benefits: the target may
+reject every proposed token. ``_resolve_pending`` checks a step's proposal
+against the *next* step's committed history (which already reflects what the
+target actually accepted) to score it as a real hit or a miss, rather than
+trusting the lookup kernel's own optimism.
+
 The verify half is unchanged: drafts are handed back via ``take_draft_token_ids``
 and verified next step by ``SpeculativeDecodeController.verify_greedy``.
 """
@@ -73,16 +79,32 @@ class NgramProposer:
     ) -> None:
         self._controller = controller
         # Per-request consecutive-miss count and, once throttled, remaining
-        # cooldown steps before the next retry. Both pruned each step against
-        # the live request set so they never grow past what's active.
+        # cooldown steps before the next retry. Pruned each step against the
+        # scheduler's own finished-id set, not against absence from
+        # request_states: vLLM can hand a finished id straight back out to a
+        # new request in the same step, and that new request repopulates
+        # request_states under the same id before this wrapper ever sees it.
         self._miss_streak: dict[str, int] = {}
         self._cooldown: dict[str, int] = {}
+        # A lookup match doesn't mean the target accepted any of it. Each
+        # non-empty draft is stashed here (position it starts at, tokens
+        # proposed) and scored against the *next* step's actual committed
+        # history in _resolve_pending, before this step's proposal is
+        # scored as a hit or a miss.
+        self._pending: dict[str, tuple[int, tuple[int, ...]]] = {}
         # Upstream reads only scalar config (prompt_lookup_min/max,
         # num_speculative_tokens, max_model_len, max_num_seqs) and runs a one-time
         # Numba JIT warmup in its constructor — keep that off the hot path.
         self._ngram = VllmNgramProposer(vllm_config)
         spec = vllm_config.speculative_config
         assert spec is not None
+        # Below this many committed tokens, the kernel structurally cannot
+        # match (not enough history to form even the shortest n-gram window)
+        # -- that's not evidence against repetition, just not enough data
+        # yet, so it must not count as a miss. Normalized to a concrete int
+        # by SpeculativeConfig.__post_init__ for method="ngram".
+        assert spec.prompt_lookup_min is not None
+        self._prompt_lookup_min = spec.prompt_lookup_min
 
         # Pre-allocate the int32 token-id buffer once. Upstream only reads
         # ``token_ids_cpu[i, :num_tokens_no_spec[i]]`` per row, so the buffer
@@ -130,7 +152,8 @@ class NgramProposer:
         if ctx.num_speculative_tokens <= 0:
             return None
 
-        self._prune_finished(ctx.request_states)
+        self._prune_finished(ctx.finished_req_ids)
+        self._resolve_pending(ctx.request_states)
 
         drafting = list(
             self._controller.draft_eligible_requests(
@@ -176,11 +199,17 @@ class NgramProposer:
 
         req_ids: list[str] = []
         draft_token_ids: list[list[int]] = []
-        for (req_id, _), draft in zip(drafting, drafts, strict=True):
+        for (req_id, state), draft in zip(drafting, drafts, strict=True):
             if not draft:
-                self._record_miss(req_id)
+                # A too-short history isn't evidence against repetition --
+                # there just isn't enough of it yet to check.
+                if len(state.token_ids) >= self._prompt_lookup_min:
+                    self._record_miss(req_id)
                 continue
-            self._miss_streak.pop(req_id, None)
+            # Don't score this as a hit yet -- the target hasn't verified it.
+            # _resolve_pending scores it next step against what was actually
+            # accepted.
+            self._pending[req_id] = (len(state.token_ids), tuple(draft))
             req_ids.append(req_id)
             # Upstream already yields Python ints via ndarray.tolist() — the
             # old ``[int(t) for t in draft]`` was redundant.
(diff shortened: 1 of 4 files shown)


LxYuan0420 (commented):
.

LxYuan0420 (changes requested):
.

vllm_metal/v1/ngram_proposer.py
```diff
@@ -106,6 +130,8 @@ def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
         if ctx.num_speculative_tokens <= 0:
             return None
 
+        self._prune_finished(ctx.request_states)
```
LxYuan0420: I don't think request_states is the right lifecycle signal here. Upstream allows a finished request and a new request with the same id in the same scheduler step, so the new request can inherit old cooldown state. Can we carry finished_req_ids through ProposeContext and clear those ids first?

vllm_metal/v1/ngram_proposer.py
```diff
@@ -154,3 +190,33 @@ def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
...
+            del self._cooldown[req_id]
+        else:
+            self._cooldown[req_id] = remaining - 1
+        return True
+
+    def _record_miss(self, req_id: str) -> None:
+        streak = self._miss_streak.get(req_id, 0) + 1
+        if streak >= _MAX_CONSECUTIVE_MISSES:
+            self._cooldown[req_id] = _COOLDOWN_STEPS
+            self._miss_streak.pop(req_id, None)
```
LxYuan0420: I think this forgets too much state. After cooldown, the same permanent miss gets another 8 full scans before throttling again. Can we keep the miss streak capped until a hit, so a failed retry goes right back to cooldown?

vllm_metal/v1/ngram_proposer.py
```diff
@@ -144,7 +178,9 @@ def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
         draft_token_ids: list[list[int]] = []
         for (req_id, _), draft in zip(drafting, drafts, strict=True):
             if not draft:
+                self._record_miss(req_id)
```
LxYuan0420: An empty draft is not always a real miss. Upstream also returns [] when the history is shorter than prompt_lookup_min. Can we avoid counting that as a miss and add a short-history-to-eligible test?

vllm_metal/v1/ngram_proposer.py
```diff
@@ -144,7 +178,9 @@ def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
...
                 continue
+            self._miss_streak.pop(req_id, None)
```
LxYuan0420: I’m not sure a match means the request actually benefits. This resets before target verification, so repeated matches with zero accepted tokens never throttle. Can we either use accepted draft count, or make the PR claim only about no-match throttling?

vllm_metal/v1/ngram_proposer.py
```diff
@@ -42,6 +49,18 @@
...
+# The match kernel scans a request's whole committed history every decode
+# step whether or not it finds anything -- an O(history length) Numba scan
+# plus a full-history copy into token_ids_cpu, paid regardless of outcome.
+# A request whose content has no exploitable repetition (free-form prose,
+# for instance) pays that tax every step for nothing. After this many
+# consecutive misses, stop attempting a request for _COOLDOWN_STEPS steps
+# rather than giving up on it forever -- generation can turn repetitive
+# partway through (e.g. a response that starts as prose and then quotes
+# earlier context), so a permanent cutoff would miss that.
+_MAX_CONSECUTIVE_MISSES = 8
```
LxYuan0420, line 66: These 8/8 constants are the actual CPU cost policy, but the numbers shown are ngram vs no-spec, not main vs this PR. Can we get raw main-vs-head results with proposal rate and acceptance rate for miss-heavy, low-acceptance, repetitive, delayed-repetition, and mixed batches?

vllm_metal/v1/ngram_proposer.py
KrxGu: Confirmed, checked SchedulerOutput and it does expose finished_req_ids directly. Threaded it through ProposeContext (scheduler_output is already in scope right where the context gets built in model_runner.py) and prune against that instead of absence from request_states. Added a regression test where a request finishes and a new one reuses the same id in the same context, confirming the new one isn't skipped as if it were still the old, throttled request.

vllm_metal/v1/ngram_proposer.py
KrxGu: Fixed. _record_miss now caps the streak at _MAX_CONSECUTIVE_MISSES instead of clearing it when cooldown triggers, so a miss on the very next retry sends it straight back to cooldown instead of earning another full eight-miss grace period. Only a genuine accepted hit clears it now. Added a test for this specifically.

vllm_metal/v1/ngram_proposer.py
KrxGu: Confirmed against upstream, it returns [] below prompt_lookup_min regardless of content. Stored prompt_lookup_min off the speculative config and only record a miss once the request's committed history has actually reached it. Added test_short_history_is_not_counted_as_a_miss.

vllm_metal/v1/ngram_proposer.py
KrxGu: This was the real gap. Reworked it so a non-empty draft doesn't reset anything by itself, it gets stashed (position, proposed tokens) and scored on the *next* step against what state.token_ids actually shows was committed, which by then reflects what verify_greedy accepted. Only a genuine accept resets the streak now; a request that keeps finding matches that are never accepted accumulates toward cooldown exactly like a true miss would. Added tests for both the real-accept and match-without-accept cases.

vllm_metal/v1/ngram_proposer.py
KrxGu: Fair, here's raw main-vs-head, same workloads, same seed, instrumented directly (VllmNgramProposer.propose call count for proposal rate, SpeculativeDecodeController.verify_greedy accepted-vs-proposed for acceptance rate), on the real vLLM 0.25.1 target:

| workload | kernel calls (main -> head) | accepted tokens (main -> head) | acceptance rate (main -> head) | decode tok/s (main -> head) |
|---|---|---|---|---|
| miss_heavy | 250 -> 59 (-76%) | 6 -> 5 | 0.065 -> 0.238 | 67.6 -> 75.7 (+12%) |
| low_acceptance | 126 -> 126 | 133 -> 133 | 0.547 -> 0.547 | 89.2 -> 88.8 (noise) |
| repetitive | 67 -> 67 | 192 -> 192 | 1.0 -> 1.0 | 69.8 -> 69.7 (noise) |
| delayed_repetition | 146 -> 70 (-52%) | 112 -> 92 | 0.794 -> 0.807 | 103.1 -> 100.5 (noise) |
| mixed_batch | 245 -> 105 (-57%) | 203 -> 197 (-3%) | 0.769 -> 0.925 | 77.1 -> 80.8 (+5%) |

low_acceptance and repetitive never cross the miss threshold, so they're untouched, which is the right behavior. miss_heavy drops kernel calls 76% and loses essentially nothing (6 vs 5 accepted is noise), and tok/s actually improves. delayed_repetition and mixed_batch are the real test of the cooldown-retry design: kernel calls roughly halve in both, but accepted tokens barely move (mixed_batch is -3%) and acceptance rate goes up, not down, meaning the throttle is cutting the wasted attempts specifically and still catching the recovery into repetition, not just cutting volume indiscriminately. That's what 8/8 is actually buying here. Happy to sweep those constants if you want a specific alternative to compare against, this was picked as a reasonable default, not tuned against this data.

workload definitions and the instrumented harness: miss_heavy is free-form prose with no self-repetition, low_acceptance is a templated list with a repeating prefix but a different value each entry, repetitive is the code-annotation task from the PR description, delayed_repetition asks for two original sentences followed by a verbatim quote of prompt content, mixed_batch runs the repetitive and miss_heavy prompts concurrently in one batch.

KrxGu:
@LxYuan0420 i have addressed all the reviews and have made the changes accordingly, CI is all green. LGMT!!

vllm_metal/v1/ngram_proposer.py
```diff
@@ -106,6 +152,9 @@ def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
         if ctx.num_speculative_tokens <= 0:
             return None
```
LxYuan0420, line 160: This is still a lifecycle issue. I tested K=0 + `finished_req_ids`, and `_cooldown`, `_miss_streak`, and `_pending` all survive because this returns before cleanup. Please prune finished ids and resolve pending before the zero-K return, and add a zero-K same-id-reuse test.

vllm_metal/v1/ngram_proposer.py
```diff
@@ -154,3 +219,57 @@ def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
...
+            # committed that many new tokens yet -- strict=False by design.
+            for proposed_id, actual_id in zip(proposed, actual, strict=False):
+                if proposed_id != actual_id:
+                    break
+                accepted += 1
+            if accepted > 0:
+                self._miss_streak.pop(req_id, None)
+            else:
+                self._record_miss(req_id)
+        self._pending.clear()
```
LxYuan0420: This still scores pending drafts from `request_states`, but that only means the request is live, not that it was scheduled or verified this step. I tested a live unscheduled request with pending draft: `actual=[]`, pending is cleared, and `miss_streak` increments. Please resolve only for requests actually verified this step.

vllm_metal/v1/ngram_proposer.py
KrxGu: Confirmed, moved both calls ahead of the num_speculative_tokens <= 0 return so cleanup always runs regardless of whether drafting is allowed that step. Added a test that reuses an id specifically through a disabled step and checks the new request isn't skipped afterward.

vllm_metal/v1/ngram_proposer.py
KrxGu: Right, request_states just means "exists", not "verified this step". Now resolves against decode_reqs specifically (the actual set verify_greedy operates on this step), and anything not in there stays pending untouched instead of getting scored. Added a test with a live-but-unscheduled request confirming its pending draft survives instead of being counted as a miss.

KrxGu:
@LxYuan0420 i have made the required changes, CI is all green

vllm_metal/v1/ngram_proposer.py
```diff
@@ -154,3 +223,63 @@ def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
...
+        # miss on the very next retry should send it right back into
+        # cooldown, not grant another full _MAX_CONSECUTIVE_MISSES-long
+        # grace period. Only a genuine hit (_resolve_pending finding real
+        # acceptance) fully clears the streak.
+        streak = min(self._miss_streak.get(req_id, 0) + 1, _MAX_CONSECUTIVE_MISSES)
+        self._miss_streak[req_id] = streak
+        if streak >= _MAX_CONSECUTIVE_MISSES:
+            self._cooldown[req_id] = _COOLDOWN_STEPS
+
+    def _resolve_pending(self, ctx: ProposeContext) -> None:
```
LxYuan0420, line 250: Could we clarify whether every request in decode_reqs is guaranteed to have a scheduled draft? If not, should we only resolve pending state when decode_segment.draft_token_ids is non-empty and add a test for the plain-row case?

KrxGu: No, confirmed it's not guaranteed. A request can be in decode_reqs with an empty draft_token_ids two ways, no draft was scheduled for it, or the scheduler padded/dropped a proposed draft on batch admission (same thing active_spec_decode_tokens already accounts for elsewhere). Now resolves against decode_segments specifically, only a request whose segment actually carried a non-empty draft gets scored, everything else stays pending untouched. Added the plain-row test.

LxYuan0420 (approved):
Thanks for working on this!

PR #534, opened 2026-07-22
Share KV reads across spec-verify window rows in paged decode

### LINKED ISSUE #465: [RFC] Redundant KV-cache reads during speculative-decode verification

ricky-chaoju:
This fixes #465. `prepare_unified` was splitting a K+1 verify window into K+1 single-token segments, so the decode kernel walked the whole KV context once per row. At short context that redundancy is basically free; past 4-8k it becomes the dominant verify cost, which is the same regime where #447 saw K=5 go net-negative under concurrency.

Per the RFC there is no third kernel. `prepare_unified` keeps the window whole and the existing per-token kernel grows a window mode: 2-row sub-windows per segment, each K/V block load shared across the rows, per-row accumulation order unchanged, so outputs stay bitwise identical to the old dispatch. @abcgco's gate commits (from my fork's PR 8) keep the split-KV family choice identical too, which fixed the last bitwise gap on his 16-core M4 Pro. YOCO fast prefill now selects every row of a verify window, and a window hint that doesn't match `cu_seqlens` is rejected up front.

For evidence, the main one is @KrxGu's M3 Ultra run: +37%/+40% at conc 16/32 (Qwen3-8B-4bit, 8k). Below is my M2 Ultra warm-decode A/B against main (a4fcf68), ngram K=5, greedy, 8192-token prefixes, output 96, median TPOT of 3 reps. The conc-16 cell holds at 1.35x when I interleave the server sessions, and acceptance matches between the two builds in every cell, which is the bitwise contract showing up end to end. Single-stream is the weak spot and I'd rather say so: 1.13-1.16x here, 0.99-1.03x for @WindChimeRan at K=3 on his M5 Pro. Plain decode doesn't regress, op-level is within 3% single-stream and a bit faster under load from the register fold.

| model | conc | main TPOT (ms) | branch TPOT (ms) | speedup | accept |
|---|---:|---:|---:|---:|---:|
| Qwen3-4B | 1 | 40.1 | 35.5 | 1.13x | 0.20 |
| Qwen3-4B | 16 | 170.7 | 121.8 | 1.40x | 0.79 |
| Qwen3-4B | 32 | 352.4 | 242.9 | 1.45x | 0.84 |
| Qwen3-0.6B | 1 | 6.7 | 5.8 | 1.16x | 1.00 |
| Qwen3-0.6B | 16 | 66.3 | 48.2 | 1.37x | 1.00 |


vllm_metal/metal/paged_ops.cpp
```diff
@@ -356,18 +381,40 @@ static void dispatch_paged_attention_v2_online(
...
+  // path in a band of shapes whose location depends on the core count
+  // (min_decode_grid).  Inside that band the two paths run different
+  // kernel families (_ps512 vs _ps0) and the bitwise contract breaks --
+  // observed on an M4 Pro 16-core for 32q/8kv single-sequence windows,
+  // invisible on 10-core M4 / M2 Ultra whose thresholds land elsewhere.
+  // (For non-window batches grid_y == total_q_tokens, so this is the
+  // same value the gate always used.)
+  const int gate_grid = num_heads * total_q_tokens;  // grid.z = 1 occupancy
+  const bool partition = (pure_decode || window_batch)
+      && gate_grid < min_decode_grid() && max_num_partitions >= 2;
```
LxYuan0420, line 425: I think this mixes two things: verify-window shape vs optimized window mode. For head_size=512 we still merge the verify window, but we do not use window mode, so it can fall into tiled prefill and lose the old decode/split route. I can repro 512 expanded vs merged not being bitwise equal. Can we split this into verify_batch and window_batch, keep all verify batches out of tiled fallback, and add 512 coverage?

vllm_metal/attention/context.py
```diff
@@ -189,5 +201,9 @@ def prepare_unified(
             cu_seqlens=cu_seqlens,
             offsets=offsets,
             num_decode_requests=len(decode_requests),
+            # Window routing applies only to pure-verification batches;
+            # any prefill segment keeps the whole batch on the tiled
+            # kernel, numerically identical to the non-speculative path.
+            verify_window_q=1 if prefill_requests else max_decode_window,
```
LxYuan0420, line 228: I’m wondering if this changes more than SDPA expected. MLA native decode only accepts one-row segments today, so spec verify seems to fall back to MLX SDPA after this metadata merge. Is that intended, or should we preserve the MLA route locally?

vllm_metal/metal/kernels_v2/pagedattention.metal
```diff
@@ -1040,36 +1124,65 @@ template <typename T, typename K_CACHE_T, typename V_CACHE_T, int HEAD_SIZE, int
...
-                             q_vecs[thread_group_offset], k_vecs);
+#pragma unroll
+      for (int r = 0; r < rows_per_tg; r++) {
+        if (r >= num_rows) break;
+        // Finalize the reduction across lanes (same tree as Qk_dot).
+        float qk = sum(qk_accs[r]);
+#pragma unroll
+        for (int mask = THREAD_GROUP_SIZE / 2; mask >= 1; mask /= 2) {
+          qk += simd_shuffle_xor(qk, mask);
+        }
```
LxYuan0420, line 1559: This also changes normal window_q == 0 decode, not only window mode. Do we have raw main-vs-this-PR numbers for plain decode, including TurboQuant and sliding-window? If not, I would rather keep the old non-window path separate.

vllm_metal/attention/context.py
LxYuan0420, line 228: Can we add a small mixed spec-window + prefill test? This branch relies on verify_window_q=1 in mixed batches to keep them off window mode.

vllm_metal/metal/paged_ops.cpp
ricky-chaoju: Fixed

vllm_metal/attention/context.py
ricky-chaoju: You're right. Same fix MLA keeps the expanded metadata, so the native decode gate admits verify batches exactly as on main. Covered hybrids too while at it.

vllm_metal/metal/kernels_v2/pagedattention.metal
ricky-chaoju: Yes. The shared body made plain TQ decode 2~5x slower, so I split the kernel on the window_q function constant and restored the per token body verbatim. Plain decode is back at parity and bitwise with main.

vllm_metal/attention/context.py
ricky-chaoju: Added.

abcgco:
**Re-validated on the 16-core M4 Pro at `02e99e5`** (post-review head, incl. the plain-decode kernel split) — the chip where the original partition-family divergence was visible:

- `test_spec_window_parity.py` + `test_paged_attention.py`: **131 passed**, including the `mx.array_equal` bitwise grid over both GQA shapes (16q/8kv and the 32q/8kv split-KV band), fp16/bf16/fp32, ctx 512/8192, K+1 ∈ {2,4,5,6,8}, and both the single-partition and forced-split kernel families. `test_yoco_fast_prefill.py`: 20 passed, 1 skipped.
- `tools/spec_window_bench.py`, both configs: **bitwise=1 in all 18 cells**. exp/win on this chip: 1.16x on the 8B shape at conc 4–32 @ 8k; 1.10–1.12x on the 0.6B shape at conc 16–32 @ 8k; everything else — all 2k cells at any concurrency, single-stream, 0.6B conc-4 — sits at 0.66–1.04. Same read as your M2 Ultra numbers: it's a concurrency + context effect, and single-stream is the weak spot, as stated.

8B-shape table (fp16, 32 layers/rep, 20-rep medians):

```
conc    ctx  K | exp/win | bitwise
   1   2048  5 |   0.81  | 1
   1   8192  5 |   0.66  | 1
   1  16384  5 |   1.04  | 1
   1   8192  3 |   0.79  | 1
   4   8192  5 |   1.16  | 1
  16   2048  5 |   0.98  | 1
  16   8192  5 |   1.16  | 1
  32   2048  5 |   0.91  | 1
  32   8192  5 |   1.16  | 1
```

vllm_metal/metal/paged_ops.cpp
```diff
@@ -329,15 +350,27 @@ static void dispatch_paged_attention_v2_online(
...
+
+  // Spec-decode verification windows: a multi-token batch whose segments
+  // are all verification windows runs the per-token kernel in window mode
+  // (kWindowRows-row sub-window threadgroups share every KV read), keeping
+  // split-KV eligibility.  The tiled kernel is wrong for this shape: its
+  // BQ-row MMA tiles waste ~80% of their compute on a K+1 window and it has
+  // no KV-length parallelism.  head_size <= kWindowMaxHeadSize holds for
+  // every constructed primitive (the binding rejects wider hints), so a
+  // verify batch here always takes window mode, never the tiled fallback.
+  const bool window_batch =
```
LxYuan0420, line 370: Can we gate this by measured profitability instead of capability only? On my M2 Max, window mode regressed the conc-32/8K/K=5 case twice, so we should keep a per-token fallback when window mode loses.

vllm_metal/metal/paged_ops.cpp
```diff
@@ -599,13 +669,66 @@ static array paged_attention_primitive_fn(
...
+    int v_bits = 3, int window_seqlen_q = 1) {
+  // window_seqlen_q must equal the longest cu_seqlens_q segment: window-mode
+  // threadgroups only exist for ceil(window_seqlen_q / kWindowRows)
+  // sub-windows per segment, so an understated value leaves the tail rows of
+  // a longer segment unwritten (garbage logits, no error), and an overstated
+  // one dispatches empty threadgroups.  Only the spec-verify path passes a
+  // window hint and its cu_seqlens_q is a small host-built array, so
+  // materializing it to check the real segment lengths is cheap, and the
+  // plain-decode path (window_seqlen_q == 1) skips the block entirely.
+  if (window_seqlen_q > 1) {
```
LxYuan0420, line 687: Can we reject window_seqlen_q < 1 here? 0 and -1 currently bypass validation and silently select non-window routing.

LxYuan0420 (changes requested):
Request changes.

Correctness looks much better now, but I do not think window mode is safe as an always-on default yet. On exact head 02e99e5, my M2 Max 8B-like conc-32/8K/K=5 benchmark was slower in two runs:

| run | expanded | windowed | exp/win |
|---|---:|---:|---:|
| full matrix | 73.84 ms | 91.54 ms | 0.81 |
| fresh-process repeat | 59.97 ms | 65.89 ms | 0.91 |

Both were bitwise-identical, so this is a performance routing issue, not correctness. The M4 Pro data also shows losing cells, including exp/win=0.66 for conc-1/8K/K=5.

I think we need to separate verify-batch correctness routing from window-mode profitability: keep the merged verify metadata, but fall back to the per-token route when window mode is not expected to win.

The non-positive window hint check is smaller, but it is an easy contract fix while touching this path.

vllm_metal/v1/model_runner.py
```diff
@@ -384,6 +386,26 @@ def is_hybrid(self) -> bool:
...
+        pure-decode check admit only one-row segments, and heads past
+        PA_WINDOW_MAX_HEAD_SIZE would leave the decode kernel for the
+        tiled one.
+        """
+        return (
+            envs.VLLM_METAL_SPEC_VERIFY_WINDOW
+            and not self.is_mla
+            and not self.is_hybrid
+            and self.model_config.get_head_size() <= PA_WINDOW_MAX_HEAD_SIZE
+        )
```
LxYuan0420, line 414: This is still an issue. I verified on this head that with `VLLM_METAL_SPEC_VERIFY_WINDOW=1`, `get_head_size()==256`, and `head_dim_per_layer=[256, 512]`, this returns `True`.

So a mixed-head model can merge a 512-wide verify window, then fail in the C++ binding instead of staying expanded. Can we gate on the resolved runtime max head dim and add that regression test?

vllm_metal/attention/context.py
```diff
@@ -129,6 +135,8 @@ def prepare_unified(
     decode_requests: list[tuple[list[int], int] | tuple[list[int], int, int]],
     prefill_requests: list[tuple[list[int], int, int]],
     block_size: int,
+    *,
+    merge_verify_windows: bool = True,
```
LxYuan0420: Since window mode is opt-in now, this helper should not default to the merged layout. Future callers can silently opt into the experimental shape without passing the runtime policy. Can we default this to `False` or require callers to pass it explicitly?

LxYuan0420 (commented):
Few fixes needed before we merge it.

ricky-chaoju:
done.

LxYuan0420 (approved):

PR #484, opened 2026-07-08
[Spec Decode] GLM-4.7-Flash native MTP head: model, loader, extraction tool

samithaj:
## What

The GLM-4.7-Flash native MTP head as a standalone, fully-tested component — **no runtime wiring yet** (the proposer + `install_drafter` registration land in the next PR, so this changes no existing behavior). Stacked on #483 (its 2 commits are included; review the last 5).

Design + measured data: https://github.com/vllm-project/vllm-metal/issues/482#issuecomment-4917989509

- **`vllm_metal/v1/mtp_heads/glm4_moe_lite_mtp.py`** — `Glm4MoeLiteMTPModel` reusing the stock `mlx_lm.models.glm4_moe_lite.Glm4MoeLiteDecoderLayer` (MLA + MoE) with upstream `glm4_moe_lite_mtp` input semantics: `eh_proj([enorm(embed(t_{p+1})), hnorm(h_p)])`, position-0 embedding zeroing, post-final-norm target hidden states, `shared_head` readout. Loader + metadata validation mirror the Gemma4 assistant loader (process cache, narrow snapshot download, fail-fast target compatibility: model_type + 9 numeric fields; consolidation with the Gemma4 loader deliberately deferred until a second native head exists — noted in the module docstring).
- **`tools/extract_glm47_mtp_head.py`** — produces a standalone head checkpoint from `zai-org/GLM-4.7-Flash` (`model.layers.47.*` lives in 3 of 48 shards ≈ 4 GB download, fetched via one targeted `snapshot_download`): flat MLX layout, quantize-aware `kv_b_proj → embed_q/unembed_out` split + expert stacking shared with `sanitize`, **`e_score_correction_bias` kept fp32** (matching stock mlx_lm `cast_predicate` / upstream vLLM noaux_tc semantics), optional `--q-bits 4`. Needed because mlx-community conversions strip the nextn layer (mlx_lm `sanitize` drops layers ≥ `num_hidden_layers`).
- **`compat.py`** — AutoConfig shim for `model_type: glm4_moe_lite_mtp` (config loading only), mirroring the existing `gemma4_assistant` shim; pinned transformers has no mapping for it.

## Numerics validation

[...]

PR #535, opened 2026-07-22
[Spec Decode] Native MTP speculative decoding for GLM-4.7-Flash

samithaj:
Native MTP speculative decoding for GLM-4.7-Flash, reopened fresh against current
main as invited when #484/#485 were closed. GLM-4.7-Flash ships one trained nextn
(MTP) layer inside the target checkpoint (`num_nextn_predict_layers: 1`); this runs
it as a native drafter on the paged path — one decoder-layer forward over the
target's post-final-norm hidden states proposes one draft token, keeping only the
head's own compressed MLA latent per request (~1.2 KB/token). It reads none of the
target's paged KV, needs no second block allocator, and drafts without a separate
draft-model prompt prefill during the normal path.

The head loads from a hosted MLX checkpoint split from a pinned target revision
(`samithaj/GLM-4.7-Flash-MTP-4bit` / `-bf16`), produced with the
`glm4_moe_lite_mtp` split tool proposed in Blaizzy/mlx-vlm#1570 — the upstream
artifact contract laid out in #482. There is no per-user extraction step. Pin the
head `revision` in `speculative_config` for reproducible runs.

What's here:
- A class-owned `NativeMTPHeadRegistry` keyed on the draft `model_type`, and
  `Glm4MoeLiteMTPHead`, which owns its validation and builds its proposer.
- The MLX head model + hosted-checkpoint loader. The drafter split nests the
  source config under `text_config` and leaves the head's architecture / nextn
  depth out of the top level, so the compat shim promotes both before vLLM's MTP
  override runs (and the promoted config survives pickling into vLLM's spawned
  engine workers), and the head's config/args parsing reads the shape from the
  nested `text_config` rather than falling back to defaults.
- `NativeMTPProposer` / `MTPSlab`, which releases its per-request KV slab on
  eviction/preemption/resume through a new `MetalProposer.release_requests` seam,
  rebuilding it during scheduler recompute (with rewind/hole checks) rather than
  pinning it while a request waits.
- `model_runner.install_drafter` dispatch for non-Gemma4 `method="mtp"` configs.

Changed vs the earlier stack (#484/#485): the hosted loader replaces the in-repo
per-user extraction tool (deleted), and the whole thing is a single integration PR
rather than a seam-split series, per the reopen-fresh request.

Constraints: greedy only (`temperature=0`), `num_speculative_tokens=1`, synchronous
scheduling (`--no-async-scheduling`), the paged text path, prefix caching off
(`--no-enable-prefix-caching`), and no pipeline parallelism — each rejected at
startup / proposer construction rather than surfacing mid-run.

Tests: `pytest -m "not slow" tests/` — 1591 passed / 15 skipped / 46 deselected
(under vLLM 0.25.1 + MLX 0.32.0, this branch). `ruff check` / `format --check`
clean on all touched files.

Benchmarks (M3 Max, GLM-4.7-Flash-4bit target + 4-bit hosted head, both
revision-pinned and recorded in the JSONs; greedy, bs1, 8 distinct prompts run
serialized x 256 tokens): SD on vs off raises end-to-end output-token
throughput by **~10%** (66.1 vs 59.8 tok/s, interleaved A/B processes with
cooldowns on an idle machine; 1.06-1.13x across all idle-machine pairings) at
**0.770** acceptance measured over the timed runs only (889/1155 post-warmup
draft tokens; identical across runs under greedy). An earlier version of this
paragraph claimed 1.35x — that paired an SD run against a baseline depressed
by steady background load (which passes an intra-run drift gate); interleaved
A/B runs corrected it. Lossless verified against a no-SD batch-shape control:
SD divergences fall in the same class as the 4-bit MLA target's intrinsic
batch-shape wobble. The earlier measurement in #482 (~1.40x at ~0.91
acceptance) used a different prompt set, token budget, and MLX/vLLM versions,
so the two are not directly comparable; single-prompt acceptance varies
substantially by prompt (0.71-0.83 observed). With K=1 at 0.77 acceptance the
no-overhead ceiling is 1.77x; the gap is per-step draft + verify overhead (the
2-token verify step runs the target's quantized matmuls at M=2 well below
2-token efficiency), which is kernel follow-up territory, not head quality.
Benchmark JSONs are linked in the first comment (token_ids trimmed).

### DIFF
--- a/docs/speculative_decoding.md
+++ b/docs/speculative_decoding.md
@@ -1,23 +1,25 @@
 # Speculative Decoding
 
-vllm-metal supports three speculative decoding methods on the paged attention
-path. All require synchronous scheduling and greedy sampling.
+vllm-metal supports the following speculative decoding methods on the paged
+attention path. All require synchronous scheduling and greedy sampling.
 
-| | MTP | Draft Model | N-gram |
-|---|---|---|---|
-| `--speculative-config` method | `mtp` | `draft_model` | `ngram` |
-| Target models | Gemma4 (paged) | Any paged-attention model | Any paged-attention model |
-| Draft source | MTP assistant checkpoint (reads target KV cache) | Separate smaller model (own KV cache) | Prompt/output token history (no model) |
-| `num_speculative_tokens` | 1 | Configurable (3–5 typical) | Configurable (3–5 typical) |
-| Extra memory | None (reads target KV cache) | Second un-budgeted KV cache | None |
+| | Gemma4 MTP | Native MTP (GLM) | Draft Model | N-gram |
+|---|---|---|---|---|
+| `--speculative-config` method | `mtp` | `mtp` | `draft_model` | `ngram` |
+| Target models | Gemma4 (paged) | GLM-4.7-Flash (paged) | Any paged-attention model | Any paged-attention model |
+| Draft source | MTP assistant checkpoint (reads target KV cache) | Hosted nextn head (own 1-layer KV slab) | Separate smaller model (own KV cache) | Prompt/output token history (no model) |
+| `num_speculative_tokens` | 1 | 1 | Configurable (3–5 typical) | Configurable (3–5 typical) |
+| Extra memory | None (reads target KV cache) | ~1.2 KB/token/request (compressed MLA latent) | Second un-budgeted KV cache | None |
+| Prefix caching | Tolerated | Must be off (`--no-enable-prefix-caching`) | Tolerated | Tolerated |
 
-All three methods:
+All methods:
 
 - Require `--no-async-scheduling` (vLLM auto-disables async for most spec-decode
   methods; pass it explicitly to be safe).
 - Only accelerate greedy requests (`temperature=0`). Non-greedy requests skip
   drafting silently.
-- Are lossless under greedy decoding.
+- Are lossless under greedy decoding, up to the target's own kernel numerics —
+  see the tie-flip note in the GLM section for MLA targets.
 
 ---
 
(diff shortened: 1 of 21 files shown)


samithaj:
Benchmark JSONs (token_ids trimmed): https://gist.github.com/samithaj/7de943b30962d7bd618bf5b355985b6b — `bench_sd_8serial.attach.json` (SD on) and `bench_base_8serial_r2.attach.json` (baseline); target and head revisions, environment versions, and per-run timings are recorded in each.

samithaj:
Corrected the benchmark paragraph — the earlier 1.35x paired an SD run against a load-depressed baseline; interleaved A/B runs give ~1.10x at 0.770 acceptance. The gist now includes the interleaved-run JSONs.

LxYuan0420 (commented):
This PR is too large, but it still does not prove the full runtime path works safely. Could you close it and split it into smaller PRs with clear tests and benchmark results?

samithaj:
Added a download-free engine-level test (`tests/test_glm4_moe_lite_mtp_e2e.py`, `slow`, ~50s): it generates a ~1 MB random `glm4_moe_lite` target and MTP head from the real model classes and drives a real engine through the normal front door — speculative output matches the non-speculative run, an accept-all fixture pins acceptance at 1.0, and a 32-block budget forces preemption while generation stays lossless and the drafter releases its per-request slabs. I checked it fails when the path is broken: dropping the preemption release fails the release test, and accepting drafts unconditionally fails the lossless test.

Splitting as asked. Series: hosted-config front door → MTP head package → runtime path (proposer + release seam + runner dispatch, carrying this test) → benchmark tool and docs. I'll close this once the first one is up.

samithaj:
Split started: #551 is the first piece — the drafter release seam on evict/preempt/resume, GLM-free, 47 production lines. The rest follow in order: hosted-config front door, the MTP head model and loader, the runtime path that runs it as a drafter (carrying the download-free engine-level test), and the benchmark helper with docs.

Closing this in favour of the series.

PR #536, opened 2026-07-22
Trim auto-fit tests to Metal-owned behavior

LxYuan0420:
This PR is:

- To stop re-testing vLLM’s upstream auto-fit logic.
- To keep only the Metal-specific auto-fit checks.
- To remove a local reimplementation of upstream fitting logic.

Context

These tests were checking behavior owned by upstream vLLM. One helper also reimplemented upstream fitting logic in the test.

In Metal, we should not re-test upstream code or copy upstream logic. We only need to check the part we own: the fitted length must leave room for BlockPool’s null block, and too-small memory should fail clearly.


ricky-chaoju (approved):
LGTM

PR #537, opened 2026-07-22
Trim duplicate pipeline dummy-forward tests

LxYuan0420:
This PR is:

- To remove duplicate stub-only PP dummy-forward tests.
- To keep the stronger real-model dummy-forward coverage.
- To keep the Metal-specific middle-stage and wire-dtype checks.

Context

The removed first-stage and last-stage stub tests duplicate behavior already covered by the micro `mlx_lm` Qwen tests, which are closer to the real production path. This keeps the suite smaller while preserving the Metal-owned checks for middle-stage zero inputs and wire-dtype fail-fast behavior.


ricky-chaoju (approved):
LGTM

PR #539, opened 2026-07-23
Add vLLM-backed MHA KV cache layout

LxYuan0420:
This PR is:

- To add a Metal layout object for standard MHA KV cache.
- To translate vLLM’s `KVCacheConfig` into Metal’s per-layer cache layout.
- To let `MetalPagedKVCache` allocate shared KV slots from that layout.

Note: this is PR 1 of 3 for #505. Gemma-style mixed full/sliding MHA needs to use the cache groups and tensor slots from vLLM. Without that, Metal can keep treating the cache as one flat shape. This PR only adds the layout and allocation path. It does not change runtime routing yet.

I will submit two follow-up PRs:

- Carry grouped paged-KV metadata through the MHA runtime path.
- Use this layout for Gemma-style full/sliding MHA cache.

### DIFF
--- a/tests/test_per_layer_kv_cache.py
+++ b/tests/test_per_layer_kv_cache.py
@@ -7,9 +7,18 @@ from types import SimpleNamespace
 
 import mlx.core as mx
 import pytest
+import torch
+from vllm.v1.kv_cache_interface import (
+    FullAttentionSpec,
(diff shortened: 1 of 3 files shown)


ricky-chaoju (approved):
Thanks @LxYuan0420. LGTM

PR #540, opened 2026-07-23
[Doc] Document VLLM_METAL_BUILD_FROM_SOURCE in the configuration reference

feiiiiii5:
### Summary

The environment-variable reference table in `docs/configuration.md` documents 11 of the 12 `VLLM_METAL_*` variables defined in `vllm_metal/envs.py`, but omits **`VLLM_METAL_BUILD_FROM_SOURCE`**.

That is the variable the Contributing guide tells developers to set for source / kernel builds — `docs/CONTRIBUTING.md` uses `VLLM_METAL_BUILD_FROM_SOURCE=1` (line 29) and explains its effect (line 44) — yet a reader of the canonical configuration reference won't find it listed.

### Change (docs-only)
Add the missing row to the env-var table, matching its definition in `vllm_metal/envs.py` (default `0`; compiles the native `_paged_ops` extension from source at runtime instead of loading the prebuilt wheel artifact; requires the Xcode command-line tools), placed in the same order as `envs.py` (after `VLLM_METAL_MLA_KERNEL`).

No behavior change — this only brings the reference table back in sync with `envs.py` and the Contributing guide.


feiiiiii5:
Thanks @ricky-chaoju — the relative `CONTRIBUTING.md` link is the right call: it matches the other rows in the table (e.g. `distributed.md`) and points at the local dev guide that actually documents `VLLM_METAL_BUILD_FROM_SOURCE`. Looks good to me, and thanks for the quick polish.

ricky-chaoju (approved):

PR #541, opened 2026-07-23
[Bugfix] Validate VLLM_MLX_DEVICE instead of silently falling back to CPU

feiiiiii5:
### Summary

`MetalConfig` validates `memory_fraction`, `multimodal_mode`, and the TurboQuant quant types, but **`mlx_device` receives no validation** — even though it is typed `Literal["gpu", "cpu"]`.

`from_env()` passes the raw `VLLM_MLX_DEVICE` string straight through (`# type: ignore[arg-type]`), and both consumers treat anything that isn't exactly `"gpu"` as CPU:

```python
# vllm_metal/v1/worker.py  and  vllm_metal/platform.py
device_type = mx.DeviceType.gpu if config.mlx_device == "gpu" else mx.DeviceType.cpu
```

So a typo or wrong backend name (`GPU`, `gpuu`, `mps`, `cuda`) **silently routes the entire engine onto the MLX CPU device** — a large, unannounced performance regression on Apple Silicon, with no error or warning. This is the one config option that doesn't follow the file's fail-fast convention.

### Change

- Add `VALID_MLX_DEVICES` and validate `mlx_device` in `MetalConfig.__post_init__`, raising a clear `ValueError` consistent with the existing `memory_fraction` / `multimodal_mode` messages.
- Normalize the env value with `.strip().lower()` in `from_env()` — mirrors the existing `.lower()` handling of `VLLM_METAL_MEMORY_FRACTION=auto`, so `GPU` / ` gpu ` work while genuinely-wrong values fail fast.
- Add tests: invalid device rejected (direct construction + from env) and case normalization.

### Testing

```
pytest tests/test_config.py   # 20 passed
ruff check vllm_metal/config.py tests/test_config.py   # All checks passed!
```

No behavior change for valid `"gpu"`/`"cpu"` configs; wrong values now fail fast instead of silently degrading to CPU.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -206,3 +206,32 @@ class TestMetalConfig:
                 k_quant="q8_0",
                 v_quant="fp16",
             )
+
+    def test_invalid_mlx_device_rejected(self) -> None:
+        """An unknown mlx_device must fail fast, not silently fall back to CPU."""
+        for device in ["mps", "cuda", "gpuu", ""]:
+            with pytest.raises(ValueError, match="Invalid VLLM_MLX_DEVICE"):
+                MetalConfig(
+                    memory_fraction=AUTO_MEMORY_FRACTION,
+                    use_mlx=True,
+                    mlx_device=device,  # type: ignore[arg-type]
+                    debug=False,
+                    use_paged_attention=True,
+                )
+
+    def test_mlx_device_from_env_normalizes_case(self, monkeypatch) -> None:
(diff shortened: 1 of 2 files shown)


LxYuan0420 (commented):
Docs already state the valid values clearly. I don’t think we need to be this defensive here, so I’m closing it.

feiiiiii5:
Thanks @LxYuan0420, fair enough — I will keep contributions focused on documented gaps and reported issues rather than speculative hardening. Appreciate the quick look.

PR #545, opened 2026-07-24
Route MHA paged KV metadata by scheduler group

LxYuan0420:
This PR is:

- To carry one block-table and slot-map set per vLLM scheduler KV group through paged forward.
- To let the MHA paged runtime adopt the vLLM-backed cache layout added in #539.
- To route hybrid SDPA layers to the scheduler group assigned by vLLM.
- To keep Gemma4 MTP, YOCO, and spec-decode metadata on the grouped contract.

Note:

#539 added the vLLM-backed MHA cache layout but did not change serving routing. This is PR 2 of the Gemma 4 KV-cache fix plan: it trusts vLLM's KVCacheConfig as the source of truth for scheduler KV groups and passes those groups through Metal runtime metadata instead of assuming group 0 everywhere.

Follow-up:

PR 3 will emit Gemma-style full/sliding MHA KV specs so vLLM can plan separate groups for issue #505. After that lands, we should ask Ricky to run the real Gemma 4 31B serve smoke on M1 Ultra 128GB before closing #505.


vllm_metal/v1/cache_policy.py
```diff
@@ -352,11 +355,110 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+            )
+            if group_indices != (0,):
+                raise NotImplementedError(
+                    "TurboQuant MHA currently supports one scheduler KV group"
+                )
+            return
+
+        layout = MHAKVCacheLayout.from_config(kv_cache_config, model_layer_names)
+        runtime.adopt_layout(layout)
+        runtime.patch_model(self._runner.model)
```
ricky-chaoju, line 401: Gemma 4 doesn't start with this branch. `from_config` here hits `_group_layout()`, which only takes `FullAttentionSpec`/`SlidingWindowSpec`, but vLLM hands us a merged spec instead:

```
group=0 spec_type=UniformTypeKVCacheSpecs nlayers=15
  ('FullAttentionSpec', 1, 256, 16, None) x12
  ('FullAttentionSpec', 1, 512, 16, None) x3
```

We emit `FullAttentionSpec` for every layer and Gemma 4 widens the global ones to head_dim 512, so same type plus different shape means vLLM merges them into one group.

`gemma-4-e2b-it-4bit` and `gemma-4-31B-it-OptiQ-4bit` both serve on main and both fail here, including after I merged current main in. Qwen3-0.6B is unaffected.

The description says this keeps Gemma4 MTP and YOCO on the grouped contract, so I expected it to still boot. PR 3 probably fixes this on its own, since `SlidingWindowSpec` won't merge with `FullAttentionSpec`, but they land separately and main can't serve Gemma 4 in between. Could this fall back to the dense path instead of raising?

ricky-chaoju (commented):
Tested this on an M1 Ultra 128GB. Qwen3-0.6B is fine, Gemma 4 does not start. One inline note.

vllm_metal/attention/runtime/mha.py
```diff
@@ -67,13 +68,35 @@ def initialize(self, num_blocks: int) -> None:
...
+        """Replace the legacy dense cache with vLLM's grouped MHA layout."""
+        self._require_initialized("adopt_layout")
+        if self._turboquant:
+            raise NotImplementedError(
+                "layout-backed MHA runtime does not support TurboQuant"
+            )
+
+        self._layout = layout
+        self._block_size = layout.group_block_sizes[0]
+        self._cache = MetalPagedKVCache.from_layout(layout, self._dtype)
```
ricky-chaoju, line 81: `adopt_layout` builds the new pool while the wrappers still hold the old one, so both are live until `patch_model()` rebinds. On Qwen3-0.6B:

```
before adopt_layout  active= 56.93 GB
after  adopt_layout  active=112.67 GB  peak=112.67 GB
after  re-patch      active= 56.93 GB
```

It does come back down, so it's transient, but the peak is 112.67 GB against a 115.45 GB `metal_limit`. Since `kv_budget` is whatever is left inside `metal_limit * fraction`, the peak is roughly `model + 2 * kv_budget`, so it gets tightest on small models where kv_budget is largest. I didn't try a higher `VLLM_METAL_MEMORY_FRACTION`.

tests/test_per_layer_kv_cache.py
```diff
@@ -309,6 +309,52 @@ def test_allocates_shared_slots_from_layout(self) -> None:
         assert cache.group_index_for_layer(1) == 1
         assert cache.block_size_for_layer(1) == 16
 
+    def test_cache_policy_adopts_engine_layout(self, monkeypatch) -> None:
```
ricky-chaoju, line 368: This goes through `_mixed_mha_config()`, which hand-builds `KVCacheGroupSpec`s holding raw `FullAttentionSpec` and `SlidingWindowSpec`. vLLM doesn't hand us that shape today, since we never emit `SlidingWindowSpec` yet and it merges same-type specs, so this passes on a config the engine can't generate. That's why the Gemma break got through.

Could it run `get_kv_cache_groups()` on the real `get_kv_cache_spec()` output, the way `test_initialize_kv_cache_selects_sdpa_scheduler_group` does for the hybrid path? A mixed 256/512 head-dim case would have caught this.

LxYuan0420:
> Tested this on an M1 Ultra 128GB. Qwen3-0.6B is fine, Gemma 4 does not start. One inline note.

Not yet ready; PR3 is where this becomes ready for Gemma4/505 testing

ricky-chaoju (approved):
LGTM. Re-tested on the latest head with vLLM 0.26: gemma-4-e2b-it-4bit, gemma-4-31B-it-OptiQ-4bit, Qwen3-0.6B and Qwen3.5-0.8B all serve now.

PR #547, opened 2026-07-24
Experimental scheduler-coherent APC for hybrid GDN models

mbrfail:
## Summary
- add opt-in scheduler-physical-block GDN convolution/recurrent-state checkpoints for hybrid APC
- preserve all scheduler cache-group block tables and restore complete Mamba groups at aligned boundaries
- account sparse GDN checkpoints in Metal cache capacity and fail closed on missing/mismatched state
- restrict the experimental path to local text-only paged execution; reject KV transfer, TurboQuant, LoRA, speculative decoding, and distributed modes

## Validation
- Python 3.12 compile pass
- affected suite: 153 passed, 10 skipped on Apple M3 Ultra
- exact Qwen/Qwen3.6-35B-A3B-FP8 real endpoint startup on current main
- 11,616-token probe: 1.096 s exact-prefix hit and 0.235 s divergent-suffix hit
- scheduler metrics: 22,176 hit tokens; GDN restores logged at 10,560 and 11,616 aligned tokens
- cache-on/cache-off controlled experiment on pinned base: 11/11 functional checks, output equivalence, 7.05x exact-repeat and 32.39x divergent-suffix median speedups

## Risk / rollout
This remains behind VLLM_METAL_EXPERIMENTAL_GDN_APC=1. External KV transfer is rejected because vLLM 0.25.1 does not provide scheduler-coherent Mamba allocation generation metadata and GDN state is not transferred. Broader eviction, preemption, churn, and concurrency coverage is still needed before removing the gate.

Reproducible evidence: https://github.com/mbrfail/benchmark/tree/main/experiments/2026-07-23-qwen36-35b-a3b-vllm-metal-prefix-cache-m3-ultra


mbrfail:
## Production-readiness qualification: **NO-GO**

I ran the approved deterministic correctness gate on the M3 Ultra against this PR head (`f791709a9312af65091bb146d29bff892ed637d8`). The gate requires zero cache-on/cache-off output mismatches before eviction, preemption, concurrency, restart, or soak testing.

Measured results:

- cache-disabled controls: **144/144 passed**;
- APC matrix (`max_num_batched_tokens=2048`): **8/216 mismatches** (3.70%);
- block-sized scheduler-cap diagnostic (`1056`): **3/75 mismatches** (4.00%);
- APC matrix after aligned-only GDN snapshot publication: **8/216 mismatches** (3.70%);
- cache-disabled five-repeat checks for implicated prompts: **40/40 stable**;
- final affected regression suite: **321 passed, 10 skipped**.

Mismatches changed generated text, finish reason, and completion length (for example, `GDN_SUFFIX_OK` vs `<think>` continuations and EOS vs continued output). Paired top-20 generated-token logprobs show winner changes with margins up to ~0.875 log-prob units, so this is not only an exact-tie artifact.

Commit `f791709` is still a useful lifecycle fix: it stops publishing partial GDN states as reusable aligned checkpoints. It skipped 1,673 unaligned publications in the final arm, but did not resolve output correctness.

Per the stop rule, I did **not** run forced eviction/preemption, concurrency/restart qualification, or the two-hour soak. The PR should remain draft and must not be promoted.

Full report, commands, environment, harness, machine-readable results, logs, metrics, and controls:

https://github.com/mbrfail/benchmark/tree/0101b72724081c09e76ec74510f08641ce66af20/experiments/2026-07-24-qwen36-gdn-apc-production-readiness

mbrfail:
## Independent-review follow-up

An independent review of the raw matrices, controls, generated-token logprobs, server hit logs, harness, and aligned-publication patch agrees that the **NO-GO is supported** for the tested candidate.

Key review conclusions:

- the original and aligned-publication arms reproduced the same eight failures at the same request positions with identical response signatures;
- server logs confirm scheduler-approved GDN snapshot hits, rather than hits inferred from latency;
- cache-off repeat controls were stable;
- the evidence proves **not production-ready**, but does not isolate one root cause;
- the aggregate 19/507 artifact count spans unlike configurations and must not be treated as a single-build failure rate.

I incorporated the review's actionable hardening in PR head `9b51e1b37e46564b02aa838a86d950536af3aa49`:

- prove an unaligned checkpoint cannot overwrite a prior valid aligned snapshot or alter its accounting/restore result;
- cover mixed aligned/unaligned checkpoint batches;
- reduce per-token unaligned-skip logging from INFO to DEBUG.

Post-review affected suite: **322 passed, 10 skipped**.

Updated report and raw test log:

https://github.com/mbrfail/benchmark/tree/c25abe33385ad018c8f7c4d4e8573811a44d3bac/experiments/2026-07-24-qwen36-gdn-apc-production-readiness

The PR should remain draft; no promotion was performed.

PR #549, opened 2026-07-25
Re-ensure the ByteLevel tokenizer patch after vLLM finishes importing

### LINKED ISSUE #548: [Bug] ByteLevel tokenizer compat patch never installs because plugin registration always skips it
## Summary

The ByteLevel tokenizer patch added in #337 and refined in #342 (fixing #336) never actually installs.

`apply_compat_patches()` runs at plugin registration time, while `vllm.utils.torch_utils` is still mid-import. The patch's own `import vllm.tokenizers.registry` therefore raises `ImportError`, the `except` branch logs a warning and returns, and nothing retries it later. The result is that `Ġ` and `Ċ` token pieces leak back into served text for exactly the model repos #336 was about.

## Reproduce

```bash
python -c "
import vllm, vllm.platforms as P
P.current_platform
import vllm.tokenizers.registry as reg
print('patch installed:', getattr(reg, '_vllm_metal_bytelevel_decoder_patch', False))
"
```

```
patch installed: False
```

The warning is printed on every startup:

```
WARNING [compat.py:504] Could not install vLLM ByteLevel tokenizer compatibility
patch because vLLM tokenizer registry is unavailable: cannot import name
'common_broadcastable_dtype' from partially initialized module
'vllm.utils.torch_utils' (most likely due to a circular import)
```

To confirm this is not just a cosmetic flag, here is the repo's own fixture (`tests/test_compat.py::_write_mismatched_bytelevel_tokenizer_repo`) driven through `get_tokenizer` in the state a real server is actually in:

```
patch installed at this point: False
decoded (UNPATCHED): 'ĠHelloĊ'      <- the #336 symptom
decoded (PATCHED):   ' Hello\n'
```

## Root cause

`vllm/utils/torch_utils.py` runs `PIN_MEMORY = is_pin_memory_available()` at module level, and that call resolves `current_platform`, which loads this plugin. So `_register()` always runs from inside a partially initialized `torch_utils`:

```
vllm/__init__.py:14
 └─ vllm/env_override.py:90          from vllm.utils.torch_utils import ...
    └─ vllm/utils/torch_utils.py:72  PIN_MEMORY = is_pin_memory_available()
       └─ vllm/utils/platform_utils.py:44   from vllm.platforms import current_platform
          └─ vllm/platforms/__init__.py:219 resolve_current_platform_cls_qualname()
             └─ vllm_metal/__init__.py:117  _register()
                └─ vllm_metal/compat.py:31  apply_compat_patches()
```

The patch then imports `vllm.tokenizers.registry`, which reaches straight back into that half-built module:

```
vllm/tokenizers/__init__.py:4          from .hf import maybe_make_thread_pool
vllm/tokenizers/hf.py:11               from vllm.transformers_utils.config import ...
vllm/transformers_utils/config.py:34   from vllm.utils.torch_utils import common_broadcastable_dtype
```

`common_broadcastable_dtype` is defined at `torch_utils.py:255`, well below the line 72 that started all of this, so it does not exist yet and the import fails. Line numbers are from vLLM 0.25.1. The same ordering holds on 0.24.0 (72/255 vs 71/254).

This is the same failure mode as #222. Worth noting that `ensure_vllm_auto_fit_null_block_patch` already documents this exact hazard in its docstring and works around it with a re-ensure hook from `MetalPlatform.update_block_size_for_backend` (#513). The ByteLevel patch has no equivalent. It is also the only one of the six patches in `apply_compat_patches()` that fails, the other five install fine.

## Why CI does not catch it

`tests/test_compat.py` calls `_patch_vllm_bytelevel_tokenizer_loading()` directly, and the `apply_compat_patches` test replaces it with a `lambda` that only records that it was called. The patch body is well covered, but installation through real registration never is, so the suite stays green while the patch is dead.

## Environment

- vllm-metal `main` @ 29f8e3f
- vLLM 0.25.1 (the version `install.sh` pins), also reproduces on 0.24.0
- macOS 15.5, Apple M4, Python 3.12

## Proposed fix

Mirror what #513 did for the sibling patch: rename to `ensure_vllm_bytelevel_tokenizer_patch` (it is already idempotent via its sentinel), demote the registration time `ImportError` to `logger.debug`, and re-ensure it from `MetalPlatform.check_and_update_config`. vLLM calls that from `VllmConfig.__post_init__`, which is after vLLM is fully imported, runs in every process that builds a config, and happens before the serving tokenizer is created (`renderers/registry.py` -> `cached_tokenizer_from_config` -> `cached_get_tokenizer`).

`check_and_update_config` rather than `update_block_size_for_backend` because the serving tokenizer is built in the front-end process, which the KV sizing hook never reaches.

I have this working locally with a regression test that fails without the hook and passes with it. Happy to open a PR if this approach looks right.

varad-more:
Fixes #548.

The ByteLevel tokenizer compatibility patch from #337, refined in #342, has never installed. Not in a corner case, not on some transformers versions. Never, in any real `vllm serve`. So the `Ġ` and `Ċ` pieces that #336 was filed about have been leaking into served text this whole time for the affected MLX-community Qwen and DeepSeek repos.

The patch body itself is fine. It just never runs.

## Why it always fails

Plugin registration is triggered from inside `vllm.utils.torch_utils`'s own import. Its module-level `PIN_MEMORY = is_pin_memory_available()` is what resolves `current_platform`, which loads this plugin, which calls `apply_compat_patches()`:

```
vllm/env_override.py:90            from vllm.utils.torch_utils import ...
 └─ vllm/utils/torch_utils.py:72   PIN_MEMORY = is_pin_memory_available()
    └─ vllm/utils/platform_utils.py:44   from vllm.platforms import current_platform
       └─ vllm/platforms/__init__.py:219 resolve_current_platform_cls_qualname()
          └─ vllm_metal/__init__.py:117  _register()
             └─ vllm_metal/compat.py:31  apply_compat_patches()
```

The patch then does `import vllm.tokenizers.registry`, which reaches back into that half-built module:

```
vllm/tokenizers/__init__.py:4          from .hf import maybe_make_thread_pool
vllm/tokenizers/hf.py:11               from vllm.transformers_utils.config import ...
vllm/transformers_utils/config.py:34   from vllm.utils.torch_utils import common_broadcastable_dtype
```

`common_broadcastable_dtype` is defined at `torch_utils.py:255`, well below the line 72 that started all of this. ImportError, every time. The `except` branch logged a warning and returned, and nothing ever retried.

Checking the sentinel after full platform activation:

```
$ python -c "
import vllm, vllm.platforms as P
P.current_platform
import vllm.tokenizers.registry as reg
print(getattr(reg, '_vllm_metal_bytelevel_decoder_patch', False))"
False
```

It is the only one of the six patches in `apply_compat_patches()` that fails. The other five install cleanly, which is probably why the warning on every startup never got a second look.

## Why CI stayed green

`tests/test_compat.py` calls the patch function directly, and the `apply_compat_patches` test replaces it with a lambda that only records that it was called. The patch body is well covered. Its installation through the real registration path was not covered at all, so the suite happily stayed green while the patch was dead.

## The fix

This is the same hazard `ensure_vllm_auto_fit_null_block_patch` already documents in its own docstring, and #513 solved it there with a re-ensure hook. This does the same thing for the ByteLevel patch rather than inventing a second mechanism:

- Renamed to `ensure_vllm_bytelevel_tokenizer_patch`, matching the sibling's `ensure_*` naming that signals "safe to call repeatedly". It was already idempotent via its sentinel.
- Re-ensured from `MetalPlatform.check_and_update_config`.
- The registration-time ImportError drops from `logger.warning` to `logger.debug`. With a re-ensure hook it is expected and benign at that point, and it was firing on every startup.

The patch body is untouched.

## On the hook choice

`check_and_update_config` rather than `update_block_size_for_backend`, which is where #513 put the sibling. vLLM calls `check_and_update_config` from `VllmConfig.__post_init__`, so it runs after vLLM is fully imported and in every process that builds a config. That matters here because the serving tokenizer is built in the front end process, via `renderers/registry.py:83` to `cached_tokenizer_from_config` to `cached_get_tokenizer`, which the KV sizing hook never reaches.

Worth noting for review: the patch rebinds both `get_tokenizer` and `cached_get_tokenizer`. That second rebinding is load bearing, because `cached_tokenizer_from_config` resolves `cached_get_tokenizer` as a module global at call time.

## Testing

New test in `tests/test_platform.py` asserts the sentinel is set on `vllm.tokenizers.registry` after `check_and_update_config`, which is the coverage gap that let this ship. I verified it fails with the `platform.py` hook reverted and passes with it, rather than just assuming it would.

`tests/test_compat.py` and `tests/test_platform.py`: 90 passed, 13 skipped. `ruff check`, `ruff format --check`, and `mypy vllm_metal` are clean.

To confirm this is not merely a flag flipping from False to True, here is the repo's own fixture (`_write_mismatched_bytelevel_tokenizer_repo`) driven through `get_tokenizer` in the state a real server is actually in:

```
patch installed at this point: False
decoded (UNPATCHED): 'ĠHelloĊ'      <- the #336 symptom
decoded (PATCHED):   ' Hello\n'
```

Reproduced against vLLM 0.25.1, the version `install.sh` pins. The same ordering holds on 0.24.0 (72/255 vs 71/254), so this is not specific to the pinned release.

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -127,7 +127,7 @@ class TestByteLevelTokenizerCompatPatch:
         from transformers.tokenization_utils_tokenizers import TokenizersBackend
 
         _write_mismatched_bytelevel_tokenizer_repo(tmp_path)
-        compat._patch_vllm_bytelevel_tokenizer_loading()
+        compat.ensure_vllm_bytelevel_tokenizer_patch()
 
         tokenizer = tokenizer_registry.get_tokenizer(tmp_path, tokenizer_mode="hf")
         decoded = tokenizer.decode([0, 1])
@@ -269,7 +269,7 @@ class TestGemma4MTPConfigCompatPatch:
         )
         monkeypatch.setattr(
             compat,
-            "_patch_vllm_bytelevel_tokenizer_loading",
+            "ensure_vllm_bytelevel_tokenizer_patch",
             lambda: calls.append("bytelevel"),
         )
         monkeypatch.setattr(
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -246,7 +246,7 @@ class TestGemma4MTPConfigCompatPatch:
         assert config.text_config.num_kv_shared_layers == 0
         assert get_hf_text_config(config).model_type == "gemma4_text"
 
-    def test_missing_gemma4_config_module_does_not_stop_other_patches(
+    def test_registration_time_failures_do_not_stop_other_patches(
         self,
         monkeypatch,
     ) -> None:
@@ -267,10 +267,15 @@ class TestGemma4MTPConfigCompatPatch:
             "_gemma4_assistant_config_class",
             _raise_missing_gemma4_config,
         )
+
+        def _raise_bytelevel_import_error() -> None:
+            calls.append("bytelevel")
+            raise ImportError("partial vLLM import")
+
         monkeypatch.setattr(
             compat,
             "ensure_vllm_bytelevel_tokenizer_patch",
-            lambda: calls.append("bytelevel"),
+            _raise_bytelevel_import_error,
         )
         monkeypatch.setattr(
             compat,
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -135,6 +135,82 @@ class TestMetalPlatform:
         finally:
             reset_config()
 
+    def test_check_and_update_config_reensures_bytelevel_tokenizer_patch(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """Plugin activation can never install the ByteLevel tokenizer patch.
+
+        ``_register()`` runs from inside ``vllm.utils.torch_utils``'s own
+        import -- its module-level ``PIN_MEMORY = is_pin_memory_available()``
+        is what resolves the platform -- so ``import vllm.tokenizers`` reaches
+        a partially initialized ``torch_utils`` and raises. The patch is
+        skipped every time. Without this post-import re-ensure, the fix from
+        #337/#342 never installs and ByteLevel pieces leak into served text.
+        """
+        import vllm.tokenizers.registry as tokenizer_registry
+
+        sentinel = "_vllm_metal_bytelevel_decoder_patch"
+
+        # Register both rebindings with monkeypatch so teardown restores them,
+        # then clear the sentinel to simulate a freshly registered process.
+        monkeypatch.setattr(
+            tokenizer_registry,
+            "get_tokenizer",
+            tokenizer_registry.get_tokenizer,
+        )
+        monkeypatch.setattr(
+            tokenizer_registry,
+            "cached_get_tokenizer",
+            tokenizer_registry.cached_get_tokenizer,
+        )
+        # setattr (not delattr) so teardown restores a previously-absent
+        # sentinel by deleting it, instead of leaving it set for later tests.
+        monkeypatch.setattr(tokenizer_registry, sentinel, False, raising=False)
+        assert not getattr(tokenizer_registry, sentinel, False)
+
+        self._patch_stt_resolution(monkeypatch, is_stt=False)
+        reset_config()
+        try:
+            vllm_config = SimpleNamespace(
+                parallel_config=SimpleNamespace(
+                    worker_cls="auto",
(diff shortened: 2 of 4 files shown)


LxYuan0420 (approved):
Thanks for fixing it

PR #550, opened 2026-07-25
[GPT-OSS] Wire attention sinks through the paged attention kernel

### LINKED ISSUE #212: [GPT-OSS] Ground truth tests + sink attention kernel support
Based on the discussion in #148, opening this issue to track GPT-OSS support:

- [ ] mlx_lm groundtruth for GPT-OSS
- [ ] Sink attention kernel for GPT-OSS

Here's my plan for the ground truth task

1. Create `tools/gen_golden_token_ids_gpt_oss.py` (similar to `tools/gen_golden_token_ids_for_deterministics.py` but adapted for GPT-OSS). Key difference is applying the harmony chat template via `AutoTokenizer` before passing prompts to `LLM.generate()`.

2. Create `tests/test_gpt_oss_smoke.py` following the `test_qwen35_smoke.py` pattern with:
   - 5 parametrized MLX-path smoke tests comparing against golden token IDs (greedy decoding, non-paged path)
   - 5 parametrized paged-path tests marked `xfail(strict=True)` since sink attention handling is unverified (#191 shows 1/5 content correctness). This creates a concrete test target for the sink attention kernel task.

**Questions**

- Is [mlx-community/gpt-oss-20b-MXFP4-Q4](https://huggingface.co/mlx-community/gpt-oss-20b-MXFP4-Q4) an acceptable reference model? The 8-bit variant needs ~20GB memory which could be an issue for CI.
- With GPT-OSS, `MAX_TOKENS=10` only captures the start of the harmony "analysis" phase (thinking tokens), not actual answer content. Should I increase this, or is 10 tokens sufficient for a deterministic smoke test?

varad-more:
Closes #212. That issue tracks two things: the ground truth tests, which landed in #221, and sink attention kernel support, which is this.

Two things worth knowing before reading the diff: GPT-OSS doesn't currently run on `main` at all, and the sink code in `pagedattention.metal` had never executed, so two of its three paths were wrong.

## GPT-OSS crashes before sinks even matter

`sdpa_forward` reads `inner.scale`, but GPT-OSS's `AttentionBlock` calls it `sm_scale`, so the engine dies during startup:

```
AttributeError: 'AttentionBlock' object has no attribute 'scale'
```

That's blocker 1 from #212. The head count probe at `sdpa.py:437` was added at some point, but the scale half never was, so nothing generates. This uses the same `getattr` pattern already sitting there for the head counts.

## The sinks were unreachable

`pagedattention.metal` has carried sink support all along: `use_sinks` as function constant 40, the sink array at buffer 18 in the main kernel and buffer 6 in the reduce, plus folding logic in the non-partitioned, split-KV reduce and window paths. None of it could run. `paged_ops.cpp:436` hardcoded `bool use_sinks = false;` and the nanobind signature had no `sinks` argument, so buffers 18 and 6 were never bound.

MLX 0.32 added `sinks=` to `mx.fast.scaled_dot_product_attention`, which gives a first class oracle to check against. On `main`:

```
max|kernel - mlx_sdpa(sinks=None)| = 3.576e-07
max|kernel - mlx_sdpa(sinks=2.0)|  = 4.535e-02
```

Plain softmax exactly, and roughly 45x the fp32 parity tolerance away from the sink case.

## Two of the three sink paths were also wrong

Wiring the argument through wasn't enough. Because the sink code had never run, the per-token kernel and its window mode both folded the sink into every warp's state before the threadgroup merge summed all `NUM_WARPS` states, so the sink landed in the denominator `NUM_WARPS` times.

Pinned by comparing against a reference that deliberately over-counts, on a 256-thread threadgroup with 32-lane simdgroups:

```
sink counted 1x -> max_abs_diff = 1.959e-01
sink counted 8x -> max_abs_diff = 1.788e-07
```

Both now fold the sink exactly once, after the merge completes. The split-KV path was already correct, since `paged_attention_v2_reduce` folds the sink into the global max and exp-sum, counting it once across partitions rather than once per partition. After the fix all three paths agree with the oracle at ~1e-07.

## Prefill routing

`pagedattention_tiled.metal` has no sink support whatsoever, and prefill batches route to it. Rather than silently dropping the sink term, a sinks batch is excluded from the tiled path and falls through to the per-token kernel, mirroring exactly what `use_turboquant` already does in that same condition. Sink support in the tiled kernel is a reasonable follow-up if GPT-OSS prefill throughput warrants it.

## Sinks with TurboQuant

Rejected up front rather than ignored. Upstream refuses the same combination (`mlx_lm/models/base.py`: "Quantized SDPA does not support attention sinks"), and the TurboQuant reduce already owns the deferred inverse FWHT, so folding a sink there would need its own derivation. Failing loudly follows #510 and #523 rather than returning quietly wrong numbers.

## Testing

`tests/test_paged_attention_sinks.py` is deterministic, needs no model download, and runs in well under a second:

- parity against the MLX oracle across fp32/fp16/bf16, GQA ratios (8:8, 8:2, 16:4), head sizes 64 and 128, and both the single-partition and split-KV paths
- per-head sink values, to catch a kernel reading head 0's sink for every head
- window mode against the expanded per-token layout, non-partitioned and split-KV
- a direct pin on the warp multiplication bug
- guards that `sinks=None` leaves the existing path unchanged, and that passing sinks actually changes the output
- fail-fast coverage: wrong head count, non-1D, non-float32, and the TurboQuant combination

Regression: `tests/test_spec_window_parity.py` (102), `tests/test_paged_deterministic.py` (6 golden token IDs against a real model), `tests/test_attention_sdpa.py`, `tests/test_primitive_and_donation.py` and `tests/test_paged_attention.py` all pass. `ruff check`, `ruff format --check` and `mypy vllm_metal` are clean.

Run on an M4 and an M4 Pro, which have different GPU core counts, so the split-KV partition mix differs between the two.

## GPT-OSS 20B end to end

`openai/gpt-oss-20b` on the M4 Pro (48GB), vllm 0.25.1, mlx 0.32.0, greedy, `MAX_TOKENS=100`, via `tools/test_gpt_oss_smoke.py`.

Before, on `main`: 0/5, no tokens generated, hard crash at the `inner.scale` AttributeError above. After: engine healthy, all five prompts generate, and the smoke test scores 2/5 against its committed goldens.

That 2/5 deserves an explanation rather than a shrug, and it isn't the FP non-determinism the file's own comment warns about. Three repeat runs produced bit-identical token IDs, so the divergence is deterministic. Regenerating the reference on the same machine through the MLX inline-cache path (`VLLM_METAL_USE_PAGED_ATTENTION=0`, which bypasses the paged kernel and uses MLX's native sink-aware SDPA) gives:

| prompt | oracle vs committed golden | paged vs oracle |
| --- | --- | --- |
| The capital of France is | identical | **identical** |
| The weather today is not | differs @ 56 | **identical** |
| One plus one equals | differs @ 49 | differs @ 19 |
| The largest planet in our solar system is | differs @ 29 | **identical** |
| Water boils at a temperature of | identical | **identical** |

So four of five are bit-identical to a same-machine MLX reference. Two of the three smoke-test failures are stale committed goldens: the oracle diverges from them at the same indices the paged run did, and the file's own note says to regenerate when that happens. Worth fixing, but it's a separate change from this one.

The remaining divergence is a semantic near-tie, `' 2.'` against `' "Two".'`, which re-syncs after three tokens and stays aligned. That argues against a sink defect: a wrong sink term biases the softmax denominator at every step, so sequences would drift apart rather than re-converge. Taken with the unit tests matching `mx.fast.scaled_dot_product_attention` at ~1e-07 across all three sink paths, this reads as marginal-logit tie-breaking.

The before/after is gated by a probe that asserts the rebuilt extension matches the checked-out branch (`sinks` absent on `main`, present here) and aborts rather than reporting a number from a stale `.so`.

### DIFF
--- a/tests/test_attention_sdpa.py
+++ b/tests/test_attention_sdpa.py
@@ -755,6 +755,77 @@ class TestSDPAForward:
 
         assert captured["num_kv_heads"] == actual_kv_heads
 
+    def test_gpt_oss_scale_and_sinks_reach_kernel(self) -> None:
+        """GPT-OSS exposes ``sm_scale`` and per-head attention sinks."""
+        cache = MetalPagedKVCache(
+            num_layers=1,
+            num_kv_heads=_N_KV_HEADS,
+            head_dim=_HEAD_DIM,
+            num_blocks=1,
+            block_size=8,
+            dtype=mx.float16,
+        )
+        expected_scale = 0.125
+        fp16_sinks = mx.arange(_N_HEADS, dtype=mx.float16)
+        inner = SimpleNamespace(
+            num_attention_heads=_N_HEADS,
+            num_key_value_heads=_N_KV_HEADS,
+            sm_scale=expected_scale,
+            sinks=fp16_sinks,
+            o_proj=lambda out: out,
+        )
+        ctx = _make_ctx(_SEQ_LEN)
+        x = mx.ones((_BATCH, _SEQ_LEN, _HIDDEN))
+        queries = mx.ones((_BATCH, _N_HEADS, _SEQ_LEN, _HEAD_DIM))
+        keys = mx.ones((_BATCH, _N_KV_HEADS, _SEQ_LEN, _HEAD_DIM))
+        values = mx.ones((_BATCH, _N_KV_HEADS, _SEQ_LEN, _HEAD_DIM))
+        captured: dict[str, object] = {}
+
+        class _FakeOps:
+            def reshape_and_cache(
+                self,
+                _key,
+                _value,
+                key_cache,
+                value_cache,
+                _slot_mapping,
+            ) -> tuple[mx.array, mx.array]:
+                return key_cache, value_cache
+
+            def paged_attention_primitive(
+                self,
+                _query,
+                _key_cache,
+                _value_cache,
+                _num_kv_heads,
+                scale,
+                *_args,
+                sinks=None,
+                **_kwargs,
+            ) -> None:
+                captured["scale"] = scale
+                captured["sinks"] = sinks
+
+        with (
+            patch.object(
+                sdpa_mod,
+                "prepare_sdpa_qkv",
+                return_value=(queries, keys, values, None, (keys, values)),
+            ),
+            patch.object(sdpa_mod, "get_ops", return_value=_FakeOps()),
+            patch.object(
+                sdpa_mod,
+                "truncate_padded_output",
+                return_value=mx.zeros((_BATCH, _SEQ_LEN, _N_HEADS * _HEAD_DIM)),
+            ),
+        ):
+            sdpa_forward(inner, x, ctx, cache, layer_idx=0)
+
+        sinks = captured["sinks"]
+        assert captured["scale"] == expected_scale
+        assert isinstance(sinks, mx.array)
+        assert sinks.dtype == mx.float32
+
     def test_shared_kv_path_does_not_rebind_cache_arrays(self) -> None:
         """Shared-KV attention must read the existing cache without writing."""
         cache = MetalPagedKVCache(
--- a/tests/test_attention_sdpa.py
+++ b/tests/test_attention_sdpa.py
@@ -615,8 +615,9 @@ class _PagedRoutingOpsSpy:
         _sliding_window: int,
         _out: mx.array,
         window_seqlen_q: int = 1,
+        sinks: mx.array | None = None,
     ) -> None:
-        del window_seqlen_q
+        del window_seqlen_q, sinks
         self.calls[-1].block_tables = block_tables.tolist()
         self.calls[-1].block_size = block_size
 
--- a/tests/test_paged_attention_sinks.py
+++ b/tests/test_paged_attention_sinks.py
@@ -109,6 +109,64 @@ def _run_decode(
     return out.reshape(num_q_heads, head_size), ref[0, :, 0, :]
 
 
+def _run_prefill(
+    ctx: int,
+    query_len: int,
+    num_q_heads: int,
+    num_kv_heads: int,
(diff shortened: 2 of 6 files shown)


LxYuan0420 (approved):
I rebased it onto current `main` and added a few focused fixes. For GPT-OSS serving, 3 of 5 prompts were token-identical. The other 2 differed from the inline/gold wording, but I checked the outputs and both look coherent: the weather prompt still asks for clarification, and the boiling-point prompt gives the same correct answer with pressure qualification. So I see those as wording differences, not evidence of a broken sink path.

I will share the output below

LxYuan0420:
**Prompt**

```text
The weather today is not
```

**Paged attention**

```text
It looks like your message got cut off. Could you let me know what you’re looking for? If you’d like a weather update or something else, just let me know!
```

**Inline attention**

```text
It sounds like you’re starting a thought about the weather—could you let me know a bit more? For example:

- Are you looking for the current weather in a specific city or region?
- Do you want a forecast for the next few days?
- Or is there something else you’d like to discuss about today’s weather?

Just let me know, and I’ll help you out!
```

Both correctly recognize that the prompt is incomplete and ask for clarification.

----

**Prompt**

```text
Water boils at a temperature of
```

**Paged attention**

```text
Water boils at **100 °C** (212 °F, 373.15 K) when the surrounding pressure is one atmosphere (≈101.3 kPa).
```

**Inline attention**

```text
Water boils at **100 °C** (212 °F, 373.15 K) under standard atmospheric pressure (1 atm).

> **Note:** The boiling point decreases with lower pressure (e.g., at higher altitudes) and increases with higher pressure.
```

Both give the same correct boiling point and pressure qualification. The difference is wording, not factual correctness.

PR #551, opened 2026-07-25
[Spec Decode] Release drafter state for evicted, preempted, and resumed requests

samithaj:
The runner's lifecycle reconcile released the paged runtime's recurrent state for evicted, preempted, and resumed requests, but the drafter's own per-request state was never released. A request that stops running keeps pinning a bounded drafter resource for as long as it waits.

I added `release_requests` to the `MetalProposer` seam and call it from `_reconcile_request_lifecycle` on the same invalidation set the runtime already uses (evicted + preempted + resumed). `DraftModelProposer` returns the request's draft KV blocks to its free pool and drops its committed-length bookkeeping, so a waiting request no longer holds draft cache; a resumed request re-ingests from the paged context and draws fresh blocks. The n-gram and Gemma4 proposers pin no bounded resource, so they are no-ops.

The pinning is real rather than theoretical: a preempted request keeps its `RequestState`, so `DraftModelProposer._prune_finished` never reclaims its blocks — only an explicit release can.

Tests:

- `tests/test_draft_model_proposer.py::test_release_requests_returns_draft_blocks_to_the_free_pool` drives allocation through `propose()` with a stub draft model until the block pool is drained, releases the waiting request, and asserts the next request is handed the same block ids. `test_draft_blocks_stay_pinned_without_release` is the paired negative: without the release the allocator fails before any draft forward runs.
- `tests/test_v1_model_runner_generate.py::TestDrafterReleaseOnLifecycle` covers the runner side — the drafter receives the union of evicted, preempted, and resumed ids.

`pytest -m "not slow" tests/` — 1622 passed, 15 skipped, 46 deselected.

This is the first of a small series splitting #535 as requested: the drafter release seam (here), the hosted-config front door, the GLM-4.7-Flash MTP head model and loader, the runtime path that runs it as a drafter (carrying a download-free engine-level test), and the benchmark helper with docs.

### DIFF
--- a/tests/test_draft_model_proposer.py
+++ b/tests/test_draft_model_proposer.py
@@ -1,0 +1,121 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the draft-model proposer's draft KV block pool.
+
+A stub draft model returns logits of the right shape, so the block allocator and
+the release path run through ``propose`` without loading any weights.
+"""
+
+from __future__ import annotations
+
+import mlx.core as mx
+import pytest
+from vllm.sampling_params import SamplingParams
+
+from vllm_metal.attention.context import OffsetCache, get_context
+from vllm_metal.v1.draft_model_proposer import DraftModelProposer
+from vllm_metal.v1.model_runner import RequestState
+from vllm_metal.v1.proposer import ProposeContext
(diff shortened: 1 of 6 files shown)


LxYuan0420 (approved):
LGTM. This is a focused lifecycle fix

PR #553, opened 2026-07-25
[Spec Decode] Register the GLM-4.7-Flash MTP head config for vLLM

samithaj:
Second of the PRs split out of #535, stacked on #551 — that PR's commit shows up here until it merges.

The hosted GLM-4.7-Flash drafter checkpoint carries its identity at the top level (`model_type: glm4_moe_lite_mtp`) and nests the source model config under `text_config`, so `architectures` and `num_nextn_predict_layers` are not at the top level at all. vLLM already knows how to normalize such a draft into its MTP wrapper (`Glm4MoeLiteMTPModel`, `num_hidden_layers=0`, `n_predict=num_nextn_predict_layers`), but it keys that override on the top-level fields, and the pinned transformers has no AutoConfig entry for `glm4_moe_lite_mtp` — so the config load raises before the override ever runs.

I registered the config with `AutoConfig` and promoted `architectures` and `num_nextn_predict_layers` out of the nested `text_config`. This mirrors the existing `_patch_vllm_gemma4_mtp_config_loading` shim next to it and covers config parsing only; it never constructs a model.

The config class is module-scope rather than a closure so vLLM's spawn-based EngineCore can pickle a config instance by qualified name into the worker process. A closure-scoped class fails only in the real engine, never in an in-process config test, so `test_compat_config_survives_pickle_round_trip` covers that round trip explicitly.

Tests, in `tests/test_compat.py`:

- `test_vllm_get_config_applies_existing_glm4_moe_lite_mtp_override` — vLLM's own `get_config` front door, the call that crashed on the raw hosted config, now produces the normalized MTP wrapper.
- `test_speculative_config_builds_from_hosted_head` — a full `SpeculativeConfig` builds from the hosted head against a `glm4_moe_lite` target.
- `test_transformers_autoconfig_loads_raw_glm4_moe_lite_mtp_config`, `test_compat_config_survives_pickle_round_trip`, and `test_missing_glm4_config_module_does_not_stop_other_patches` cover the raw load, the pickle round trip, and a missing `glm4_moe_lite` module leaving the surrounding patches intact.

The one change outside the new code is `TestGemma4MTPConfigCompatPatch`: its expected patch-call list gains the new entry, and it stubs the new patch so an unrelated Gemma4 test does not register `glm4_moe_lite_mtp` with `AutoConfig` as a side effect.

Tests: `pytest -m "not slow" tests/` — 1627 passed, 15 skipped, 46 deselected. `scripts/lint.sh` clean.

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -7,6 +7,7 @@ import builtins
 import importlib.util
 import json
 import os
+import pickle
 import sys
 from types import ModuleType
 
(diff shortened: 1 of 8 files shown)


LxYuan0420:

PR #554, opened 2026-07-25
Trim duplicate GGUF parity tests

LxYuan0420:
This PR is:

- To keep GGUF quantized math parity in the native tensor tests.
- To make GGUF wrapper tests cover only wrapper-owned module behavior.
- To share repeated Qwen/Llama loader cases through parameterized tests.

Note: `GGUFLinear` and `GGUFEmbedding` do not own quantized GGUF math; they delegate to the `GGUFTensor` contract. The native tensor tests still cover real GGUF files, MLX repack, qtype behavior, dense-oracle parity, unsupported Q4_1, and real generation. Loader tests still cover wrapper install, tied and untied heads, bias policy, q/k un-permutation, and fail-fast cases.

### DIFF
--- a/tests/test_gguf_loader.py
+++ b/tests/test_gguf_loader.py
@@ -168,11 +168,25 @@ def _gguf_module_histogram(model: nn.Module) -> dict[str, int]:
     return counts
 
 
+def _assert_forward_vocab_shape(model: nn.Module) -> None:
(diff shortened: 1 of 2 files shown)


ricky-chaoju (approved):
LGTM

PR #555, opened 2026-07-26
[Spec Decode] Add the GLM-4.7-Flash MTP head model and checkpoint loader

samithaj:
GLM-4.7-Flash (`glm4_moe_lite`) ships a single trained nextn layer that predicts one extra token. This adds the pieces needed to run that layer as a standalone MTP head — the MLX model, the checkpoint weight transform, and the loader — without wiring anything into the runtime.

`vllm_metal/v1/mtp_heads/glm4_moe_lite_mtp_model.py`

- `Glm4MoeLiteMTPArgs` — model args for the extracted head. `from_dict` reads the shape out of the nested `text_config` of the drafter-split config rather than silently falling back to the GLM-4.7-Flash defaults, and rejects a `first_k_dense_replace` / `moe_layer_freq` pair that would build a dense `mtp_block` where the trained nextn layer is MoE.
- `Glm4MoeLiteMTPModel` — `build_slot_inputs` projects `[embed(t_{p+1}), h_p]` into one row per slot and zeroes the absolute-position-0 embedding; `forward_slots` runs one stock `Glm4MoeLiteDecoderLayer` over the slab and takes an optional `expected_offset` so a cache/position anchor desync fails loud instead of drafting at the wrong position; `compute_logits` reads out through the untied `lm_head`.
- `convert_nextn_weights` — the absorbed-MLA `kv_b_proj` split plus MoE expert stacking, idempotent so it accepts either the pre- or post-`sanitize` layout. The mlx-vlm drafter split applies the same transform, so there is one source of truth for the layout.

`vllm_metal/v1/mtp_heads/glm4_moe_lite_mtp.py`

- `Glm4MoeLiteMTPHeadMetadata` validates the head config against the drafter-split schema and fails loud when the head and target disagree on any shared MLA / embedding hyperparameter, or when the target is not a `glm4_moe_lite` backbone.
- `Glm4MoeLiteMTPHeadLoader` downloads (or reads from disk), preflights `config.json` before touching weights, loads through `mlx_lm.load_model` with the head's own model classes, asserts the checkpoint carries `lm_head.weight` and `model.embed_tokens.weight`, and caches per (model, revision).

The loader imports the MLX classes only inside `_get_model_classes`, so it stays importable without mlx / mlx_lm and config detection never constructs Metal-backed modules. Verified with mlx and mlx_lm blocked at the import-system level: `vllm_metal.v1.mtp_heads.glm4_moe_lite_mtp` imports with zero mlx modules loaded, while `vllm_metal.v1.mtp_heads.glm4_moe_lite_mtp_model` fails on `mlx`.

The load-bearing test is `test_default_loader_loads_real_head_from_disk`: it writes a real safetensors checkpoint plus the hosted drafter-split `config.json`, loads it through the default loader path with no injected fakes, and checks the loaded model produces finite logits over a slot stream. `test_glm4_moe_lite_mtp_model.py` pins the forward against a self-contained numpy reference of the upstream nextn math (enorm/hnorm/eh_proj, position-0 masking, absorbed MLA with traditional RoPE, noaux_tc sigmoid gating, shared expert) and checks that incremental slab forwards equal a single multi-row forward.

Tests: `pytest -m "not slow" tests/` — 1682 passed, 15 skipped, 46 deselected.

Stacked on #553 (and #551 behind it); their commits show up in this diff until they merge.

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -7,6 +7,7 @@ import builtins
 import importlib.util
 import json
 import os
+import pickle
 import sys
 from types import ModuleType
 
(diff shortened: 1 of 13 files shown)


LxYuan0420:

PR #556, opened 2026-07-26
[Spec Decode] Run the GLM-4.7-Flash MTP head as a drafter on the paged path

samithaj:
## [Spec Decode] Run the GLM-4.7-Flash MTP head as a drafter on the paged path

This is the runtime half of the GLM-4.7-Flash MTP work: it takes the head model and loader and actually runs them as a drafter on the paged decode path, end to end through a real engine.

### What it adds

- `NativeMTPHeadRegistry`, keyed by draft `model_type`. A head owns its own validation and builds its own proposer from a frozen `NativeMTPBuildContext`, so `install_drafter` stays a lookup-and-delegate.
- `Glm4MoeLiteMTPHead`, registered for `glm4_moe_lite_mtp`. It rejects prefix caching, more than one speculative token, and pipeline parallelism, each with a message naming the offending value and the supported alternative.
- `NativeMTPProposer`, which runs the head as a tiny autoregressive 1-layer model over the target's slot stream, keeps a per-request KV slab, and hands that slab back through the drafter release seam when a request is preempted or evicted.
- Dispatch in `install_drafter` for `method == "mtp"`, below the existing Gemma4 MTP, draft-model, and n-gram branches. `Gemma4MTPAssistantSource.is_gemma4_mtp` also matches `method == "mtp"` and discriminates further on the assistant config, so the native branch has to stay under it. An unregistered draft `model_type` fails loud and names the registered heads.

### Tests

The engine-level test builds a tiny `glm4_moe_lite` target and a matching MTP head on disk from the real model classes and drives a real `LLM`, so **it needs no checkpoint download** and runs in CI in about 50 seconds. It asserts:

- drafts actually flow (`drafted > 0`) and the speculative run's output tokens are identical to the non-speculative run's;
- an accept-all fixture (zeroed lm_head, so every draft is trivially correct) makes `accepted` equal `drafted`;
- a 32-block cache override forces real preemptions while generation stays token-identical to the unpressured baseline;
- under that same pressure the proposer's release seam fires, so a preempted request hands its slab back instead of pinning it while it waits — output equality cannot show this, since rejected drafts leave generation unchanged either way, so the run is observed through the seam itself;
- accepted drafts survive preemption: `accepted` still equals `drafted` and every sequence still produces its full 160 tokens.

Unit coverage sits alongside it for the registry, the head's rejection paths, the proposer's slot-stream and slab lifecycle, and the `install_drafter` dispatch.

```
pytest -m "not slow" tests/                                          # 1713 passed, 15 skipped
VLLM_LOGGING_LEVEL=ERROR pytest -m slow tests/test_glm4_moe_lite_mtp_e2e.py   # 5 passed
```

Stacked on #555 and the ones behind it; their commits show up in this diff until they merge.

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -7,6 +7,7 @@ import builtins
 import importlib.util
 import json
 import os
+import pickle
 import sys
 from types import ModuleType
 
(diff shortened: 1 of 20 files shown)


LxYuan0420:

PR #557, opened 2026-07-26
[Spec Decode] Document GLM-4.7-Flash native MTP and add its benchmark helper

samithaj:
## [Spec Decode] Document GLM-4.7-Flash native MTP and add its benchmark helper

Stacked on `mtp/glm4-native-runtime`, which is the PR that adds the runtime path this documents. Final PR of the split of #535.

### `docs/speculative_decoding.md`

- Adds a Native MTP (GLM) column to the method comparison table, plus a prefix-caching row, since the four methods no longer agree on it.
- New "GLM-4.7-Flash MTP (native head)" section: what the head is, where to get the hosted revision-pinned checkpoint, how to serve it, and how to confirm from the server log that it is actually active.
- The constraints the runtime enforces at startup, each with the reason: greedy-only with `num_speculative_tokens=1`, prefix caching must be off, an actual MTP head checkpoint is required (pointing `model` at the raw target repo fails loud), and no pipeline parallelism.
- A tie-aware parity note for this MLA target: greedy token identity is not bit-exact across batch shapes even with speculative decoding off, so a divergence is only evidence of a verify bug once it is judged against a no-SD calibration at the same batch shapes.

No performance numbers in the docs. The benchmark writes JSON that attaches to performance PRs instead, which is the same convention the Gemma4 section already uses.

### `tools/benchmark/native_mtp_benchmark.py`

A thin sibling of `gemma4_mtp_benchmark.py`. It reuses that tool's prompt selection and measurement helpers and adds only what the native-MTP contract forces:

- `enable_prefix_caching=False` is always passed, because cached prompt tokens skip the target forward that produces the hidden states the head consumes.
- `--chat` renders each prompt through the tokenizer chat template; raw continuation prompts drive thinking models into degenerate loops that distort acceptance.
- The engine's `spec_decode` draft/accepted counters land in the output JSON, snapshotted after warmup and reported as the measured-runs delta so warmup drafts do not pollute the acceptance rate.
- An MTP run whose engine reports no draft/accepted counters raises instead of emitting a result that is missing the acceptance evidence it promises.

The tool imports nothing from `vllm_metal` and defers `from vllm import LLM` into the function body, so it stands alone. `tests/test_native_mtp_benchmark.py` covers the parser extension and the acceptance-rate and fail-loud behavior of the counter reader without loading an engine.

Benchmarks (M3 Max, GLM-4.7-Flash-4bit target + the 4-bit hosted head, both revision-pinned and recorded in the JSONs; greedy, bs1, 8 distinct prompts run serialized x 256 tokens, interleaved A/B processes with cooldowns on an idle machine): SD on vs off raises end-to-end output-token throughput by ~11% (66.3 vs 59.4 tok/s; 1.107-1.117x across three independent quiet-machine pairs) at 0.770 acceptance measured over the timed runs only (889/1155 post-warmup draft tokens, identical across runs under greedy).

Stacked on #556 and the ones behind it; their commits show up in this diff until they merge.

### DIFF
--- a/docs/speculative_decoding.md
+++ b/docs/speculative_decoding.md
@@ -1,23 +1,25 @@
 # Speculative Decoding
 
-vllm-metal supports three speculative decoding methods on the paged attention
-path. All require synchronous scheduling and greedy sampling.
+vllm-metal supports the following speculative decoding methods on the paged
+attention path. All require synchronous scheduling and greedy sampling.
 
-| | MTP | Draft Model | N-gram |
-|---|---|---|---|
-| `--speculative-config` method | `mtp` | `draft_model` | `ngram` |
-| Target models | Gemma4 (paged) | Any paged-attention model | Any paged-attention model |
-| Draft source | MTP assistant checkpoint (reads target KV cache) | Separate smaller model (own KV cache) | Prompt/output token history (no model) |
-| `num_speculative_tokens` | 1 | Configurable (3–5 typical) | Configurable (3–5 typical) |
-| Extra memory | None (reads target KV cache) | Second un-budgeted KV cache | None |
+| | Gemma4 MTP | Native MTP (GLM) | Draft Model | N-gram |
+|---|---|---|---|---|
+| `--speculative-config` method | `mtp` | `mtp` | `draft_model` | `ngram` |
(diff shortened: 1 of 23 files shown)


LxYuan0420:

PR #558, opened 2026-07-26
Trim TurboQuant tests

LxYuan0420:
This PR is:

- To remove the embedded TurboQuant manual benchmark and serve harness from the pytest file.
- To keep the real TurboQuant unit and kernel-contract tests.
- To make the remaining test data deterministic and easier to review.
- To avoid keeping benchmark-style code inside the normal test suite.

Note:

The deleted harness was not production runtime code and did not belong in `tests/test_turboquant.py`. If we want live benchmark coverage later, it should live under `tools/`, not inside the pytest suite.


WindChimeRan (approved):
Verified all low value test deletion. 

Thanks for the cleanup!

PR #560, opened 2026-07-26
[Spec Decode] Ride the verify window kernel for small draft ingests

abcgco:
Follow-up to #502, as offered [in its thread](https://github.com/vllm-project/vllm-metal/pull/502#issuecomment-5073309780), building on #534.

#502 already submits merged decode specs for the ≤16-token draft ingest, but the context expands them to per-token segments, so every ingest row still walks the KV context independently. With the operator's `VLLM_METAL_SPEC_VERIFY_WINDOW` opt-in and a draft the window kernel serves (head size within the decode-kernel bound; this proposer is MHA-only, so the runner gate's MLA/GDN arms cannot arise), each plan's ingest now stays ONE merged segment — the same #534 window layout the target's verify batches use, sharing each K/V block load across the K+1 rows. Without the opt-in the expanded per-token metadata is bit-for-bit unchanged.

Evidence: generated tokens are identical either way in every configuration tried (greedy, 0.6B draft==target, 8k prefix — 14 hash-instrumented engine runs across both layouts and both proposer versions, one ids-hash); single-stream TPOT is within run-to-run noise of the expanded ingest (interleaved A/B medians 53.1 vs 53.4 ms/token on the 16-core M4 Pro; runs degraded by thermal throttling or background GPU load were re-run after cooldowns or neutralized by the medians — every degraded run still produced the same ids-hash). The concurrency win carries over from #534's op-level grid — ingest rows are the same kernel shape as verify windows — and I couldn't measure concurrent e2e on this 24 GB box, so I'm not claiming a number for it. The window-parity suite's large-window case now covers 16 (`_DECODE_INGEST_MAX_TOKENS`, the widest structurally pinned window), and a new zero-context case pins the one shape this ingest produces that verify never does — a merged window over an empty prefix (first ingest of a short prompt; recurrent after #551's release-and-resume). Docs updated in both places the env is described.


LxYuan0420 (approved):
LGTM

PR #561, opened 2026-07-26
[Perf] Build kernel format paged metadata once per forward, not per layer

abcgco:
`sdpa_forward` reserializes the context's Python metadata lists into mx arrays on every layer: `slot_mapping`, `context_lens`, `cu_seqlens`, and the padded block tables. All of them are invariant within one forward pass, so every layer after the first repeats identical work. The block table conversion is O(rows x blocks) pure Python, and the waste grows with batch size, speculation width, and context length.

Isolated numbers on a 16 core M4 Pro (median of 60 reps, calling the real `_build_block_tables` from main): 16 us per call at rows=4, blocks=512, which is a single stream K=3 verify batch at 8k context. The cost grows linearly to about 500 us per call at rows=128 (concurrency 32 with K=3 expanded verify at 8k) and about 2.1 ms at rows=128, blocks=2048 (the same at 32k). Multiplied by 28 to 36 layers that is roughly 0.6 ms per step single stream and 18 ms per step at the concurrency 32 shape. That is the same envelope where the verify window kernel (#534, #560) earns its win, so the two compose.

This PR builds the converted arrays once per forward pass and KV group and reuses them across layers: a small memo dict on `PagedAttentionContext` keyed by (group index, cache block size), populated by the first layer of a group. The context lives for exactly one forward pass and nothing mutates its metadata lists after `prepare_grouped` installs it (checked across the callers, including YOCO, which swaps in its own reduced context), so entries cannot go stale and no invalidation logic is needed.

Verification on the engine (0.6B draft pair, CONC=4, K=3, vLLM 0.26.0, interleaved A/B, 3 runs per side): generated token ids are bitwise identical in all six runs, so the change is semantically a no op. Throughput at this small shape is parity within run to run noise (medians 69.1 vs 68.7 tok/s). I am deliberately not claiming an engine speedup from that data: the isolated scaling above is the sizing argument, and the shapes where it reaches double digit milliseconds per step need more memory than my 24 GB box can host.

The same per layer pattern exists in the MLA wrapper's per request slow path (`impls/mla.py`); left untouched here since that path is the documented fallback. Happy to follow up if there is interest.

Tests: `TestKernelMetadataMemo` pins reuse within one context, bitwise agreement with direct conversion, fresh context isolation, key independence by block size, and key independence by group at equal block size (guards hybrid models against a key regression). Full attention suites pass on vLLM 0.26.0; the one red test on my box fails identically on clean main, so it is unrelated.

### DIFF
--- a/tests/test_block_size_translation.py
+++ b/tests/test_block_size_translation.py
@@ -10,9 +10,16 @@ from __future__ import annotations
 
 import pytest
 
+from vllm_metal.attention.context import (
+    clear_context,
+    get_context,
+    prepare_grouped,
+    prepare_unified,
+)
 from vllm_metal.attention.impls.sdpa import (
     _KERNEL_BLOCK_SIZES,
     _build_block_tables,
+    _kernel_metadata,
     _pick_kernel_block_size,
 )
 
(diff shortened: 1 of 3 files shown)


LxYuan0420 (approved):
LGTM

PR #563, opened 2026-07-27
Fix Gemma4 mixed KV cache planning

LxYuan0420:
This PR is:

- To emit Gemma4 full/sliding MHA KV specs instead of charging every layer as full attention.
- To let vLLM build the authoritative grouped `KVCacheConfig` before Metal allocates cache tensors.
- To initialize Metal paged attention from the returned grouped layout.
- To keep unsupported modes on the existing dense path.

This fixes issue #505 overcharge where Gemma4 31B's 50 sliding layers and 10 full-attention layers were all planned as full attention. Metal now trusts vLLM's grouped cache contract and only translates the returned layout into Metal cache tensors.

Note:

@ricky-chaoju could you test this on the real Gemma4 hardware path after this PR is up? The local tests cover the config/startup path, but the final proof should be a real 26B/31B long-context serve.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 from types import SimpleNamespace
 from typing import Any
 
+import mlx.core as mx
 import torch
 
 import vllm_metal.v1.model_runner as mr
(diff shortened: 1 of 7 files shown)


ricky-chaoju (approved):
LGTM. Ran it on the M1 Ultra with no --max-model-len, so the 262144 case from #505. Boots and serves it, with 331,003 tokens of KV cache. A fact planted at the start of the prompt comes back correctly at 50, 8,792, and 26,502 tokens. Capped at 32768 for a like-for-like read, 31B goes 35,552 -> 143,327 tokens and 26B 171,536 -> 664,225, same KV budget.

PR #564, opened 2026-07-27
Publish stable wheels to PyPI via Trusted Publishing

WindChimeRan:
Last step toward `pip install vllm-metal` (#436 Step 3). Adds a PyPI publish path for the plugin; the wheel itself already needs no compiler since #423 and #508.

**Remaining blockers, all upstream in vLLM:**

- vLLM publishes no macOS wheel to PyPI — only manylinux wheels and an sdist.
- That sdist doesn't even resolve on macOS: it pins `pynvvideocodec`, which is NVIDIA-only.
- The macOS vLLM wheel exists only as a GitHub release asset, and PyPI forbids direct-URL dependencies, so `install.sh`'s pin can't move into package metadata.
- So vllm-metal still can't declare a hard `vllm` dependency; users install vLLM core separately until upstream publishes a macOS wheel to PyPI.
- vLLM's macOS wheel is cp312-only, so we can't offer newer Python either even though mlx already ships cp310–cp314.

**Handled here:** our PyPI project is stale at `0.1.0` and nothing publishes to it. `requires-python` also claimed 3.13 while only a cp312 wheel exists, which turned "requires Python 3.12" into a confusing "no matching distribution" for anyone on a default 3.13/3.14 interpreter — now tightened to match the artifacts. Builds are wheel-only, so pip can never fall back to a source build that would need the Metal toolchain.

**Proposed release mechanism**

Two channels, both building the same wheel via a shared `build_wheel` in `scripts/lib.sh`. `release.yml` keeps pushing per-commit `0.3.0.devTIMESTAMP` wheels to GitHub Releases on every merge to main. The new `publish-pypi.yml` is manual: it builds the release-versioned wheel on macOS, then uploads it with PyPI Trusted Publishing (OIDC, no stored token), targeting TestPyPI or PyPI as chosen at dispatch so a release can be rehearsed first. Manual because a PyPI version can only be uploaded once, so publishing on every merge would burn version numbers. Setup steps are in `docs/releasing.md`.

@mgoin this needs a one-time Trusted Publisher registration on the PyPI `vllm-metal` project (fields in `docs/releasing.md`) before it can be rehearsed on TestPyPI.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,0 +1,66 @@
+name: Publish to PyPI
+
+# Manual, unlike release.yml's per-push dev wheels: a PyPI version can only be
+# uploaded once, so publishing on every merge would burn version numbers.
+# Publishes whatever version pyproject.toml holds — bump it first.
+#
+# Needs a one-time Trusted Publisher registration; see docs/releasing.md.
+
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: "Index to publish to"
+        type: choice
+        options:
+          - testpypi
+          - pypi
+        default: testpypi
+
+jobs:
+  build:
+    name: Build macOS arm64 wheel
+    # macOS runner: the wheel bundles compiled .metallib shaders.
+    runs-on: macos-15
+    permissions:
+      contents: read
+    steps:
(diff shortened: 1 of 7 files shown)


WindChimeRan:
close as not planned

PR #566, opened 2026-07-28
Simplify Gemma4 MTP metadata handling

LxYuan0420:
This PR is:

- To use loaded MLX model args as the source of truth for Gemma4 MTP target metadata.
- To keep strict validation on the assistant checkpoint config.
- To remove duplicate HF-vs-MLX target metadata checks and mismatch-only tests.
- To simplify Gemma4 MTP runtime state and parsing code.

The target model is already loaded before MTP is wired, so we should trust its loaded args instead of reconciling them again with HF config. The assistant config is still external JSON, so it keeps the strict validation boundary.

### DIFF
--- a/tests/test_gemma4_mtp_assistant.py
+++ b/tests/test_gemma4_mtp_assistant.py
@@ -327,11 +327,13 @@ def test_validate_accepts_assistant_layer_types_tail_matching_target_non_shared_
 def test_validate_rejects_layer_types_not_tail_matching_target_non_shared_layers() -> (
     None
 ):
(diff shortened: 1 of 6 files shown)


ricky-chaoju (approved):
LGTM. 
One thing. `docs/speculative_decoding.md` still says unsupported speculative decode "fails early instead of falling back silently", and with `GEMMA4_TARGET_MODEL_TYPES` gone that is no longer true. A llama-shaped target passes `from_model_args` and only fails after the assistant download at the vocab check; a qwen3-shaped one raises a bare `KeyError: 'layer_types'`.

PR #568, opened 2026-07-30
Restore prefix cache e2e spy target

ricky-chaoju:
This test has verified nothing since #534. It spies on the function that receives per-prefill `start_pos`, but the spy hardcoded a three-argument signature. #534 added `merge_verify_windows` and it started raising `TypeError`. #545 renamed `prepare_unified` to `prepare_grouped` and it went quiet instead.

So the cache-hit assertion fails on an empty list and blames prefix caching, and the determinism check behind it never runs. That check is the only thing covering "a cache hit gives the same tokens as the priming pass". The unit tests mock the logits and the sampler, so they cannot cover it.

The spy only needs the first two arguments, so bind those and pass the rest through. Also a separate assertion for the spy never firing, so next time the message says the harness broke.

### DIFF
--- a/tests/test_paged_prefix_caching_e2e.py
+++ b/tests/test_paged_prefix_caching_e2e.py
@@ -7,9 +7,12 @@ first ``generate`` populates the cache; the second triggers cache hits
 that walk the model_runner's ``start_pos > 0`` path because the upstream
 scheduler reports ``num_computed_tokens > 0``.
 
-Two assertions (in code order):
+Three assertions (in code order):
+  0. Spy reach — the ``prepare_grouped`` patch actually fired.  Without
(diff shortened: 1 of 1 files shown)


LxYuan0420 (approved):
Thanks. Could you please run and share the slow test result, just to confirm the repaired test works correctly?

ricky-chaoju:
Hi @LxYuan0420, sure. Ran it on an M1 Ultra 128GB with vLLM 0.26.0 and VLLM_METAL_USE_PAGED_ATTENTION=1.

On main:
截圖 2026-07-31 晚上8 53 23



On this branch:

截圖 2026-07-31 晚上8 55 15

Issue #569, opened 2026-07-31
[Bug]: llm segfaults on macOS arm64 with xgrammar==0.2.3 and apache-tvm-ffi==0.1.13


jhou1:
# Environment
```
.venv-vllm-metal ❯ python3 -c "import platform; print(platform.machine())"
arm64
.venv-vllm-metal ❯ file "$(which python3)"
/Users/hjw/.venv-vllm-metal/bin/python3: Mach-O 64-bit executable arm64
.venv-vllm-metal ❯ uname -r
25.5.0
```

# Steps
1. Clean installation: `rm -rf ~/.venv-vllm-metal && curl -fsSL https://raw.githubusercontent.com/vllm-project/vllm-metal/main/install.sh | bash`
2. Run `vllm`: 
```
.venv-vllm-metal ❯ vllm
INFO 07-31 13:49:40 [__init__.py:52] Available plugins for group vllm.platform_plugins:
INFO 07-31 13:49:40 [__init__.py:54] - metal -> vllm_metal:register
INFO 07-31 13:49:40 [__init__.py:57] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 07-31 13:50:01 [__init__.py:237] Platform plugin metal is activated
INFO 07-31 13:50:06 [importing.py:88] Triton not installed or not compatible; certain GPU-related functions will not be available.
objc[5923]: Class AVFFrameReceiver is implemented in both /Users/hjw/.venv-vllm-metal/lib/python3.12/site-packages/cv2/.dylibs/libavdevice.61.3.100.dylib (0x12ca303a8) and /opt/homebrew/Cellar/ffmpeg/8.1.1/lib/libavdevice.62.3.101.dylib (0x12ee50328). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
objc[5923]: Class AVFAudioReceiver is implemented in both /Users/hjw/.venv-vllm-metal/lib/python3.12/site-packages/cv2/.dylibs/libavdevice.61.3.100.dylib (0x12ca303f8) and /opt/homebrew/Cellar/ffmpeg/8.1.1/lib/libavdevice.62.3.101.dylib (0x12ee50378). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
!!!!!!! Segfault encountered !!!!!!!
  File "<unknown>", line 0, in TVMFFIEnvRegisterCAPI
[...]

PR #570, opened 2026-07-31
Cap apache-tvm-ffi below the xgrammar ABI break

ricky-chaoju:
A clean install segfaults before vLLM starts. `apache-tvm-ffi` published 0.1.13, xgrammar's prebuilt wheels are compiled against an older tvm-ffi C++ ABI, and xgrammar declares only `apache-tvm-ffi>=0.1.9`, so the resolver takes 0.1.13 and `import xgrammar` dies in `xgrammar::__TVMFFIStaticInitFunc0()`. Reported in #569.

This does not come from our dependency chain. Installing just the vLLM wheel, which is what `install.sh` does first, already lands 0.1.13, and `import xgrammar` segfaults right there, before vllm-metal is installed at all. `VLLM_PLUGINS=` does not help either, so nothing on the Metal path is involved. The cap bounds the resolver and nothing else, the same way we already bound llguidance, fastapi, and starlette without importing any of them.

Evidence from a fresh venv on an M1 Ultra 128GB. After the vLLM wheel alone, `import xgrammar` exits 139. Installing our dependency set on top downgrades tvm-ffi 0.1.13 to 0.1.12, and both `import xgrammar` and `vllm --help` then exit 0. Putting 0.1.13 back in that same venv brings the segfault straight back.

The version that matters is tvm-ffi, not xgrammar. 0.2.3 and 0.2.4 both crash on 0.1.13 and both work on 0.1.12, so the xgrammar 0.2.4 upgrade suggested in the issue thread does not fix it on its own, and 0.2.5 has no cp312 macOS arm64 wheel on PyPI. Scoped to macOS arm64 because that is where I verified it. Removing the cap is gated on xgrammar bounding its own range.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,13 @@ dependencies = [
     "mlx==0.32.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
     "mlx-lm>=0.31.3; platform_system == 'Darwin' and platform_machine == 'arm64'",
     "llguidance>=1.7.0,<1.8.0; platform_machine == 'x86_64' or platform_machine == 'arm64' or platform_machine == 'aarch64' or platform_machine == 'ppc64le'",
+    # Temporary cap: xgrammar's prebuilt wheels are compiled against a specific
+    # apache-tvm-ffi C++ ABI but declare only `apache-tvm-ffi>=0.1.9`, so a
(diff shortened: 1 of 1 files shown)


jhou1:
Thank you, this fixes the problem.

LxYuan0420 (approved):
Thanks for the quick fix!

PR #572, opened 2026-08-01
Remove dead MLX backend switch

LxYuan0420:
This PR is:

- To remove the unused `VLLM_METAL_USE_MLX` environment switch.
- To make MLX device initialization unconditional in the Metal worker and platform path.
- To keep `VLLM_MLX_DEVICE` as the only MLX device selector.

The old switch did not select another backend. Setting it to `0` only skipped required MLX initialization, while the runtime still depended on MLX. This removes the dead config surface and the related test scaffolding.


ricky-chaoju (approved):
LGTM

PR #573, opened 2026-08-01
[Bugfix] Route MLX-quantized VL wrappers to the text backbone

### LINKED ISSUE #571: [Bug] Qwen3.6-35B-A3B MLX-quantized MoE generates incoherent output (mlx_lm on the same checkpoint is correct)
## Summary

`Qwen3.6-35B-A3B` in MLX 4-bit loads without error under vllm-metal and then generates incoherent token soup. The same checkpoint generates correct text under `mlx_lm` directly, and the documented example checkpoint for this family (`Qwen/Qwen3.5-0.8B`) is correct under vllm-metal — so this is specific to the quantized MoE path, not to the install or the checkpoint.

`supported_models/` lists **Qwen3.5 / 3.6 → ✅** with example checkpoint `Qwen/Qwen3.5-0.8B`, which is 0.8B and dense; the "3.6 adds MoE" note appears only in the attention-kernel column.

## Environment

| | |
|---|---|
| Hardware | Apple M4 Max, 64 GB unified memory |
| macOS | 26.5.2 |
| vllm-metal | `0.3.0.dev20260730125022` |
| vLLM | `0.26.0+cpu` |
| mlx / mlx-lm | `0.32.0` / `0.31.3` |

## Reproduction

```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 vllm serve \
  lmstudio-community/Qwen3.6-35B-A3B-MLX-4bit \
  --served-model-name Qwen3.6-35B-A3B \
  --host 0.0.0.0 --port 8000 \
  --max-model-len 32768
```

```bash
curl -s localhost:8000/v1/chat/completions -H 'content-type: application/json' -d '{
  "model":"Qwen3.6-35B-A3B","max_tokens":150,"temperature":0,
  "messages":[{"role":"user","content":"用一句话说明什么是境外投资备案。"}]}'
```

Startup is clean — no unmapped parameters, no warnings beyond the expected hybrid-model notice:

```
Hybrid model (e.g., Qwen3.5) with paged attention enabled. Using block-size
translation (PR #235) to convert vLLM blocks into kernel blocks.
Paged attention memory breakdown: metal_limit=64.42GB, fraction=0.92,
usable_metal=59.27GB, model_memory=20.40GB
```

**Actual output** (greedy, `temperature=0`):

```
.gaz ZvMEG understuggyza 领 CONSTRت1相信我cipe?, ". V Urla 馆 dropED  `例子!,ament. 4
aftert5.:ت った_o辞y部分or..: at s面外的:   pec?1，_的iveeta case  Iig 1. utt ? ت
Subsryp, page تmenand.2I对新 asking-为建设年99iel颜色 O?-对这个11 到 9" Year11 pec/
```

## Controls

**1. Same checkpoint under `mlx_lm` — correct.**

```python
from mlx_lm import load, generate
model, tok = load("<snapshot path of lmstudio-community/Qwen3.6-35B-A3B-MLX-4bit>")
prompt = tok.apply_chat_template(
    [{"role": "user", "content": "用一句话说明什么是境外投资备案。"}],
    add_generation_prompt=True)
print(generate(model, tok, prompt=prompt, max_tokens=120))
```

```
Here's a thinking process:

1.  **Analyze User Input:**
   - **Topic:** 境外投资备案 (Outbound Investment Filing/Registration)
   - **Requirement:** 用一句话说明 (Explain in one sentence)
   - **Language:** Chinese
```

**2. The documented example checkpoint under vllm-metal — correct.**

`vllm serve Qwen/Qwen3.5-0.8B --max-model-len 8192 --reasoning-parser qwen3`, same prompt:

> 境外投资备案是指企业或个人在境外设立机构、场所或进行投资活动时，向所在国或地区主管机关提交的文件和资料，以确认其投资行为的合法性、合规性及所需信息的完整性。

## The output does not depend on any runtime flag

Three configurations produce **byte-identical** output, which suggests the corruption happens before the attention path is reached rather than in the kernel:

| configuration | result |
|---|---|
| default (paged attention, lazy GDN) | garbage `A` |
| `--language-model-only` | byte-identical to `A` |
| `VLLM_METAL_GDN_LAZY_DECODE=0` (eager GDN) | byte-identical to `A` |
| `VLLM_METAL_USE_PAGED_ATTENTION=0`, `MEMORY_FRACTION=auto` | will not start — `ValueError: To serve at least one request with the model's max seq len (4096), (0.14 GiB KV cache is needed, which is larger than the available…)`, so the non-paged path could not be used as a comparison |

## Checkpoint layout

`lmstudio-community/Qwen3.6-35B-A3B-MLX-4bit` stores experts pre-stacked in mlx-lm's own layout rather than per-expert, so `_stack_qwen36_moe_per_expert_weights` is a no-op here:

```
language_model.model.layers.0.mlp.switch_mlp.gate_proj.{weight,scales,biases}
language_model.model.layers.0.mlp.switch_mlp.up_proj.{weight,scales,biases}
language_model.model.layers.0.mlp.switch_mlp.down_proj.{weight,scales,biases}
language_model.model.layers.0.mlp.shared_expert.{gate,up,down}_proj.{weight,scales,biases}
language_model.model.layers.0.mlp.gate.{weight,scales,biases}
language_model.model.layers.0.mlp.shared_expert_gate.{weight,scales,biases}
```

Quantization is `{group_size: 64, bits: 4, mode: affine}` with 80 per-layer overrides, all of them 8-bit on the routing gates (`mlp.gate`, `mlp.shared_expert_gate`).

## Related data point: the AWQ MoE build fails to load outright

`QuantTrio/Qwen3.6-35B-A3B-AWQ` (`awq / bits=4 / group_size=128 / zero_point=true`, i.e. within the documented AWQ constraints) fails at load:

```
ValueError: Received 89856 parameters not in model:
  language_model.model.layers.9.mlp.experts.99.gate_proj.qweight,
  language_model.model.layers.9.mlp.experts.99.gate_proj.qzeros,
  language_model.model.layers.9.mlp.experts.99.gate_proj.scales, …
```

Possibly unrelated, but it is the same architecture and the same "expert weights are not where the loader expects them" theme, so noting it here rather than opening a second issue.

## Other checkpoints tried

| checkpoint | result |
|---|---|
| `lmstudio-community/Qwen3.6-35B-A3B-MLX-4bit` | loads, incoherent output |
| `Jundot/Qwen3.6-35B-A3B-oQ4e-mtp` (mixed-precision MLX affine) | loads, incoherent output |
| `QuantTrio/Qwen3.6-35B-A3B-AWQ` | load fails, 89856 unmapped parameters |
| `cyankiwi/Qwen3.6-35B-A3B-AWQ-4bit` | rejected at config validation, `group_size=32` (documented constraint is 128) |
| `Qwen/Qwen3.6-35B-A3B-FP8` | not testable here — `_dequantize_qwen35_fp8_weights` dequantizes to bf16, so 35B needs ~70 GB and does not fit in 64 GB |

The FP8 path is the one `compat.py` explicitly patches for (`_patch_mlx_lm_qwen35_fp8_sanitize`, whose docstring names `Qwen/Qwen3.6-35B-A3B-FP8`), so it may well be the only variant of this model that has been exercised. If that is the case, it would be worth narrowing the ✅ in `supported_models/` for Qwen3.6 MoE to the checkpoint formats that are actually covered — the current entry reads as though any Qwen3.6 checkpoint works, and a silent wrong-output failure is much harder to diagnose than a load error.

Happy to run further diagnostics on this machine if that would help.

ricky-chaoju:
Fixes #571. MLX-quantized Qwen3.5/3.6 MoE checkpoints load fine but generate garbage, while the same checkpoint works under `mlx_lm`. The config is marked multimodal, so the load goes through `mlx_vlm` and the runner ends up driving the language model with mrope state that never gets set. The fp8 checkpoints already get force-routed to the `mlx_lm` text loader; the MLX affine ones slipped past that gate.

This PR makes the same override match MLX quantization dicts, but only for architectures that have no native multimodal adapter, so MLX-quantized Qwen3.5 VL and Qwen3-VL keep image serving.

Verified with #571's repro on `lmstudio-community/Qwen3.6-35B-A3B-MLX-4bit`: main produces token soup, with this change the greedy output matches `mlx_lm` on the same snapshot for all 120 generated tokens. `mlx-community/Qwen3.6-35B-A3B-4bit` behaves the same. Added ten tests for the predicate truth table. The AWQ MoE load failure from #571 is a separate problem.

### DIFF
--- a/tests/test_model_adapter.py
+++ b/tests/test_model_adapter.py
@@ -54,6 +54,98 @@ class TestShouldForceTextBackbone:
         result = adapter.should_force_text_backbone(hf_config)
         assert result is True
 
+    @pytest.mark.parametrize("bits", [4, 8])
+    @pytest.mark.parametrize(
+        ("model_type", "architecture"),
+        [
+            # The real pair of the motivating checkpoints
+            # (Qwen3.6-35B-A3B MLX 4-bit, issue #571).
+            ("qwen3_5_moe", "Qwen3_5MoeForConditionalGeneration"),
+            ("qwen3_6", "Qwen3_6ForConditionalGeneration"),
+        ],
+    )
+    def test_mlx_quant_conditional_generation_uses_auto_override(
+        self, model_type: str, architecture: str, bits: int
+    ) -> None:
+        hf_config = SimpleNamespace(
+            model_type=model_type,
+            architectures=[architecture],
+            quantization={"group_size": 64, "bits": bits, "mode": "affine"},
+        )
+        adapter = DefaultModelAdapter()
+        result = adapter.should_force_text_backbone(hf_config)
+        assert result is True
+
+    @pytest.mark.parametrize(
+        ("model_type", "architecture"),
+        [
+            # The flagship native-multimodal checkpoint family
+            # (mlx-community Qwen3-VL MLX conversions carry the same
+            # top-level quantization dict) must keep the native path.
+            ("qwen3_vl", "Qwen3VLForConditionalGeneration"),
+            # A natively-adapted Qwen3.5 VL wrapper keeps image serving.
+            ("qwen3_5", "Qwen3_5ForConditionalGeneration"),
+            # An arbitrary non-allowlisted multimodal model.
+            ("llava", "LlavaForConditionalGeneration"),
+        ],
+    )
+    def test_mlx_quant_natively_adapted_arch_skips_auto_override(
+        self, model_type: str, architecture: str
+    ) -> None:
+        hf_config = SimpleNamespace(
+            model_type=model_type,
+            architectures=[architecture],
+            quantization={"group_size": 64, "bits": 4, "mode": "affine"},
+        )
+        adapter = DefaultModelAdapter()
+        result = adapter.should_force_text_backbone(hf_config)
+        assert result is False
+
+    def test_multimodal_native_mode_disables_mlx_quant_override(
+        self, monkeypatch
+    ) -> None:
+        monkeypatch.setenv("VLLM_METAL_MULTIMODAL_MODE", "multimodal-native")
+        reset_config()
+
+        hf_config = SimpleNamespace(
+            model_type="qwen3_6",
+            architectures=["Qwen3_6MoeForConditionalGeneration"],
+            quantization={"group_size": 64, "bits": 4, "mode": "affine"},
+        )
+        adapter = DefaultModelAdapter()
+        result = adapter.should_force_text_backbone(hf_config)
+        assert result is False
+
+    def test_omitted_quantization_key_skips_auto_override(self) -> None:
+        # The common dense-checkpoint case: no `quantization` key at all.
+        hf_config = SimpleNamespace(
+            model_type="qwen3_6",
+            architectures=["Qwen3_6ForConditionalGeneration"],
+        )
+        adapter = DefaultModelAdapter()
+        result = adapter.should_force_text_backbone(hf_config)
+        assert result is False
+
+    def test_non_dict_quantization_value_skips_auto_override(self) -> None:
+        hf_config = SimpleNamespace(
+            model_type="qwen3_6",
+            architectures=["Qwen3_6ForConditionalGeneration"],
+            quantization="int4",
+        )
+        adapter = DefaultModelAdapter()
+        result = adapter.should_force_text_backbone(hf_config)
+        assert result is False
+
+    def test_quantization_dict_without_bits_skips_auto_override(self) -> None:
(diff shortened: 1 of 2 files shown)


LxYuan0420 (approved):
Thanks for the fix. The old FP8-only rule missed MLX affine quantization, so multimodal-looking text checkpoints could be routed through the wrong VL path.

As a follow-up, we should make loader selection more robust with validated format/capability detection instead of extending architecture heuristics, so future checkpoint variants cannot silently enter an incompatible VLM path.

PR #574, opened 2026-08-02
Remove custom Metal debug config

LxYuan0420:
This PR is:

- To remove the repo-specific `VLLM_METAL_DEBUG` setting.
- To use normal debug logging for Metal config and model-args dumps.
- To reduce config, env, docs, and test plumbing that did not affect runtime behavior.

Note: `VLLM_METAL_DEBUG` no longer works. Use `VLLM_LOGGING_LEVEL=DEBUG` to show the Metal config and model-args debug logs.


ricky-chaoju (approved):
LGTM. Thanks!

PR #575, opened 2026-08-02
[Bugfix] Count padded mamba pages in the non-paged admission estimate

ricky-chaoju:
On the MLX (non-paged) path, hybrid GDN models fail admission at any `--max-model-len`: startup dies with "X GiB KV cache is needed, which is larger than the available X GiB". The #571 reporter hit this too when trying the non-paged path as a control. Same estimate-vs-requirement principle as #505: the worker reports one max-length sequence using the raw linear-state bytes, but vLLM checks against the `MambaSpec` we report, which carries `mamba_page_size_padded`, so the budget always lands one padding margin short.

The fix makes the estimate count what the spec charges: when `cache_config.mamba_page_size_padded` is set, linear layers are priced at the padded page; otherwise, the raw bytes as before. Attention layers were already aligned.

Before, on main, `VLLM_METAL_USE_PAGED_ATTENTION=0 VLLM_METAL_MEMORY_FRACTION=auto vllm serve mlx-community/Qwen3.6-35B-A3B-4bit --max-model-len 8192` fails with "0.22 GiB is needed, which is larger than the available 0.22 GiB". With this change, the same command boots and generates correctly. Two tests pin the padded and non-hybrid cases.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -34,6 +34,7 @@ def make_stub_runner(
 
     defaults: dict[str, Any] = {
         "vllm_config": SimpleNamespace(speculative_config=None),
+        "cache_config": SimpleNamespace(mamba_page_size_padded=None),
         "model_config": SimpleNamespace(
             runner_type="generate", get_head_size=lambda: 128
         ),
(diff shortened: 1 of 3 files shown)


LxYuan0420 (approved):
Thanks for fixing it

PR #576, opened 2026-08-03
Share LoRA slot logic between LoRA and QLoRA

LxYuan0420:
This PR is:

- To move shared LoRA/QLoRA slot setup, weight install, mapping, and reset logic into one base wrapper.
- To keep plain LoRA and QLoRA wrappers focused on their real differences.
- To remove duplicated QLoRA tests that now repeat the shared LoRA slot contract.

Validation:

Real offline LoRA smoke passed with base `Qwen/Qwen2.5-0.5B-Instruct` and LoRA `taronklm/Qwen2.5-0.5B-Instruct-lora-chatbot`.

The smoke loaded the real model, wrapped 168 LoRA modules, activated LoRA 1, and completed base-vs-LoRA generation. The outputs were close but not identical, which confirms the adapter path was active without breaking base generation.

### DIFF
--- a/tests/test_qlora.py
+++ b/tests/test_qlora.py
@@ -170,13 +170,6 @@ def test_can_wrap_qlora_rejects_non_unary_quantized_projection() -> None:
 # MLXQuantizedLinearWithLoRA wrapper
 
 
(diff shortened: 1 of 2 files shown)


ricky-chaoju (approved):
LGTM. Thanks!

PR #577, opened 2026-08-03
Remove unused logits processor wiring

LxYuan0420:
This PR is:

- To reject logits processor options that Metal does not support.
- To remove logits processor state that was passed around but not actually used correctly.
- To reject custom logits processor plugins at startup instead of ignoring them.
- To keep the empty logits processor object that vLLM still requires for sampling metadata.

Metal does not currently support `min_p`, `logit_bias`, `min_tokens`, or custom logits processors. Before this PR, some of that plumbing existed, but the behavior was incomplete. This could make unsupported options look like they worked.

After this PR, supported sampling still works, and unsupported logits processor options fail clearly.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -67,7 +67,6 @@ def make_stub_runner(
         "use_async_scheduling": True,
         "device": torch.device("cpu"),
         "_sampler": None,
-        "_logitsprocs": None,
         "_structured_output_applier": MetalStructuredOutputApplier(),
         "_lora": MetalLoRARuntime(),
         "_yoco_cache_mapping": None,
--- a/tests/test_draft_model_proposer.py
+++ b/tests/test_draft_model_proposer.py
@@ -74,7 +74,6 @@ def _context(
         cu_seqlens=[],
         num_decode_segments=1,
         num_speculative_tokens=1,
-        logitsprocs=None,
         finished_req_ids=set(),
     )
 
--- a/tests/test_ngram_spec_decode.py
+++ b/tests/test_ngram_spec_decode.py
@@ -86,7 +86,6 @@ def _context(
         cu_seqlens=[],
         num_decode_segments=len(decode_reqs),
         num_speculative_tokens=num_speculative_tokens,
-        logitsprocs=None,
         finished_req_ids=finished_req_ids or set(),
     )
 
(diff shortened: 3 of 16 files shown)


ricky-chaoju (commented):
Premise is right. I checked the three builtins with the same construction `build_logitsprocs` uses: without `update_state` they all leave logits untouched, and we never call `update_state` anywhere. So min_p, logit_bias and min_tokens were doing nothing.

The per-request check is in the wrong place though. It sits inside `execute_model`, so the raise goes straight into the engine core loop. I served the head with Qwen3-0.6B and sent one request with min_p:
1.plain request                    200, " Paris. The capital of France is also"
2.same prompt with "min_p": 0.1    500, EngineCore encountered an issue
3.plain request again              connection refused
```
NotImplementedError: vllm-metal does not support sampling controls backed by vLLM logits processors (min_p); request='cmpl-a80bb2ed2c5fcce9-0-b4685659'.
File ".../vllm_metal/v1/model_runner.py", line 1857, in _handle_new_requests
ERROR [core.py:1332] EngineCore encountered a fatal error.
ERROR [async_llm.py:704] vllm.v1.engine.exceptions.EngineDeadError
```
One bad request and the server.

Can we put this in `MetalPlatform.validate_request` instead? `input_processor.py:296` calls it for every request before it reaches the engine, on both the async and sync paths, and we do not override it today. The startup plugin check is fine where it is.

ricky-chaoju (commented):
Thanks. Move looks right, engine stays up now. Plain request 200, min_p / logit_bias / min_tokens all rejected, plain request again 200.

Missed something last time though. The client gets nothing back:

```json
{"error":{"message":"","type":"InternalServerError","code":500}}
```

Not really the type's fault. `create_error_response` maps `NotImplementedError` to 501 fine, it just isn't in the handler list at `api_server.py:287`, so it never gets there and Starlette returns the generic 500. Looks like an upstream gap.

I swapped it for `VLLMValidationError` locally:

```json
{"error":{"message":"vllm-metal does not support ... (min_p). (parameter=min_p)","type":"BadRequestError","param":"min_p","code":400}}
```

Server still up, and the client can see which option was rejected. Can we use that? It is a `ValueError` subclass from `vllm.exceptions`. `test_platform.py` needs the same swap. WDYT?

LxYuan0420:
@ricky-chaoju  My bad and thanks for flagging it out. Please give it a try and help me test it out again 🙌

ricky-chaoju (approved):
LGTM. Three come back as client errors now and the server keeps going:

```
min_p=0.1              400  "param":"min_p"
logit_bias={"5":2.0}   400  "param":"logit_bias"
min_tokens=3           400  "param":"min_tokens"
```
Plain request 200 before and after.
Thanks for fixing it !

PR #578, opened 2026-08-03
Default MLX command-buffer limits on macOS

ricky-chaoju:
A vLLM decode step on a large MoE model builds thousands of lazy ops per submit, and MLX's command-buffer split defaults are sized for small generate loops, so the step submit spends much of its time on per-buffer commits. This defaults `MLX_MAX_OPS_PER_BUFFER` and `MLX_MAX_MB_PER_BUFFER` to `2000` during plugin registration on macOS, via `setdefault` so a value the user exports always wins; the two limits default independently.

Standalone on main this is worth 42.4 to 44.8 tok/s single-stream (`Qwen3.6-35B-A3B-4bit`, M1 Ultra, greedy 256 tokens, temperature 0), and greedy output is byte-identical. 2000 sits on a measured plateau (1000 and 2000 equal within noise), so the smaller value ships. The effect compounds with the decode pipelining change I'll send next, which is where most of the win in my branch came from.

The defaults reach every process: `_register` re-runs during platform resolution in the EngineCore, and spawned workers inherit the parent environment either way. Tests pin the exact default strings, per-variable user override, independent defaulting, and the non-macOS no-op.

### DIFF
--- a/vllm_metal/__init__.py
+++ b/vllm_metal/__init__.py
@@ -60,6 +60,28 @@ def _apply_macos_defaults() -> None:
     )
 
 
+# MLX splits each lazy evaluation into command buffers after this many ops /
+# this much memory; its defaults suit small generate loops, while a vLLM
+# decode step builds thousands of lazy ops per submit. 2000 sits on the
+# measured plateau, and ``setdefault`` keeps user values authoritative.
+_MLX_BUFFER_ENV_DEFAULTS = {
+    "MLX_MAX_OPS_PER_BUFFER": "2000",
+    "MLX_MAX_MB_PER_BUFFER": "2000",
+}
+
+
+def _apply_mlx_buffer_defaults() -> None:
+    """Default MLX command-buffer limits to decode-step-sized values."""
+    if sys.platform != "darwin":
+        return
+    for name, value in _MLX_BUFFER_ENV_DEFAULTS.items():
+        if name not in os.environ:
+            logger.debug(
+                "Defaulting %s=%s (set it explicitly to override).", name, value
+            )
+        os.environ.setdefault(name, value)
+
+
 # Lazy imports to avoid loading vLLM dependencies when just importing the Rust extension
 def __getattr__(name):
     """Lazy import module components."""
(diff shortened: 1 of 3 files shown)


vllm_metal/__init__.py
```diff
@@ -60,6 +60,28 @@ def _apply_macos_defaults() -> None:
...
+    if sys.platform != "darwin":
+        return
+    for name, value in _MLX_BUFFER_ENV_DEFAULTS.items():
+        if name not in os.environ:
+            logger.debug(
+                "Defaulting %s=%s (set it explicitly to override).", name, value
+            )
+        os.environ.setdefault(name, value)
+
+
```
LxYuan0420, line 84: Im okay with this 2000 default together with the set default behaviour

but before merging, could you provide a little more justification for choosing 2000 here?

LxYuan0420 (approved):

vllm_metal/__init__.py
ricky-chaoju: Ran a sweep to back the number. Same setup as the PR (`Qwen3.6-35B-A3B-4bit`, M1 Ultra, greedy 256 tokens, temp 0, median of 3), MB fixed at 2000 while sweeping OPS:

| config | tok/s |
|---|---:|
| native defaults (both unset) | 42.4 |
| OPS=500 | 44.8 |
| OPS=1000 | 44.9 |
| OPS=2000 (this PR) | 44.8 |
| OPS=4000 | 45.3 |
| OPS=8000 | 44.6 |
| OPS=2000, MB unset | 43.9 |

So it's a step, not a slope: you get basically the whole win the moment you leave the native default, and everything from 500 to 8000 is the same within noise (~±0.4 here). MB matters on its own too; leaving it at the native default gives back about half the win, which is why the PR sets both.

Why 2000 specifically: it's been on the plateau everywhere I've measured, this sweep and my pipelined branch (there 1000, 2000, and 6000 came out equal). Any value in that range would be fine; I took a low one so buffers stay smaller, since going higher never bought anything.

PR #579, opened 2026-08-05
Fix hybrid MambaSpec cache accounting

WindChimeRan:
## Summary

- **Problem:** Hybrid GDN linear attention keeps one fixed-size convolution + recurrent state per request, but its scheduler-visible `MambaSpec` described that state like a token-growing attention cache. This silently wasted pool capacity and under-declared the recurrent state's fp32 storage.
- **Before:** The spec used the smaller attention `block_size` and declared both state tensors at the model dtype. Because the scheduler allocates `ceil(sequence_length / spec.block_size)` blocks, a 1,301-token request with a 544-token attention block reserved three Mamba blocks per cache group even though only one state block was needed. The runtime also stored the recurrent matrix in fp32 while the spec commonly billed it as fp16/bf16.
- **After:** The spec uses `cache_config.mamba_block_size` (`max_model_len` while prefix caching is disabled), declares the recurrent matrix as fp32, identifies the backend as `GDN_ATTN`, and propagates `mamba_cache_mode`. Any valid request therefore reserves one Mamba state block per group in `none` mode, and scheduler memory accounting matches the runtime layout.
- **Core fix:** `MambaSpec(block_size=mamba_block_size, dtypes=(torch_dtype, torch.float32), mamba_type=GDN_ATTN, mamba_cache_mode=cache_config.mamba_cache_mode)`.
- **Scheduler compatibility:** When attention and Mamba block sizes differ, grouping follows upstream's LCM alignment and fine-hash fallback behavior.
- **Regression coverage:** A general, non-TurboQuant cache-spec test captures the three-block-to-one-block fix; the existing TurboQuant integration test pins the affected grouping behavior.

- Future PR: this will unlock the support to Qwen3.5-3.8, LFM2.5 models, and prefix caching support. 

## Testing

- `pytest -q tests/test_kv_cache_spec_capacity.py tests/test_turboquant_hybrid_sizing.py` (14 passed)
- `ruff check` on all changed Python files

### DIFF
--- a/vllm_metal/attention/runtime/hybrid.py
+++ b/vllm_metal/attention/runtime/hybrid.py
@@ -17,6 +17,7 @@ import mlx.core as mx
 import mlx.nn as nn
 import torch
 from vllm.logger import init_logger
+from vllm.v1.attention.backends.registry import MambaAttentionBackendEnum
 from vllm.v1.kv_cache_interface import MambaSpec
 
 from vllm_metal.attention.caches.gdn_cache import GDNPagedStateCache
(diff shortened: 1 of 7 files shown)


vllm_metal/attention/runtime/hybrid.py
```diff
@@ -46,25 +47,24 @@ def _build_linear_layer_spec(
...
         shapes=(
             (conv_kernel_dim - 1, conv_dim),
             (num_v_heads, value_head_dim, key_head_dim),
         ),
-        dtypes=(torch_dtype, torch_dtype),
-        block_size=block_size,
+        # Match the fp32 recurrent pool used for kernel accumulation.
+        dtypes=(torch_dtype, torch.float32),
+        block_size=mamba_block_size,
         page_size_padded=page_size_padded,
```
LxYuan0420, line 65: Can we resolve this at the platform config boundary too? This now declares the recurrent state as fp32, but upstream calculates mamba_page_size_padded from cache_config.mamba_ssm_cache_dtype before this spec is built. If that is still auto/non-fp32, the padded page can be smaller than this fp32 spec. I think Metal should set auto -> float32 and reject explicit non-fp32 for hybrid GDN in check_and_update_config().

LxYuan0420 (approved):
LGTM overall. I left one config-boundary question to consider.

WindChimeRan:
@LxYuan0420 Thanks! addressed the comments.

PR #581, opened 2026-08-05
Route Qwen MLX quant text wrappers through mlx_lm

### LINKED ISSUE #580: [Bug] Qwen3.6-27B dense (Qwen3_5ForConditionalGeneration) MLX 4-bit generates garbled output on native mlx_vlm path — same class as #571 but not covered by #573
## Summary

`mlx-community/Qwen3.6-27B-4bit` (dense 27B, `Qwen3_5ForConditionalGeneration`, MLX affine 4-bit, `quantization: {bits: 4, group_size: 64, mode: affine}`) loads without error under vllm-metal and then generates incoherent token soup. The **same checkpoint generates correct text under `mlx_lm` directly**, and `Qwen/Qwen3.5-0.8B` (BF16) is correct under vllm-metal — so this is specific to the MLX-quantized dense wrapper, mirroring #571 (which covered the MoE variants).

## Root cause

PR #573 force-routes MLX-affine checkpoints through the `mlx_lm` text backbone only for architectures in `_MLX_QUANT_TEXT_BACKBONE_ARCHITECTURES`:

```python
_MLX_QUANT_TEXT_BACKBONE_ARCHITECTURES = frozenset({
    "Qwen3_5MoeForConditionalGeneration",
    "Qwen3_6ForConditionalGeneration",
    "Qwen3_6MoeForConditionalGeneration",
})
```

`Qwen3_5ForConditionalGeneration` was deliberately excluded ("has a native multimodal adapter, keeps the native path and image serving"). However, the dense Qwen3.6-27B 4-bit checkpoint **also garbles on the native mlx_vlm path** when served text-only — the server log shows `MLX-VLM model loaded`, and output is token soup for both `/v1/completions` and `/v1/chat/completions`.

## Reproduction

```bash
vllm serve mlx-community/Qwen3.6-27B-4bit --port 8100 --max-model-len 32768
curl http://localhost:8100/v1/completions -H 'Content-Type: application/json' -d \
  '{"model":"mlx-community/Qwen3.6-27B-4bit","prompt":"The capital of France is","max_tokens":20,"temperature":0}'
```

**Observed (native path):** `izu.).s:== c |，:_D!`

**Expected (and observed with `mlx_lm generate` on the same checkpoint):** ` Paris.` — 24.5 tok/s, peak memory 15.4 GB.

## Workaround verified locally

Adding `"Qwen3_5ForConditionalGeneration"` to `_MLX_QUANT_TEXT_BACKBONE_ARCHITECTURES` in `vllm_metal/v1/model_adapter.py` fixes generation completely — both endpoints produce coherent output, 8.5k-token prompts generate correctly, and 4-way concurrent serving scales ~2.3x over single-stream. Tradeoff: vision input is bypassed (text-only serving).

## Environment

- vllm-metal `v0.3.0.dev20260803164600` + vLLM `0.26.0+cpu` (macOS arm64 wheel)
- MacBook Pro M3 Max, 48 GB unified memory, macOS 26.6
- Native arm64 Python 3.12.12
- Checkpoint: `mlx-community/Qwen3.6-27B-4bit`

If the dense 27B is expected to work on the native multimodal path, I am happy to help debug the mrope state handling there instead — but until then, routing it to the text backbone like its MoE siblings would at least make text serving correct.

LxYuan0420:
This PR is:

- To route dense Qwen3.5 MLX affine quant text wrappers through `mlx_lm`.
- To keep real Qwen VL MLX quant checkpoints on the native multimodal path when `vision_config` is present.
- To replace the architecture-only check with a clearer config-shape check.

Fixes #580.

Qwen3.5 dense text wrappers and real Qwen VL checkpoints can share `Qwen3_5ForConditionalGeneration`, so the architecture string alone is not enough. The routing now uses the checkpoint shape: MLX quant without `vision_config` uses the text path, while MLX quant with `vision_config` stays multimodal.

### DIFF
--- a/tests/test_model_adapter.py
+++ b/tests/test_model_adapter.py
@@ -87,11 +87,6 @@ class TestShouldForceTextBackbone:
                 "Qwen3VLForConditionalGeneration",
                 SimpleNamespace(spatial_merge_size=2),
             ),
-            (
-                "qwen3_5",
-                "Qwen3_5ForConditionalGeneration",
-                SimpleNamespace(spatial_merge_size=2),
-            ),
             # An arbitrary non-allowlisted multimodal model.
             ("llava", "LlavaForConditionalGeneration", None),
         ],
@@ -108,6 +103,20 @@ class TestShouldForceTextBackbone:
         adapter = DefaultModelAdapter()
         result = adapter.should_force_text_backbone(hf_config)
         assert result is False
+
+    def test_mlx_quant_qwen35_wrapper_uses_auto_override_with_vision_config(
+        self,
+    ) -> None:
+        hf_config = SimpleNamespace(
+            model_type="qwen3_5",
+            architectures=["Qwen3_5ForConditionalGeneration"],
+            quantization={"group_size": 64, "bits": 4, "mode": "affine"},
+            vision_config=SimpleNamespace(spatial_merge_size=2),
+            text_config=SimpleNamespace(model_type="qwen3_5_text"),
+        )
+        adapter = DefaultModelAdapter()
+        result = adapter.should_force_text_backbone(hf_config)
+        assert result is True
 
     def test_multimodal_native_mode_disables_mlx_quant_override(
         self, monkeypatch
@@ -408,6 +417,7 @@ class TestNormalizeModelConfig:
         [
             ("qwen3_5", "Qwen3_5ForConditionalGeneration"),
             ("qwen3_5_moe", "Qwen3_5MoeForConditionalGeneration"),
+            ("qwen3_6", "Qwen3_6ForConditionalGeneration"),
         ],
     )
     def test_clears_multimodal_config_for_mlx_quant_wrapper_in_auto_mode(
--- a/tests/test_model_adapter.py
+++ b/tests/test_model_adapter.py
@@ -58,8 +58,8 @@ class TestShouldForceTextBackbone:
     @pytest.mark.parametrize(
         ("model_type", "architecture"),
         [
-            # The real pair of the motivating checkpoints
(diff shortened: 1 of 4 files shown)


ricky-chaoju (commented):
This doesn't fix it. Served `mlx-community/Qwen3.6-27B-4bit` on this branch:

MLX-VLM model loaded
"The capital of France is"      -> '辅������〕|or>(4�%}� =#hf |:'
"2 + 2 ="                       -> '.……0\n\n. ; –,.)11型7/0技给二\n]intढ्र'
"The largest ocean on Earth is" -> ",,.\n\n��#�� )**���_^与私))')(*� Tarij"

`should_force_text_backbone` returns False for that checkpoint on `602e8f0` and on this head, so its routing is unchanged.

Forcing `_has_native_qwen_vl_config` to False locally and re-serving the same model with the same prompts:

MLX-LM model loaded
"The capital of France is"      -> ' Paris.\n\n<think>\n\n</think>\n\nThat is correct. Paris is the capital and most populous city of France. It is'
"2 + 2 ="                       -> " 5\n\n<think>\nHere's a thinking process:..."
"The largest ocean on Earth is" -> ' the Pacific Ocean. It covers an area of approximately 63.8 million square miles (165.2'

LxYuan0420:
My bad. I used `vision_config` as the native-VL signal, but that is not correct here.

I fixed MLX-affine Qwen3.5/Qwen3.6 text-wrapper architectures to use the `mlx_lm` path even when `vision_config` exists. Real Qwen3-VL still stays native via `Qwen3VLForConditionalGeneration`.

@ricky-chaoju

ricky-chaoju (approved):
LGTM. Thanks for the quick fix.

PR #583, opened 2026-08-06
Reject eager checkpoint loaders under pipeline parallelism

LxYuan0420:
This PR is:

- To reject pipeline parallelism with text AWQ and GGUF checkpoints before workers allocate model weights.
- To keep lazy MLX-LM and MLX-VLM load paths available under pipeline parallelism.
- To document the loader limitation and cover the admission behavior.

Note:

AWQ and GGUF text loaders materialize the full checkpoint before stage slicing, so PP can still OOM before it reduces per-stage memory. Users should use `pipeline_parallel_size=1` or an MLX-LM safetensors checkpoint.


ricky-chaoju (approved):
LGTM

PR #584, opened 2026-08-06
Prefix caching for hybrid GDN models

WindChimeRan:
## Summary

Implements upstream's `mamba_cache_mode="align"` for the Metal/MLX worker. `--enable-prefix-caching` now works on hybrid GDN models such as Qwen3.5, which interleave SDPA and gated-DeltaNet layers.

- The worker indexes GDN state by scheduler block id, so a cached block also identifies its recurrent-state checkpoint.
- `mamba_cache_mode="none"` keeps the existing request-keyed allocator. Caching-off deployments do not construct the new block-id path.
- The cache-on/off parity matrix passes 80/80 comparisons, including prefix hits and block-aligned chunked prefill.
- Shared-prefix requests improve median served TTFT by 2.4×. All-cold unique prompts run at 0.38× throughput at concurrency 1 and 0.80× at concurrency 8; a CUDA control points to the MLX worker's lazy state pool.

Hybrid prefix caching remains opt-in upstream, and this PR does not change that default.

## Design

A GDN layer keeps a fixed-size recurrent state rather than per-token KV. Align mode checkpoints that state at scheduler block boundaries. Before each forward, the worker performs two operations per mamba cache group:

- **Zero-init:** clear the destination slab for a fresh request because a reused block may contain old state.
- **Copy-forward:** when a request enters a new block, copy the previous block's state into the new slab and leave the old slab unchanged as a checkpoint.

A prefix hit starts at a checkpointed block, so its first copy-forward restores the state. Requests do not own slabs; the scheduler's block lifecycle does. The engine's striped mamba cache groups are planned independently, and layers that share a physical pool retain separate group rows.

Unlike #547, this design has one store: live state and cached checkpoints both reside in scheduler-indexed slabs. There is no request-keyed live store to synchronize with a block-keyed snapshot store.

## Validation

Experiments used `Qwen/Qwen3.5-0.8B` (18 GDN and 6 SDPA layers, resolved block size 544) with greedy sampling. Metal results were measured on an Apple M5 Pro with 64 GB.

### Correctness

- The in-tree e2e test passes token identity against cache-off and asserts that align mode and a real prefix hit were reached.
- The parity matrix passes 40/40 comparisons with default scheduling and 40/40 with `max_num_batched_tokens=1088`.

The default gate pins prefill-produced checkpoints. A separate `--shared-corpus` case documents the numerical caveat for a checkpoint completed by decode: 17 tokens matched before the restored path reached an exact greedy tie and selected a different token. See the reproduction details for the tested matrix and caveat.

### Shared-prefix latency

The served workload contains 16 requests at concurrency 1, each with a 1,088-token shared prefix, a 112-token unique suffix, and 32 output tokens. The first request is cold and the remaining requests hit the prefix.

| metric | cache-off (`none`) | cache-on (`align`) | |
|---|---:|---:|---|
| Median TTFT | 128.8 ms | 52.8 ms | **2.4× faster** |
| Mean TTFT | 142.7 ms | 80.0 ms | 1.8× |
| P99 TTFT | 319.7 ms | 326.3 ms | ≈ cold + compile request |

An in-process `generate(max_tokens=1)` probe isolates engine time:

| case (1,200-token prompts) | cache-off | cache-on | |
|---|---:|---:|---|
| hot: exact repeat | 118 ms | 29.5 ms | **4.0×** |
| hot: shared prefix, new suffix | 126 ms | 26.6 ms | **4.7×** |
| clean cold prefill | 121 ms | 144 ms | **+19%** |

### All-cold throughput

Unique 1,200-token prompts show the current cost when there are no prefix hits:

| concurrency | cache-off | cache-on | on/off |
|---|---:|---:|---:|
| 1 | 2764 tok/s | 1037 tok/s | **0.38** |
| 8 | 5528 tok/s | 4428 tok/s | **0.80** |

Most of this regression occurs during decode. The state pool materializes through the highest block id in use, and each decode step walks the growing pool. A cold 32-token request moves from 446 ms to 1188 ms, while prefill accounts for only 23 ms of the 742 ms gap.

A CUDA control on an NVIDIA GB10 used the same model, block size, and all-cold workload. Its worker preallocates mamba state rather than growing MLX slabs lazily:

| conc | metric | cache-off | cache-on | on/off |
|---|---|---:|---:|---:|
| 1 | total throughput | 3658 tok/s | 3526 tok/s | **0.96** |
| 1 | median TTFT | 56.2 ms | 65.2 ms | 1.16 |
| 1 | median TPOT | 8.90 ms | 9.01 ms | **1.01** |
| 8 | total throughput | 14663 tok/s | 14259 tok/s | **0.97** |
| 8 | median TTFT | 240.3 ms | 306.0 ms | 1.27 |
| 8 | median TPOT | 12.88 ms | 12.44 ms | **0.97** |

The CUDA on/off ratios suggest that the large Metal regression is worker-side rather than inherent to align mode. Absolute CUDA and Metal numbers are not comparable.

## Limits

- Hybrid prefix caching cannot be combined with speculative decoding because draft-state rollback across mamba state blocks is not implemented.
- `mamba_cache_mode="all"` is rejected. Upstream also falls back to align for models without `SupportsMambaPrefixCaching`.
- The memory plan charges every block for state bytes, while resident state grows lazily to the high-water block id. This keeps admission worst-case safe but reduces the planned block count from 3162 to 1585 on the tested configuration.
- Checkpoints completed by decode are mathematically equivalent but not bitwise identical to prefill-produced checkpoints, matching the numerical caveat of upstream decode-produced KV reuse.

Enable caching for workloads with repeated prefixes. The lazy-pool decode cost on unique-prompt streams is follow-up work.

## Review guide

The highest-risk paths are:

1. `attention/state/align.py`: destination/source block math and zero/copy planning.
2. `attention/caches/gdn_cache.py`: lazy batched slab operations and shared-pool aliasing.
3. `v1/cache_policy.py`: state-pool grouping, validation, and per-block memory accounting.
4. `v1/model_runner.py`: state block-id lifecycle across admission, append, resume, and finish.

<details>
<summary>Full file-by-file map</summary>

| File | What to check |
|---|---|
| `attention/state/align.py` | Pool growth, zero/copy planning, and per-group slab mappings. |
| `attention/caches/gdn_cache.py` | `copy_slots`/`zero_slots`, source immutability, and shared-pool storage. |
| `attention/runtime/hybrid.py` | Align/none mode fork and scheduler-group adoption. |
| `v1/cache_policy.py` | Mamba groups, physical pools, validation, and memory accounting. |
| `v1/model_runner.py` | State block-id lifecycle and per-step positions. |
| `impls/linear.py`, `impls/gdn_lazy.py`, `context.py` | Mapping consumption and state write-back. |
| `platform.py` | Paged-attention, `all`-mode, and speculative-decoding gates. |
| `state/gdn.py`, `runtime/base.py`, `protocol.py`, `state/__init__.py` | Signature plumbing and unchanged none-mode behavior. |
| Tests | State motion, shared pools, growth bounds, block lifecycle, and platform gates. |
| `tools/hybrid_apc_parity_matrix.py` | Cache-on/off parity and reach assertions. |

</details>

<details>
<summary>Reproduction commands and experiment settings</summary>

Enable the feature:

```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 vllm serve Qwen/Qwen3.5-0.8B --enable-prefix-caching
# Expected log: "Hybrid cache initialized: ... mamba_cache_mode=align"
```

Run the correctness gates:

```bash
VLLM_METAL_E2E=1 QWEN35_MODEL_PATH=Qwen/Qwen3.5-0.8B \
  pytest tests/test_hybrid_prefix_caching_e2e.py -q

QWEN35_MODEL_PATH=Qwen/Qwen3.5-0.8B \
  python tools/hybrid_apc_parity_matrix.py
```

The 80-comparison matrix covers prompts at `k * 544 +/- 1`, divergent-suffix hits at two depths, exact repeats, a concurrent batch, and a `max_num_batched_tokens=1088` arm. It asserts configuration reach, prefix-hit reach, and token identity. `--shared-corpus` reproduces the decode-completed checkpoint caveat described above.

Run the shared-prefix serving benchmark:

```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 vllm serve Qwen/Qwen3.5-0.8B \
  --max-model-len 2048 --max-num-seqs 4 --enable-prefix-caching --port 8123
# Second arm: --no-enable-prefix-caching

vllm bench serve --base-url http://127.0.0.1:8123 \
  --model Qwen/Qwen3.5-0.8B --dataset-name random \
  --random-input-len 112 --random-prefix-len 1088 \
  --random-range-ratio 0 --random-output-len 32 \
  --num-prompts 16 --max-concurrency 1 --seed 0
```

For the all-cold Metal workload, use `--random-prefix-len 0 --random-input-len 1200`, 32 prompts, and `--max-num-seqs 8`. The CUDA control used vLLM `c56f169d9a`; each arm ran against a fresh server and reported zero prefix-cache hits.

</details>


vllm_metal/v1/model_runner.py
```diff
@@ -1118,7 +1139,36 @@ def _start_paged_forward(
...
+                    ]
+                    step_positions.extend(
+                        (pr.start_pos, len(pr.token_ids)) for pr in prefill_reqs
+                    )
+                runtime.populate_step_context(
+                    req_ids=step_req_ids,
+                    ctx=ctx,
+                    state_block_ids=step_state_ids,
+                    step_positions=step_positions,
+                )
```
LxYuan0420, line 1188: Apply scheduler_output.kv_cache_block_copies before populate_step_context. vLLM uses this for partial prefix-hit copy-on-write; without it, the rewritten table already points at the destination block and align can leave SDPA KV / GDN state stale.

vllm_metal/attention/runtime/hybrid.py
```diff
@@ -146,31 +156,66 @@ def initialize(self, num_blocks: int) -> None:
...
-            max_seqs=self._max_num_seqs,
+            max_seqs=state_slots,
             conv_kernel_dim=self._linear_conv_kernel_dim,
             conv_dim=self._linear_conv_dim,
             num_v_heads=self._linear_num_v_heads,
             value_head_dim=self._linear_value_head_dim,
             key_head_dim=self._linear_key_head_dim,
-            initial_seqs=0,
+            initial_seqs=initial_slots,
             dtype=self._dtype,
```
LxYuan0420, line 181: Start align mode with zero initial slots. This allocates before set_layer_layout() adopts shared_by, so it materializes the identity layout while the planner budgets the final shared layout.

vllm_metal/attention/caches/gdn_cache.py
```diff
@@ -132,25 +156,25 @@ def ensure_capacity(self, num_seqs: int) -> None:
...
-            recurrent_states.append(recurrent)
 
-        self.conv_states = conv_states
-        self.recurrent_states = recurrent_states
+            # Materialize now so each old pool is released as it is copied;
+            # the transient peak stays ~one pool above the new size (the
+            # plan budgets the final size only).
+            mx.eval(conv, recurrent)
+            self.store_conv_state(layer_idx, conv)
+            self.store_recurrent_state(layer_idx, recurrent)
```
LxYuan0420, line 175: Budget one old physical pool for growth overlap. ensure_capacity() holds the old pool while materializing its replacement, but the planner currently charges only steady-state memory.

LxYuan0420 (approved):
Approving to unblock the progress; but please address those comments before merging.

anthonyalayo:
Hi @WindChimeRan and @LxYuan0420,

Thank you for the work on hybrid prefix-caching! 

I’ve been attempting to benchmark this PR branch locally on an Apple M4 Max (48GB Unified Memory, with sysctl limits adjusted to expose 40GB to Metal). I am hitting significant issues with memory allocation requirements that I shouldn't be hitting.

I believe (I could be wrong) this branch has https://github.com/vllm-project/vllm-metal/issues/585 and https://github.com/vllm-project/vllm-metal/pull/586 included in it, but if that isn't the case, please correct me. Regardless, I think the following are issues.

---

## Reproduction

### Successfully running Qwen3.6 27B on llama.cpp with 100k context window
Using `llama.cpp`, I can load this model with a full 100K context window. By specifying `-ub 2048`, `llama.cpp` restricts active GPU computations to a 2,048-token microbatch, keeping the compute scratchpad at a constant, low footprint of **976.18 MiB**:

```bash
./llama-server \
  -m <path-to-Qwen3.6-27B-UD-Q4_K_XL.gguf> \
  --no-mmproj -c 102400 -fa on -np 1 \
  -b 2048 -ub 2048 \
  --spec-type draft-dflash \
  --spec-draft-model <path-to-Qwen3.6-27B-DFlash-Q8_0.gguf> \
  --spec-draft-ngl 999 -ctk q8_0 -ctv q8_0 --cache-reuse 256 \
  -lv 4 --port 8080
```
**Result:** Successfully boots and runs within a total memory footprint of **22.56 GB** (16.39 GB weights + 5.22 GB context cache + 0.95 GB compute buffer).

### Failing to run Qwen3.6 27B on this PR with 80K context window
Here is my attempt to run the same model as a text-only target with an 80K limit and a 2080 batched token limit (required by the block-alignment constraint):

```bash
VLLM_LOGGING_LEVEL=DEBUG HF_HUB_OFFLINE=1 vllm serve \
  --model mlx-community/Qwen3.6-27B-4bit \
  --max-model-len 80K \
  --enable-prefix-caching \
  --kv-cache-dtype turboquant_k8v4 \
  --max-num-seqs 1 \
  --max-num-batched-tokens 2080 \
  --language-model-only \
  --port 8080
```
**Result:** After the profiling run completes, the server aborts with a false-positive OOM validation failure:
```text
ValueError: To serve at least one request with the model's max seq len (81920), 5.84 GiB KV cache is needed, which is larger than the available KV cache memory (4.06 GiB). Based on the available memory, the estimated maximum model length is 52000.
```

---

## Analyzing Results

I am an amateur to this space, so please take this with a grain of salt, but I believe the root cause of this failure may be a memory double-counting issue in how the `cache_policy.py` planner evaluates model memory bounds?

When running with `--max-model-len 80K`, I believe the PyTorch Inductor (`torch.compile`) compiles an optimized execution graph, which is up to the maximum context window, about 17.08 GB. 

By monitoring the `mlx.core` memory allocator state immediately after `profile_run()` finishes and before `setup_paged_attention()`, I saw:
* **Active Memory (Permanent Weights):** 15.13 GB
* **Cache Memory (Reclaimable Workspace):** 1.18 GB

However, during budget estimation, the paged attention memory breakdown logs:
`usable_metal=39.51GB, model_memory=32.21GB`

Because `get_model_memory_usage()` simply queries active memory allocated on the device, I think **it mistakenly classifies the JIT compiler's compiled execution workspace (17.08 GB) as permanent model weight memory.** 

This 32.21 GB is permanently subtracted from the 39.51 GB limit, leaving only **4.06 GiB** for the KV cache block pool. This starves the budget and causes the engine to throw a false-positive `ValueError` at startup, even though the transient prefill compilation activations and the static block pool never need to coexist in RAM. 

This constraint is further tightened by the block tax introduced in this PR (`per_block_bytes += self._hybrid_align_state_bytes_per_block()`), which pre-charges the budget for all block-aligned Mamba states up-front instead of dynamically.

## Discussion

What are your thoughts about this? Am I missing something, or is this genuinely an issue? Thank you for your time!

WindChimeRan:
Thanks for reporting the issue! @anthonyalayo 

We are not using pytorch (we use mlx instead). and that's why precise memory management is more difficult. 

Following your recipe, I run some probing tests. quick updates:
* this PR indeed has block double counting issue: every block id is charged for both gdn and normal kv cache. This is not ideal. But I'll probably have to defer it to future PRs (because it might requires some kernel work for finegrained memory read and write). 
* Our memory accounting is indeed confusing. 

I'll investigate these issues further.

anthonyalayo:
Makes sense, thanks @WindChimeRan !

vllm_metal/v1/model_runner.py
WindChimeRan: Thanks for catching this. I added `copy_blocks()` support and now apply `scheduler_output.kv_cache_block_copies` before `populate_step_context()`, including both the SDPA KV cache and align-mode GDN state.

vllm_metal/attention/runtime/hybrid.py
WindChimeRan: Thanks! Implemented the fix in commit `21eb4b3`

vllm_metal/attention/caches/gdn_cache.py
WindChimeRan: Thanks for catching this. commit `f0af111` now budgets one additional physical GDN pool for the temporary old/new overlap during ensure_capacity() growth.

anthonyalayo:
@WindChimeRan is there any place tracking this follow up?

> this PR indeed has block double counting issue: every block id is charged for both gdn and normal kv cache. This is not ideal. But I'll probably have to defer it to future PRs (because it might requires some kernel work for finegrained memory read and write).

Also, mind sharing your idea for the fix? I can try out implementations locally to see if any look stable.

WindChimeRan:
@anthonyalayo I think the block double counting issue has been fixed by #778 , by a big refactoring (Now different kinds of kv cache has a unified management) 

image

anthonyalayo:
Huge! Thank you!

PR #586, opened 2026-08-07
Stop defaulting MLX buffer memory limit

### LINKED ISSUE #585: [Bug] MLX_MAX_MB_PER_BUFFER=2000 default from #578 inflates the profile_run overhead 19x and blocks startup
On an Apple M4 Pro with 48 GB of unified memory, `openai/gpt-oss-20b` fails to
start on current `main` when `max_num_batched_tokens` is 8192. It starts on the
commit before #578. Dropping the new `MLX_MAX_MB_PER_BUFFER` default while
keeping the op default takes the measured overhead from 37.86 GB to 2.02 GB and
the server comes up.

Scope first, so this is not read as broader than it is: `vllm serve` at its own
default of 2048 batched tokens is **not** affected and starts cleanly. 8192 is
the default for offline `LLM(...)` (`arg_utils.py:2472-2475`: `LLM_CLASS` 8192,
`OPENAI_API_SERVER` 2048), so the people who hit this are offline-inference
users, who need no unusual flags, and anyone who raises
`--max-num-batched-tokens` on the server.

### Reproduction

On current `main`:

```bash
vllm serve openai/gpt-oss-20b --max-model-len 8192 --max-num-seqs 1 \
  --max-num-batched-tokens 8192
```

```
metal_limit=40.20GB, fraction=0.92, usable_metal=36.98GB,
model_memory=13.76GB, overhead=37.86GB, kv_budget=-14.63GB.
Mitigations: increase VLLM_METAL_MEMORY_FRACTION; use a smaller or more
quantized model.
```

Offline `LLM(...)` reaches the same place with no flag at all, since
`max_num_batched_tokens` defaults to 8192 there (`arg_utils.py:2472-2475`).
That is where I first hit this. On one of those runs it failed earlier still,
inside the profile forward itself:

```
RuntimeError: [METAL] Command buffer execution failed: Insufficient Memory
(00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
  ... model_runner.py:668 in profile_run -> mx.eval(*self._dummy_forward_outputs(...))
```

Neither mitigation in that message works. `base_kv_budget = metal_limit *
fraction - model_memory - overhead` (`cache_policy.py:938-946`), so even at
`fraction=1.0` the budget is 40.20 - 13.76 - 37.86 = -11.42 GB. No value of
`VLLM_METAL_MEMORY_FRACTION` can rescue it, because the overhead alone exceeds
what is left after the model.

Workaround for anyone hitting this today: `MLX_MAX_MB_PER_BUFFER=100`.

### Why the batched-token count is the trigger

`model_runner.py:656-671`:

```python
def profile_run(self) -> int:
    warmup_len = self.scheduler_config.max_num_batched_tokens
    mx.clear_cache()
    cache_before = mx.get_cache_memory()
    dummy_tokens = mx.zeros((1, warmup_len), dtype=mx.int32)
    mx.eval(*self._dummy_forward_outputs(dummy_tokens))
    overhead = mx.get_cache_memory() - cache_before
    mx.set_cache_limit(overhead)
    return overhead
```

Three things here interact with a raised `MLX_MAX_MB_PER_BUFFER`:

1. `warmup_len` is `max_num_batched_tokens`, so the profile forward is 4x
   larger at 8192 than at the server default.
2. `overhead` is a **buffer-cache delta**, not a peak and not active memory. A
   larger per-buffer ceiling means fewer command-buffer splits, so intermediates
   are not recycled mid-eval and far more distinct buffers are allocated, all of
   which land in the cache when the eval finishes. Note this figure is then
   subtracted from the same budget as `model_memory`, which is
   `mx.get_active_memory()` (`cache_policy.py:882-885`), a different pool.
3. Line 670 installs that same figure as MLX's serving cache limit. That is the
   guard added for #234. So on machines where startup does **not** fail, #578
   still raises this ceiling by whatever factor it inflates the measurement.
   That part is not specific to 48 GB Macs.

The command-buffer OOM above is a real allocation failure during the eval, not
only an accounting artifact, so both effects are in play.

### Why I think this was missed

#578 validated the change on decode throughput (42.4 to 44.8 tok/s,
`Qwen3.6-35B-A3B-4bit`, M1 Ultra) and on greedy output equality. Both hold up.
Outputs really are unaffected and I am not disputing the decode win.

Neither check touches startup sizing, which is where this bites, and an M1 Ultra
ships with 64 or 128 GB, so a 2000 MB per-buffer ceiling is a much smaller share
of the working set there. `docs/configuration.md:28` currently ends with
"Outputs are unaffected", which is accurate but reads as "this is safe"; the
memory footprint is the part that bites.

### Measurements

Same machine, same model, back to back, all with
`--max-model-len 8192 --max-num-seqs 1`. The last row is the untouched `vllm
serve` default and is included to show where the boundary sits:

| base | `max_num_batched_tokens` | `MLX_MAX_MB_PER_BUFFER` | overhead | kv_budget | result |
| --- | ---: | --- | ---: | ---: | --- |
| `46fc8d8` (parent of #578) | 8192 | MLX native default | 2.06 GB | +21.16 GB | starts |
| `c35fe7b` (current main) | 8192 | 2000 (#578) | **37.86 GB** | **-14.63 GB** | **fails to start** |
| `c35fe7b` | 8192 | 100 (user override) | 2.21 GB | +21.01 GB | starts |
| `c35fe7b` + candidate fix below | 8192 | not defaulted | **2.02 GB** | **+21.20 GB** | starts |
| `c35fe7b` | 2048 (serve default) | 2000 (#578) | 9.32 GB | +12.18 GB | starts |

`model_memory` was 13.76 GB in every 8192 row, so the overhead column is the
only thing moving.

It is the megabyte limit, not the op limit. The candidate-fix row keeps
`MLX_MAX_OPS_PER_BUFFER=2000` at #578's value and only stops defaulting the MB
limit, and the overhead drops from 37.86 GB to 2.02 GB.

### Candidate fix

```python
_MLX_BUFFER_ENV_DEFAULTS = {
    "MLX_MAX_OPS_PER_BUFFER": "2000",
}
```

That is the whole change, plus the corresponding assertion in
`tests/test_macos_defaults.py` and the note in `docs/configuration.md`. It
unbreaks startup here and costs nothing on large-memory machines.

What I cannot test is the half you measured: whether
`MLX_MAX_OPS_PER_BUFFER=2000` alone still carries the 42.4 to 44.8 tok/s win, or
whether the MB limit was doing real work for it. You have the M1 Ultra and the
`Qwen3.6-35B-A3B-4bit` setup, so that is a few minutes for you and not
reproducible for me.

If the MB limit does carry part of the decode win, the alternative is to cap it
against `mx.device_info()["max_recommended_working_set_size"]` rather than fix it
at 2000. Happy to test either shape on this machine and report back.

### Environment

- vllm-metal `main` @ `c35fe7b`, vLLM 0.26.0, MLX 0.32.0, mlx-lm 0.31.3
- Apple M4 Pro, 48 GB unified memory, macOS 26.5.1, Python 3.12.13
- `openai/gpt-oss-20b` (mxfp4), `VLLM_METAL_BUILD_FROM_SOURCE=1`
- `metal_limit=40.20GB` is `mx.device_info()["max_recommended_working_set_size"]`
  on this box, which reads 40.200896512 GB

### Scope of what I verified

- One model, one machine. I have not tested smaller models or other memory
  sizes. This needs the model plus an inflated profile measurement to crowd the
  budget, so a small model on the same machine would very likely be fine.
- I confirmed the vllm-metal side from source (`model_runner.py:656-671`,
  `cache_policy.py:882-946`), and the link from the MB default to the overhead
  figure is a direct A/B: four runs, one variable, `model_memory` identical at
  13.76 GB in all of them. What I have not done is read MLX's allocator, so the
  *explanation* for why a higher ceiling produces more distinct buffers is my
  reading of the numbers rather than of MLX source.
- The `set_cache_limit` consequence in point 3 is code-visible, but I have not
  measured what it costs during serving on a machine that starts successfully.

LxYuan0420:
This PR is:

- To keep the `MLX_MAX_OPS_PER_BUFFER=2000` default from #578.
- To stop defaulting `MLX_MAX_MB_PER_BUFFER`.
- To avoid inflating startup profile memory and shrinking the KV budget.

Fixes #585.

Note:

`MLX_MAX_MB_PER_BUFFER=2000` can make `profile_run()` measure a much larger MLX buffer-cache delta at high `max_num_batched_tokens`. That inflated overhead is then subtracted from the paged KV budget, so valid offline or high-batched-token runs can fail startup. Users can still set `MLX_MAX_MB_PER_BUFFER` manually if they want to tune it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,13 +19,13 @@
 
 ## MLX Command-Buffer Defaults
 
-On macOS the plugin defaults `MLX_MAX_OPS_PER_BUFFER` and
-`MLX_MAX_MB_PER_BUFFER` to `2000` via `setdefault`, so a value you export
-yourself always wins; the two limits default independently, so setting
-one keeps the plugin default for the other. MLX's own defaults are sized for small generate
-loops; a vLLM decode step on a large MoE model builds thousands of lazy
-ops per step, and the resulting per-buffer commit overhead slows the step
-submit. `2000` sits on the measured plateau. Outputs are unaffected.
+On macOS the plugin defaults `MLX_MAX_OPS_PER_BUFFER` to `2000` via
+`setdefault`, so a value you export yourself always wins. MLX's own default is
+sized for small generate loops; a vLLM decode step on a large MoE model builds
+thousands of lazy ops per step, and the resulting per-buffer commit overhead
+slows the step submit. `2000` sits on the measured plateau. The plugin does not
+default `MLX_MAX_MB_PER_BUFFER`; large global MB limits can inflate startup
+profile memory and reduce the KV budget. Outputs are unaffected.
 
 ## Multimodal Serve Modes
 
--- a/tests/test_macos_defaults.py
+++ b/tests/test_macos_defaults.py
@@ -6,8 +6,6 @@ import logging
 import os
 import sys
 
-import pytest
-
 import vllm_metal as vm
 
 
@@ -35,7 +33,7 @@ def test_apply_macos_defaults_noop_on_non_macos(monkeypatch) -> None:
     assert "VLLM_WORKER_MULTIPROC_METHOD" not in os.environ
 
 
-def test_apply_mlx_buffer_defaults_sets_both_limits(monkeypatch) -> None:
+def test_apply_mlx_buffer_defaults_sets_ops_limit_only(monkeypatch) -> None:
     monkeypatch.setattr(sys, "platform", "darwin")
     monkeypatch.delenv("MLX_MAX_OPS_PER_BUFFER", raising=False)
     monkeypatch.delenv("MLX_MAX_MB_PER_BUFFER", raising=False)
@@ -43,7 +41,7 @@ def test_apply_mlx_buffer_defaults_sets_both_limits(monkeypatch) -> None:
     vm._apply_mlx_buffer_defaults()
 
     assert os.environ["MLX_MAX_OPS_PER_BUFFER"] == "2000"
-    assert os.environ["MLX_MAX_MB_PER_BUFFER"] == "2000"
+    assert "MLX_MAX_MB_PER_BUFFER" not in os.environ
 
 
 def test_apply_mlx_buffer_defaults_respects_user_values(monkeypatch) -> None:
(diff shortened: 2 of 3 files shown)


ricky-chaoju (approved):
LGTM. Thanks for the quick fix.

PR #587, opened 2026-08-07
Batch structured-output bitmask application

LxYuan0420:
This PR is:

- To apply structured-output grammar bitmasks in one xgrammar call per batch.
- To remove the per-row xgrammar loop.
- To keep non-contiguous grammar rows mapped correctly.

Before: each constrained logits row called `xgr.apply_token_bitmask_inplace()` separately.

After: selected logits rows and matching bitmask rows are compacted into aligned CPU batches, then passed to xgrammar once.


ricky-chaoju (approved):
LGTM. Thanks~

PR #588, opened 2026-08-07
Add one-step-ahead decode sampling pipeline

ricky-chaoju:
Decode currently blocks on the sampling eval every step, so the CPU sits idle while the GPU runs and the GPU idles while the next graph gets built. This PR defers the sampling sync one step, the same overlap mlx_lm's generate loop uses: an eligible step submits a lazy argmax and returns an async output, and the next step feeds those lazy tokens straight in as its decode input while the GPU is still working.

The gate is owned by the new DecodePipeline and is strict: pure single-token decode, whole batch native greedy, paged path with async scheduling. Spec decode, structured output, logprobs, multimodal decode, pooling, PP and the mp executor all go through the synchronous path as before, and a configured speculative_config turns the pipeline off for the whole serve. `VLLM_METAL_DECODE_PIPELINE=0` is the kill switch. Desyncs raise instead of falling back silently.

Measured on my M1 Ultra, kill switch off vs on on this branch (`VLLM_METAL_MEMORY_FRACTION=0.5`, greedy, tok/s):

| model | off | on |
|---|---:|---:|
| Qwen3.6-35B-A3B-4bit | 44.2 | 57.1 |
| Qwen3.6-35B-A3B-4bit, concurrency 8 | 181.4 | 208.2 |
| gemma-4-26B-A4B-it-OptiQ-4bit | 46.9 | 53.5 |
| gemma-4-e2b-it-4bit | 56.6 | 84.2 |
| Qwen3-0.6B | 109.7 | 146.5 |


vllm_metal/v1/model_runner.py
```diff
@@ -1196,6 +1230,73 @@ def _start_paged_forward(
...
+            use_async_scheduling=self.use_async_scheduling,
+            paged_runtime_active=self._paged_attention_runtime is not None,
+            is_pooling=self._is_pooling,
+            pp_active=self.pp is not None and self.pp.size > 1,
+            hybrid_without_lazy_gdn=(
+                self.is_hybrid and not envs.VLLM_METAL_GDN_LAZY_KERNELS
+            ),
+            has_new_requests=(
+                bool(scheduler_output.scheduled_new_reqs) or states_missing
+            ),
```
LxYuan0420, line 1285: This misses `cached_reqs.all_token_ids`. vLLM uses that when a cached request comes back after being absent from the previous step. In that case there is no pending row, so this can crash later in `assemble_decode_input_ids`. Please block this and use the normal sync path.

vllm_metal/v1/model_runner.py
```diff
@@ -1196,6 +1230,73 @@ def _start_paged_forward(
...
+            has_structured_output=scheduler_output.has_structured_output_requests,
+            has_mm_decode=has_mm_decode or mm_forward_forced,
+            native_greedy=native_greedy,
+            has_prompt_logprobs=any(
+                sp.prompt_logprobs is not None for sp in decode_params
+            ),
+            multiproc_executor=(
+                self.vllm_config.parallel_config.distributed_executor_backend == "mp"
+            ),
```
LxYuan0420: Can we make this a positive UniProc check instead of only blocking `mp`? This pipeline depends on local async queue/thread ownership, and Ray/custom executors do not have the same contract.

vllm_metal/envs.py
```diff
@@ -58,6 +59,13 @@
     "VLLM_METAL_GDN_LAZY_KERNELS": lambda: (
         os.getenv("VLLM_METAL_GDN_LAZY_KERNELS", "1") == "1"
     ),
+    # One-step-ahead decode pipelining (default on). Eligible pure-decode
+    # greedy steps defer the sampling sync one step so the next step's graph
+    # build and submit overlap the in-flight GPU forward. Set to "0" to force
+    # the fully synchronous per-step sample path.
+    "VLLM_METAL_DECODE_PIPELINE": lambda: (
+        os.getenv("VLLM_METAL_DECODE_PIPELINE", "1") == "1"
+    ),
```
LxYuan0420, line 68: I am okay with this being default-on later, but not before the contract is tighter. Please keep it opt-in until the cached re-entry path, executor gate, and more real-model parity/perf results are covered.

vllm_metal/v1/decode_pipeline.py
```diff
@@ -0,0 +1,297 @@
...
+    has_prefill_phase_requests: bool
+    has_resumed_requests: bool
+    has_preempted_requests: bool
+    has_encoder_inputs: bool
+    has_spec_decode: bool
+    has_structured_output: bool
+    has_mm_decode: bool
+    native_greedy: bool
+    has_prompt_logprobs: bool
+    multiproc_executor: bool
```
LxYuan0420: This flat gate input is already hard to maintain. Can we split it into smaller owned shapes, like runner capability, scheduler step shape, and sampling shape, and let each expose its own block reason? Then `ModelRunner` only collects facts, and the pipeline policy is easier to extend safely.

LxYuan0420 (approved):
Approving to unblock progress. Thanks for working on this important feature.

Issue #525, opened 2026-07-19
[RFC] Selective logits projection for paged text prefill


HermanZeng:
## Motivation

The paged text path packs all scheduled tokens into one sequence in this order:

```text
[all decode/spec-verification rows][all rows of prefill 0][all rows of prefill 1]...
```

The target forward currently materializes logits for every packed row. Prefill
sampling then keeps only the final row of each prefill segment, while decode and
speculative verification consume the decode prefix:

- [`model_runner.py`: the target forward produces full logits](https://github.com/vllm-project/vllm-metal/blob/5f4e50d1fcff299b9668177e7fba97955853437e/vllm_metal/v1/model_runner.py#L1097-L1104)
- [`sampling_batch.py`: prefill sampling extracts only each segment's last row](https://github.com/vllm-project/vllm-metal/blob/5f4e50d1fcff299b9668177e7fba97955853437e/vllm_metal/v1/sampling_batch.py#L436-L499)
- [`structured_output.py`: grammar masking targets the same decode spans and prefill-final rows](https://github.com/vllm-project/vllm-metal/blob/5f4e50d1fcff299b9668177e7fba97955853437e/vllm_metal/v1/structured_output.py#L136-L187)

For a prefill-heavy batch this constructs a tensor of shape
`[1, total_scheduled_tokens, vocab_size]` although only
`num_decode_tokens + num_prefill_requests` rows are currently consumed.

This was already identified as a separate follow-up in
[#482](https://github.com/vllm-project/vllm-metal/issues/482):

> prefill logits are materialized for every position before row extraction
> repo-wide — both the target forward and the draft ingest do this; that's its
> own optimization issue

I could not find a standalone issue or pull request implementing that follow-up
as of 2026-07-20.

I am filing this RFC before writing an implementation to ask whether maintainers
consider the optimization worthwhile and whether the proposed boundary fits the
project. The design below is a starting point for discussion, not an assumed
decision.

Upstream vLLM uses a relevant high-level contract: it computes a
one-dimensional `logits_indices`, selects the required hidden-state rows, and
only then calls the model-owned logits projection:

```python
sample_hidden_states = hidden_states[logits_indices]
logits = self.model.compute_logits(sample_hidden_states)
```

See [`gpu_model_runner.py`](https://github.com/vllm-project/vllm/blob/e6d1310b2ac52e56e266450693af9de12a871a51/vllm/v1/worker/gpu_model_runner.py#L4428-L4429).

[...]

PR #590, opened 2026-08-09
Project logits only for sampled rows on the paged prefill path

robertomeroni:
## Summary

The paged text path projects the output head over **every** packed row, but sampling reads only one row per
prefill segment (`_build_paged_row_targets` takes `cu_seqlens[num_decode + j + 1] - 1`). For a 2048-token prompt
that is 2047 wasted rows of `[hidden_dim × vocab_size]` matmul and a `[2048 × 151936]` tensor materialized to
sample one row.

`ModelAdapter.target_forward` gains an optional `logits_indices`: prefill steps gather the sampled rows and run
the head on those alone; pure-decode steps pass `None` and are byte-for-byte unchanged. `_PagedForwardState`
carries a new `logits_cu_seqlens`, since logits are no longer row-aligned with the input — hidden-state consumers
keep `cu_seqlens` for the spec drafter. Implements #525.

**A model's head is not always reproducible outside its `__call__`, and getting that wrong yields silently wrong
logits, not a crash** (Cohere's `logit_scale`, Granite's `logits_scaling`). Rather than an `mlx_lm` allowlist,
`supports_selective_logits` probes once at load and demands bit-exact equality with the model's own output;
mismatches keep full logits and log why.

## Benchmarks

`vllm bench serve --model Qwen/Qwen3-0.6B --dataset-name random --random-input-len 2048 --random-output-len 16
--num-prompts 40`, M3 Pro, vllm 0.26.0. 5 iterations per arm, **interleaved** with a restart per arm, medians;
40/40 requests OK across all 10 runs.

| metric | before | after | delta |
|---|---|---|---|
| Output throughput | 20.26 tok/s | 23.46 tok/s | **+15.8%** |
| Mean TTFT | 17644.89 ms | 15375.80 ms | **−12.9%** |
| Mean TPOT | 571.96 ms | 485.36 ms | **−15.1%** |
| Duration | 31.58 s | 27.28 s | **−13.6%** |

CV ≤ 3.2% and the arms' ranges are **disjoint** on all four. The machine was not idle — hence the interleaving,
so load drift could not land in one arm — so absolute numbers are low for this hardware and the ratio is the
result. Isolated head projection at 2048 rows: 107.00 → 2.51 ms, 593.5 MiB avoided.

## Testing
```
pytest tests/test_model_adapter.py tests/test_v1_model_runner_generate.py tests/test_grammar_bitmask.py tests/test_decode_pipeline.py
224 passed
```

Covers index selection, `logits_cu_seqlens` compaction, selective and full projections agreeing on sampled rows,
the probe rejecting a model that scales its head output, and bitmask application under the compacted layout.
Prompt logprobs are unimplemented here (`prompt_logprobs_dict={}` is hardcoded), so dropping non-last prefill
rows regresses nothing. Also updates #588's new `_paged_state` helper and `fake_target_forward` stub (6 lines) to
match the file's eight others.

---

AI assistance was used in developing this change. The problem analysis, design, benchmarks and tests were
reviewed and verified by me, and I am responsible for all submitted changes.

### DIFF
--- a/vllm_metal/v1/model_runner.py
+++ b/vllm_metal/v1/model_runner.py
@@ -158,6 +158,22 @@ def _create_request_generator(
     return generator
 
 
+def _compact_logits_cu_seqlens(
+    cu_seqlens: list[int],
+    *,
+    num_decode_segments: int,
+) -> list[int]:
+    """Return ``cu_seqlens`` rewritten for selectively projected logits.
+
+    The decode/spec prefix is projected unchanged, so its boundaries carry over
+    verbatim; each prefill segment collapses to its single last-token row.
+    """
+    compact = cu_seqlens[: num_decode_segments + 1]
+    for _ in cu_seqlens[num_decode_segments + 1 :]:
+        compact.append(compact[-1] + 1)
+    return compact
+
+
 @dataclass
 class RequestState:
     """State for an ongoing request with KV cache."""
@@ -293,6 +309,9 @@ class _PagedForwardState(NamedTuple):
     cu_seqlens: list[int]
     decode_segments: tuple[PagedDecodeSegment, ...]
     num_decode_tokens: int
+    # Boundaries in ``logits``; equals ``cu_seqlens`` unless the forward
+    # projected only the sampled rows.
+    logits_cu_seqlens: list[int]
     # ``{req_id: mrope_position_delta}`` from paged mm prefill;
     # ``_sample_paged_batch`` stashes each onto ``RequestState``.
     mm_prefill_deltas: dict[str, int]
@@ -396,6 +415,10 @@ class MetalModelRunner:
         # Async forward state: stashed by execute_model, consumed by
         # sample_tokens (mirrors upstream's execute_model_state pattern).
         self._execute_model_state: _PagedForwardState | None = None
+
+        # Resolved in load_model by probing the output head; False until then so
+        # a partially initialized runner keeps full logits.
+        self._selective_logits_supported: bool = False
 
         # Pipeline-parallel group (set by the worker when pipeline_parallel_size
         # > 1). None means single-stage: the forward and sampling paths run
@@ -578,6 +601,14 @@ class MetalModelRunner:
             dtype=self.kv_cache_dtype or mx.float16,
             max_position_embeddings=max_position_embeddings,
         )
+        # Probed here because it runs a cacheless one-token forward, which is
+        # only safe before a paged context is installed. PP and multimodal take
+        # other forward branches that never request selection.
+        self._selective_logits_supported = (
+            self.pp is None
+            and not self._is_vlm
+            and self._model_adapter.supports_selective_logits(self._forward_model)
+        )
         if self._is_pooling:
             self._model_lifecycle.install_pooling_backend()
 
@@ -684,13 +715,46 @@ class MetalModelRunner:
         *,
         cache: Any | None = None,
         collect_hidden_states: bool = False,
+        logits_indices: mx.array | None = None,
     ) -> TargetModelForwardOutput:
         return self._model_adapter.target_forward(
             self._forward_model,
             input_ids,
             cache=cache,
             collect_hidden_states=collect_hidden_states,
+            logits_indices=logits_indices,
         )
+
+    def _paged_logits_indices(
+        self,
+        cu_seqlens: list[int],
+        *,
+        num_decode_tokens: int,
+        num_decode_segments: int,
+    ) -> mx.array | None:
+        """Return the sampled rows of the packed sequence, or None for all rows.
+
+        The packed order is ``[decode/spec rows][prefill 0 rows]...``. Every
+        decode row is consumed (spec verification reads its whole span), but
+        prefill sampling only ever reads each segment's last row, so the head
+        need not project the rest.
+
+        ``None`` means every row is already needed — any pure-decode step — so
+        those keep the model's own ``__call__`` and pay no gather.
+        """
+        if not self._selective_logits_supported:
+            return None
+        num_rows = cu_seqlens[-1]
+        selected = [
+            *range(num_decode_tokens),
+            *(
+                cu_seqlens[num_decode_segments + i + 1] - 1
+                for i in range(len(cu_seqlens) - num_decode_segments - 1)
+            ),
+        ]
+        if len(selected) == num_rows:
+            return None
+        return mx.array(selected, dtype=mx.int32)
 
     def _target_input_embeddings(self, input_ids: mx.array) -> mx.array:
         return self._model_adapter.target_input_embeddings(
@@ -1189,7 +1253,16 @@ class MetalModelRunner:
         for pr in prefill_reqs:
             prefill_info.append((pr.block_ids, len(pr.token_ids), pr.start_pos))
 
+        # ---- build cu_seqlens for logit extraction ----
+        # Before the forward: the sampled-row selection derives from these.
+        cu_seqlens: list[int] = [0]
+        for segment in decode_segments:
+            cu_seqlens.append(cu_seqlens[-1] + segment.num_query_tokens)
+        for pr in prefill_reqs:
+            cu_seqlens.append(cu_seqlens[-1] + len(pr.token_ids))
+
         logits: mx.array | None = None
+        logits_indices: mx.array | None = None
         target_hidden_states: mx.array | None = None
         pooling_hidden_states: mx.array | None = None
         intermediate_hidden: mx.array | None = None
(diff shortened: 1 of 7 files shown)


WindChimeRan (commented):
Thanks for the contribution! The optimization is real.

One issue on LoRA compatibility: 

This breaks the LoRA path: `_lora.prepare_step()` creates one Punica index per packed token, while this path gathers only `logits_indices` before invoking `lm_head`. With the default `target_modules=None`, `lm_head` is wrapped, so a multi-token prefill reduced to one logits row still passes the full-token routing map to a one-row head input.

<details>
<summary>Reproduction</summary>

```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 vllm serve \
  mlx-community-staging/TinyLlama-1.1B-Chat-v1.0-mlx-4Bit \
  --enable-lora \
  --max-lora-rank 32 \
  --lora-modules tiny=zjudai/flowertune-general-nlp-lora-tinyllama-1.1b-chat-v1.0 \
  --max-model-len 128 \
  --max-num-batched-tokens 128
```

```bash
  -d '{"model":"tiny","prompt":"The quick brown fox jumps over the lazy dog.","max_tokens":1,"temperature":0}'
```

The request returns HTTP 500:

```text
ValueError: [reshape] Cannot reshape array of size 416000
into shape (1,1,32000).
```

</details>

Could you either make the LoRA routing follow logits_indices, or fall back to the full-logits path when LoRA is enabled?

e.g., MetalLoRARuntime.enabled already exists at `lora/runtime.py` . Fall back is the easy fix.

WindChimeRan:
Also found this optimization will change deterministic numerics. This is acceptable since it does not hurt the model intelligence, but we'll probably need to update the tests.

vllm_metal/v1/model_runner.py
```diff
@@ -617,14 +648,47 @@ def _target_forward(
...
+        """
+        if not self._selective_logits_supported:
+            return None
+        num_rows = cu_seqlens[-1]
+        selected = [
+            *range(num_decode_tokens),
+            *(
+                cu_seqlens[num_decode_segments + i + 1] - 1
+                for i in range(len(cu_seqlens) - num_decode_segments - 1)
+            ),
```
LxYuan0420: This still treats every prefill segment as a sampled row. Intermediate chunks should not sample, so the row policy needs to know final vs intermediate prefill instead of only looking at cu_seqlens.

vllm_metal/v1/model_runner.py
```diff
@@ -1459,7 +1541,7 @@ def _sample_paged_batch(
         prefill_result = sample_prefill_tokens(
```
LxYuan0420, line 1741: Same issue here: intermediate chunks still enter prefill sampling and the result is discarded later. Please pass only final prefill rows into sampling and keep the mapping explicit.

vllm_metal/v1/model_runner.py
```diff
@@ -160,6 +160,22 @@ def _create_request_generator(
...
+    """Return ``cu_seqlens`` rewritten for selectively projected logits.
+
+    The decode/spec prefix is projected unchanged, so its boundaries carry over
+    verbatim; each prefill segment collapses to its single last-token row.
+    """
+    compact = cu_seqlens[: num_decode_segments + 1]
+    for _ in cu_seqlens[num_decode_segments + 1 :]:
+        compact.append(compact[-1] + 1)
+    return compact
+
```
LxYuan0420: I’d avoid making this a standalone cu_seqlens helper. The hard part is not the math; it is the request meaning. Keeping this near paged batch construction will make the code easier to reason about and harder to misuse later.

LxYuan0420 (commented):
Can you also rerun one clean comparison with `--no-enable-prefix-caching`? Even if both branches use the same cache setting, prefix hits can shrink the prefill work this PR targets, so a no-prefix-cache run makes the selective-logits effect easier to isolate.

vllm_metal/v1/model_runner.py
robertomeroni: but I think the row policy can't act alone

with `cu_seqlens=[0,1,4,6]` and chunk A intermediate, skipping it yields 2 logits rows while `sample_prefill_tokens` still reads row 2 of 2 -> `IndexError`

vllm_metal/v1/model_runner.py
robertomeroni: thanks! I have updated with your suggestion

robertomeroni:
@LxYuan0420 
> Can you also rerun one clean comparison with `--no-enable-prefix-caching`? Even if both branches use the same cache setting, prefix hits can shrink the prefill work this PR targets, so a no-prefix-cache run makes the selective-logits effect easier to isolate.

I can't produce that one here (no served model on this box, and I've previously measure inconsistent values so I don't want to give you noise)

If you or CI can run the `--no-enable-prefix-caching` A/B I'll fold it into the body


---
*Edited*

nevermind, I managed to run them

Baseline 25b608a (parent of my first commit) vs 7c4570b. Qwen3-0.6B, prefix caching off on both arms and
confirmed off in each server log, random 1024-in/16-out, 24 prompts at concurrency 8, temperature 0,
servers restarted between arms, one discarded warmup then 5 measured runs per arm. 

Medians:

```
                     25b608a      7c4570b
total throughput   2337 tok/s   2668 tok/s   +14.1%
median TTFT           2052 ms      1642 ms   -20.0%
median TPOT           43.7 ms      39.9 ms    -8.7%
```

24/24 requests succeeded in every run and the throughput ranges do not overlap: baseline spans 2329-2381, treatment 2571-2737.

vllm_metal/v1/model_runner.py
robertomeroni: correct, though it's unrelated to this PR

main already samples the full prefill list and discards intermediates at :1548, and my diff touches 0 lines of sampling_batch.py

it's also not a pure cleanup: the seeded generator is shared across chunks, so dropping the discarded
sample changes sampled tokens for any seeded request spanning >1 chunk

I would do it as its own PR, with that behaviour change called out

robertomeroni:
@WindChimeRan 

thank you

I've taken the fallback you suggested. `MetalLoRARuntime.enabled` is now part of the load-time
`_selective_logits_supported` probe, so a LoRA-enabled server keeps the full-logits path and everyone
else keeps the optimization

On numerics: agreed. Which tests would you like updated? The ones asserting bit-exact logits, or is there a tolerance convention you'd prefer?

WindChimeRan:
@robertomeroni  No worry on the numeric tests. I'll probably open a new PR for it and maybe improve the CI.

WindChimeRan (commented):
> treats every prefill segment as a sampled row
I think this is a non-blocker for this PR. 

Sampling reads the selected logits as input, so after this PR, logits smaller --> sampling smaller. 

Sample-then-discard is fine and it's upstream vllm's choice too. (from my understanding, optimizing chunked prefilling is in low priority for upstream vllm because they have good async scheduling pipeline to hide the bubble, but I'm not sure if our vllm-metal has a good async ... )

WDYT? @LxYuan0420

WindChimeRan:
I pushed two small commits for code simplifications.

fdaluiso:
I found a profile-run case that complements this PR; details in #638, and I
have opened #640 for it, reusing this PR's `logits_indices` extension point
so the two changes can land without an API conflict.

On clean main at 211a1e4, `tools/hybrid_apc_parity_matrix.py --quick`
profiles a `(1, 8192)` dummy for Qwen/Qwen3.5-0.8B on the tested 16 GB
M-series Mac. In the prefix-caching-enabled child, profiling with full
logits reports 6.46 GB of overhead and startup fails with
`kv_budget=-1.81 GB`.

As an A/B, I kept the 8192-token dummy and changed only the profile path to
project logits for its final row. Cache-on overhead fell from 6.46 GB to
2.39 GB, a measured 4.07 GB reduction on this host; the final KV budget was
+2.26 GB, and the APC matrix passed 20/20 comparisons with zero mismatches
(`PARITY PASS`). The 4.07 GB figure is the measured delta for this A/B, not
a host-independent estimate.

There is also a coverage gap for this model: Qwen3.5-0.8B loads through
mlx-vlm, while `_selective_logits_supported` currently excludes `_is_vlm`.
The mlx-vlm cacheless dummy needs the transformer body and explicit text
`position_ids` rather than the public wrapper, which persists RoPE state,
so a follow-up would need to stay gated on the same bit-exact capability
probe.

#640 touches only the profile path and includes the mlx-vlm cacheless
body/`position_ids` case, gated on the same bit-exact capability probe and
retaining the full-forward fallback for unsupported models, pooling, and
pipeline parallelism. I am happy to rebase it onto this PR as it evolves,
or fold any overlapping adapter code into yours once this lands.

WindChimeRan (approved):

vllm_metal/v1/model_runner.py
WindChimeRan: Treating this as a non-blocking follow-up for #590. The PR preserves main’s current intermediate-chunk sampling/discard behavior and its seeded-generator advancement; changing the request mapping here would be a separate behavioral change. The selective projection itself remains correct because it retains one sampled row for every prefill segment.

WindChimeRan:
LGTM. Merged based on the clear reduction in prefill projection work and the measured end-to-end speedup.

Selective logits remain disabled for VLM-backed models. I’ll review the findings in #640 and follow up on that limitation separately.

Thanks for the optimization! @robertomeroni

PR #591, opened 2026-08-10
Fix prefix cache E2E spy target

LxYuan0420:
This PR is:

- To patch the `prepare_grouped` symbol used by the production model runner.
- To make the prefix-cache E2E fail clearly if the spy does not observe the priming pass.
- To keep the runtime implementation unchanged.

The old test patched `vllm_metal.attention.context.prepare_grouped`, but production imports that function into `vllm_metal.v1.model_runner`. That could make the test observe the wrong symbol.


ricky-chaoju (approved):
LGTM. Thanks for tightening this up.

PR #592, opened 2026-08-10
Run body-only forwards for intermediate prefill chunks

ricky-chaoju:
On a chunked prefill, every chunk except the last samples nothing: the step still runs the full forward including the lm_head projection over the whole chunk, then throws the logits away. This PR runs the transformer body only on those steps. The hidden states still go through the forward-output submit, so the KV and GDN cache writes evaluate exactly as before; there is just no vocab projection and no sampling sync.

The step qualifies only when every prefill entry is an intermediate chunk, there are no decode rows, and no drafter is installed (`propose()` runs unconditional per-step bookkeeping the skip would bypass). Body resolution lives on the model adapter next to `text_model`; a model whose structure cannot be resolved logs one warning and keeps the full forward. Reaching the sampling step without logits while any row must sample raises instead of guessing. This is flagless on purpose: on any step that produces output the path is structurally unchanged, and the fallback is the existing full forward.

Measured on my M1 Ultra, branch vs base (`VLLM_METAL_MEMORY_FRACTION=0.5 vllm serve <model> --no-enable-prefix-caching`, greedy, prefill tok/s):

| model | prompt | base |this pr |
|---|---:|---:|---:|
| Qwen3.6-35B-A3B-4bit | 16.2k tokens | 686.7 | 742.6 |
| gemma-4-26B-A4B-it-OptiQ-4bit | 6k tokens | 389.9 | 413.8 |
| gemma-4-e2b-it-4bit | 6k tokens | 1363.9 | 1556.8 |
| Qwen3-0.6B | 6k tokens | 3654.5 | 4018.8 |


vllm_metal/v1/model_adapter.py
```diff
@@ -351,6 +354,16 @@ def text_model(self, model: Any) -> Any:
...
 
+    def transformer_body(self, model: Any) -> Any | None:
+        """Return the transformer body (no lm_head) of *model*, or ``None``.
+
+        mlx_lm text models expose it at ``model.model``; conditional-
+        generation wrappers served text-only nest it under the
+        ``language_model`` sub-model that :meth:`text_model` resolves.
+        """
+        body = getattr(self.text_model(model), "model", None)
+        return body if callable(body) else None
```
LxYuan0420: This leaks raw model internals to the runner. `.model` is not a stable “body-only forward” contract across mlx-lm models. Make this an explicit adapter-owned intermediate forward with a typed return shape.

vllm_metal/v1/model_runner.py
```diff
@@ -1238,21 +1261,51 @@ def _start_paged_forward(
...
+                body = (
+                    self._transformer_body_for_intermediate()
+                    if intermediate_only
+                    else None
                 )
-                logits = target_output.logits
-                target_hidden_states = target_output.hidden_states
-                del target_output
+                if body is not None:
+                    intermediate_hidden = body(input_ids, cache=offset_caches)
```
LxYuan0420: The optimization is good, but this adds a second model execution path in the runner. Keep the runner thin: it should request zero logits, and the adapter should own how to run the model without projection.

tests/test_v1_model_runner_generate.py
```diff
@@ -2785,6 +2785,335 @@ def test_prefill_rows_in_deferred_submit_raise(self) -> None:
...
+        # submitted (forcing the KV/GDN cache writes through the lazy graph),
+        # and no logits exist for the sampling step.
+        assert captured["body_input_ids"] == [[5, 6]]
+        assert any(out is hidden for out in captured["submitted"])
+        state = runner._execute_model_state
+        assert state is not None
+        assert state.logits is None
+        assert state.target_hidden_states is None
+        assert state.cu_seqlens == [0, 2]
+
```
LxYuan0420, line 2873: This test mostly proves routing. It mocks the cache setup and submit boundary, so it does not prove real KV/GDN writes or the next final chunk. Add one real-model or runtime-backed multi-chunk parity test.

vllm_metal/v1/model_runner.py
```diff
@@ -593,6 +597,24 @@ def apply_pipeline_split(self, pp: PipelineGroup) -> None:
...
+        """
+        forward_model = self._forward_model
+        body = self._model_adapter.transformer_body(forward_model)
+        if body is None and not self._warned_no_intermediate_body:
+            self._warned_no_intermediate_body = True
+            logger.warning(
+                "Metal: could not resolve a transformer body on %s; "
+                "intermediate prefill chunks run the full lm_head forward.",
+                type(forward_model).__name__,
+            )
```
LxYuan0420: I would remove this warning state. Falling back to full forward is expected when the optional fast path is unavailable. Resolve the capability once, then silently use the full path when unsupported.

Aryan4443:
Thanks for this  the intermediate body-only path looks like a solid win.

I saw LxYuan's review notes (adapter-owned intermediate forward, thinner runner, multi-chunk parity test, silent fallback). Happy to help if useful  especially on the real multi-chunk KV/GDN parity coverage, or the adapter-owned forward cleanup  just say the word and I can open a PR against your branch or send a patch.

ricky-chaoju:
done.

vllm_metal/v1/model_runner.py
```diff
@@ -1401,7 +1436,23 @@ def _sample_paged_batch(
             )
             return batch, scheduler_output
 
-        assert logits is not None
+        if logits is None:
```
LxYuan0420: All-intermediate batches should skip sampling regardless of adapter support; otherwise unsupported models can advance seeded RNG on a discarded token. Split skip-sampling from skip-projection.

vllm_metal/v1/model_adapter.py
```diff
@@ -351,6 +370,42 @@ def text_model(self, model: Any) -> Any:
             return model.language_model
         return model
 
+    def supports_intermediate_forward(self, model: Any) -> bool:
```
LxYuan0420, line 375: A callable `.model` is not proof of a projection-free backbone. Reuse the existing backbone helper and keep this capability narrow.

tests/test_intermediate_prefill_parity_e2e.py
```diff
@@ -0,0 +1,43 @@
...
+from pathlib import Path
+
+import pytest
+
+TOOL = (
+    Path(__file__).resolve().parent.parent / "tools" / "intermediate_prefill_parity.py"
+)
+
+
+@pytest.mark.slow
```
LxYuan0420, line 28: Good proof, but CI skips it. Rebase and post `python tools/intermediate_prefill_parity.py --quick` from current main.

LxYuan0420 (changes requested):
Requesting changes: split skip-sampling from skip-projection, narrow the adapter capability, rebase, and share rerunnable baseline-vs-PR benchmark numbers.

vllm_metal/v1/model_runner.py
ricky-chaoju: Done. _PagedForwardState carries intermediate_only now and sampling keys off that instead of the missing logits, so unsupported models skip sampling too. Added a test that a seeded generator's state stays untouched with logits present.

vllm_metal/v1/model_adapter.py
ricky-chaoju: It goes through pooling's sequence_model on the text sub-model now.

tests/test_intermediate_prefill_parity_e2e.py
ricky-chaoju: Rebased onto current main (0.27.1). From this branch: python tools/intermediate_prefill_parity.py --quick
PARITY PASS: 3/3 prompts token-identical Full matrix is 9/9.
截圖 2026-08-13 晚上8 01 01

ricky-chaoju:
> Requesting changes: split skip-sampling from skip-projection, narrow the adapter capability, rebase, and share rerunnable baseline-vs-PR benchmark numbers.

Rerunnable numbers, M1 Ultra:
```bash
VLLM_METAL_MEMORY_FRACTION=0.5 vllm bench latency \
  --model mlx-community/Qwen3.6-35B-A3B-4bit --max-model-len 32768 \
  --no-enable-prefix-caching --batch-size 1 --input-len 16000 --output-len 8 \
  --num-iters 3 --num-iters-warmup 1
```
 main 23.95s, this branch 22.71s.

LxYuan0420 (approved):
Thanks for working on this!

PR #562, opened 2026-07-26
Support for Laguna family of models

tzachicohen:
This patch introduces support for [Laguna](https://huggingface.co/poolside/Laguna-XS-2.1) familly of models by Poolside.ai .
This patch is built upon [this](https://github.com/ml-explore/mlx-lm/pull/1334) mlx_lm PR. 
Tests conducted:

- ./scripts/lint.sh - run successfully
- ./scripts/test.sh - run  successfully.
- successfully running agentic sessions with the server initialized with:
`vllm serve     --model poolside/Laguna-XS-2.1-NVFP4-mlx     --tool-call-parser poolside_v1    --reasoning-parser poolside_v1     --enable-auto-tool-choice     --served-model-name laguna  --port 8888 --max_model_len 65536`

Architectural difference:

- Laguna models attention blocks are based on Qwen 3.x attention block with additional per head gating.

PR #595, opened 2026-08-11
Pin mlx-lm to a git SHA to unblock Laguna support (#562)

WindChimeRan:
## Why

mlx-lm's last PyPI release is **0.31.3 (2026-04-22)**, but development continues on `main`, so any architecture merged after that date never reaches PyPI. We depend on `mlx-lm>=0.31.3`, which resolves to that April release.

Laguna is the immediate blocker for #562. It landed in `e5baded8` (2026-07-26) and is absent from 0.31.3, so loading `model_type: laguna` raises:

```text
ValueError: Model type laguna not supported.
```

This PR pins mlx-lm to `254d153f` (2026-08-05), which includes Laguna. The pinned revision still reports `__version__ == "0.31.3"`, so the version string does not distinguish it from the PyPI release.

## Changes

- Pin mlx-lm to `254d153f`.
- Remove the local Gemma 4 KV-shared sanitize patch and its tests. The pinned upstream revision now handles these weights natively.
- Keep the Qwen3.5 FP8 sanitize patch, which is not yet handled upstream.

## RFC: Vendor unreleased model modules

This SHA pin solves the immediate gap, but it couples vllm-metal to the full unreleased state of mlx-lm. If mlx-lm releases continue to lag behind `main`, an alternative is to vendor only the model modules we need and register them under `mlx_lm.models.*` until they ship in a release.

[oMLX](https://github.com/jundot/omlx) already uses this pattern without forking mlx-lm or maintaining a parallel model registry. At `350dc08b`, its vendored `*model.py` files under `omlx/patches/` total 9,229 lines; the largest is [`deepseek_v4_model.py`](https://github.com/jundot/omlx/blob/350dc08b757225531fecd76c8d552ae761d08082/omlx/patches/deepseek_v4/deepseek_v4_model.py) at 2,345 lines.

Its [Laguna registration helper](https://github.com/jundot/omlx/blob/350dc08b757225531fecd76c8d552ae761d08082/omlx/patches/laguna/__init__.py#L30-L51) loads a local file under the qualified name `mlx_lm.models.laguna`. The [fallback path](https://github.com/jundot/omlx/blob/350dc08b757225531fecd76c8d552ae761d08082/omlx/patches/laguna/__init__.py#L91-L105) installs it only when that exact upstream module is missing, so it becomes a no-op once mlx-lm provides the module.

This would avoid pinning the entire dependency to an unreleased commit. The tradeoff is that vllm-metal would own review, testing, and synchronization for each vendored module.

## End-to-end validation

Compared `main` with PyPI `mlx-lm==0.31.3` against this PR with `mlx-lm@254d153f`, using real `vllm serve` runs with Metal paged attention on an Apple M5 Pro. Every benchmark run completed 100/100 requests successfully.

### Correctness

Eight fixed prompts per model at `temperature=0, top_p=1, seed=0` produced identical completion text, token strings, and finish reasons:

| Model | Identical prompts | Tokens compared |
|---|---:|---:|
| `mlx-community/gemma-4-e2b-it-bf16` | 8 / 8 | 388 |
| `Qwen/Qwen3-0.6B` | 8 / 8 | 768 |

The two mlx-lm revisions also produced identical 540-key weight sets through the real Gemma 4 load path.

The available Gemma 4 checkpoint does not contain the phantom K/V keys handled by the removed patch. That case was validated separately by replaying the deleted tests' nine vectors against upstream `Model.sanitize`.

### Performance

Sonnet 1024+128, 100 prompts, rate 10, concurrency 32. Results are the mean of two warm-cache runs 

| Model | Output tok/s, main → PR | TTFT, main → PR | TPOT, main → PR |
|---|---:|---:|---:|
| Gemma 4 E2B | 346.98 → 347.35 | 250.16 → 250.84 ms | 76.81 → 76.70 ms |
| Qwen3-0.6B | 669.36 → 681.91 | 133.70 → 132.57 ms | 38.21 → 37.31 ms |

No regression.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,10 @@ dependencies = [
     # a wheel built against one MLX is only ABI-safe against that exact MLX.
     # Rebuild and re-release the wheel on every MLX bump.
     "mlx==0.32.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
-    "mlx-lm>=0.31.3; platform_system == 'Darwin' and platform_machine == 'arm64'",
+    # Git SHA, not a version: architectures merged after the 0.31.3 release
+    # (2026-04-22) never reach PyPI, and Laguna is one of them. main still
+    # self-reports "0.31.3", so this looks like a no-op upgrade -- it is not.
+    "mlx-lm @ git+https://github.com/ml-explore/mlx-lm@254d153fdeb6f150edd4fc5a54f9828638481fa8 ; platform_system == 'Darwin' and platform_machine == 'arm64'",
     "llguidance>=1.7.0,<1.8.0; platform_machine == 'x86_64' or platform_machine == 'arm64' or platform_machine == 'aarch64' or platform_machine == 'ppc64le'",
     # Temporary cap: xgrammar's prebuilt wheels are compiled against a specific
     # apache-tvm-ffi C++ ABI but declare only `apache-tvm-ffi>=0.1.9`, so a
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,9 +34,9 @@ dependencies = [
     # a wheel built against one MLX is only ABI-safe against that exact MLX.
     # Rebuild and re-release the wheel on every MLX bump.
     "mlx==0.32.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
-    # Git SHA, not a version: architectures merged after the 0.31.3 release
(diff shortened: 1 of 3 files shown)


LxYuan0420 (approved):
Yes, I noticed the delayed release issue. We can go with this SHA-pinning approach, but it will take extra time and effort to check dependency compatibility for each revision.

PR #597, opened 2026-08-11
Trust resolved draft config for Gemma4 MTP assistant loading

LxYuan0420:
This PR is:

- To use vLLM’s resolved `draft_model_config` as the source of truth for Gemma4 MTP assistant loading.
- To remove fallback reads from the outer speculative config.
- To keep the assistant cache key tied to the draft model revision vLLM actually resolved.

vLLM 0.27 builds the draft `ModelConfig` before Metal sees the speculative config, so the Gemma4 assistant loader should trust that typed contract directly.

Local smoke prompts confirmed the Gemma4 target and MTP assistant load together and produce coherent chat outputs on Metal.

| Prompt | Output |
|---|---|
| `Write one short sentence about the Moon.` | `The Moon orbits the Earth, casting a silvery glow across the night sky.` |
| `Answer with one word: What color is grass?` | `Green` |
| `Complete this sentence naturally: Apple pie tastes` | `Here are several ways to complete the sentence, depending on the desired tone and focus: ...` |

### DIFF
--- a/tests/test_gemma4_mtp_assistant.py
+++ b/tests/test_gemma4_mtp_assistant.py
@@ -93,6 +93,7 @@ def _validate_assistant_config(
 def _speculative_config(*, hf_config: object | None = None) -> SimpleNamespace:
     draft_config = SimpleNamespace(
         model="/assistant",
+        revision=None,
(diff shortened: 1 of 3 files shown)


ricky-chaoju (approved):
LGTM. Thanks.

PR #594, opened 2026-08-10
Add dense BGE-M3 / XLM-RoBERTa embed via native Apache encoder

Aryan4443:
## Summary
- Load encoder embedding checkpoints (`XLMRobertaModel` / RoBERTa / BGE-M3) through a **native Apache-2.0 MLX XLM-RoBERTa backend** in `vllm_metal/v1/encoder/`.
- `EncoderEmbeddingAdapter` owns loading, segment-wise bidirectional forward, CLS pooling defaults, and paged-attention patch skipping (single ownership boundary).
- Serve dense `embed` with CLS + L2. Sparse `token_classify` is out of scope here.

Partial progress on #589 (dense path only). Sparse lexical weights are being pursued separately in #599.

## Review feedback addressed
1. **Structure:** encoder special-cases live behind one adapter (load / forward / pooling defaults / patch skip).
2. **License:** dropped GPLv3 `mlx-embeddings`; native in-tree Apache-2.0 XLM-R/RoBERTa backend instead (not added to package extras).

## Test plan
- [x] `pytest tests/test_encoder_embeddings.py tests/test_v1_pooling.py`
- [x] Native load smoke for `mlx-community/bge-m3-mlx-8bit`
- [x] Offline e2e `LLM(..., runner="pooling").embed(["hello metal"])` with tight memory limits → `dim=1024`, `backend: XLMRobertaModel`

Issue #589, opened 2026-08-09
[Feature] Support BGE-M3 sparse embeddings with token-level pooling


rocke2020:
BGE-M3 sparse embeddings need a lexical weight for each input token. vllm-metal currently supports sequence-level pooling, which discards that token dimension.

I tested `mlx-community/bge-m3-mlx-8bit` on an M4 Pro with vllm-metal `0.3.0.dev20260809035703` and vLLM `0.26.0`. vLLM resolves `BgeM3EmbeddingModel` with `token_classify`, but the Metal loader fails with:

```text
ValueError: Model type xlm-roberta not supported.
```

Could Metal V1 support:

- XLM-RoBERTa/BGE-M3 model loading
- upstream-compatible `token_classify` pooling
- a BGE-M3 smoke test comparing token IDs and lexical weights with upstream vLLM or FlagEmbedding

This would allow BGE-M3 to serve sparse lexical embeddings, not only dense embeddings.

Related: #361, #371, #391.

LxYuan0420:
I suspect this is because vllm-metal doesnt yet support XLM-RoBERTa loading or `token_classify` pooling. Would you like to tackle this and submit a PR?

LxYuan0420:
Closing this as #612 fixes it

PR #599, opened 2026-08-11
[Pooling] Add BGE-M3 sparse token classification

rocke2020:
TL;DR: This adds BGE-M3 sparse lexical weights through vLLM's existing `token_classify` pooling contract. It is stacked on #594 (native Apache-2.0 XLM-RoBERTa encoder), so the reviewed diff below is the sparse increment; once #594 merges this becomes sparse-only. **Update (2026-08-13): the sparse head now builds on the native Apache encoder under `vllm_metal/v1/encoder/` — no GPLv3 `mlx-embeddings` dependency.**

## What changed

The encoder adapter now owns one validated sparse head and the pooling path keeps token/request boundaries explicit.

- Load the official `BAAI/bge-m3` `sparse_linear.pt` from a pinned revision only for the exact `mlx-community/bge-m3-mlx-8bit` checkpoint with model-level `token_classify` enabled. The sparse head is attached to `NativeEncoderModel` via `EncoderEmbeddingAdapter.load(model_config=...)`.
- Project packed hidden states once, then slice per request, apply logit calibration and optional ReLU, remove matching boundary BOS/EOS rows, and return CPU tensors in request order.
- Advertise `token_classify` only after the head loads successfully; dense-only BGE/XLM-R loads do not fetch it.
- Reject partial, prefix-cached, or resumed token-classify prefills before LoRA/cache/request ownership changes.
- Document offline and `/pooling` usage, plus the distinction between model-level capability and request-level dispatch.

The dense-encoder base is unchanged from #594; this PR's increment is the `BgeM3SparseHead` + `pool_bge_m3_token_classification` + token-classify scheduling/validation wiring.

## Verification

The sparse logic (`BgeM3SparseHead` projection / BOS-EOS filtering / ReLU, `_load_bge_m3_sparse_head` revision-pinning, and the `NativeEncoderModel` adapter wiring through `EncoderEmbeddingAdapter.load(model_config=)`) was re-verified after porting onto the native encoder.

- `BgeM3SparseHead` math (`project_token_logits` / `filter_token_weights`) matches the earlier reference outputs; invalid-shape and non-finite-weight rejection pass.
- Adapter load with `task="token_classify"` fetches the sparse head (`supports_token_classify is True`); dense-only load does not fetch it; unrelated encoder models are rejected.
- `ruff`-level compile check passes on all touched files.

⚠️ The figures below were captured on the earlier `mlx-embeddings`-based head. After porting to the native encoder, the sparse pooling math is byte-identical (it operates purely on MLX hidden states, loader-agnostic), but the full numeric suite has not been re-run on this environment (vllm 0.27.0 unavailable on this macOS toolchain). I will re-confirm parity on the M4 Pro and update:

- `pytest tests/test_encoder_embeddings.py tests/test_model_lifecycle.py tests/test_v1_pooling.py -q`: previously 102 passed, 1 skipped.
- Pinned, cache-only real-model parity test: `Hi` tokenizes to `[0, 2673, 2]`; 8-bit lexical weight `0.2698793411` vs upstream full-precision reference `0.2671086192` (1.04% relative error).
- Real `LLM.encode` batch: empty input returns `[]`; `Hi` returns `[0.2698793411]`.
- Temporary server: `/health`, `/tokenize`, `/pooling` returned HTTP 200; `/pooling` preserved the empty/`Hi` batch order and shapes.

The source checkout has no prebuilt `_paged_ops` and this Mac has Command Line Tools rather than full Xcode, so the runtime smokes bypassed only the unused native-kernel warm-up; the engine reported zero encoder layers patched.

### DIFF
--- a/docs/supported_models.md
+++ b/docs/supported_models.md
@@ -36,6 +36,7 @@ validation guidance. The reranker requires Qwen3 sequence-classification
 | --- | --- | --- | --- |
 | Qwen3-Embedding | 🔵 | `pooling` / `embed` (paged) | `mlx-community/Qwen3-Embedding-0.6B-8bit` |
 | Qwen3-Reranker | 🔵 | `pooling` / `classify` (paged) | `mku64/Qwen3-Reranker-0.6B-mlx-8Bit` |
+| BGE-M3 / XLM-RoBERTa | 🔵 | `pooling` / dense `embed` only (`vllm-metal[embeddings]`) | `mlx-community/bge-m3-mlx-8bit` |
 
 ## Multimodal Language Models
 
--- a/docs/supported_models.md
+++ b/docs/supported_models.md
@@ -36,7 +36,7 @@ validation guidance. The reranker requires Qwen3 sequence-classification
 | --- | --- | --- | --- |
 | Qwen3-Embedding | 🔵 | `pooling` / `embed` (paged) | `mlx-community/Qwen3-Embedding-0.6B-8bit` |
 | Qwen3-Reranker | 🔵 | `pooling` / `classify` (paged) | `mku64/Qwen3-Reranker-0.6B-mlx-8Bit` |
-| BGE-M3 / XLM-RoBERTa | 🔵 | `pooling` / dense `embed` only (`vllm-metal[embeddings]`) | `mlx-community/bge-m3-mlx-8bit` |
+| BGE-M3 / XLM-RoBERTa | 🔵 | `pooling` / dense `embed`; BGE-M3 sparse `token_classify` (`vllm-metal[embeddings]`) | `mlx-community/bge-m3-mlx-8bit` |
 
 ## Multimodal Language Models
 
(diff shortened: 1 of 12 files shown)


LxYuan0420:
Holding review for now. This depends on #594, but it targets `main` and still includes dense-loader changes. Please wait for #594 first, then rebase this as a sparse-only PR.

Aryan4443:
Thanks for picking up the sparse `token_classify` follow-up ,this lines up well with the #589 series (your PR2 on top of our dense PR1).

Heads-up for the rebase: #594 no longer depends on GPLv3 `mlx-embeddings`. Per maintainer feedback it now uses an in-tree Apache-2.0 MLX XLM-RoBERTa backend under `vllm_metal/v1/encoder/`, with `EncoderEmbeddingAdapter` still owning load / forward / CLS defaults / patch-skip.

Once you're rebasing after #594, please stack the sparse head on that native encoder path (not `mlx-embeddings`). Happy to help if anything in the adapter surface is unclear.

LxYuan0420:
Closing this as well because it depends on #594’s foundation and is not a sparse-only follow-up. Sparse BGE-M3 token classification is still useful, but it should come after the encoder backend interface is settled and reopen as a small PR on top of that base.

PR #602, opened 2026-08-12
Trim obsolete Metal platform capability glue

LxYuan0420:
This PR is:

- To trust vLLM 0.27 platform defaults for device capability and quantization validation.
- To remove Metal’s fake SM80 device capability.
- To delete a test that only asserted the fake capability shape.

vLLM now returns `None` for non-CUDA device capability, and an empty `supported_quantization` list already means there is no platform-side quantization filter. The Metal overrides were redundant and the fake CUDA capability was misleading.

### DIFF
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -828,12 +828,6 @@ class TestMetalPlatform:
         finally:
             reset_config()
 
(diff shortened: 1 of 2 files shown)


ricky-chaoju (approved):
LGTM. thanks.

PR #606, opened 2026-08-13
Refactor pooling behind a decoder backend

LxYuan0420:
This PR is:

- To move the current decoder pooling path behind a small backend interface.
- To keep `model_runner.py` responsible for scheduling and output attachment.
- To move packed hidden-state forward and pooling into the decoder backend.
- To keep Qwen3 reranker/classify behavior model-owned.

Note:
This is a behavior-preserving refactor. The current pooling path mixes generic decoder embedding, Qwen3 reranker logic, validation, and runner output handling in one module. This creates a clean seam before adding encoder pooling, without changing the existing embedding behavior.

Follow-up plan:
- PR 2: Add the encoder pooling backend and no-KV lifecycle path.
- PR 3: Add native XLM-R dense embedding support.
- PR 4: Add BGE-M3 dense and sparse pooling on top of the encoder backend.


ricky-chaoju (approved):
LGTM. Ran embed and rerank on base and head with the two checkpoints from the docs, outputs bitwise identical both ways.

PR #608, opened 2026-08-14
Add encoder pooling backend foundation

LxYuan0420:
This PR is:

- To add the encoder pooling runtime path without using decoder KV cache or paged attention.
- To add a minimal native XLM-R/RoBERTa backbone as the foundation for BGE-M3 pooling.
- To keep decoder pooling on the existing generation loader while encoder pooling owns its model-family loader.
- To preserve vLLM’s configured tokenizer and resolved model config contracts.

Note:

This is the PR2 foundation for BGE-M3 support. It does not claim broad XLM-R/RoBERTa model support as a product feature. The local backbone exists because `mlx-lm` / `mlx-vlm` do not currently provide a license-compatible encoder implementation for this path.

Follow-up plan:

- PR3: Add BGE-M3 dense embedding behavior on top of this encoder backend.
- PR4: Add BGE-M3 sparse `token_classify` / lexical-weight output.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -39,6 +39,7 @@ def make_stub_runner(
     defaults: dict[str, Any] = {
         "vllm_config": SimpleNamespace(
             speculative_config=None,
+            lora_config=None,
             # The value MetalPlatform resolves for the in-process executor.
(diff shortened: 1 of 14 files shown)


ricky-chaoju (approved):
LGTM, thanks for working on this!

PR #613, opened 2026-08-16
Trim weak platform block-size tests

LxYuan0420:
This PR is:

- To remove platform block-size tests that only assert declarations or no-op state.
- To tighten non-hybrid coverage so it proves delegation to vLLM's base block-size update.
- To keep Metal block-size runtime behavior unchanged.


ricky-chaoju (approved):
LGTM

PR #619, opened 2026-08-16
Add split MLA decode path

rohash123:
### Summary

Implements the Phase 3 item from #360: long-context split/reduce decode for the opt-in absorbed MLA Metal path.

Before this change, the opt-in Metal MLA decode path had one kernel shape: each query/head group handled the whole cached context in one pass. This PR adds a second shape for long contexts. It splits the cached context into ranges, computes one partial attention output per range, and reduces those partial outputs into the final result.

MLX remains the default MLA path. This PR only changes behavior when both env vars are enabled:

- `VLLM_METAL_MLA_KERNEL=1`
- `VLLM_METAL_MLA_SPLIT_KV=1`

### Implementation

The one-pass Metal MLA path is still present. It is used when split decode is disabled or the split plan rejects the request.

For split decode, the C++ dispatcher builds a plan from the request shape. When the plan selects split, the first Metal pass computes attention over each context range and writes:

- one partial output
- one log-sum-exp value

The reduce pass combines those partial outputs into the same result shape as the one-pass kernel.

The split decision lives in `mla_split_plan_for_shape`. It is exposed to Python as `metal_mla_split_plan` so tests and benchmarks can assert which path was selected. The Python MLA wrapper only checks whether the request is eligible for the opt-in Metal MLA path; it does not skip the one-pass path when split is off.

### Review feedback addressed

| Feedback | Change in this PR |
|---|---|
| Prove 32K+ split Metal beats default MLX. | Added the benchmark below using a real absorbed-MLA GLM layer on an AWS M1 Ultra. Split Metal is faster than MLX for every 32K-128K row measured. |
| Include MLX vs one-pass Metal vs split Metal, with repeated latency numbers. | The benchmark table includes all three paths. Each number is the median of 5 trials, with 20 timed iterations per trial. |
| Provide batch size > 1 results. | Added a B=2 long-context GLM benchmark. Split Metal is faster than MLX for every 32K-128K row measured. Also added a B=4 check showing the planner rejects split and falls back to MLX. |
| Preserve the existing one-pass Metal path. | The Python wrapper still admits the opt-in Metal MLA path without requiring split. The C++ dispatcher chooses one-pass or split from the split plan. For GLM-style long-context decode with split requested, the wrapper only enters Metal when the planner selects split. |
| Make the split decision device-aware. | GLM split now sizes the target partition count from `min_decode_grid() / gate_grid`. Wider GPUs select the 60-64 partition plan; narrower GPUs reject GLM split instead of forcing a slow route. |
| Move hard-coded policy into a small helper. | Split selection now goes through `mla_split_plan_for_shape`, exposed to Python as `metal_mla_split_plan`. |
| Tests could pass even if split never runs. | Tests now call `metal_mla_split_plan` and assert the selected split plan before checking output parity. |
| Cover bf16, block16, and one mixed long/short batch. | Added bf16 split cases, block-size-16 coverage, and a mixed `[40960, 2048]` decode batch. |
| Remove the unreachable scratch-budget guard. | The old scratch-budget calculation is gone. The planner keeps the decode-grid and partition-count checks. |
| Update stale docs. | `docs/configuration.md` now documents `VLLM_METAL_MLA_SPLIT_KV` and 32K+ split/reduce eligibility. |
| Keep the existing MLA cache contract. | The PR uses the existing absorbed-MLA latent cache. It does not change upstream vLLM scheduling, cache layout, or prefill behavior. |

### Benchmark

The benchmark times the full MLA attention call, not just the Metal kernel. It includes the Python and MLX work around the kernel: query projection, cache write, attention, unembed, and output projection.

Device:

- AWS EC2 `mac2-m1ultra.metal`, Apple M1 Ultra, 128 GB unified memory, 64-core GPU
- macOS 15.7.8

Model and shape:

- `mlx-community/GLM-4.7-Flash-4bit-mxfp4`
- absorbed-MLA layer 0 through `MLAPagedAttentionWrapper`
- `H=20`, `kv_lora_rank=512`, `qk_rope_head_dim=64`, `qk_nope_head_dim=192`, `v_head_dim=256`
- `bf16`, batch 1, block size 32, one decode token
- 5 trials, 20 timed iterations per trial, median of trial medians

Command:

```bash
.venv-vllm-metal/bin/python tools/benchmark/mla_split_vs_mlx_wrapper.py \
  --model mlx-community/GLM-4.7-Flash-4bit-mxfp4 \
  --contexts 32768 40960 49152 57344 61440 65536 81920 98304 131072 \
  --dtype bf16 \
  --block-size 32 \
  --batches 1 \
  --trials 5 \
  --iters 20 \
  --warmup 6 \
  --csv-out /tmp/mla_glm_32k_128k_device_gate_m1ultra.csv
```

Split MLA decode benchmark

| ctx | parts | MLX ms | one-pass Metal ms | split Metal ms | split vs MLX | max abs diff |
|---:|---:|---:|---:|---:|---:|---:|
| 32768 | 64 | 1.208 | 4.503 | 1.121 | +7.8% | 0.03125 |
| 40960 | 64 | 1.347 | 5.546 | 1.246 | +8.1% | 0.03906 |
| 49152 | 64 | 1.505 | 6.580 | 1.438 | +4.7% | 0.02148 |
| 57344 | 64 | 1.681 | 7.628 | 1.577 | +6.6% | 0.01172 |
| 61440 | 60 | 1.744 | 8.131 | 1.678 | +3.9% | 0.04688 |
| 65536 | 64 | 1.817 | 8.675 | 1.695 | +7.1% | 0.04199 |
| 81920 | 64 | 2.070 | 10.724 | 1.951 | +6.1% | 0.01758 |
| 98304 | 64 | 2.377 | 12.769 | 2.202 | +7.9% | 0.02344 |
| 131072 | 64 | 2.884 | 16.902 | 2.735 | +5.5% | 0.03125 |

Multi-batch check, same device/model/settings, `batch=2`:

```bash
.venv-vllm-metal/bin/python tools/benchmark/mla_split_vs_mlx_wrapper.py \
  --model mlx-community/GLM-4.7-Flash-4bit-mxfp4 \
  --contexts 32768 40960 49152 57344 61440 65536 81920 98304 131072 \
  --dtype bf16 \
  --block-size 32 \
  --batches 2 \
  --trials 5 \
  --iters 20 \
  --warmup 6 \
  --csv-out /tmp/mla_glm_b2_32k_128k_m1ultra.csv
```

| ctx | batch | parts | MLX ms | split Metal ms | split vs MLX | max abs diff |
|---:|---:|---:|---:|---:|---:|---:|
| 32768 | 2 | 64 | 1.945 | 1.812 | +7.3% | 0.01025 |
| 40960 | 2 | 64 | 2.254 | 2.068 | +9.0% | 0.03125 |
| 49152 | 2 | 64 | 2.600 | 2.365 | +9.9% | 0.02344 |
| 57344 | 2 | 64 | 2.879 | 2.610 | +10.3% | 0.02344 |
| 61440 | 2 | 60 | 2.999 | 2.846 | +5.4% | 0.03516 |
| 65536 | 2 | 64 | 3.129 | 2.879 | +8.7% | 0.02832 |
| 81920 | 2 | 64 | 3.587 | 3.415 | +5.0% | 0.03125 |
| 98304 | 2 | 64 | 4.231 | 3.958 | +6.9% | 0.02539 |
| 131072 | 2 | 64 | 5.209 | 5.023 | +3.7% | 0.02734 |

For `batch=4`, the same H=20 shape has a wider normal decode launch
(`gate_grid=16` on this device), so the GLM split planner rejects split. The
selected path falls back to MLX and matches MLX output exactly:

| ctx | batch | selected route | MLX ms | selected path ms | selected vs MLX | max abs diff |
|---:|---:|:---|---:|---:|---:|---:|
| 32768 | 4 | MLX fallback | 3.511 | 3.511 | +0.0% | 0.00000 |
| 65536 | 4 | MLX fallback | 5.728 | 5.729 | -0.0% | 0.00000 |
| 131072 | 4 | MLX fallback | 9.876 | 9.881 | -0.0% | 0.00000 |

### Device-aware split gate

The GLM H=20 path does not always force split when `VLLM_METAL_MLA_SPLIT_KV=1`.
It first asks how many partitions this device would need to make the small decode
launch wide enough:

```text
gate_grid = (num_heads / heads_per_tg) * batch
target_partitions = ceil(min_decode_grid / gate_grid)
```

For the benchmarked GLM shape, `H=20`, `heads_per_tg=5`, and `batch=1`, so
`gate_grid=4`. The current GLM split policy requires a near-max partition plan
(`target_partitions >= 56`) before entering Metal split. In Mac terms, that
means the H=20/B=1 split path is aimed at Max/Ultra-width Apple GPUs and is
rejected on Pro/base-width GPUs where the measured split overhead is too high.

Examples:

- AWS M1 Ultra: `min_decode_grid=512`, H=20/B=1 GLM decode selects split with 60-64 partitions for 32K-128K.
- Local M5 Pro: `min_decode_grid=160`, the same H=20/B=1 GLM decode rejects split for 32K-128K, so split-requested wrapper runs the MLX fallback instead of forcing the slow split route.
- M5 Max / M4 Max class Macs with 32-40 GPU cores are expected to select split by this planner.
- M3 Ultra / M1/M2 Ultra class Macs are expected to select split by this planner.

### Tests

Local:

```bash
.venv-vllm-metal/bin/python -m pytest \
  tests/test_mla_kernel.py \
  tests/test_mla_paged_backend.py \
  tests/test_mla_materialized_prefill.py
```

Result: `105 passed in 3.95s`

AWS M1 Ultra:

```bash
.venv-vllm-metal/bin/python -m pytest \
  tests/test_mla_kernel.py \
  tests/test_mla_paged_backend.py
```

Result: `102 passed in 2.07s`

```bash
.venv-vllm-metal/bin/python -m pytest tests/test_mla_materialized_prefill.py
```

Result: `3 passed in 1.42s`

Lint:

```bash
scripts/lint.sh
```

Result: `Success: no issues found in 123 source files`

Not run: `scripts/test.sh`. It builds the wheel, runs serving smoke tests, and then runs the full non-slow pytest suite. For this restart I ran lint, the focused MLA correctness suite, and the 32K+ benchmark above.

### Scope

This PR does not make Metal MLA the default. The broader set of model shapes requested in #360 is still the right bar before changing the default route or support-status docs.

This PR also does not pull in unrelated MLX optimization work. The change is limited to the vllm-metal absorbed-MLA split/reduce decode path.


WindChimeRan (commented):
higher level review (I didn't check the code detail)

**UX**: we should provide the best kernel for users automatically. (i.e., good optimization flags should not be opt-in. It can be opt-out for developer comparison)

The two required environment variables expose an implementation choice that should be automatic. Upstream vLLM auto-selects an MLA backend from model/device capabilities, and the selected backend chooses its split count from runtime shape and hardware. 

**Eval**: please provide batch_size > 1 results, because vllm-metal is mainly designed for concurrent workflow.

rohash123:
Thanks, addressed the batch-size > 1 eval point.

I added a multi-batch section to the PR body and reran the real absorbed-MLA GLM wrapper benchmark on the same AWS M1 Ultra setup:

- `batch=2`, H=20, bf16, 32K-128K: split Metal is selected and is faster than MLX for every measured row (`+3.7%` to `+10.3%`).
- `batch=4`, same shape at 32K/64K/128K: the planner rejects split because the normal decode launch is already wider; the selected route falls back to MLX and matches MLX output exactly.

I also fixed the benchmark script so its split-plan query matches the wrapper's GLM routing for multi-batch cases (`heads_per_tg=5` for H=20 split), then rebased onto current `main`.

Latest local validation after the rebase:

```text
python -m pytest tests/test_mla_kernel.py tests/test_mla_paged_backend.py tests/test_mla_materialized_prefill.py
105 passed in 3.95s
scripts/lint.sh
Success: no issues found in 123 source files
```

PR #621, opened 2026-08-17
Compile stateless MLP blocks for decode

ricky-chaoju:
Profiling the decode step showed it is op-count-bound: the per-layer MLP/MoE chains cost ~10 ms of the ~17.5 ms step on Qwen3.6-35B, spread over hundreds of tiny kernels. The blocks are stateless, so this PR routes decode-shaped calls (up to 4 tokens) through per-module `mx.compile` traces. No custom kernel. Opt-in via `VLLM_METAL_COMPILED_MLP=1`.

M1 Ultra, kill switch off vs on:

|  | off | on |
|---|---:|---:|
| Qwen3.6-35B serve decode (tok/s) | 54.2 | 56.0–57.5 |
| `vllm bench latency --batch-size 1 --input-len 32 --output-len 256` | 7.41 s | 7.28 s |


tests/test_compiled_mlp_parity_e2e.py
```diff
@@ -0,0 +1,41 @@
...
+    os.environ.get("VLLM_METAL_E2E", "1") != "1",
+    reason="e2e disabled",
+)
+def test_compiled_mlp_dispatch_is_lossless_and_engages() -> None:
+    result = subprocess.run(
+        [sys.executable, str(TOOL), "--quick"],
+        capture_output=True,
+        text=True,
+        timeout=1800,
+    )
```
LxYuan0420: This spawns vLLM and may run/download Qwen3.5. Keep it as a tool with PR evidence, or make it default-skipped behind an explicit local model path.

vllm_metal/compiled_mlp.py
```diff
@@ -0,0 +1,197 @@
...
+            raise RuntimeError(
+                "mlx_vlm Qwen3_5MLP.__call__ signature changed "
+                f"({params}); update CompiledMLPBlocks' plain-call policy "
+                "before wrapping it."
+            )
+        return {
+            Qwen3NextSparseMoeBlock: cls._PLAIN_CALL,
+            Qwen3NextMLP: cls._PLAIN_CALL,
+            Qwen3_5MLP: cls._PLAIN_CALL_TARGET_VERIFY,
+        }
```
LxYuan0420, line 143: `Qwen3_5MoeSparseMoeBlock` exists but is not registered, so native mlx-vlm MoE can be only partially compiled. Add the outer MoE target or reject that layout.

vllm_metal/compiled_mlp.py
```diff
@@ -0,0 +1,197 @@
...
+                elif isinstance(child, nn.Module):
+                    walk(child)
+                elif isinstance(child, dict):
+                    for sub in child.values():
+                        walk_container(mod, sub)
+                elif isinstance(child, list):
+                    for sub in child:
+                        walk_container(mod, sub)
+
+        walk(model)
```
LxYuan0420: This hand-rolls module traversal. Use `named_modules()` plus `update_modules(tree_unflatten(...))`, like the LoRA wrapper already does.

vllm_metal/compiled_mlp.py
LxYuan0420, line 143: The `(args, kwargs)` call-policy table is hard to maintain. Use explicit wrappers for unary calls and `target_verify=False` calls.

LxYuan0420 (changes requested):
Requesting changes. The opt-in `mx.compile` idea is reasonable, but this needs cleaner MLX module replacement, full native MoE handling or explicit rejection, and the heavy parity check should stay as a tool/PR proof instead of an enabled pytest wrapper.

ricky-chaoju:
One thing I wanted to check with you: this starts at 0 for now. My plan is to let it get some serve test first, then flip it default-on in a follow-up, same pattern as the decode pipeline (#588 -> #611). WDYT?

LxYuan0420 (approved):
LGTM!

LxYuan0420:
> One thing I wanted to check with you: this starts at 0 for now. My plan is to let it get some serve test first, then flip it default-on in a follow-up, same pattern as the decode pipeline (#588 -> #611). WDYT?

Yes, Agree to start with `0` first.  Let’s merge it behind the flag first, then flip it default-on in a small follow-up after we have repeated benchmark and correctness proof across different models (basically more proof that this is working well with other models)

Amazing work Ricky!

PR #624, opened 2026-08-18
Support Hugging Face config sources for local GGUF weights

LxYuan0420:
This PR is:

- To support serving local `.gguf` weights with a Hugging Face config/tokenizer source.
- To resolve non-local GGUF config sources before building the MLX model skeleton.
- To keep existing local config-directory GGUF loading unchanged.

This enables the normal GGUF serve shape:

```bash
vllm serve /path/to/Qwen3-0.6B-Q8_0.gguf \
  --tokenizer Qwen/Qwen3-0.6B
```

Before this change, the GGUF loader treated `Qwen/Qwen3-0.6B` as a local directory and failed while looking for:

```text
Qwen/Qwen3-0.6B/config.json
```

After this change, the loader downloads only the config files needed to build the MLX model skeleton. Tokenizer loading continues through the existing MLX-LM tokenizer path.

Validation:

- `tests/test_gguf_loader.py`: 42 passed.
- Ruff, compileall, and `git diff --check` passed.
- Real smoke: local `Qwen3-0.6B-Q8_0.gguf` weights with `Qwen/Qwen3-0.6B` config/tokenizer loaded successfully.

### DIFF
--- a/vllm_metal/gguf/loader.py
+++ b/vllm_metal/gguf/loader.py
@@ -8,6 +8,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, cast
 
+from huggingface_hub import snapshot_download
+
 try:
     import gguf
 except ImportError as exc:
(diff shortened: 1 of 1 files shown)


WindChimeRan (approved):
Please consider adding a GGUF documentation page. GGUF weights must currently be local, and remote GGUF references such as repo_id:Q8_0 remain unsupported. Since this PR adds remote config/tokenizer support, this distinction may be confusing to users.

LxYuan0420:
Remote GGUF weight references like `repo_id:Q8_0` are out of scope here. I will handle that in a separate follow-up PR later so this one stays focused on local `.gguf` weights with remote config/tokenizer.

PR #626, opened 2026-08-18
Use loopback for the single-process rendezvous address

### LINKED ISSUE #625: Startup hangs indefinitely when a VPN is active: get_ip() selects the tunnel interface and gloo rendezvous never binds
## Summary

With a VPN active, `vllm serve` on the Metal backend **hangs forever during startup**. It emits only a repeating `WARNING`, never an error, and never exits — so it looks like a slow model load rather than a failure.

The cause is IP auto-detection, not anything Metal-specific: `get_ip()` resolves the local address by opening a UDP socket toward `8.8.8.8` and reading back `getsockname()`. Under a VPN that returns the **tunnel interface** address, which gloo then cannot bind for its local rendezvous.

Single-process serving on Apple Silicon (`world_size=1`) has no reason to rendezvous on a routable address at all, which is what makes this pure friction.

## Repro

1. Connect a VPN.
2. `vllm serve mlx-community/Qwen3.8-27B-8bit --port 11439 --host 127.0.0.1`

Observed — repeating indefinitely, no error, no exit:

```
The server socket on [::ffff:10.4.6.125]:63617 has timed out, will retry.
The server socket on [::ffff:10.4.6.125]:63617 has timed out, will retry.
The server socket on [::ffff:10.4.6.125]:63617 has timed out, will retry.
```

`10.4.6.125` is the VPN interface. With the VPN off, the same command starts normally and logs:

```
world_size=1 rank=0 local_rank=0 distributed_init_method=tcp://192.168.12.165:65334 backend=gloo
```

## Workaround

```bash
VLLM_HOST_IP=127.0.0.1 vllm serve ...
```

Startup then proceeds immediately and logs `distributed_init_method=tcp://127.0.0.1:49549`. This works with the VPN still connected, because `VLLM_HOST_IP` short-circuits the lookup before the `8.8.8.8` probe.

## Source

`vllm/utils/network_utils.py`:

```python
def get_ip() -> str:
    host_ip = envs.VLLM_HOST_IP
    ...
    if host_ip:
        return host_ip
    # IP is not set, try to get it from the network interface
    try:
        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
            s.connect(("8.8.8.8", 80))  # Doesn't need to be reachable
            return s.getsockname()[0]
```

The comment "Doesn't need to be reachable" is true for route selection but the *returned address* still has to be bindable, which the VPN interface is not in this configuration.

## Suggested fix

Any of these would resolve it; the first seems most in the spirit of the plugin:

1. **Default to loopback on the Metal platform when `world_size == 1`.** There is no multi-node case to serve here, so the routable address buys nothing.
2. **Fail fast** — bound the rendezvous retry and raise with the selected address named, instead of warning in a loop forever.
3. **Document `VLLM_HOST_IP`** in the install/troubleshooting docs as the fix for VPN and multi-interface hosts.

Happy to open a PR for (1) or (2) if you have a preference.

## Environment

- `vllm-metal`: `0.3.0.dev20260817081527`
- `vllm`: `0.27.1`
- macOS 26.5.2, arm64, Apple M4 Max
- Python 3.12, installed via the repo's `install.sh`

## Note

This is distinct from #58 (closed) — that one is about reaching the server's own listening address after startup; this one prevents startup from completing at all.

evemcgivern:
Fixes #625.

## Problem

`get_ip()` resolves the local address by opening a UDP socket toward `8.8.8.8` and returning `getsockname()`. With a VPN active that yields the **tunnel interface**, which gloo cannot bind for its local rendezvous. Startup then retries forever, emitting only:

```
The server socket on [::ffff:10.4.6.40]:52437 has timed out, will retry.
```

No error, no exit — so it reads as a slow model load rather than a failure.

## Change

A single-process engine has no peer to reach, so the rendezvous address only has to be bindable locally. This wraps `UniProcExecutor._distributed_args` to rewrite the host to `127.0.0.1`, preserving the chosen port.

The wrapper is installed through a **post-import hook** on `vllm.v1.executor.uniproc_executor`. Three simpler placements were tried first and each provably fails:

1. **Directly at `apply_compat_patches()`** — plugin registration runs while vLLM is partially initialized, so the import raises and the patch silently skips:
   ```
   cannot import name 'is_quantized_kv_cache' from partially initialized module
   'vllm.utils.torch_utils' (most likely due to a circular import)
   ```
2. **Re-ensuring in `MetalPlatform.check_and_update_config`** — that runs in the **APIServer** process, but the executor is constructed in the **EngineCore** subprocess, which does not re-run it. Confirmed by logging the install site: `(APIServer pid=15257) ... patch INSTALLED`, while the rendezvous was computed under `(EngineCore ...)`.
3. **Patching `network_utils.get_ip`** — it is called **outside** any `set_current_vllm_config()` context, so the guard conditions cannot be evaluated there:
   ```
   Current vLLM config is not set. This typically means get_current_vllm_config()
   was called outside of a set_current_vllm_config() context
   ```
   The config is only reachable as `self.vllm_config` inside `_distributed_args`.

The hook fires in whichever process imports the module, before the class is used, and keeps `self.vllm_config` in scope for the guards. If the module is already imported, it patches directly and installs no `sys.meta_path` entry.

## Scope

Per the guidance on #625, the rewrite applies only when **all** of these hold:

- the executor is **exactly** `UniProcExecutor`
- `world_size == 1`
- executor backend is `uni` or unset
- `VLLM_HOST_IP` is unset — an explicit address always wins

Nothing else can reach it:

| Path | Why unaffected |
|---|---|
| `ExecutorWithExternalLauncher` (torchrun / multi-node) | subclasses `UniProcExecutor` but **overrides** `_distributed_args` to return `env://` — never reaches this code |
| Ray | builds its init method in `ray_executor.py` |
| Multiproc (PP / DP / TP) | builds its init method in `multiproc_executor.py` |

Ray, pipeline-parallel, data-parallel and multi-node behaviour are unchanged.

## Verification — end to end, against a live VPN

Host on a tunnel interface (`inet 10.4.6.40`), model `HuggingFaceTB/SmolLM2-135M-Instruct`:

| Condition | `distributed_init_method` | Result |
|---|---|---|
| Unpatched | `tcp://10.4.6.40:52437` | retry loop — **#625 reproduced on demand** |
| **Patched** | **`tcp://127.0.0.1:58367`** | **clean start, 0 retry lines** |
| Patched + explicit `VLLM_HOST_IP=10.4.6.40` | `tcp://10.4.6.40:59191` | operator's address honored, not overridden |

**Control** (before enabling the VPN): unpatched starts cleanly on the LAN address, `tcp://192.168.12.165:49892`, 0 retries — so the hang is attributable to the VPN rather than the model, the port, or the build.

## Tests

`tests/test_uniproc_loopback_rendezvous.py` — 9 cases: the rewrite and port preservation, explicit `VLLM_HOST_IP` passthrough, each excluded backend (`ray`, `mp`, `external_launcher`) and `world_size > 1`, the `ExecutorWithExternalLauncher` subclass, idempotency, and the already-imported path installing no finder.

```
9 passed
```

## Checks

- `ruff check` — clean
- `ruff format --check` — clean
- `mypy vllm_metal/compat.py` — reports 4 errors, **all pre-existing and unrelated to this PR**. They are in `vllm_metal/gguf/mlx_native.py` (`GGMLQuantizationType` "not valid as a type", lines 81 and 145), reached via import following. Verified identical on an unmodified checkout at `4edfe6a` by stashing this change and re-running. This PR touches neither that file nor its imports.

Environment: `vllm-metal 0.3.0.dev20260817081527`, `vllm 0.27.1`, macOS 26.5.2 arm64, Apple M4 Max.

### DIFF
--- a/tests/test_uniproc_loopback_rendezvous.py
+++ b/tests/test_uniproc_loopback_rendezvous.py
@@ -1,0 +1,123 @@
+"""Tests for the single-process loopback rendezvous patch (vllm-metal#625)."""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from typing import Any
+
+import pytest
+
+from vllm_metal.compat import (
+    _install_uniproc_loopback_rendezvous,
+    ensure_vllm_uniproc_loopback_rendezvous_patch,
+)
+
+uniproc_executor = pytest.importorskip("vllm.v1.executor.uniproc_executor")
+
+
+ORIGINAL_URI = "tcp://192.168.12.165:65334"
+
+
+def _make_executor(
+    cls: type,
+    *,
+    world_size: int = 1,
+    backend: str | None = "uni",
+) -> Any:
+    """A stand-in carrying only the attributes the patch inspects."""
+    obj = object.__new__(cls)
+    obj.vllm_config = SimpleNamespace(
+        parallel_config=SimpleNamespace(
+            world_size=world_size,
+            distributed_executor_backend=backend,
+        )
+    )
+    return obj
+
+
+@pytest.fixture
+def patched(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Apply the patch over a stub ``_distributed_args`` returning a routable IP."""
+    monkeypatch.setattr(
+        uniproc_executor.UniProcExecutor,
+        "_distributed_args",
+        lambda self: (ORIGINAL_URI, 0, 0),
+        raising=False,
+    )
+    monkeypatch.setattr(
+        uniproc_executor, "_vllm_metal_uniproc_loopback_patched", False, raising=False
+    )
+    monkeypatch.delenv("VLLM_HOST_IP", raising=False)
+    _install_uniproc_loopback_rendezvous(uniproc_executor)
+
+
+def _call(obj: Any) -> tuple[str, int, int]:
+    return uniproc_executor.UniProcExecutor._distributed_args(obj)
+
+
+def test_rewrites_to_loopback_preserving_port(patched: None) -> None:
+    method, rank, local_rank = _call(_make_executor(uniproc_executor.UniProcExecutor))
+    assert method == "tcp://127.0.0.1:65334"
+    assert (rank, local_rank) == (0, 0)
+
+
+def test_explicit_vllm_host_ip_is_preserved(
+    patched: None, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    monkeypatch.setenv("VLLM_HOST_IP", "10.0.0.5")
+    method, _, _ = _call(_make_executor(uniproc_executor.UniProcExecutor))
+    assert method == ORIGINAL_URI
+
+
+@pytest.mark.parametrize(
+    ("world_size", "backend"),
+    [(2, "uni"), (1, "ray"), (1, "mp"), (1, "external_launcher")],
+)
+def test_untouched_outside_single_process_uni(
+    patched: None, world_size: int, backend: str
+) -> None:
+    obj = _make_executor(
+        uniproc_executor.UniProcExecutor, world_size=world_size, backend=backend
+    )
+    assert _call(obj)[0] == ORIGINAL_URI
+
+
+def test_external_launcher_subclass_untouched(patched: None) -> None:
+    """``ExecutorWithExternalLauncher`` must keep its own ``env://`` rendezvous."""
+    obj = _make_executor(uniproc_executor.ExecutorWithExternalLauncher)
+    assert _call(obj)[0] == ORIGINAL_URI
+
+
+def test_patch_is_idempotent(patched: None) -> None:
+    _install_uniproc_loopback_rendezvous(uniproc_executor)
+    _install_uniproc_loopback_rendezvous(uniproc_executor)
+    assert _call(_make_executor(uniproc_executor.UniProcExecutor))[0] == (
+        "tcp://127.0.0.1:65334"
+    )
+
+
+def test_hook_installs_when_module_already_imported(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """With the module already in sys.modules, patch it directly rather than
+    installing a redundant meta_path finder."""
+    import sys
+
+    monkeypatch.setattr(
+        uniproc_executor.UniProcExecutor,
+        "_distributed_args",
+        lambda self: (ORIGINAL_URI, 0, 0),
+        raising=False,
+    )
+    monkeypatch.setattr(
+        uniproc_executor, "_vllm_metal_uniproc_loopback_patched", False, raising=False
+    )
+    monkeypatch.delenv("VLLM_HOST_IP", raising=False)
+    before = len(sys.meta_path)
+
+    ensure_vllm_uniproc_loopback_rendezvous_patch()
+
+    assert len(sys.meta_path) == before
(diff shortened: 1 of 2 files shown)


evemcgivern:
Updated — the first version of this PR **did not actually work**, and finding out why changed the implementation. Verified end-to-end against a live VPN this time rather than reasoning from the mechanism.

## Why the direct approach fails

Two constraints, neither obvious until instrumented:

1. **Registration is too early.** `register()` → `apply_compat_patches()` runs while vLLM is partially initialized, so `from vllm.v1.executor import uniproc_executor` raises there:
   ```
   cannot import name 'is_quantized_kv_cache' from partially initialized module
   'vllm.utils.torch_utils' (most likely due to a circular import)
   ```
   The patch silently skipped.

2. **The executor is built in the EngineCore subprocess**, which does not re-run `check_and_update_config`. Re-ensuring there installs the patch in the **APIServer** process only — confirmed by logging the install site: `(APIServer pid=15257) ... patch INSTALLED`, while the rendezvous was still computed in `(EngineCore ...)`.

I also tried patching `network_utils.get_ip` directly. That cannot work: it is called **outside** any `set_current_vllm_config()` context, so the guard conditions are unevaluable there —
```
Current vLLM config is not set. This typically means get_current_vllm_config()
was called outside of a set_current_vllm_config() context
```
The config is only reachable as `self.vllm_config` inside `_distributed_args`.

## What this version does

Installs the wrapper through a **post-import hook** on `vllm.v1.executor.uniproc_executor`, so it lands in whichever process imports the module, before the class is used — and keeps full access to `self.vllm_config` for the guards. If the module is already imported, it patches directly and installs no finder.

Guards and scope are unchanged from the original description.

## End-to-end verification (VPN active, `inet 10.4.6.40` on a tunnel interface)

| Condition | `distributed_init_method` | Result |
|---|---|---|
| Unpatched | `tcp://10.4.6.40:52437` | retry loop — **#625 reproduced on demand** |
| **Patched** | **`tcp://127.0.0.1:58367`** | **clean start, 0 retry lines** |
| Patched + explicit `VLLM_HOST_IP=10.4.6.40` | `tcp://10.4.6.40:59191` | operator's address honored, not overridden |

Control, before enabling the VPN: unpatched starts cleanly on the LAN address (`tcp://192.168.12.165:49892`, 0 retries) — so the hang is attributable to the VPN, not the model or the port.

Tests are now 9 cases (added one asserting the already-imported path patches directly without adding a `sys.meta_path` entry). `ruff check` and `ruff format --check` clean. The mypy note from the original description still applies unchanged — those 4 errors are pre-existing in `vllm_metal/gguf/mlx_native.py`.

LxYuan0420 (commented):
Thanks for the repro and VPN validation. I agree the issue is real, but this is a high-maintenance fix for a narrow VPN footgun.

The scope I suggested was much smaller: default single-process path, `distributed_executor_backend=uni`, `world_size=1`, and `VLLM_HOST_IP` unset. This PR goes further by adding a global import hook, patching a private vLLM executor method, and testing hook machinery. That is too much surface area for this problem.

I’m going to close this PR in its current form. If you want to continue, please open a smaller PR for either a troubleshooting doc note, or a narrow platform-level change that sets `VLLM_HOST_IP=127.0.0.1` only for the default single-process Metal path while preserving explicit `VLLM_HOST_IP` and leaving Ray, PP, DP, and multi-node untouched.

PR #627, opened 2026-08-18
Correctly log dense KV cache size from previously allocated array nbytes

### LINKED ISSUE #628: Gemma 4 E2B and Gemma 4 E4B incorrectly log the allocated amount of bytes
Found this while serving Gemma 4 E4B on Metal with M2 Mac. Startup prints a `KV cache: … MB` line that does not match the planner or what was actually allocated.

## Issue

E2B/E4B unique cache layers are mixed width: sliding is 256, global is 512. The allocator already uses `kv_heads_per_layer` / `head_dim_per_layer`. `_log_dense_cache` instead multiplies every unique layer by the scalar `self.head_dim`, which is `max(head_dim, global_head_dim) = 512`.

So it bills every unique layer as if it were a global layer.

- E4B: 24 unique (20 sliding + 4 global) → log is **12/7** of the real bytes
- E2B: 15 unique (12 sliding + 3 global) → **5/3**

Block count does not matter; the ratio is fixed. Absolute MB will move with `--gpu-memory-utilization` and the warmup overhead, so I am not hanging an issue on one 3.5 GB figure.

The planner (`get_cache_block_size_bytes` / `_kv_layer_size_sum`) and the arrays themselves are fine. Serving size is right. The one line a user reads at startup is not.

This is the dense path. YOCO on E2B/E4B takes that path, so a normal `vllm serve` of those checkpoints prints the bad line. Default 12B/26B/31B use the layout logger (`layout.total_bytes`), which is already correct.

## Reproduction

Construct `MetalPagedKVCache` with E4B unique shapes (24 layers, 20×256 + 4×512, 2 KV heads, bf16, 3 blocks, block_size 16):

- allocated: **2,752,512** bytes
- current log product: **4,718,592** bytes
- log line: `KV cache: 4.7 MB …`

Same thing shows up on a real E4B serve: planner reports `N × 917504` bytes, the `KV cache:` line reports `N × 1572864`.

## Suggested fix
Found at #627.

Astrowebdeving:
Tl;DR: Gemma 4 E2B and E4B mix the 256-wide sliding window KV with 512-wide global KV. The manually calculated logger assumed every unique layer as 512-wide, so it overstated E4B by 12/7 and E2B by 5/3. Allocation was already right. The log now dynamically sums the allocated arrays' nbytes. No allocation changes made.

## Summary
Gemma 4 unique cache layers are not one width. Sliding layers store K/V at `head_dim=256`; global/full layers store `global_head_dim=512`. E4B YOCO keeps 24 unique layers (20 sliding + 4 full). E2B keeps 15 (12 + 3). `_allocate_dense_caches` already builds each layer from `kv_heads_per_layer` / `head_dim_per_layer`.

`_log_dense_cache` ignored those lists and multiplied `num_layers × blocks × 16 × num_kv_heads × self.head_dim × 2 × dtype`. `self.head_dim` is `max(256, 512) = 512` from `resolve_max_head_dim` (kernel dispatch, not sizing). Every unique layer was billed at 4096 bytes/token. Real E4B is `20×2048 + 4×4096` per token (12/7). Real E2B is 5/3. The planner's `get_cache_block_size_bytes()` / `_kv_layer_size_sum()` already matched allocation, so serving size was fine; only the startup `KV cache: … MB` line was wrong.

The logger now sums `key_caches` and `value_caches` `.nbytes` after allocate, so it follows the arrays instead of a second formula. The `dtype` argument is gone because nbytes already includes element size. `_log_layout_cache` still uses `layout.total_bytes`: those per-layer entries are views over shared slots, and summing nbytes would double-count. Default 12B/26B/31B take that layout path and were not on the bad line.

## Tests
- `test_heterogeneous_log_matches_allocation` — E4B-shaped 24 layers, 3 blocks: log 2,752,512 B, not the old 4,718,592 B product
- `test_uniform_log_matches_allocation` — uniform models still match the old product
- `test_layout_log_reports_physical_bytes` — layout path still logs `layout.total_bytes`, not the per-layer nbytes sum

No model or kernel change; no performance claim. CONTRIBUTING golden-token and benchmark checks do not apply.

Checked and fixed with AI assistance (GPT-5.6 Sol, Opus 5, Grok 4.6). I first raised the issue and reviewed the patch and take responsibility for it. Fixes #628.

### DIFF
--- a/tests/test_per_layer_kv_cache.py
+++ b/tests/test_per_layer_kv_cache.py
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 import logging
-import re
 from types import SimpleNamespace
 
 import mlx.core as mx
@@ -52,26 +51,6 @@ def config_from_vllm_groups(
         available = groups[0].kv_cache_spec.page_size_bytes * num_blocks * group_size
     return get_kv_cache_config_from_groups(
         vllm_config_for_kv_grouping(), groups, available
-    )
-
-
-def logged_kv_cache_mb(caplog: pytest.LogCaptureFixture) -> float:
-    """Return the MB figure from the single ``KV cache:`` startup log line."""
-    messages = [
-        record.message
-        for record in caplog.records
-        if record.message.startswith("KV cache:")
-    ]
-    assert len(messages) == 1, messages
-    match = re.search(r"KV cache: ([\d.]+) MB", messages[0])
-    assert match is not None, messages[0]
-    return float(match.group(1))
-
-
-def allocated_cache_bytes(cache: MetalPagedKVCache) -> int:
-    """Return the bytes held by a cache's per-layer key/value arrays."""
-    return sum(array.nbytes for array in cache.key_caches) + sum(
-        array.nbytes for array in cache.value_caches
     )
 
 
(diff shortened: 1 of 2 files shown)


tests/test_per_layer_kv_cache.py
```diff
@@ -155,6 +177,80 @@ def test_length_mismatch_raises(self) -> None:
             )
```
LxYuan0420, line 156: Can we keep this to one focused E4B-shaped regression that asserts the log reports `2.8 MB`, not the old `4.7 MB`? The layout path is untouched, and the extra helpers/tests make this logging-only fix heavier than needed.

LxYuan0420 (commented):
The fix looks correct. Please trim the tests to the one heterogeneous dense-cache regression; existing allocation/layout tests already cover the unchanged paths.

tests/test_per_layer_kv_cache.py
Astrowebdeving: With the second commit, I removed the extraneous tests and only have 1 focused E4B test now. Thanks!

LxYuan0420 (approved):
Thanks!

PR #629, opened 2026-08-19
[Bugfix] Pin torch sampling to CPU to stop random token corruption

### LINKED ISSUE #622: Sampling-path corruption: single garbage token injected into ~11% of completions when temperature > 0 (Metal backend, Apple Silicon)
### Summary

With any non-zero temperature, `vllm serve` on the Metal backend injects a
single stray token into roughly **1 in 9 completions** (11.1%, 5/45 in our
largest run; small-n, so treat the exact percentage as an order-of-magnitude
estimate, not a precise rate — see Evidence for the full run and two smaller
independent runs landing in the same 10–20% band). The
substitution can land in a tool-call argument value, in a tool-call
**function name** (producing a call to a function that does not exist), or
in free-text reasoning output. Greedy decoding (`temperature=0`) is clean
over 90 samples total across several separate runs on the same host.

This looks like a decode-path bug, not a tool-call-parser bug: it
reproduces before any parser sees the output (confirmed against the raw
`content` field with no `--tool-call-parser` configured at all), it hits
plain reasoning text as well as structured arguments, and it disappears
entirely at `temperature=0`.

**Impact:** any application relying on non-greedy sampling (i.e. following
this model family's own recommended operating point:
`temperature=0.7, top_p=0.8, top_k=20, repetition_penalty=1.05`) gets
silently corrupted tool calls at production scale. We only caught it
because our scoring harness checks tool-call arguments field-by-field
against an expected value; a fuzzier judge (or no judge at all) would have
let this through as "looks plausible."

### Environment

- `vllm`: `0.27.1+cpu` (reported by `/version` and `vllm --version`)
- `vllm-metal`: `0.3.0.dev20260816085229`
- Platform plugin: `metal` (`vllm.platform_plugins` -> `metal -> vllm_metal:register`)
- OS: macOS 26.5.1 (BuildVersion 25F80), arm64
- Hardware: Apple M3 Ultra, 96 GB unified memory
- Model: `mlx-community/Qwen3.8-27B-8bit` (MLX-quantized, served directly by
  vllm-metal — no conversion step)
- Server flags: `--enable-auto-tool-choice --tool-call-parser qwen3_xml
  --reasoning-parser qwen3 --max-num-seqs 1`
- No `--api-key` set (any bearer token accepted)

Full server command used for every run below:

```bash
vllm serve mlx-community/Qwen3.8-27B-8bit \
  --served-model-name qwen3.8-27b-8bit \
  --host 0.0.0.0 --port 8000 \
  --enable-auto-tool-choice \
  --tool-call-parser qwen3_xml \
  --reasoning-parser qwen3 \
  --max-num-seqs 1
```

### Cross-check that rules out model / quantization / harness

Same model weights, same prompts, same harness, two other inference
engines, zero corruption:

| Engine | Quantization | temperature=0.7, n=45 |
|---|---|---|
| **vllm-metal** | MLX 8bit | **40/45 clean (89%)** |
| rapid-mlx (MLX's own server) | MLX 8bit | 45/45 clean |
| llama.cpp (`llama-server`, `--jinja`) | GGUF Q8_0 | 45/45 clean |

Full numbers: `BEFUND-engine-sampling-campaign-0817.md` in the repo this
issue was drafted from.

### Reproduction (for an agent or script, not just a human)

Two parts: start the server, then run a scripted probe that fires N
requests and reports the corruption rate. Expect **~10–20% at
temperature=0.7**, **0% at temperature=0**.

**1. Install and start the server** (Apple Silicon host, call it `<HOST>`):

```bash
curl -fsSL https://raw.githubusercontent.com/vllm-project/vllm-metal/main/install.sh | bash
huggingface-cli download mlx-community/Qwen3.8-27B-8bit   # ~28 GB

~/.venv-vllm-metal/bin/vllm serve mlx-community/Qwen3.8-27B-8bit \
  --served-model-name qwen3.8-27b-8bit \
  --host 0.0.0.0 --port 8000 \
  --enable-auto-tool-choice \
  --tool-call-parser qwen3_xml \
  --reasoning-parser qwen3 \
  --max-num-seqs 1
# wait for "Application startup complete" / GET /v1/models to return 200
```

**2. Run this probe** (stdlib only, no dependencies — save as `probe.py` and
run `python3 probe.py http://<HOST>:8000 50`):

```python
#!/usr/bin/env python3
"""Fires N chat-completion requests with a tool schema at temperature 0.7,
flags any completion containing a non-ASCII stray token in a tool-call
argument, tool-call function name, or the reasoning/content text. Expects
a corruption rate near 0% at --sampling greedy (edit TEMPERATURE below to
0.0 for the control run) and roughly 10-20% at temperature 0.7."""
import json, sys, urllib.request

ENDPOINT = sys.argv[1] if len(sys.argv) > 1 else "http://localhost:8000"
N = int(sys.argv[2]) if len(sys.argv) > 2 else 50
TEMPERATURE = 0.7   # set to 0.0 for the greedy control run

PAYLOAD = {
    "model": "qwen3.8-27b-8bit",
    "messages": [
        {"role": "system", "content": "Du bist Hermes. Nutze die bereitgestellten Tools."},
        {"role": "user", "content": "Zeig mir die letzten 50 Logzeilen des Containers hermes."},
    ],
    "tools": [{
        "type": "function",
        "function": {
            "name": "docker_logs",
            "description": "Liest Logzeilen eines Docker-Containers.",
            "parameters": {
                "type": "object",
                "properties": {
                    "container": {"type": "string"},
                    "lines": {"type": "integer", "description": "Anzahl der Zeilen"},
                },
                "required": ["container", "lines"],
            },
        },
    }],
    "tool_choice": "auto",
    "temperature": TEMPERATURE,
    "top_p": 0.8,
    "top_k": 20,
    "repetition_penalty": 1.05,
    "max_tokens": 512,
}

VALID_FN = {"docker_logs"}

def has_stray_token(s: str) -> bool:
    # crude but effective: the injected tokens observed so far are CJK/Thai/
    # box-drawing characters that have no business in an English/German
    # numeric-argument completion (e.g. U+65E2 "既", U+7684 "的", U+0E1E "ฯ").
    return any(ord(c) > 0x2E80 for c in s)

bad = 0
for i in range(N):
    req = urllib.request.Request(
        f"{ENDPOINT}/v1/chat/completions",
        data=json.dumps(PAYLOAD).encode(),
        headers={"Content-Type": "application/json", "Authorization": "Bearer dummy"},
    )
    with urllib.request.urlopen(req, timeout=60) as resp:
        msg = json.load(resp)["choices"][0]["message"]
    tool_calls = msg.get("tool_calls") or []
    reasoning = msg.get("reasoning") or msg.get("content") or ""
    corrupt = has_stray_token(reasoning)
    for tc in tool_calls:
        fn = tc["function"]["name"]
        args = tc["function"]["arguments"]
        if fn not in VALID_FN or has_stray_token(args):
            corrupt = True
    marker = "CORRUPT" if corrupt else "ok"
    if corrupt:
        bad += 1
        print(f"[{i+1}/{N}] {marker}  tool_calls={tool_calls!r}")
    else:
        print(f"[{i+1}/{N}] {marker}")

print(f"\n{bad}/{N} corrupted ({100*bad/N:.1f}%) at temperature={TEMPERATURE}")
```

### Expected vs. actual

- **Expected:** at `temperature=0.7` (this model family's documented
  recommended operating point), 0/N completions contain stray tokens —
  same as at `temperature=0`.
- **Actual:** ~10–20% of completions (11.1% at n=45, our largest run so
  far) contain exactly one injected token, either in a tool-call argument
  value, a tool-call function name, or free-text reasoning.

### Evidence: three raw completions from our n=45 run

Server started exactly as in "Environment" above, request sampling
`{temperature: 0.7, top_p: 0.8, top_k: 20, repetition_penalty: 1.05}`.

**1. Argument value corrupted** (prompt: "Zeig mir die letzten 50
Logzeilen des Containers hermes.", tool `docker_logs(container, lines:
integer)`):

```json
{
  "tool_calls": [{
    "function": {
      "name": "docker_logs",
      "arguments": "{\"container\": \"hermes\", \"lines\": \"50\\n</既定\"}"
    }
  }],
  "reasoning": "The user is asking for the last 50 log lines of the container \"hermes\". I should use the docker_logs tool with container=\"hermes\" and lines=50.\n"
}
```

Note `lines` was supposed to be the integer `50`; instead it's the string
`"50\n</既定"` — `既定` (U+65E2 U+5B9A) has no plausible source in this
context (German/English prompt, no CJK anywhere in the conversation).
`reasoning` here is clean, so the corruption is localized to the tool-call
argument serialization, not a general text-quality problem in this sample.

**2. Reasoning text ALSO corrupted, not just structured output** (prompt:
"Suche alle Rechnungen vom 1. Juli 2026.", tool `search_invoices(date_iso:
string)`):

```json
{
  "tool_calls": [{
    "function": {
      "name": "search_invoices",
      "arguments": "{\"date_iso\": \"2026-的共同07-01\"}"
    }
  }],
  "reasoning": "The user wants to search for kolesterol invoices from July 1, 2026. Let's use the search_invoices tool with the date 2026-07-01.\n"
}
```

Two independent injections in the same completion: `的共同` (U+7684 U+5171
U+540C) inside the date argument, and the nonsense word "kolesterol"
spliced into otherwise-fluent English reasoning ("search for kolesterol
invoices" — not a real English word, no relation to the domain). This is
why we don't think this is specific to tool-call serialization: it also
corrupts plain generated prose in the same response.

**3. Function name itself corrupted, not just an argument** (prompt: "Wie
ist der aktuelle Zustand von light.kueche?", tools `hactl_get_state(entity_id)`
/ `hactl_call_service(service, entity_id)`):

```json
{
  "tool_calls": [{
    "function": {
      "name": "hactl_getcoles",
      "arguments": "{\"entity_id\": \"light.kueche\"}"
    }
  }],
  "reasoning": "The user is asking about the current state of light.kueche (kitchen light). I should use the hactl_get_state tool to read the state of that entity.\n"
}
```

`reasoning` correctly names `hactl_get_state` (the only tool that matches
the description). The emitted `tool_calls[0].function.name` is
`hactl_getcoles` — not `hactl_get_state`, not any other tool in the
request's `tools` array, not a hallucinated-but-plausible name. This is a
call to a function that does not exist; a client would get a hard failure,
not a wrong-but-parseable answer. We read this as the clearest sign that
the substitution happens at the token-sampling level, after the correct
plan was already formed (visible in `reasoning`), rather than being a
higher-level model mistake.

### Additional notes

- **Not a tool-call-parser artifact.** An earlier run with no
  `--tool-call-parser` configured at all showed the same single-token
  injection sitting in the raw `content` field, before any parsing:
  `<parameter=lines>\n50\n</既定\n</parameter>` instead of
  `<parameter=lines>\n50\n</parameter>`. Parser and client-side harness are
  both ruled out.
- **Not top_p/top_k/repetition_penalty specifically.** Isolated each
  parameter earlier (9 samples each): `temperature=0.7` alone already
  corrupts at ~1/9; adding `top_p`, `top_k`, or `repetition_penalty`
  individually stays in the same 1–2/9 range. It's `temperature > 0` in
  general, not one particular sampling knob.
- **Not (obviously) deterministic per-request.** Two early runs on
  identical requests came back byte-identical including the corruption,
  which briefly looked like a deterministic indexing bug — that did not
  hold up: it's the expected outcome of a low-probability event on short,
  strongly-peaked completions, not a fixed seed artifact.
- We're happy to run further probes against this host if useful — the
  model and harness that produced this report are still in place.

ricky-chaoju:
Fixes #622. With `temperature > 0`, about 10% of completions came back with one garbage CJK token dropped into otherwise normal text, while greedy was clean. The cause is the vLLM torch sampler running on MPS: torch's `exponential_()` on MPS sometimes returns exact zeros (I counted 873 in 29 billion draws on an M1 Ultra; CPU gives none), and the sampler computes `probs.div(q).argmax()`, so a zero draw becomes inf or NaN and argmax lands on a random vocab id. That also matches the `top_k=1` result in the issue thread, since masked positions turn into `0/0 = NaN`. In a synthetic harness, the zero draws and the bad tokens matched 1:1, and clamping just the zeros made every violation disappear, so this is the whole mechanism.

The fix pins the torch sampling path to CPU. `SamplingBatch` owns the device as a class constant and the device parameters are deleted from the sampling entry points, so nothing can route sampling back to MPS. Greedy stays on native MLX. Seeded outputs change because the CPU RNG stream differs from MPS. This also stops an EngineCore abort seen with Metal API validation enabled, where torch's MPS radix sort kernel trips an assertion on the first non-greedy request.

For evidence, replaying the reporter's request shape on `mlx-community/Qwen3.8-27B-4bit` at temperature 0.7, 60 completions per run: 6/60 corrupted before, 0/60 after. Sampling on CPU is also slightly faster since it skips the per-step MPS transfer. `vllm bench latency --batch-size 1 --input-len 32 --output-len 128` on the same model:

| | avg latency |
|---|---:|
| main | 6.14 s |
| this PR | 5.86 s |

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -7,7 +7,6 @@ from types import SimpleNamespace
 from typing import Any
 
 import mlx.core as mx
-import torch
 
 import vllm_metal.v1.model_runner as mr
 from vllm_metal.v1.cache_policy import ModelCachePolicy
@@ -78,7 +77,6 @@ def make_stub_runner(
         "head_dim_per_layer": None,
         "sliding_window_per_layer": None,
         "use_async_scheduling": True,
-        "device": torch.device("cpu"),
         "_sampler": None,
         "_structured_output_applier": MetalStructuredOutputApplier(),
         "_lora": MetalLoRARuntime(),
--- a/tests/test_paged_prefix_caching.py
+++ b/tests/test_paged_prefix_caching.py
@@ -247,7 +247,7 @@ class TestSamplingMetadataWithPenalties:
 
         captured_batches: list = []
 
-        def spy_sample(logits_2d, batch, sampler, device):
+        def spy_sample(logits_2d, batch, sampler):
             captured_batches.append(batch)
             return _SamplingResult([99])
 
--- a/tests/test_sampling.py
+++ b/tests/test_sampling.py
@@ -283,7 +283,7 @@ class TestV1SeededSamplingGenerator:
         )
 
         sp = SamplingParams(temperature=1.0, seed=123)
-        generator = _create_request_generator(runner.device, sp)
+        generator = _create_request_generator(sp)
         assert generator is not None
 
         state = RequestState(
@@ -304,6 +304,69 @@ class TestV1SeededSamplingGenerator:
         assert not torch.equal(after_second, after_first)
 
 
+class TestSamplerDevicePolicy:
+    """Regression tests for #622: torch sampling must never run on MPS.
+
+    ``exponential_()`` on MPS can return exact zeros, which the vLLM
+    sampler's Gumbel-style division turns into inf/NaN rows whose argmax
+    lands on an arbitrary vocab id.
+    """
+
+    def test_sampling_metadata_tensors_live_on_cpu(self) -> None:
+        batch = SamplingBatch(
+            [
+                SamplingParams(temperature=0.8, top_k=20, top_p=0.9),
+                SamplingParams(temperature=1.0, presence_penalty=0.5),
+                SamplingParams(temperature=0.7, allowed_token_ids=[1, 2]),
+            ],
+            [[1, 2], [3], [4]],
+            [[], [], []],
+            vocab_size=VOCAB_SIZE,
+        )
+
+        metadata = batch.make_sampling_metadata()
+
+        assert metadata.temperature is not None
+        assert metadata.temperature.device.type == "cpu"
+        assert metadata.top_k is not None
+        assert metadata.top_k.device.type == "cpu"
+        assert metadata.top_p is not None
+        assert metadata.top_p.device.type == "cpu"
+        assert metadata.prompt_token_ids is not None
+        assert metadata.prompt_token_ids.device.type == "cpu"
+        assert metadata.frequency_penalties.device.type == "cpu"
+        assert metadata.presence_penalties.device.type == "cpu"
+        assert metadata.repetition_penalties.device.type == "cpu"
+        assert metadata.allowed_token_ids_mask is not None
+        assert metadata.allowed_token_ids_mask.device.type == "cpu"
+
+    def test_seeded_generator_lives_on_cpu(self) -> None:
+        generator = _create_request_generator(SamplingParams(temperature=1.0, seed=7))
+
+        assert generator is not None
+        assert generator.device.type == "cpu"
+
+    def test_bridged_logits_reach_sampler_on_cpu(self) -> None:
+        logits = mx.array([[0.0, 1.0, 4.0, 2.0]], dtype=mx.float32)
+        batch = SamplingBatch(
+            [SamplingParams(temperature=0.8, top_k=2)],
+            [[1]],
+            [[]],
+            vocab_size=4,
+        )
+        seen_devices: list[torch.device] = []
+
+        class SpySampler(Sampler):
+            def forward(self, logits_torch, sampling_metadata):  # noqa: ANN001
+                seen_devices.append(logits_torch.device)
+                return super().forward(logits_torch, sampling_metadata)
+
+        result = sample_from_logits(logits, batch, SpySampler())
+
+        assert [d.type for d in seen_devices] == ["cpu"]
+        assert 0 <= result.token_ids[0] < 4
+
+
 class TestV1SamplingBatch:
     @staticmethod
     def _batch(params_list: list[SamplingParams]) -> SamplingBatch:
@@ -312,7 +375,6 @@ class TestV1SamplingBatch:
             [[1]] * len(params_list),
             [[]] * len(params_list),
             vocab_size=VOCAB_SIZE,
-            device=torch.device("cpu"),
         )
 
     def test_prefill_requests_share_one_sampler_batch(self, monkeypatch) -> None:
@@ -344,7 +406,6 @@ class TestV1SamplingBatch:
             cu_seqlens=[0, 2, 4, 6],
             num_decode=1,
             sampler=Sampler(),
-            device=torch.device("cpu"),
             vocab_size=4,
         )
 
@@ -368,7 +429,6 @@ class TestV1SamplingBatch:
             cu_seqlens=[0],
             num_decode=0,
             sampler=Sampler(),
-            device=torch.device("cpu"),
             vocab_size=4,
         )
 
@@ -416,9 +476,8 @@ class TestV1SamplingBatch:
             [[1, 2, 3]],
             [[]],
             vocab_size=4,
-            device=torch.device("cpu"),
         )
-        result = sample_from_logits(logits, batch, Sampler(), torch.device("cpu"))
+        result = sample_from_logits(logits, batch, Sampler())
         assert result.token_ids[0] in [2, 3]
 
     def test_allowed_token_ids_mixed_batch(self) -> None:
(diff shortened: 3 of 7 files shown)


LxYuan0420 (approved):
Thanks for fixing this!

LxYuan0420:
This maps to upstream PyTorch issue pytorch/pytorch#192577, with the fix tracked in pytorch/pytorch#192621. Please link both in the PR description. We should keep torch sampling on CPU until the vllm-metal environment uses a PyTorch release or pinned commit containing that fix and the #622 repro passes on MPS again.

PR #630, opened 2026-08-19
[Spec Decode] Register draft-model KV as a scheduler-managed cache group

mhdimo:
CC: @ricky-chaoju 

### Problem
`method="draft_model"` speculative decoding re-prefills a request's entire prompt into the draft model's KV on every new request. The proposer's private block allocator has no content-based reuse, so even a resubmitted identical prompt pays a full draft prefill again (#482). PR #500 fixed this with a proposer-private prefix cache (vLLM's BlockPool for hashing, refcounting, and eviction), but was closed: the cache lived in the proposer/worker path while the real cache policy, `cache_salt`, per-request cache-read rules (`skip_reading_prefix_cache`), and the actual scheduler admission/allocation limit live in the scheduler/request layer, which a proposer-private allocator structurally cannot see.

### Fix
Split the draft model's KV by where cross-request reuse actually happens:

- **Committed portion (`[0, committed_len)`)**: Registered as a real second scheduler-owned `KVCacheSpec` group (`cache_policy.py::_draft_layer_specs`), sized and budgeted alongside the target's own groups. The scheduler hashes, matches, admits, and evicts it exactly like the target's, so `cache_salt`, `skip_reading_prefix_cache`, and the real admission limit apply automatically. `draft_model_proposer.py`'s `_make_decode_plan` and `_make_prefill_plan` read block IDs straight off `state.block_ids[self._committed_group_index]` (the scheduler's own assignment) rather than a private pool.
- **Speculative lookahead tail**: Positions drafted ahead of `committed_len` (not yet verified) have no scheduler concept, so it stays a small proposer-local scratch reservation over-provisioned into the draft's physical backend beyond the scheduler-visible block count.
- **Validation fail-fast**: `resolve_draft_dims` now rejects draft models with sliding-window or hybrid attention (e.g., Gemma4) at config-resolution time, since `_draft_layer_specs` unconditionally emits `FullAttentionSpec` and a mismatch would silently corrupt KV cache sizing.
- **Ingest alignment**: `DraftModelProposer` now ingests every active decode and prefill row each step (chunked or not, greedy-eligible or not), not just drafting-eligible ones. The scheduler advances `num_computed_tokens` and marks committed-group blocks "cached" uniformly per step, trusting whatever it scheduled was actually computed; skipping ingest for excluded rows would leave the scheduler believing blocks hold real KV that was never written. Only drafting (producing returned tokens, running extra lookahead steps) stays restricted to the greedy-eligible, non-intermediate subset.

### Behavior Changes
The draft cache's memory is now counted against the shared Metal KV budget rather than allocated after the target's budget was computed and left untracked. Concretely, each scheduler block now reserves room for the draft layers as well, so `per_block_bytes` doubles (1,835,008 -> 3,670,016 on a 128 GB Mac) and `num_blocks` halves (30,398 -> 15,189) for the same `VLLM_METAL_MEMORY_FRACTION`. This is a hard requirement, not a latent risk: on base at 0.5 the target takes the entire budget (`kv_budget` 55.78 GB, `num_blocks` 30,398) and drafting dies with a Metal out-of-memory inside `mx.eval(drafts)` — it only fits at 0.3. With this PR the same 0.5 fraction works end to end. `RequestState` gains a `num_computed_tokens` field set at request creation from the scheduler's own reconciled cache-hit boundary.

### Verification
Same methodology as the original #500 validation thread (`Qwen/Qwen3-0.6B` draft==target, greedy, resubmitting an identical prompt under a fresh request ID). The resubmit's first draft plan goes from full-prompt ingest with zero reuse to ingest 16 / reuse 8176 at 8k (ingest 16 / reuse 2032 at 2k); first propose drops from 3525 ms to 111 ms and e2e from 66.8 to 31.2 ms/tok at 8k (438 to 64 ms and 21.9 to 19.5 at 2k).

Losslessness is shown by comparing spec-decode output against a non-speculative reference run on the same tree (rather than cross-tree SHA comparison, which is unreliable when the base path is broken). The validation harness is in-tree at `tools/benchmark/draft_resubmit_benchmark.py`.

### Tests & Documentation
- E2E coverage that the draft group honors `cache_salt` and `skip_reading_prefix_cache` (the cache-policy gaps named when the proposer-private draft cache was rejected).
- Stub-level unit test that the scratch reserve covers `max_num_seqs` concurrent drafters.
- Regression tests verifying sliding-window and hybrid-attention draft models are rejected at config resolution time.
- Updated `docs/speculative_decoding.md`: documents scheduler-managed/budgeted draft KV cache, SWA/hybrid draft rejection at startup, and the block-alignment safety invariant for lookahead writes into scheduler-owned blocks.

Addresses #482 direction 1 (per-request re-ingest). The steady-state proposer overhead (#482 Problem 2) is unaffected and left as a follow-up.

Credits:
@KrxGu  for the first PR #500

### DIFF
--- a/tools/benchmark/draft_resubmit_benchmark.py
+++ b/tools/benchmark/draft_resubmit_benchmark.py
@@ -1,0 +1,251 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Benchmark for draft-model KV cache cross-request reuse and losslessness.
+
+Two things are measured on a *single* tree (no cross-tree comparison):
+
+**Reuse.** Two speculative-decode generates of the same prompt, under fresh
+request ids each time:
+
+  gen1_cold      -- fresh request, cold draft cache
+  gen2_resubmit  -- identical prompt, new request id
+
+A tree without cross-request reuse re-ingests the full prompt into the draft
+model's KV inside every first propose; with the scheduler-managed draft cache
+the resubmit reuses the committed prefix blocks and ingests only the uncached
+suffix. Greedy throughout.
+
+**Losslessness.** A separate non-speculative reference generate of the same
+prompt establishes ground-truth output tokens. The two spec-decode runs are
+compared against it; ``lossless`` is ``true`` when the token ids match exactly.
+This proves the spec-decode path does not alter greedy output without relying
+on a comparison across different code trees (where a broken base path can
+produce degenerate output, making the comparison meaningless).
+
+Structural numbers come from monkeypatching ``DraftModelProposer``'s plan
+construction: ``first_plan_reused_tokens`` / ``first_plan_ingest_tokens`` are
+the first plan's ``draft_seq_len`` (where existing KV starts) and its ingest
+length, read directly from internal proposer state.
+
+Notes on methodology:
+
+- Requires ``VLLM_ENABLE_V1_MULTIPROCESSING=0`` so the monkeypatch reaches
+  the engine; the script sets it if unset. Set ``VLLM_METAL_MEMORY_FRACTION``
+  as needed for the machine (0.4 is a reasonable default).
+- The warm 1-token generate heats the *target* prefix cache only on trees
+  where ``propose()`` is skipped for prefill-only steps. On the
+  scheduler-managed tree the runner calls ``propose()`` every step, so the
+  warm request also ingests draft KV and gen1_cold is draft-warm there. Use
+  ``--skip-warm`` for a draft-cold gen1.
+- ``propose_first_ms`` times the first ``propose()`` call, which on the
+  scheduler-managed tree can be an empty call that precedes the first plan;
+  prefer the plan columns and wall/tpot for conclusions.
+- The reference run creates a separate non-speculative ``LLM`` instance.
+  Pass ``--skip-lossless`` to skip it if memory is tight.
+
+Output: one JSON object with ``reference`` (or ``null``) and ``spec_runs``
+(array of per-generate records), prefixed with ``RESULT ``.
+
+Usage:
+
+    VLLM_METAL_MEMORY_FRACTION=0.4 \\
+        python tools/benchmark/draft_resubmit_benchmark.py --prefix 8192
+"""
+
+from __future__ import annotations
+
+import argparse
+import gc
+import json
+import os
+import sys
+import time
+
+os.environ.setdefault("VLLM_ENABLE_V1_MULTIPROCESSING", "0")
+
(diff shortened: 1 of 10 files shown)


ricky-chaoju (commented):
The fail-fast and the harness rework both look right. gemma-4-e2b trips the sliding_window branch and Qwen3.5-0.8B trips the layer_types branch, so both paths are reachable with real checkpoints. Reuse works here too: ingest 16 / reuse 2032 at prefix 2048, TPOT 16.4 ms, same cold and resubmit.

One thing for the description. On base at VLLM_METAL_MEMORY_FRACTION=0.5 the target takes the whole budget (kv_budget=55.78GB, num_blocks=30398) and drafting dies with a Metal out-of-memory in mx.eval(drafts); it only runs at 0.3. Here the same fraction works because per_block_bytes goes 1835008 -> 3670016 and num_blocks drops to 15189. Worth putting that in Behavior Changes instead of calling it a latent risk.

tools/benchmark/draft_resubmit_benchmark.py
```diff
@@ -0,0 +1,251 @@
...
+        enable_prefix_caching=True,
+        async_scheduling=False,
+    )
+    sp = SamplingParams(temperature=0, max_tokens=args.gen)
+    t0 = time.perf_counter()
+    out = llm.generate([prompt], sp)
+    dt = time.perf_counter() - t0
+    ids = list(out[0].outputs[0].token_ids)
+    del llm
+    gc.collect()
```
ricky-chaoju: This teardown is not enough, so the losslessness path never runs. On a clean 128 GB machine at VLLM_METAL_MEMORY_FRACTION=0.5 the full harness fails with model_memory=58.13GB, kv_budget=-1.19GB — the reference LLM's KV is still resident when the spec run profiles memory. `mx.clear_cache()` does not release it, and `gpu_memory_utilization` has no effect since this backend sizes KV from VLLM_METAL_MEMORY_FRACTION. So --skip-lossless is effectively mandatory, which disables the comparison this rework was for. Its own subprocess would fix it.

mhdimo: Fixed: the reference generate now runs in its own --reference-only subprocess so its Metal memory is gone before the spec run profiles; i also reproduced your exact kv_budget=-1.19GB failure with the old in-process flow on a 32 GB Mac at 0.4.

mhdimo:
Fixed as well the description

ricky-chaoju (approved):
LGTM. Thanks for the work!

PR #633, opened 2026-08-20
Use loopback for default UniProc rendezvous

LxYuan0420:
This PR is:

- To avoid VPN-selected tunnel IPs during default single-process Metal startup.
- To set `VLLM_HOST_IP=127.0.0.1` only for `uni`, `world_size=1`, and `data_parallel_size=1` when the user did not configure it.
- To preserve explicit `VLLM_HOST_IP` and leave mp, Ray, PP, DP, and multi-node paths unchanged.

Note:

This fixes issue #625. Pinned vLLM 0.27.1 still uses `get_ip()` for UniProc gloo rendezvous, which can select a VPN tunnel address and hang startup. The workaround is marked for removal after the pinned vLLM includes vLLM #50999.

### DIFF
--- a/vllm_metal/platform.py
+++ b/vllm_metal/platform.py
@@ -782,7 +782,7 @@ class MetalPlatform(Platform):
             parallel_config.distributed_executor_backend == "uni"
             and parallel_config.world_size == 1
             and parallel_config.data_parallel_size == 1
-            and "VLLM_HOST_IP" not in os.environ
+            and not os.environ.get("VLLM_HOST_IP")
         ):
             os.environ["VLLM_HOST_IP"] = "127.0.0.1"
 
--- a/vllm_metal/platform.py
+++ b/vllm_metal/platform.py
@@ -2,6 +2,7 @@
 """Metal Platform implementation for vLLM."""
 
 import logging
+import os
 import platform as py_platform
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


vllm_metal/platform.py
```diff
@@ -773,6 +774,18 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
 
+        # COMPAT(vLLM 0.27.1): UniProc uses get_ip() for its local gloo
+        # rendezvous. Use loopback so a VPN tunnel address cannot be selected.
+        # Remove after pinned vLLM includes vllm#50999, which uses a file://
+        # store for UniProc rendezvous.
+        if (
+            parallel_config.distributed_executor_backend == "uni"
+            and parallel_config.world_size == 1
+            and parallel_config.data_parallel_size == 1
+            and "VLLM_HOST_IP" not in os.environ
```
ricky-chaoju: `get_ip()` treats an empty value as unset (`envs.VLLM_HOST_IP` is `os.getenv("VLLM_HOST_IP", "")`, then `if host_ip:`), so `VLLM_HOST_IP=` still falls through to the 8.8.8.8 probe. This reads that same value as configured and skips, so the one case upstream would still resolve off the default route is the one case this leaves alone.

Ran your branch on vLLM 0.27.1, `vllm serve mlx-community/Llama-3.2-1B-Instruct-4bit --max-model-len 512`:

```
unset          ->  distributed_init_method=tcp://127.0.0.1:53013
VLLM_HOST_IP=  ->  distributed_init_method=tcp://172.31.45.28:53063
```

On a tunnel-owning host that second address is the one #625 is about.

```suggestion
            and not os.environ.get("VLLM_HOST_IP")
```

ricky-chaoju (approved):
Approving to unblock progress. Thanks for fixing it!

vllm_metal/platform.py
LxYuan0420: Thanks for this.

PR #634, opened 2026-08-20
Write align-mode GDN state rows in place with a Metal kernel

WindChimeRan:
## Summary

Prefix caching should help when requests share a prefix and add essentially no cost when they do not. On hybrid GDN models, `main` did neither: shared-prefix traffic improved by only 1.03×, while traffic without shared prefixes became 2.24× slower.

Qwen3.5+ prefix caching `align` will be the default in the next vllm version (currently it's off by default): https://github.com/vllm-project/vllm/pull/50991

This PR makes it behave like prefix caching:

| Model / mode | Shared prefix | No shared prefix |
| --- | ---: | ---: |
| Qwen3-0.6B, dense reference | **1.32× faster** | 1.02×, effectively free |
| Qwen3.5-0.8B, align, `main` | 1.03×, barely faster | **2.24× slower** |
| Qwen3.5-0.8B, align, this PR | **1.26× faster** | 0.98×, effectively free |

The dense row is the behavioral control: shared prefixes help, while unrelated traffic stays neutral. The PR restores that behavior for the hybrid model.

<details>
<summary>Benchmark setting</summary>

Each cell compares APC enabled with disabled for the same model and workload in the same benchmark session. All runs use 100 prompts, concurrency 16, request rate `inf`, 2304 input tokens, 128 output tokens, `--ignore-eos`, and seed 1234. The shared-prefix workload uses 10 repeated 2048-token prefixes with 256-token suffixes, producing a 61% cache-hit rate; the no-sharing workload uses random prompts.

The server runs on an Apple M5 Pro with 64 GB unified memory, macOS 26.2, vllm 0.27.0, mlx 0.32.0, `--max-model-len 4096`, paged attention enabled, and `VLLM_METAL_MEMORY_FRACTION=0.5`.

Each cell is one run. Repeat `main` runs for the no-sharing workload were 153.10 s and 156.73 s, so changes within approximately 3% should be treated as noise.

</details>

## Roadmap

* #584 prototype of align-mode GDN prefix caching
* #634 align mode kernel: after this PR, prefix caching becomes actually very useful, if you have enough memory. 
* future work: fix kv cache double count issue. 
* then we can mark this feature as support (currently experimental)
* future work: extend the support to LFM
* not planned: `all` mode prefix caching

## What was wrong & Why we need a `Primitive` metal kernel

Under `mamba_cache_mode="align"`, 18 linear layers share 6 physical state pools indexed by scheduler block id. At 1650+ cached blocks, each recurrent pool is about 1.6 GiB.

Every state write used MLX assignment. The available MLX 0.32.0 paths behave as follows:

| Write path | What `eval_gpu` does | Result for this pool |
| --- | --- | --- |
| `pool[ids] = rows` | [`Scatter`](https://github.com/ml-explore/mlx/blob/7a1d4f5c12ac82f4b4d0a6e71538d89ca0605247/mlx/backend/metal/indexing.cpp#L247-L256) calls `copy_gpu(pool, out)` before scattering | Full-pool copy when donation fails |
| `pool[a:b] = rows` | [`SliceUpdate`](https://github.com/ml-explore/mlx/blob/7a1d4f5c12ac82f4b4d0a6e71538d89ca0605247/mlx/backend/metal/indexing.cpp#L731-L764) calls the same `copy_gpu` path before copying the slice | Same full-pool copy |
| `mx.fast.metal_kernel` | [`CustomKernel`](https://github.com/ml-explore/mlx/blob/7a1d4f5c12ac82f4b4d0a6e71538d89ca0605247/mlx/backend/metal/custom_kernel.cpp#L21-L38) creates fresh output storage | Cannot express an aliased pool write |
| This PR: MLX `Primitive` | Aliases the pool with `copy_shared_buffer`, then writes selected rows | No full-pool pre-copy, with a graph dependency |

For a contiguous pool, `copy_gpu` avoids the copy only by donating the source buffer. [`array::is_donatable()`](https://github.com/ml-explore/mlx/blob/7a1d4f5c12ac82f4b4d0a6e71538d89ca0605247/mlx/array.h#L293-L296) requires both the array descriptor and its data buffer to have `use_count() == 1`. `store_conv_state` and `store_recurrent_state` alias each physical pool across its sibling layers, so donation is unavailable. A row update therefore rewrote up to all 1650+ rows.

The cache owned three affected paths: the deferred-state drain, `zero_slots`, and `copy_slots`. Three more writers reached into the pools directly. The hot one was `gdn_lazy.try_conv_decode`, which wrote the conv pool once per linear layer on every pure-decode step. This made a fresh request slower as unrelated requests filled the prefix cache (9.2 ms to 112.5 ms decode ITL).

## What changed

- Add `gdn_state_scatter`, an MLX `Primitive` that aliases the pool, writes only selected rows with a 2D/vectorized Metal kernel, and returns the post-write graph handle.
- Route all six writers through `write_conv_rows` or `write_recurrent_rows`, which scatter the rows and rebind every sibling layer to that handle.

This adds no `mx.eval` barriers. Recurrence, cache placement, and pending-state transitions are unchanged.

## Validation

- `tools/hybrid_apc_parity_matrix.py` on Qwen/Qwen3.5-0.8B: 6 runs, 480 comparisons, 0 mismatches. The matrix covers engine defaults and `max_num_batched_tokens=1088` to force aligned chunked splits.
- Focused GDN kernel, wrapper, and state-manager tests: 92 passed.

## Build

This changes C++ and Metal sources. Run `python -m vllm_metal.metal.build` to refresh prebuilt artifacts, or set `VLLM_METAL_BUILD_FROM_SOURCE=1` for a source build.

## Scope

This changes state-write traffic, not align-mode memory usage. The remaining 1.7% no-sharing overhead and the 8.2 ms versus 6.5 ms decode gap are not addressed here.

### DIFF
--- a/tests/test_gdn_lazy_kernels.py
+++ b/tests/test_gdn_lazy_kernels.py
@@ -710,7 +710,10 @@ class TestLazyConvPrefill:
         # Assert
         assert result is not None
         assert cache.pending_conv_state(0, [3, 1]) is None
-        assert decode_kernel.state_input is cache.conv_states[0]
+        # The kernel must read the stable pool, not the compact pending rows.
+        # Object identity no longer distinguishes them: the in-place scatter
+        # mints a fresh handle onto the same buffer, so compare shape.
+        assert decode_kernel.state_input.shape == cache.conv_states[0].shape
         mx.eval(cache.conv_states[0], decode_kernel.slot_mapping)
         np.testing.assert_array_equal(np.array(decode_kernel.slot_mapping), [1, 3])
         expected_state = np.array(initial_state)
--- a/tests/test_gdn_lazy_kernels.py
+++ b/tests/test_gdn_lazy_kernels.py
@@ -204,29 +204,27 @@ class TestGDNPagedStateCache:
         np.testing.assert_array_equal(np.array(view.cache_slot_ids), [3, 1])
         np.testing.assert_array_equal(np.array(view.state_slot_ids), [0, 1])
 
-    def test_pending_states_batch_matches_per_layer_drain(self) -> None:
-        # The batched drain must land exactly what draining each layer in turn
-        # would have landed.
-        def build() -> GDNPagedStateCache:
-            cache = _make_state_cache(num_layers=2, max_seqs=4)
-            cache.set_layer_layout([0, 1], [0, 0])
-            cache.set_pending_conv_state(
-                0, [1], mx.full((1, 2, 4), 4, dtype=mx.float32)
-            )
-            cache.set_pending_conv_state(
-                1, [3], mx.full((1, 2, 4), 6, dtype=mx.float32)
-            )
-            return cache
+    def test_copy_slots_gathers_overlapping_sources_before_writing(self) -> None:
+        cache = _make_state_cache(
+            max_seqs=4,
+            conv_kernel_dim=2,
+            conv_dim=1,
+            value_head_dim=1,
+            key_head_dim=1,
+        )
+        cache.conv_states[0] = mx.arange(4, dtype=mx.float32).reshape(4, 1, 1)
+        cache.recurrent_states[0] = mx.arange(4, dtype=mx.float32).reshape(4, 1, 1, 1)
 
-        batched = build()
-        batched.apply_pending_conv_states()
-
-        per_layer = build()
-        per_layer.apply_pending_conv_state(0)
-        per_layer.apply_pending_conv_state(1)
+        # Destination 2 is also the second source. Both source rows must be
+        # gathered before the in-place writes begin.
+        cache.copy_slots([1, 2], [2, 3], [0])
+        mx.eval(cache.conv_states[0], cache.recurrent_states[0])
 
         np.testing.assert_array_equal(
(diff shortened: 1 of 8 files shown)


tests/test_gdn_lazy_kernels.py
```diff
@@ -1730,3 +1774,185 @@ def fake_make_kernel(name: str, *_: Any) -> object:
...
         ]
+
+
+class TestNativeGDNStateScatter:
+    """The in-place row scatter that replaces MLX's whole-pool copy."""
+
+    @staticmethod
+    def _scatter_fn() -> Any:
+        ops = _get_native_ops_or_skip()
+        fn = getattr(ops, "gdn_state_scatter", None)
```
LxYuan0420: non-blocking-cleanup: Since `gdn_state_scatter` is required in production now, can we access `_get_native_ops_or_skip().gdn_state_scatter` directly here? The current `getattr(..., None)` can skip these tests for a stale extension even though runtime would fail.

LxYuan0420 (approved):
Approving. The inline cleanup comment is non-blocking.

PR #637, opened 2026-08-23
[Perf] Conditionally restore the MLX buffer memory default

ricky-chaoju:
Fixes the silent 36% single-stream regression that came with #586. Dropping the MLX_MAX_MB_PER_BUFFER default fixed the #585 startup failure, but nobody measured decode afterwards: on Qwen3.6-35B-A3B-4bit (M1 Ultra) main does 58.2 tok/s where the same code with the old default does 91.4. MLX splits each decode step into command buffers by memory touched, and a large MoE step walks about 2 GB of weights, so without the limit raised every step pays the commit overhead many times. I traced it by rebuilding the campaign-era tree in an isolated vLLM 0.26.0 venv (97.8 tok/s), running today's code under 0.26 (57.6, so not a vLLM regression), then flipping the single env var on main (91.0).

Both sides of the trade are real, so the default is conditional and single-tier: 2000 when the usable budget (total memory times the effective memory fraction) is at least 90 GiB, and nothing below that, on Ray executors, or above 4096 batched tokens. The #585 shape is pinned two ways: it never gets a default, and a second engine with a disqualifying shape removes the plugin's own earlier default instead of inheriting it. A manual export always wins. Per review, the extrapolated 512 tier is dropped; I can follow up with it once I have serving numbers from a real mid-memory machine.

| MLX_MAX_MB_PER_BUFFER | decode tok/s | profiled overhead |
|---|---|---|
| unset (main today) | 58.2 | 0.99 GB |
| 512 (manual export) | 88.0 | 7.6 GB |
| 2000 | 91.4 | 13.6 GB |

Serve command for all rows: `VLLM_METAL_MEMORY_FRACTION=0.5 vllm serve mlx-community/Qwen3.6-35B-A3B-4bit --max-num-seqs 8`, decode-only tok/s from streamed completions; only 2000 ships as a default. At 8192 batched tokens, 2000 inflates the overhead to 42 GB and startup fails, reproducing #585 exactly. Greedy output is byte-identical across values. Known limits: offline LLM() defaults to 8192 batched tokens so it stays unchanged, and in-process engines (VLLM_ENABLE_V1_MULTIPROCESSING=0) never adopt the default at all since MLX initializes before the config hook runs, verified live, so they get neither the win nor any new risk; the speedup applies through spawned engine processes.


vllm_metal/platform.py
```diff
@@ -34,6 +34,28 @@ class MetalPlatform(Platform):
     device_type: str = "cpu"  # PyTorch device type (use CPU for compatibility)
     dispatch_key: str = "CPU"  # PyTorch dispatch key
 
+    # ``MLX_MAX_MB_PER_BUFFER`` tiers by USABLE memory (GiB): total unified
```
LxYuan0420: This comment is too much for production code. Keep the invariant here and move benchmark numbers to the PR description.

vllm_metal/platform.py
```diff
@@ -34,6 +34,28 @@ class MetalPlatform(Platform):
...
+    # memory times the effective KV budget fraction. Larger command buffers
+    # cut per-step commit overhead but raise the transient profile/prefill
+    # peak that profile_run charges against the KV budget, so each tier
+    # requires its measured overhead to stay a small share (~15%) of the
+    # budget. Sweep on Qwen3.6-35B-A3B-4bit, M1 Ultra, serve defaults —
+    # decode tok/s / profiled overhead: unset 58.2/1.0GB, 512 88.0/7.6GB,
+    # 2000 91.4/13.6GB. No default at all above
+    # ``_MB_BUFFER_MAX_BATCHED_TOKENS`` (#585: 2000 at 8192 batched tokens
+    # inflates the profiled overhead ~19x and fails startup).
+    _MB_BUFFER_TIERS_USABLE_GIB: ClassVar[tuple[tuple[int, str], ...]] = (
```
LxYuan0420: I do not think we should default the `512` tier without real serving proof. Keep only the measured `2000` path for now, or add benchmark/startup evidence for this tier.

vllm_metal/platform.py
```diff
@@ -815,6 +842,59 @@ def _find_non_ssm_backend(
 
         return MetalBackend
 
+    @classmethod
+    def _default_mb_per_buffer(cls, vllm_config: "VllmConfig") -> None:
```
LxYuan0420, line 860: This private classmethod is becoming the feature API. I would keep `MetalPlatform` as the vLLM hook and move the buffer-default decision into a small module-level helper with clear inputs.

tests/test_macos_defaults.py
```diff
@@ -94,3 +101,163 @@ def test_apply_macos_defaults_logs_when_setting(monkeypatch, caplog) -> None:
...
+            max_num_seqs=8,
+            max_model_len=4096,
+            is_encoder_decoder=False,
+        ),
+        cache_config=SimpleNamespace(gpu_memory_utilization=gpu_memory_utilization),
+        parallel_config=SimpleNamespace(distributed_executor_backend=backend),
+    )
+
+
+class TestDefaultMbPerBuffer:
```
LxYuan0420: This tests too much implementation detail. Keep the user-path cases only: eligible default, manual export wins, #585-style large batch clears the plugin default, and non-local executor skips it.

LxYuan0420 (commented):
Direction looks right, but tighten the scope before merge. Keep only measured defaults, trim private-helper tests, and keep benchmark detail in the PR description.

vllm_metal/platform.py
ricky-chaoju: Done, trimmed to the invariant and pointed at the PR for the numbers.

vllm_metal/platform.py
ricky-chaoju: Dropped, kept only the measured 2000 path. Happy to bring 512 back as a follow-up once I have real serving numbers on a mid-memory machine.

vllm_metal/platform.py
ricky-chaoju: Done, the decision is a module-level _pick_mb_buffer_default(total_memory_bytes, memory_fraction, max_num_batched_tokens, executor_backend) now, and the classmethod only gathers inputs and installs or removes the env value.

tests/test_macos_defaults.py
ricky-chaoju: Trimmed to the four user-path cases through check_and_update_config: eligible default, manual export wins, large-batch clears the plugin default, non-local executor skips.

ricky-chaoju:
Addressed in the last two commits.

tests/test_platform.py
```diff
@@ -1567,6 +1581,124 @@ def fake_sync() -> None:
...
 
+    def test_check_and_update_config_installs_mb_default_after_resolution(
+        self, monkeypatch
+    ) -> None:
+        """Threading + ordering pin: the public hook installs the MB default,
+        and the allowlist sees the RESOLVED executor backend (the config
+        enters with the unset default and resolves to "uni" in the same
+        call)."""
+        import vllm_metal.config as vm_config
+        import vllm_metal.platform as platform_module
```
LxYuan0420, line 1592: Nit: these repeated local imports do not seem necessary in tests

LxYuan0420 (approved):

PR #639, opened 2026-08-23
Keep the hybrid APC parity harness bootable on 16 GB hosts

### LINKED ISSUE #638: [Bug] Hybrid APC parity harness fails during profiling on a 16 GB Mac when the dummy uses 8192 tokens
`tests/test_hybrid_prefix_caching_e2e.py` invokes
`python tools/hybrid_apc_parity_matrix.py --quick` with
Qwen/Qwen3.5-0.8B. On the tested 16 GB M-series Mac, on clean main at
211a1e4, the cache-off child completes, but the prefix-caching-enabled
child fails inside `LLM()` while sizing the paged KV cache.

This is an initialization-capacity failure in the parity harness, not an
on/off parity mismatch. The child exits before `queue.put()`. The parent
currently waits in `queue.get(timeout=1500)` before checking the child exit
code, so the original `ValueError` is easy to miss; the parent would later
raise `_queue.Empty`.

Relevant configuration:

- The tool sets `VLLM_METAL_MEMORY_FRACTION=0.5` with `setdefault()`; this
  is a tool default, not the library default.
- The quick/defaults arm passes `max_model_len=2048` but no
  `max_num_batched_tokens`, so the LLM-class default resolves to 8192.
- `profile_run()` executes a `(1, 8192)` dummy and evaluates the full
  logits tensor.
- Qwen3.5-0.8B has `vocab_size=248320` and bfloat16 logits.

On this host (`max_recommended_working_set_size` approximately 12.71 GB),
the cache-on child reports:

    usable_metal = 12.71 GB * 0.5 = 6.36 GB
    model_memory = 1.71 GB
    profiled overhead = 6.46 GB
    kv_budget = -1.81 GB

and fails with:

    metal_limit=12.71GB, fraction=0.5, usable_metal=6.36GB,
    model_memory=1.71GB, overhead=6.46GB, kv_budget=-1.81GB.

The cache-off child initializes with 2.38 GB of profiled overhead and a
final 2.21 GB KV budget.

As a causal A/B check, I kept the 8192-token dummy and changed only the
profile path to project logits for its final row. In the cache-on child,
profiled overhead fell from 6.46 GB to 2.39 GB: a measured 4.07 GB
reduction on this host. The matrix then completed 20/20 comparisons with
zero mismatches (`PARITY PASS`). The residual 2.39 GB remains after
last-row projection; the 4.07 GB figure is the measured cache-on delta
for this A/B, not a host-independent estimate.

Reproduction (local weights required):

    env -u VLLM_METAL_MEMORY_FRACTION \
        -u MLX_MAX_MB_PER_BUFFER \
        python tools/hybrid_apc_parity_matrix.py --quick

The e2e test is marked `slow`, is not run in CI, and requires local model
weights.

Suggested harness-scoped fix:

1. Bound the defaults arm's `max_num_batched_tokens` to its
   `max_model_len` of 2048, so the profiling dummy matches the maximum
   single-sequence shape exercised by this tool.
2. Poll for child termination while waiting for the result and report its
   exit code immediately if it dies before `queue.put()`.

With that tool-only change on the tested host, cache-on overhead fell from
6.46 GB to 0.73 GB and `kv_budget` changed from -1.81 GB to +3.92 GB. The
quick matrix completed 20/20 comparisons with zero mismatches
(`PARITY PASS`). The cache-off child reported 0.76 GB of overhead and a
+3.83 GB final KV budget. This changes neither serving behavior nor library
defaults.

#585 and #586 addressed a different source of `profile_run()` inflation:
the `MLX_MAX_MB_PER_BUFFER=2000` default. #637 proposes restoring that
default conditionally, but leaves it unset below 50 GiB of usable memory;
this host is well below that threshold, so that default is not involved in
this measurement.

RFC #525 and #590 cover selective logits on the paged serving path. They do
not currently cover this profiling dummy. This checkpoint loads through
mlx-vlm, while #590 currently excludes `_is_vlm`. I am sharing the
profile-run A/B on #590, and will propose the profile-run change as a small
complementary PR that reuses #590's `logits_indices` extension point rather
than introducing a competing adapter API.

fdaluiso:
Fixes #638.

## Summary

The `--quick` gate (and the defaults arm in general) passed no
`max_num_batched_tokens`. `LLM()` therefore took the 8192 LLM-class default,
and `profile_run()` evaluated a single `(1, 8192)` dummy. In the reproduced
configuration on the tested 16 GB M-series Mac, the cache-on child's
profiled overhead reached 6.46 GB, its KV budget reached -1.81 GB, and it
died during `LLM()` initialization. The cache-off child on the same revision
initialized with 2.38 GB of overhead.

This change is limited to `tools/hybrid_apc_parity_matrix.py`:

- The defaults arm now passes `max_num_batched_tokens=2048`, matching its
  existing `max_model_len`. The arm no longer exercises the LLM-class 8192
  default. That default was not needed by the short parity prompts; it made
  the profile dummy four times longer than any legal sequence in this
  harness. The explicit `mnbt1088` arm is unchanged.
- The parent polls the queue in five-second intervals and fails immediately
  with the child exit code if the child terminates before `queue.put()`,
  instead of waiting up to 1500 seconds and then raising `queue.Empty`.

This PR changes no serving behavior or library default. A serving-shaped
profile path is separate work being discussed in RFC #525 and #590.
#585/#586 concerned the former MLX per-buffer default; #637's proposed
conditional default does not apply below 50 GiB usable memory and is not
involved on the tested host.

## Testing

- `python tools/hybrid_apc_parity_matrix.py --quick` — 20 comparisons,
  0 mismatches, `PARITY PASS`
- Cache-on profile: overhead 6.46 GB -> 0.73 GB; KV budget -1.81 GB ->
  +3.92 GB
- Cache-off bounded profile: overhead 0.76 GB; final KV budget +3.83 GB

The end-to-end wrapper is marked `slow`, is not run in CI, and requires
local Qwen/Qwen3.5-0.8B weights.

## AI assistance

AI assistance was used in developing this change. The problem analysis,
the patch, and the measurements were reviewed and verified by me, and I
am responsible for all submitted changes.

### DIFF
--- a/tools/hybrid_apc_parity_matrix.py
+++ b/tools/hybrid_apc_parity_matrix.py
@@ -49,7 +49,9 @@ import argparse
 import json
 import multiprocessing as mp
 import os
+import queue as queue_mod
 import sys
+import time
 
 MODEL_DEFAULT = os.environ.get("QWEN35_MODEL_PATH", "Qwen/Qwen3.5-0.8B")
 
@@ -144,9 +146,15 @@ def run_child(model, enable_prefix_caching, mnbt, shared_corpus, quick, queue):
             "max_model_len": 2048,
             "max_num_seqs": 4,
             "enable_prefix_caching": enable_prefix_caching,
+            # profile_run builds a single (1, max_num_batched_tokens) dummy
+            # sequence. Without a cap the LLM-class default (8192) makes that
+            # dummy four times longer than max_model_len, and its full-vocab
+            # logits alone can push the measured overhead past the KV budget
+            # on 16 GB hosts (negative kv_budget -> startup rejection). Keep
+            # the profile within the shapes this harness actually runs.
+            "max_num_batched_tokens": 2048 if mnbt is None else mnbt,
         }
         if mnbt is not None:
-            kwargs["max_num_batched_tokens"] = mnbt
             kwargs["enable_chunked_prefill"] = True
         llm = LLM(**kwargs)
         cache_config = llm.llm_engine.vllm_config.cache_config
@@ -189,7 +197,25 @@ def run_pair(model, mnbt, shared_corpus, quick=False):
         )
         proc.start()
         try:
-            per_mode[enable] = queue.get(timeout=1500)
+            # Poll instead of a single long get: if the child dies during
+            # engine startup (e.g. a capacity rejection) it never reaches
+            # queue.put, and a bare get(timeout=1500) would sit for 25
+            # minutes before surfacing an unrelated queue.Empty. Fail fast
+            # with the child's exit code instead.
+            deadline = time.monotonic() + 1500
+            while True:
+                try:
+                    per_mode[enable] = queue.get(timeout=5)
+                    break
+                except queue_mod.Empty:
+                    if proc.exitcode is not None:
+                        raise RuntimeError(
+                            f"child (prefix_caching={enable}) exited with "
+                            f"{proc.exitcode} before reporting results; "
+                            "see its traceback above"
+                        ) from None
+                    if time.monotonic() >= deadline:
+                        raise
         finally:
             proc.join(timeout=60)
             if proc.is_alive():


LxYuan0420 (approved):
Looks good. The fix is scoped to the parity harness, keeps serving behavior unchanged, and addresses the 16 GB startup failure without adding test noise.

PR #640, opened 2026-08-23
Avoid full-vocabulary logits in profile runs for VLM-wrapped models (complements #590)

fdaluiso:
This PR complements #590 rather than replacing or competing with it. #590 adds
`logits_indices` for the paged serving prefill path and intentionally falls back
for VLM-wrapped models; this change uses the same API name at a different call
site—the cacheless dummy `profile_run()`—and includes the mlx-vlm-wrapped case
that triggers the 16 GB startup failure reported in #638. It does not change
paged serving or #590's row-layout work. On the reproduced configuration,
selecting only the final dummy row reduces profiled overhead from 6.46 GB to
2.47 GB and preserves parity (20/20); I am happy to rebase onto #590 or reduce
overlapping adapter code once its contract lands.

### Scope

- #590: serving paged-prefill producer/consumer layout; text-only safe path; no
  `profile_run()` change.
- This PR: profile dummy's final-row projection; VLM-wrapped model support with
  fallback; no serving layout change.
- PR1: tool-only 2048-token cap and fail-fast behavior; makes the parity harness
  usable but does not fix general profiling.

### Summary

- Align the adapter extension point with #590's `logits_indices` name and
  flattened-row semantics.
- Use the same `supports_selective_logits` capability style: compare a cacheless
  one-row public forward with the split body/head projection and opt in only on
  bit-exact agreement.
- Keep the cacheless transformer-body path used by mlx-vlm wrappers, including
  explicit text `position_ids` for bodies that declare that parameter.
- During a single-stage profile run, project only the final dummy-token row when
  the capability probe succeeds. Unsupported adapters and models retain the
  full-forward fallback. Pipeline-parallel and pooling profiling are unchanged.
- Preserve the public-wrapper serving path when `logits_indices` is absent and
  omit the new keyword for legacy adapters.

### Testing

- CPU unit selection (44 cases): `tests/test_model_adapter.py
  tests/test_v1_model_runner_generate.py -k "TargetForward or
  IntermediateForward or DummyForward"`
- Live gates to run separately: parity `--quick` and smoke.

### Relationship to #590

This change deliberately adopts #590's adapter contract instead of introducing
a parallel selective-logits API. It adds only the profile-run consumer and the
VLM-wrapped cacheless execution needed for the reproduced startup failure. I
welcome guidance from the #590 author and reviewers and will rebase or trim the
small overlapping adapter surface after #590 lands.

### AI assistance

AI assistance was used in developing this change. The problem analysis, the
patch, and the measurements were reviewed and verified by me, and I am
responsible for all submitted changes.

### Measured on this branch

- Parity harness `--quick` (with the 8192 dummy retained): cache-on child now
  boots — profiled overhead 6.46 GB -> 2.47 GB, KV budget -1.81 GB -> +2.18 GB;
  20 comparisons, 0 mismatches (`PARITY PASS`).
- Qwen3.5 smoke: unchanged against the baseline on this host.
- 195 unit tests pass.

## AI assistance

AI assistance was used in developing this change. The problem analysis,
the patch, and the measurements were reviewed and verified by me, and I
am responsible for all submitted changes.

### DIFF
--- a/tests/test_model_adapter.py
+++ b/tests/test_model_adapter.py
@@ -217,6 +217,422 @@ class TestTargetForward:
         assert output.logits.tolist() == [[[1.0, 2.0]]]
         assert output.hidden_states is None
 
+    def test_plain_forward_with_backbone_still_uses_public_wrapper(self) -> None:
+        class BackboneMustNotRun:
+            def __call__(self, input_ids, *, cache=None):
+                del input_ids, cache
+                raise AssertionError("serving forward must not call the body")
+
+        class Model:
+            def __init__(self) -> None:
+                self.model = BackboneMustNotRun()
+                self.lm_head = lambda hidden_states: hidden_states
+
+            def __call__(self, input_ids, *, cache=None):
+                del input_ids, cache
+                return mx.array([[[9.0]]])
+
+        output = DefaultModelAdapter().target_forward(
+            Model(),
+            mx.array([[1]]),
+            cache=[],
+            collect_hidden_states=False,
+        )
+
+        assert output.logits.tolist() == [[[9.0]]]
+        assert output.hidden_states is None
+
+    def test_logits_indices_projects_selected_hidden_rows_only(self) -> None:
+        head_inputs: list[mx.array] = []
+
+        class Backbone:
+            def __call__(self, input_ids, *, cache=None):
+                del cache
+                values = mx.array(input_ids, dtype=mx.float32)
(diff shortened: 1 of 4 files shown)


LxYuan0420:
Closing this because it under-profiles real serving memory and can turn a startup OOM into a runtime OOM. It also conflicts with #590 and needs a redesign, not incremental fixes.

PR #642, opened 2026-08-25
[Test] Hoist buffer default test imports

ricky-chaoju:
Follow-up to the #637 nit: the four buffer-default seam tests each re-imported vllm_metal.config and vllm_metal.platform inside the test body. Hoisted both to the module top.


LxYuan0420 (approved):
Thanks!

PR #641, opened 2026-08-24
Fix LoRA adapter cache eviction

LxYuan0420:
This PR is:

- To separate registered LoRA adapters from resident execution slots.
- To reload inactive adapters through LRU eviction instead of treating adapter registration as a lifetime limit.
- To preserve pinned adapters while allowing normal requested adapters to rotate through available slots.
- To reject invalid LoRA cache sizing where `max_cpu_loras < max_loras`.

Before, serving one LoRA could permanently fill the adapter cache and block later adapters after the first one became inactive.

After, multiple registered adapters can rotate through a smaller number of active slots.

<details>
<summary>Smoke command</summary>

```bash
vllm serve Qwen/Qwen2.5-0.5B-Instruct \
  --served-model-name qwen-base \
  --enable-lora \
  --max-loras 1 \
  --max-cpu-loras 4 \
  --lora-modules \
    chatbot=taronklm/Qwen2.5-0.5B-Instruct-lora-chatbot \
    xsum=MarkFirst/qwen2.5-0.5B-instruct-xsum-lora-r4 \
    sentiment=Carlo88/qwen05b-sentiment-lora-imdb \
    injection=aditya02acharya/luna2-qwen2.5-0.5b-prompt-injection-lora \
  --max-model-len 1024 \
  --max-num-seqs 1 \
  --port 8000 2>&1 | tee /tmp/vllm-metal-lora-lru-serve.log
```

`max_loras=1` and `max_cpu_loras=4` intentionally force eviction/reload during testing.

</details>

<details>
<summary>Example curl requests</summary>

Base model:

```bash
curl http://127.0.0.1:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  --data-binary @- <<'JSON'
{
  "model": "qwen-base",
  "messages": [
    {
      "role": "user",
      "content": "Summarize this product news in two short bullets: Apple released a new student-focused laptop with a brighter screen, longer battery life, a lighter aluminum body, faster on-device AI features, and a lower education price."
    }
  ],
  "temperature": 0,
  "max_tokens": 96
}
JSON
```

Chatbot LoRA:

```bash
curl http://127.0.0.1:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  --data-binary @- <<'JSON'
{
  "model": "chatbot",
  "messages": [
    {
      "role": "user",
[...]

PR #647, opened 2026-08-27
Optimize pure LoRA prefill routing

LxYuan0420:
This PR is:

- To optimize pure LoRA prefill by applying adapter deltas per active slot instead of expanding LoRA weights per token.
- To reduce long-prompt prefill latency and temporary adapter-weight traffic.
- To keep decode and mixed decode/prefill batches on the existing routing path.

Before: each LoRA token gathered full A/B adapter matrices, so prefill copied the same adapter weights once per token.

After: pure prefill groups rows by adapter slot and reuses each adapter A/B matrix for the whole row range or row group.

Scope note: this intentionally optimizes pure prefill only. If any one-token decode row is present, the batch keeps the existing expanded route to avoid decode regression.

Local hot-path microbenchmark, lower ms is better:

| Tokens | Before old expanded route | After pure-prefill route | Speedup |
|---:|---:|---:|---:|
| 64 | 0.773 ms | 0.563 ms | 1.37x |
| 512 | 1.006 ms | 0.260 ms | 3.87x |
| 2048 | 1.175 ms | 0.406 ms | 2.90x |
| 4096 | 2.490 ms | 0.449 ms | 5.54x |

### DIFF
--- a/tests/test_lora.py
+++ b/tests/test_lora.py
@@ -92,6 +92,45 @@ def test_punica_handles_fragmented_routing() -> None:
     )
 
 
+def test_punica_output_dtype_matches_base_for_contiguous_and_fragmented() -> None:
+    a_stacked = mx.array(np.array([[[1.0]], [[2.0]]], dtype=np.float32))
+    b_stacked = mx.array(np.array([[[3.0]], [[4.0]]], dtype=np.float32))
+    x = mx.ones((4, 1), dtype=mx.float16)
+    y = mx.zeros((4, 1), dtype=mx.float16)
+
+    for index_mapping in ((11, 11, 22, 22), (11, 22, 11, 22)):
+        wrapper = punica_mod.PunicaWrapperMLX(
+            max_num_batched_tokens=4, max_batches=4, max_loras=2
+        )
+        wrapper.update_metadata(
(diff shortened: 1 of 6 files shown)


ricky-chaoju (commented):
Correctness looks fine, outputs are bitwise identical in fp16 and bf16, and the memory win reproduces.

Decode is slower though. On your #641 smoke setup, M1 Ultra 128G, Qwen2.5-0.5B with the chatbot and xsum adapters, 8 sequences, greedy, 64 tokens each: alternating adapters 546 -> 345 tok/s, four-and-four 548 -> 413 tok/s. With LoRA off the same run is 0.680s, so the LoRA-attributable part goes from 0.261s to 0.803s and 0.558s. 

Your table starts at 64 tokens, where it is already back to 0.91x. Could you add the smaller decode batches, or fall back to the old routing below some token count?

LxYuan0420:
I missed the mixed-batch decode risk in the earlier version. I refactored this to optimize only pure prefill now. Decode and mixed decode/prefill batches keep the existing expanded route, so this PR only claims the long-prompt prefill memory/latency win.

ricky-chaoju (approved):
LGTM. Thanks for turning it around so fast.

Issue #646, opened 2026-08-26
[Bug] Mixed batches with disabled and enabled top_k crash the Metal sampler


Lavmee:
## Summary

`vllm-metal` crashes the entire `EngineCore` when one batch contains both a
request with top-k disabled (`top_k <= 0`) and a request with top-k enabled
(for example, `top_k=64`).

The disabled row is passed to vLLM's sampler as `0`. The PyTorch top-k/top-p
implementation calculates the gather index as `vocab_size - top_k`, producing
an out-of-bounds index equal to `vocab_size` for that row.

## Environment

- Apple M3 Ultra, arm64
- macOS 26.6.2
- `vllm-metal 0.3.0.dev20260826134128`
- `vllm 0.28.0+cpu`
- The production failure used a model with vocabulary size `262144`

The same failure also occurred with the preceding dev build paired with
`vllm 0.27.1+cpu`.

## Minimal reproduction

```python
import torch

from vllm.sampling_params import SamplingParams
from vllm.v1.sample.sampler import Sampler
from vllm_metal.v1.sampling_batch import SamplingBatch

batch = SamplingBatch(
    [
        SamplingParams(temperature=0.0),
        SamplingParams(temperature=1.0, top_k=64, top_p=0.95),
    ],
    [[], []],
    [[], []],
    vocab_size=128,
)
logits = torch.arange(256, dtype=torch.float32).reshape(2, 128)

Sampler().forward(logits, batch.make_sampling_metadata())
```

## Actual result

```text
RuntimeError: index 128 is out of bounds for dimension 1 with size 128
```

With the production vocabulary the same exception was:

```text
RuntimeError: index 262144 is out of bounds for dimension 1 with size 262144
```

This exception terminates `EngineCore`; the in-flight request returns HTTP 500
and the API server shuts down.

## Expected result

The row with disabled top-k should sample from the full vocabulary, while the
other row should retain its requested `top_k=64`. Mixed sampling parameters
must not terminate the engine.

## Root cause

`SamplingBatch.no_top_k` correctly detects that the batch as a whole needs
top-k processing because one request has `top_k=64`. However, `_make_top_k()`
then constructs a tensor containing the raw per-request values, such as
`[0, 64]`.

The upstream vLLM sampler interprets the per-row disabled sentinel as
`vocab_size`, not `0`. Its PyTorch path calculates:

```python
top_k_mask = logits_sort.size(1) - k
```

[...]

PR #648, opened 2026-08-27
[Bugfix] Normalize disabled top_k rows to vocab_size in mixed sampling batches (#646)

mhdimo:
This PR is:

- To normalize disabled top_k rows to the vocab_size sentinel in mixed 
- To match vLLM's per-row top_k convention in GPUInputBatch and the v1 sampler 
- To keep disabled rows sampling from the full vocabulary while enabled rows keep their 
- To stop a mixed top_k=0 / top_k=N batch from crashing EngineCore.

Before: _make_top_k() passed raw per-request top_k values to vLLM's sampler, so a disabled row carried 0. The PyTorch top-k path computes a gather index of vocab_size - top_k, indexing out of bounds and terminating EngineCore (issue #646).

After: _make_top_k() sends top_k when 0 < top_k < vocab_size and the vocab_size sentinel otherwise. A regression test covers the mixed batch through the real Sampler.forward() path.

### DIFF
--- a/tests/test_sampling.py
+++ b/tests/test_sampling.py
@@ -498,6 +498,38 @@ class TestV1SamplingBatch:
         assert result.token_ids[0] in [2, 3]
         assert result.token_ids[1] == 1  # unconstrained, should pick highest logit
 
+    def test_mixed_disabled_and_enabled_top_k(self) -> None:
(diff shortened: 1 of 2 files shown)


ricky-chaoju (approved):
Thanks for the fix!

PR #649, opened 2026-08-28
[Perf] Native non-greedy sampling for the decode pipeline

ricky-chaoju:
Follow-up to #343, with the opt-in shape we agreed there. Qwen's generation_config makes real requests non-greedy by default, and on main those pay twice: the torch sampler round-trip every step, plus losing the pipeline deferral since the gate only admits greedy. That costs most of the decode speed, 26.4 tok/s against 89.9 greedy on the 35B.

With the flag on, the pipeline's deferred sampler gets one more graph (a top-k/top-p mask plus mx.random.categorical) and the gate admits plain temperature/top-k/top-p batches. Seeds, penalties, logprobs, token constraints and mixed batches keep the torch path, and the synchronous sampling functions are byte-for-byte identical to main, I checked by diff. Mask semantics are pinned against apply_top_k_top_p including boundary ties, and the #622 probe on the new path came back 0/60.

| model | greedy | non-greedy on main | non-greedy native |
|---|---:|---:|---:|
| Qwen3.6-35B-A3B | 89.9 | 26.4 | 85.1 |
| Qwen3.8-27B | 33.3 | 18.6 | 32.5 |

Streamed single-stream decode, M1 Ultra, temp 0.7 / top_k 20 / top_p 0.95, fraction 0.5, `MLX_MAX_MB_PER_BUFFER=2000`, `--no-enable-prefix-caching` to keep the hybrid APC cost  out of the comparison. Measured on 0.27.1, the greedy baseline reproduces at 91.4 on 0.28. `tools/native_sampling_microbench.py` has the sampling-only numbers, 14.01 ms torch (including the mlx_to_torch bridge) vs 1.64 ms native at batch 1. Sampled token streams differ from torch's RNG, that comes with changing the sampler; seeded requests keep the torch path and are unaffected.


vllm_metal/v1/model_runner.py
```diff
@@ -2808,7 +2833,28 @@ def _submit_deferred_decode_sample(self) -> MetalAsyncModelRunnerOutput:
         assert logits is not None
         self._draft_token_ids = None
 
-        tokens = mlx_greedy_tokens(logits[0, : len(decode_reqs), :])
+        deferred_params = [state.sampling_params for _, state in decode_reqs]
```
LxYuan0420: Too much sampling policy is moving into `model_runner.py`. The runner should manage RNG key, async submit, and backfill only. `SamplingBatch` should own greedy/random selection, vocab slicing, and sampling graph construction.

vllm_metal/v1/sampling_batch.py
```diff
@@ -192,6 +192,44 @@ def params_allow_native_greedy(
             for sampling_params in sampling_params_list
         )
 
+    @staticmethod
+    def params_allow_native_random(
```
LxYuan0420, line 196: Keep one native-sampling capability check inside `SamplingBatch`. Platform validation already rejects unsupported request options, so this method should only decide whether this batch can use the native sampler.

vllm_metal/v1/sampling_batch.py
```diff
@@ -358,6 +396,69 @@ def mlx_greedy_tokens(logits_2d: mx.array) -> mx.array:
     return mx.argmax(logits_2d, axis=-1)
 
 
+def mlx_top_k_top_p_masked_logits(
```
LxYuan0420: Do not expose `mlx_top_k_top_p_masked_logits()` and `mlx_random_tokens()` as public module functions. These are implementation details of native sampling and should sit behind a `SamplingBatch` method.

vllm_metal/envs.py
```diff
@@ -74,6 +75,15 @@
     # elementwise glue. Bitwise-identical on the quantized serving path;
     # set to "1" to enable, the default keeps the eager per-op dispatch.
     "VLLM_METAL_COMPILED_MLP": lambda: os.getenv("VLLM_METAL_COMPILED_MLP", "0") == "1",
+    # MLX-native non-greedy sampling for the decode pipeline (opt-in): the
```
LxYuan0420, line 78: If this path is proven safe and consistently faster on vLLM 0.28 serving workloads, it should move toward default-on. Keep the env var as a kill switch, not as a permanent half-enabled feature.

tools/native_sampling_microbench.py
```diff
@@ -0,0 +1,87 @@
...
+from vllm_metal.v1.sampling_batch import mlx_random_tokens
+
+VOCAB_SIZE = 151936
+TOP_K = 20
+TOP_P = 0.95
+WARMUP_ITERS = 5
+TIMED_ITERS = 50
+
+
+def _time_torch(logits: torch.Tensor, batch: int) -> float:
```
LxYuan0420: This benchmark does not include the MLX-to-Torch bridge before timing, so it does not measure the full production cost being removed. Either benchmark the real production sampling paths or label this clearly as sampling-kernel-only.

tests/test_v1_model_runner_generate.py
```diff
@@ -2459,6 +2500,89 @@ def _paged_state(
...
+        return mr.RequestState(
+            token_ids=[3, 9],
+            prompt_len=1,
+            cache=[],
+            sampling_params=SamplingParams(temperature=0.7, top_k=1, top_p=1.0),
+            generator=None,
+            generated_tokens=1,
+        )
+
+    def test_deferred_random_step_samples_natively_and_backfills(
```
LxYuan0420: Avoid spying on the helper here. Test through the public `sample_tokens()` path and assert the sampled token, key advancement, and async backfill.

vllm_metal/v1/sampling_batch.py
```diff
@@ -358,6 +396,69 @@ def mlx_greedy_tokens(logits_2d: mx.array) -> mx.array:
...
+            mx.topk(scaled_logits, k=top_k, axis=-1), axis=-1, keepdims=True
+        )
+        scaled_logits = mx.where(scaled_logits < kth_largest, -mx.inf, scaled_logits)
+    if top_p < 1.0:
+        order_desc = mx.argsort(scaled_logits, axis=-1)[..., ::-1]
+        sorted_desc = mx.take_along_axis(scaled_logits, order_desc, axis=-1)
+        sorted_probs = mx.softmax(sorted_desc, axis=-1)
+        leading_mass = mx.cumsum(sorted_probs, axis=-1) - sorted_probs
+        keep = leading_mass < top_p
+        # The leading sorted position has zero leading mass, so at least
```
LxYuan0420: vLLM rejects `top_p=0`, and valid `top_p > 0` already keeps at least one token. Remove this extra operation and the `top_p=0` tests.

LxYuan0420 (commented):
To make this a good foundation for future sampling work, we should keep sampling policy inside `SamplingBatch` and keep `model_runner.py` focused on RNG state, async submit, and backfill. Please also make the benchmark measure the real production path, so we can decide whether this should move toward default-on with the env var as a kill switch.

vllm_metal/v1/model_runner.py
ricky-chaoju: Done, the deferred site is a single SamplingBatch.native_decode_tokens() call now; the runner only keeps the key chain, submit, and backfill.

vllm_metal/v1/sampling_batch.py
ricky-chaoju: Dropped the min_p/logit_bias re-checks, the predicate only decides batch capability now.

vllm_metal/v1/sampling_batch.py
ricky-chaoju: Both moved behind SamplingBatch as private statics, native_decode_tokens is the only entry.

vllm_metal/envs.py
ricky-chaoju: Agreed, updated the env comment to say exactly that: default-on once it has mileage, var stays as the kill switch.

tools/native_sampling_microbench.py
ricky-chaoju: The torch arm now times the full path including the mlx_to_torch bridge: 14.01 ms vs 1.64 ms native at batch 1, the bridge itself is zero-copy so the cost is the CPU sort.

tests/test_v1_model_runner_generate.py
ricky-chaoju: Reworked all three through sample_tokens(), the spy is gone; they assert the sampled token, key advancement, and the async backfill.

vllm_metal/v1/sampling_batch.py
ricky-chaoju: Removed, along with the top_p=0 tests.

LxYuan0420 (approved):
LGTM

PR #650, opened 2026-08-28
[Spec Decode] Add grammar-forced drafting for structured output and tool calls

MizbaulHaqueMaruf:
## Who this helps, and who it doesn't

Up front, because it decides whether the rest is worth your time.

**Helps:** greedy (`temperature=0`) structured-output requests on the paged path — a
JSON schema via `structured_outputs.json`, or tool calling with grammar enforcement
turned on. JSON-extraction and agent/tool-call workloads are the shape this targets.
Gain measured at **1.22x–1.82x**, larger at higher batch and on smaller models.

**Does nothing for:** plain chat, free-form string values, the model's own choice of
which tool to call, `temperature > 0`, non-paged runs, and anything routed to
`outlines`/`guidance`/`lm-format-enforcer`. In those cases it returns no draft and the
step stays exactly as it is today.

**Costs, if you don't enable it:** nothing. This is a `custom_class` proposer — no new
`SpeculativeMethod` literal, no upstream change, no default path touched.

**Honest caveats before the numbers:**

- **Greedy only.** That is a pre-existing constraint of
  `SpeculativeDecodeController.verify_greedy`, shared by all three existing Metal
  methods, not something this PR introduces — but it does bound the audience.
- **`disable_any_whitespace: true` is effectively required**, and it changes output
  format rather than being a free knob. With free whitespace, determinism on the same
  workload falls from 69% to 27%.
- **Tool calling needs a three-flag opt-in** (`VLLM_ENFORCE_STRICT_TOOL_CALLING=1`,
  `"strict": true` on a tool, and a parser declaring a `structural_tag_model`). Plain
  JSON-schema requests need none of that. Out of the box, a tool-calling deployment
  gets no benefit until it opts in.
- **The gain shrinks with model size** — see the 4B row below.
- **`--reasoning-parser` is refused at construction**, which rules out thinking models.

## The problem

When vllm-metal serves a constrained request, most of the tokens it spends a full
forward pass on were **never in question**. The braces, quotes, key names and
separators are fixed by the grammar before the model runs.

Two measurements, both with speculative decoding **off**:

**1. Decode dominates these requests.** Prefill is noise at these shapes:

| Model | Batch | Prefill (TTFT) | Decode | Decode share of wall |
|---|---|---|---|---|
| Qwen3-0.6B | 1 | 20.6 ms | 265.2 ms | 92.8% |
| Qwen3-0.6B | 4 | 40.6 ms | 632.9 ms | 94.0% |
| Gemma-4 E2B | 1 | 42.5 ms | 761.4 ms | 94.7% |
| Gemma-4 E2B | 4 | 86.0 ms | 2050.9 ms | 96.0% |

**2. Half of that decode was already decided.** On **150 Berkeley Function-Calling
Leaderboard entries** (`simple`, `multiple`, `live_simple`), Qwen3-0.6B, greedy, no
speculative decoding — each entry carrying its own tools, so each compiles to its own
grammar:

```
emitted tokens       5859
grammar-determined   2934 (50%)
forced runs          557 (mean 5.3, max 22 tokens)
schemas skipped      0
traces off-grammar   0
emitted a tool call  142/150
```

Composed with the decode share, roughly **47% of end-to-end time on a constrained
tool-calling request goes into generating text the grammar had already fixed.**

For reference, the same measurement on a hand-written schema reads higher — 69% (JSON)
and 63% (structural tag) on eight prompts. **BFCL's 50% is the number to trust**; the
hand-written workload was more uniform than reality.

### Reproducing the baseline

`tools/grammar_determinism.py` measures the *workload*, not this implementation. It
runs plain greedy decoding with no speculative decoding, replays the emitted token ids
through a fresh xgrammar matcher, and counts positions where
`find_jump_forward_string()` was already non-empty. It imports nothing from
`vllm_metal` and **runs unchanged on a checkout without the proposer** — so it is not
circular, and it is worth running against `main` before reviewing anything else here:

```bash
BFCL=$(huggingface-cli download --repo-type dataset \
  gorilla-llm/Berkeley-Function-Calling-Leaderboard)

VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_MEMORY_FRACTION=0.8 \
python tools/grammar_determinism.py --model Qwen/Qwen3-0.6B --no-thinking \
  --num-prompts 150 --max-tokens 128 --max-model-len 2048 \
  --bfcl-file "$BFCL/BFCL_v3_simple.json" \
  --bfcl-file "$BFCL/BFCL_v3_multiple.json" \
  --bfcl-file "$BFCL/BFCL_v3_live_simple.json"
```

It reports the *ceiling* on what any grammar-aware drafter could remove, not what this
one achieves. The prefill/decode split comes from the `baseline` arm of
`tools/benchmark/grammar_spec_decode_report.py`, which also needs no proposer.

## Summary

Adds a fourth speculative-decoding proposer behind the existing `MetalProposer` seam.
It drafts the text the grammar has already determined and stops at every genuine
decision point. Like n-gram it loads no model and keeps no KV cache; unlike n-gram it
needs no repetition in the text. The verify half is untouched.

This works because vLLM already applies the grammar bitmask to **every** verification
row, not just the bonus row (`MetalStructuredOutputApplier.apply_paged` runs before
`verify_greedy`), so a drafted token is grammar-legal by construction.

### Why not the in-tree n-gram proposer?

Measured, same workload, same K:

| Model | Batch | baseline TPOT | n-gram | grammar-forced |
|---|---|---|---|---|
| Qwen3-0.6B | 1 | 2.82 ms | 2.90 ms (0.97x) | 1.98 ms (**1.42x**) |
| Qwen3-0.6B | 4 | 6.66 ms | 6.77 ms (0.98x) | 3.68 ms (**1.81x**) |
| Qwen2.5-0.5B-Instruct | 1 | 1.98 ms | 2.07 ms (0.96x) | 1.16 ms (**1.71x**) |
| Qwen2.5-0.5B-Instruct | 4 | 4.65 ms | 4.66 ms (1.00x) | 2.34 ms (**1.98x**) |

n-gram fails here structurally, not just from sparse repetition. On one batch-1 run of
279 output tokens, `NgramProposer.propose()` returned 3 drafts totalling 24 tokens —
only 2 were ever verified and none accepted. `Scheduler.update_draft_token_ids` filters
drafts through the active grammar's `validate_tokens()`, and a grammar-blind drafter's
history-derived guesses are mostly illegal at that position.

## Why the obvious implementation doesn't work

"Draft a token whenever the grammar leaves exactly one legal choice" fires on **zero of
23 positions** on a real tool-call schema. The *string* is forced but the *tokenization*
never is: after `{` the grammar demands `"name"`, and `"`, `"n`, `"na` and `"name` are
all legal tokens.

Instead this walks xgrammar's `find_jump_forward_string()` with greedy longest-token
matching. That **approximates** the canonical tokenization rather than reproducing it —
BPE merges by rank, not by length, and the two disagree on e.g. `celsius` (walk:
`cel|si|us`, tokenizer: `c|elsius`). Acceptance is the number that says whether the
approximation is holding.

**The boundary rule is what makes it safe.** The walk stops one token short of the end
of each forced string, because a token consuming all of it may merge into the free text
that follows (the grammar forces `"`, but the model emits `" P` as one token). Without
it, offline precision drops from 100% to 85.6% and every mismatch is a run's last token.

**Which token spells the text also matters.** Several vocabulary entries can decode to
the same text, and picking the wrong id drafts a token the model never emits — legal by
construction, rejected in practice. SentencePiece vocabularies carry a byte-fallback
token per byte value (`<0x29>` → `)`) at a *lower* id than the real token, so a "lowest
id wins" tiebreak picks it for every ASCII character: all 73 a JSON skeleton is built
from, on Gemma-4 E2B. Real tokens outrank byte-fallback ones.

## Benchmarks

Apple M4 Pro, vLLM 0.28.0, greedy. **Token ids are bit-identical to baseline in every
configuration.**

**Constrained JSON generation** — `structured_outputs.json`:

| Model | Batch | TPOT baseline | TPOT drafting | Speedup | Tokens from drafts | Acceptance |
|---|---|---|---|---|---|---|
| Qwen3-0.6B | 1 | 2.8 ms | 1.9 ms | **1.50x** | 47% | 100% |
| Qwen3-0.6B | 4 | 6.7 ms | 3.7 ms | **1.82x** | 49% | 83% |
| Gemma-4 E2B | 1 | 12.1 ms | 7.7 ms | **1.58x** | 56% | 82% |
| Gemma-4 E2B | 4 | 32.6 ms | 18.4 ms | **1.77x** | 52% | 71% |
| Gemma-3-4B | 1 | 17.1 ms | 14.0 ms | **1.22x** | 36% | 67% |

**The gain shrinks as the model grows** — 1.22x at 4B against ~1.5x below 2B. A wider
verification step costs proportionally more when the forward pass is bigger, and
acceptance is lower. I could not test 8B: at bf16 that is ~17 GB of weights against
15.26 GB usable on a 24 GB M4 Pro (the 4B model measures 8.60 GB), so it does not fit
beside a KV cache. A 4-bit 8B would fit and I am happy to run one if that is the size
you care about.

**Real tool calling** — hermes parser, structural tag. Measured on vLLM 0.27.1; Qwen3
has no byte-fallback tokens, so its drafting counters are unchanged by the tokenizer
fix above:

| Model | tok/s baseline | tok/s drafting | Speedup | Acceptance | Tool calls |
|---|---|---|---|---|---|
| Qwen3-0.6B | 117.8 | 148.2 | **1.26x** | **96/96 = 100%** | 8/8, identical to baseline |

`propose()` costs 0.6–1.8% of wall time. The vocabulary prefix table is built once at
construction (160 ms on Qwen3, ~350 ms on a 262k vocabulary), not per step.

I did **not** use `vllm bench serve` with the sonnet dataset: sonnet is free-form prose
where this proposer correctly drafts nothing, so it would measure only the overhead.
The BFCL determinism run above is the substitute — it characterises the workload
independently of this implementation. Happy to add a sonnet no-regression run.

### Why it wins at batch 1

Metal's decode step cost is governed by query-row count alone (measured here: 1 row =
23.0 ms, 2 rows = 50.4 ms, 8 rows = 52.4 ms). A fixed-K drafter is punished at batch 1
because a wrong guess forces it off the one-row fast path. This proposer returns **no
draft at all** when nothing is forced, so those steps stay on the fast path.

## What is and isn't guaranteed

Verification is lossless, so output is unaffected either way. Acceptance is
**empirical, not guaranteed**: several tokenizations of a forced string are legal, so a
drafted token is grammar-legal by construction yet only *probably* the target's argmax.

`GrammarProposerStats.altered_drafts` is the hard invariant, and is **zero in every
run**. It counts drafts that came back with their prefix rewritten — nothing benign
does that, so it is unambiguous evidence the worker's matcher and the engine's have
drifted. `truncated_drafts` (prefix intact, tail missing) is a *diagnostic*, not an
invariant: the engine's grammar rejecting the tail and the scheduler clipping against
the token budget or `max_model_len` look identical from the worker.

The design keeps a second xgrammar matcher in the worker that must stay in lockstep
with the engine's. Matcher state is a pure function of the committed output prefix, so
preemption, resume, request-id reuse and prefix-cache hits all resolve through one
resync path rather than special cases — but this is the part of the change that carries
ongoing maintenance cost, and `altered_drafts` exists to make a divergence loud.

## Configuration

```bash
vllm serve Qwen/Qwen3-8B --no-async-scheduling \
  --structured-outputs-config '{"backend":"xgrammar","disable_any_whitespace":true}' \
  --speculative-config '{"method":"custom_class","model":"vllm_metal.v1.grammar_proposer.GrammarProposer","num_speculative_tokens":8}'
```

## Limitations (all documented in `docs/speculative_decoding.md`)

- Greedy only, paged path only, synchronous scheduling only — same as the other three
  methods.
- **Gemma-4's own tool calling is not accelerated.** Its parser deliberately skips
  structured outputs (native `<|tool_call>call:...` syntax) and declares no
  `structural_tag_model`. Explicit JSON-schema requests to Gemma-4 *are* accelerated.
- A configured `--reasoning-parser` is refused at construction: while reasoning is
  unfinished the engine stops advancing its grammar matcher, which the worker cannot
  observe, so the two would silently desynchronize.

## Testing

- `tests/test_grammar_spec_decode.py` — 29 tests using a real tokenizer and real
  compiled grammars (a mocked matcher would test nothing here). Covers the forced-run
  walk, the boundary rule, matcher rollback, resync after preemption, request-id reuse,
  backend filtering, and the drift detectors. Two worth calling out: one asserts every
  JSON character resolves to the id the tokenizer itself uses, parametrized over Qwen3
  and Gemma-4; the other replays a pinned real generation through the grammar with no
  model loaded, so a change to the walk is decidable without a benchmark run.
- `tools/test_grammar_spec_decode_e2e.py` — losslessness against a pinned greedy
  golden, non-zero acceptance, and inertness on a request with no grammar.
- Full gate green on **vLLM 0.28.0**: `ruff check`, `ruff format --check`, `shellcheck`,
  `mypy vllm_metal` (124 files clean), and `pytest -m "not slow" tests/`.


LxYuan0420 (commented):
Before reviewing the implementation, can you clarify what concrete vllm-metal problem this solves and share a reproducible baseline showing it is a material bottleneck?

MizbaulHaqueMaruf:
Thanks — fair question to ask first. I've measured it rather than argue it, and rewritten the description around the numbers.

## What problem this solves

Under a JSON schema or a tool-call structural tag, a large share of decode steps produce text the grammar had *already* fixed — the braces, quotes, key names and separators. Those forward passes are pure overhead.

## How material, measured independently of this PR

`tools/grammar_determinism.py` imports nothing from `vllm_metal`, so it runs on a clean checkout with none of this code in it. It decodes greedily with no speculative decoding, replays the emitted token trace through a fresh xgrammar matcher, and counts positions where `find_jump_forward_string()` was non-empty. That is the ceiling on what *any* grammar-aware drafter could remove — it measures the workload, not my implementation.

Two public corpora, 150 entries each, Qwen3-0.6B, greedy:

```
                     BFCL          API-Bank
emitted tokens       5859          7414
grammar-determined   2934 (50%)    2714 (37%)
forced runs          557           618
traces off-grammar   0             0
schemas skipped      0             0
```

So the ceiling is **37–50% depending on corpus**, and BFCL is the flattering end. API-Bank is multi-turn with more free text in its arguments, which is the more realistic shape, so plan on the lower end. Both numbers came down as the corpus got more realistic — I first measured 63% on eight hand-written prompts, which was not a workload, it was a demo.

Runnable against `main`, before reviewing anything else here:

```bash
BFCL=$(huggingface-cli download --repo-type dataset \
  gorilla-llm/Berkeley-Function-Calling-Leaderboard)

VLLM_ENABLE_V1_MULTIPROCESSING=0 VLLM_METAL_MEMORY_FRACTION=0.8 \
python tools/grammar_determinism.py --model Qwen/Qwen3-0.6B --no-thinking \
  --num-prompts 150 --max-tokens 128 --max-model-len 2048 \
  --bfcl-file "$BFCL/BFCL_v3_simple.json" \
  --bfcl-file "$BFCL/BFCL_v3_multiple.json" \
  --bfcl-file "$BFCL/BFCL_v3_live_simple.json"
```

`--api-bank-file` takes API-Bank the same way.

## What it buys

1.22x–1.82x across models and batch sizes. Losslessness is gated by `tools/test_grammar_spec_decode_e2e.py`, which asserts SD reproduces *plain greedy* token-for-token **and** that the verify half actually accepted drafts — output equality alone would pass an inert drafter.

**Why not the in-tree n-gram proposer:** measured on the same prompts at the same K, 0.96x–1.00x. A single structured response does not repeat itself, so suffix matching finds nothing while still paying its scan, and `Scheduler.update_draft_token_ids` discards history-derived drafts the active grammar rejects.

Two caveats I'd rather state than have you find:

- **The gain shrinks with model size** — 1.22x on Gemma-3-4B against ~1.5x below 2B, since a wider verification step costs proportionally more when the forward pass is bigger. 4B at bf16 is the largest that fits my 24 GB M4 Pro beside its KV cache, so 8B+ is untested.
- **Acceptance is empirical, not guaranteed.** Drafts are grammar-*legal* by construction (vLLM applies the bitmask to every verification row), but several tokenizations of a forced string are legal and the walk only approximates the canonical one. Verification is lossless either way; a miss costs a wider step.

## Heads-up on scope

The branch also now carries a second proposer: a port of ToolSpec's retrieval-augmented drafting ([arXiv 2604.13519](https://arxiv.org/abs/2604.13519)), which targets the argument *values* the grammar cannot force — the other 50–63%. It composes the grammar proposer rather than replacing it, and wires in through the existing `custom_class` hatch, so `model_runner.py` is untouched.

Three arms, Qwen3-0.6B, 32 sequential requests, K=8:

| Workload | baseline TPOT | grammar | +retrieval |
|---|---|---|---|
| API-Bank tool calls | 6.6 ms | 5.3 ms (1.26x) | 5.1 ms (**1.31x**) |
| Sonnet under a JSON schema | 7.4 ms | 7.1 ms (1.05x) | 6.7 ms (**1.11x**) |
| Sonnet, no grammar | 7.1 ms | 7.0 ms | 7.1 ms (**1.00x**) |

That last row is the one I'd check first: unstructured traffic proposes zero drafts and stays within noise of baseline, so nothing pays for a feature it cannot use. Getting there needed a miss-streak back-off — retrieval searches whether or not it finds anything, and on prose it found nothing in 1977 attempts while still costing ~1%.

Honest limits on the retrieval half:

- **Linear drafts, not trees.** Metal's verify half is linear and shared with the MTP and draft-model proposers, so this takes the single best candidate. Part of the paper's headline speedup is tree breadth, so it does not reproduce it.
- **The API-Bank arm is realistic prompts against a synthetic union schema**, not the hermes structural-tag path. The 1.26x tool-call figure above it is the real structural-tag measurement.
- **0.6B only** for these three arms, and the gain shrinks with model size.
- **The memory is shared across requests** — one request's output can seed another's draft. Verification is unchanged, so that changes what is *guessed*, never what is *emitted*.
- Retrieval acceptance is 27% against the grammar's 67%, which is expected: a grammar draft is legal by construction, a retrieved one is a guess. It wins because the drafts it lands are longer (7.0 tokens against 4.4).

That makes this PR considerably larger than when you looked. **Happy to split it** — the grammar-forced half plus the measurement tooling stands alone and answers your question on its own, with the ToolSpec port as a follow-up. Say the word and I'll cut it that way.

Also rebased onto current main: the doc conflicted with #658's restructure, and the e2e gate's pinned greedy output needed refreshing (the final stop token changed from `<|im_end|>` to `<|endoftext|>` in 0.28.0 — the 23 content tokens are unchanged, and plain greedy with no speculation emits the same thing, so drafting is unaffected). Spec-decode suite green at 344 passing.

LxYuan0420 (commented):
Can we take a step back and narrow this to one clearly defined problem and one proposer, with a minimal implementation and reproducible before/after benchmark? 

Please also move ToolSpec to a separate proposal. That will make the ownership, correctness, and performance win much easier to review.

PR #652, opened 2026-08-29
[Release] Split dev and stable release channels & bump to 0.4.0

WindChimeRan:
## Problem

Every push to `main` publishes a GitHub release, but current dev builds are not marked as prereleases. GitHub's `/releases/latest` therefore returns a dev build, and `install.sh` has no stable channel.

## Changes

| Channel | Source | Trigger | GitHub release type |
| :--- | :--- | :--- | :--- |
| dev | `main` | Every merge | Prerelease |
| stable | `releases/vX.Y.Z` | `workflow_dispatch` | Final, except `rcN` |

- `scripts/release.sh` validates stable versions, builds both channels through the same native-artifact checks, and tags the exact commit that built the wheel.
- The workflow rejects stable releases from the wrong branch or with a mismatched version.
- Stable releases are created as drafts so the release owner can review the wheel and curate the notes before publication.
- `install.sh` defaults to dev, adds `--stable` plus `VLLM_METAL_CHANNEL=dev|stable`, and rejects non-stable tags from the stable channel.
- Each source revision records its compatible upstream vLLM release in `.github/vllm-release-tag.commit`. The installer reads that file from the selected vllm-metal tag before installing vLLM. Existing releases fall back to their pinned `install.sh` value.
- The package version moves to `0.4.0`.

## Validation

- `scripts/lint.sh`: shellcheck, ruff, formatting, and mypy all pass.
- Live GitHub selection resolves `v0.3.0.dev20260829020049` to its matching vLLM `v0.28.0` wheel.

A real publish remains untested until the workflow is merged.

## Rollout

1. Merge so new dev builds publish as prereleases.
2. Cut `releases/v0.4.0`.
3. Create a draft `0.4.0` from the tested commit, curate its notes, and publish it; use an RC first only when useful.
4. Mark the existing dev releases as prereleases after the final release exists.


install.sh
```diff
@@ -46,33 +51,76 @@ fetch_latest_release() {
...
-            break
-except Exception as e:
-    print('', file=sys.stderr)
-"
+    data = json.load(sys.stdin)
+except Exception:
+    sys.exit(0)
+
+# /releases/latest returns one release; /releases returns a list, newest first.
+if isinstance(data, dict):
```
LxYuan0420, line 70: Stable should explicitly reject `.dev` tags. Right now `--stable` selects the live dev release, so the channel split does not work yet.

.github/workflows/release.yml
```diff
@@ -19,7 +25,31 @@ jobs:
...
+    - name: Validate stable release
+      if: github.event_name == 'workflow_dispatch'
+      run: |
+        if [[ ! "$RELEASE_REF" =~ ^releases/v([0-9]+\.[0-9]+\.[0-9]+)$ ]]; then
+          echo "::error::Stable releases must run from releases/vX.Y.Z, got ${RELEASE_REF}."
+          exit 1
+        fi
+        branch_version="${BASH_REMATCH[1]}"
+        branch_pattern="${branch_version//./\\.}"
+        if [[ ! "$RELEASE_VERSION" =~ ^${branch_pattern}(rc[0-9]+|\.post[0-9]+)?$ ]]; then
```
LxYuan0420, line 37: If we keep release branches, this should also validate `pyproject.toml`. Otherwise `releases/v9.9.9` with input `9.9.9` can build while the package version is still `0.4.0`.

LxYuan0420:
I want to discuss the release process before we lock it in. I like the discipline of dev/RC/stable, but since this is a community project, I also want the process to stay light and easy to operate. Can we clarify who owns RC validation, what checks prove a release is ready, and whether we can start with main as dev prereleases plus manual promotion of a tested commit to stable?

WindChimeRan:
@LxYuan0420 
- **Core motivation**
  - Every build from `main` currently appears as a normal GitHub release, making the release history noisy and less informative.
  - Dev builds should be marked as prereleases so numbered stable releases and their release notes are easy to find.

- **Short-term goal for this PR**
  - Continue publishing dev builds automatically from `main`.
  - Keep dev as the installer default, so user behavior does not change. Stable releases is optional.
  - Publish `0.4.0` manually with curated release notes.

- **Release validation**
  - Any maintainer with write access can own validation and publish the release.
  - The release branch is only a fixed pointer to the tested commit and a place for post-release fixes.
  - Ideally most of the work should be done by CI, and maintainer should only do minimum manual work. But it's a bit difficult due to Github action resource limitation.  I'll check with bulidkite later ...

- **Long-term versioning goal (maybe in 1 month)**
  - Align vllm-metal versions with upstream vLLM, such as `0.30.0` instead of `0.5.0`.
  - Follow upstream’s roughly two-week stable release cycle.
  - It's difficult because recently the upstream vllm introduced some breaking change on kv cache layout, and we need to fix it before the next big version.
 
the release for the user is the same with before: they'll install the latest dev from main. The major version is a nice-to-have feature, user will still install the latest version even if we don't do manual release.

install.sh
WindChimeRan: Thanks for the good catch. `--stable` now rejects dev, RC, legacy timestamp, and unrelated tags.

.github/workflows/release.yml
WindChimeRan: the workflow now requires the version in `pyproject.toml` to match the base version in releases/vX.Y.Z before building.

LxYuan0420 (approved):
Looks good. Let’s keep it simple for now: `main` publishes dev prereleases, stable is manually promoted after draft-wheel review, and RC stays optional. We can formalize more later when we have a clear need and stronger CI coverage.

reneleonhardt:
https://github.com/vllm-project/vllm/releases/tag/v0.29.0 🚀

PR #658, opened 2026-08-30
[Bug fix][Spec Decode] Multi-step Gemma4 MTP drafting

WindChimeRan:
**Before**: Gemma4 MTP accepted `num_speculative_tokens > 1` but always returned one draft token. Output stayed correct because verification handled that single token, but the configured width had no effect.

**After**: This PR reuses the assistant module recurrently for K steps. Each step embeds the previous draft with the target's backbone-width table and feeds back the assistant's projected backbone hidden state. Because the assistant is Q-only and reads the target's paged KV without writing to it, every recurrence uses the same target position and one `prepare_grouped` metadata build. The existing `verify_greedy` path already handles K-token drafts.

The PR also removes the local `n_predict == 1` checkpoint validation. Upstream overwrites Gemma4 draft configs with `n_predict=1`, while Gemma4 model depth comes from `text_config.num_hidden_layers`. Unsupported checkpoint layouts remain covered by strict weight loading.

Additional cleanup:

- Moves the MLX assistant model to `vllm_metal.spec_decode.gemma4.model`; the V1 runtime and loader remain in `vllm_metal.v1`.
- Keeps `docs/speculative_decoding.md` focused on the Metal capability table, constraints, tuning, benchmarking, and one runnable example per method; deeper usage links to upstream vLLM.

## Validation

- K=1, K=2, and K=3 generated identical tokens on all 18 `spec_bench` prompts.
- Branch K=2 matched `main` K=1 on all 18 prompts.
- K=2 matched no speculative decoding on 17/18 prompts. The same prompt diverges on `main` K=1 and n-gram SD K=1, isolating it to the target's batch-shape-dependent argmax behavior rather than this drafter.
- The relevant Gemma4 MTP test set passes locally (206 tests). Ruff and mypy also pass.

## Performance

Gemma4 E4B bf16 with its E4B assistant on an M5 Pro, using natural `spec_bench` prompts, greedy sampling, `--ignore-eos`, and 128 output tokens. Ratios are output-token throughput versus no speculative decoding.

| K | Concurrency 1 | Concurrency 32 | Mean acceptance length |
|---|---:|---:|---:|
| 1 | 0.99x | 1.27x | 1.73 |
| 2 | 1.20x | 1.30x | 2.18 |
| 3 | 1.38x | 1.30x | 2.51 |
| 5 | 1.53x | 1.13x | 2.91 |

K=3 matches the best saturated result in this sweep while improving single-stream throughput by 1.38x, so the docs recommend starting there.

The existing greedy-only and synchronous-scheduling limits remain. The 18-prompt parity sweep remains a manual check; this PR does not add a heavyweight end-to-end test.

<details>
<summary>Reproduction</summary>

### 1. Prepare the dataset and environment

```bash
# 198 natural prompts, 22 per category
python tools/build_spec_bench_dataset.py

export TARGET=mlx-community/gemma-4-e4b-it-bf16
export ASSISTANT=mlx-community/gemma-4-E4B-it-assistant-bf16
export VLLM_METAL_USE_PAGED_ATTENTION=1
export VLLM_METAL_MEMORY_FRACTION=0.55
export VLLM_ENABLE_V1_MULTIPROCESSING=0
```

### 2. Check greedy parity

Run each width in a separate process:

```bash
for K in 0 1 2 3; do
  PYTHONPATH=$PWD python - "$K" <<'PY'
import itertools
import json
import os
import sys
from pathlib import Path

k = int(sys.argv[1])
rows = [json.loads(line) for line in Path("spec_bench_sample.jsonl").read_text().splitlines()]
by_category = {}
for row in rows:
    by_category.setdefault(row["category"], []).append(row["turns"][0])

# Select 18 prompts round-robin across the nine categories.
prompts = list(itertools.islice(
    (prompt for group in zip(*(by_category[name] for name in sorted(by_category))) for prompt in group),
    18,
))

speculative_config = None if k == 0 else {
    "method": "mtp",
    "model": os.environ["ASSISTANT"],
    "num_speculative_tokens": k,
}
llm = LLM(
    model=os.environ["TARGET"],
    max_model_len=2048,
    speculative_config=speculative_config,
    async_scheduling=False,
)
outputs = llm.generate(prompts, SamplingParams(temperature=0.0, max_tokens=64))
Path(f"parity_k{k}.json").write_text(json.dumps({
    output.prompt: list(output.outputs[0].token_ids) for output in outputs
}))
PY
done
```

Compare the generated token IDs:

```bash
python - <<'PY'
import json
from pathlib import Path

outputs = {k: json.loads(Path(f"parity_k{k}.json").read_text()) for k in (0, 1, 2, 3)}
for k in (2, 3):
    matches = sum(outputs[1][prompt] == outputs[k][prompt] for prompt in outputs[1])
    print(f"K=1 vs K={k}: {matches}/{len(outputs[1])}")
matches = sum(outputs[0][prompt] == outputs[2][prompt] for prompt in outputs[2])
print(f"K=2 vs no SD: {matches}/{len(outputs[2])}")
PY
```

### 3. Measure throughput and acceptance

Start one server per width. Drop `--speculative-config` for the baseline:

```bash
vllm serve "$TARGET" --max-model-len 2048 --no-async-scheduling \
  --speculative-config "{\"method\":\"mtp\",\"model\":\"$ASSISTANT\",\"num_speculative_tokens\":3}"
```

Run the single-stream benchmark:

```bash
vllm bench serve --backend vllm --base-url http://127.0.0.1:8000 \
  --model "$TARGET" --endpoint /v1/completions \
  --dataset-name spec_bench --dataset-path spec_bench_sample.jsonl \
  --spec-bench-output-len 128 \
  --num-prompts 30 --request-rate 10 --max-concurrency 1 \
  --temperature 0 --top-p 1.0 --top-k -1 --ignore-eos --seed 0
```

- For the saturated run, use `--num-prompts 100 --max-concurrency 32`.
- Read per-position acceptance from the server log's `SpecDecoding metrics` lines.

</details>


ricky-chaoju (approved):
LGTM

PR #659, opened 2026-08-30
[Bugfix] Honour Hunyuan's query_layernorm/key_layernorm in paged SDPA

ritunjaym:
## Summary

`prepare_sdpa_qkv` applies the per-head Q/K norms by attribute name, probing only `q_norm`/`k_norm`. `mlx_lm/models/hunyuan_v1_dense.py` names them **`query_layernorm`/`key_layernorm`**, so under `use_qk_norm: true` the normalization was skipped entirely — the model loads, serves, returns HTTP 200 and generates garbage, with nothing raising.

This adds a name resolver so both spellings are honoured.

**Scope: this fixes the catastrophic case, and leaves a small ordering error that @WindChimeRan identified in review (see below). It is a large improvement, not a complete fix.**

## What this fixes, quantified

Both figures measured identically — bf16 (the real compute dtype), on the actual pre-norm Q tensor captured from a live paged run, layer 0, against the `mlx_lm` reference `RMSNorm(RoPE(x))`:

| | max abs | mean abs | relative |
| --- | --- | --- | --- |
| `main` today (norm skipped entirely) | 6.3838 | 0.931429 | **97.19%** |
| this PR (norm applied, ordering wrong) | 0.0469 | 0.001139 | **0.1189%** |

~818× error reduction. Output goes from degenerate gibberish to matching the `mlx_lm` reference token-for-token on every prompt tested.

## The remaining gap — credit to @WindChimeRan

The review catch is correct, and I confirmed it **bit-exactly** rather than by reading source. Capturing the real post-RoPE Q tensor from a live paged run and comparing against both candidate orders recomputed in bf16:

| pipeline vs | max | mean |
| --- | --- | --- |
| `RoPE(RMSNorm(x))` | **0.000000** | **0.00000000** |
| `RMSNorm(RoPE(x))` | 0.046875 | 0.00113948 |

Exact zero against the first. The pipeline provably computes `RoPE(RMSNorm(x))`; Hunyuan's reference is `RMSNorm(RoPE(x))` (`hunyuan_v1_dense.py` applies `self.rope` at L54-58, then `query_layernorm` at L61-62).

The ordering error is **~0.12%**, and it is *not* dampened by quantization — bf16 amplifies it roughly 20× relative to an fp32 estimate.

**In isolated `mlx_lm` this is enough to flip greedy output.** Swapping only the order (controls: stock-vs-stock and a correct-order re-implementation both give max delta exactly 0.00000) yields a max logit delta of **0.805** against a top-1/top-2 margin of **0.188** — the argmax changes.

**It did not reproduce in vllm-metal's actual output.** Across ten constructed cases — the 5 golden prompts, a 3,089-token long-context prompt, and 4 deliberately low-margin prompts — the paged path produced tokens identical to the `mlx_lm` reference. I also implemented the corrected order in `prepare_sdpa_qkv` and re-ran: output was unchanged.

**Why that gap exists is unproven.** I am not asserting an explanation. Resolving it is the first task in #660.

## Why the ordering fix isn't in this PR

The obvious approach — key the ordering off the attribute spelling — does not work. Spelling does not predict order:

| Model | Spelling | Order |
| --- | --- | --- |
| qwen3, gemma3_text | `q_norm` | norm → RoPE |
| **stablelm** | `q_layernorm` | **norm → RoPE** |
| **lfm2** | `q_layernorm` | **norm → RoPE** |
| **hunyuan_v1_dense** | `query_layernorm` | **RoPE → norm** |

`stablelm` and `lfm2` use a non-standard *name* with the standard *order*. Hunyuan is an outlier on both axes independently, so a correct fix needs an explicit per-architecture ordering signal in `prepare_sdpa_qkv` — a larger change to the hot path, and one whose end-to-end effect I cannot currently demonstrate. Filed as #660.

## Docs

`docs/supported_models.md` gains a Hunyuan row per review. Marked 🔵 rather than ✅: "tested and passing" here means short-context greedy parity with the `mlx_lm` reference, and the ordering question above is still open. Happy to move it either way.

## Tests

`tests/test_hunyuan_golden.py`, following `tests/test_gemma4_golden.py` exactly: dual `GOLDEN_MLX_LM`/`GOLDEN_PAGED`, `HUNYUAN_MODEL_PATH` env gate, class-scoped paged-env fixture, `@pytest.mark.slow`, and a module-level consistency test that needs no model. `_MIN_COMMON_PREFIX = MAX_TOKENS - 1`, justified because the two sets came out byte-identical. Plus `tools/gen_hunyuan_golden.py`, mirroring `gen_gemma4_golden.py`.

```
with the fix:      5 passed, 1 deselected in 14.90s
fix reverted:      5 failed
CI mode (-m "not slow"):  1 passed, 5 deselected   # never loads a model
no checkpoint:     1 passed, 5 skipped
full suite:        1791 passed, 7 failed (all pre-existing: stt/gguf extras absent)
```

Without the fix the paged path diverges from `mlx_lm` at token 0 on four of five prompts (token 1 on the fifth).

**This test does not cover the ordering error** — see #660; ten constructed cases failed to expose it.

## Environment

Apple M2 / 8 GB, macOS 26.6.2, Python 3.12.2 (arm64), vllm 0.28.0+cpu, mlx 0.32.0, mlx-lm 0.31.3. Checkpoint `mlx-community/Hunyuan-1.8B-Instruct-4bit` @ `5a121aef6911ff1c99a3b3f1f7b028414cdfa618`.


WindChimeRan (commented):
Nit: for tested new model support. We should update `docs/supported_models.md` as well.

WindChimeRan (commented):
This PR is incomplete: 

I checked the mlx_lm reference:
```
PR:        RoPE(RMSNorm(x))
reference: RMSNorm(RoPE(x))
```
But it's hard to have a clean fix, because vllm-metal hard code the order:
```
mlx-lm transformer layer
  → SDPAPagedAttentionWrapper
  → sdpa_forward
  → prepare_sdpa_qkv
      project Q/K/V
      normalize / apply RoPE
  → write K/V into paged cache
  → Metal attention kernel
  → output projection
```

---

One possible refactoring is: 
```
class QKNormPlacement(Enum):
    BEFORE_ROPE = auto()
    AFTER_ROPE = auto()
```
and add a small model-specific patch file, similar to omlx.

ritunjaym:
Thanks — good catch, and you're right. I confirmed it bit-exactly rather than by reading the source, and the numbers came out worse than my first estimate.

**Confirming the ordering.** I instrumented `prepare_sdpa_qkv` to dump `queries` before the norm block and again after `apply_attention_rope`, served the model, then recomputed both candidate orders offline from the captured pre-norm tensor using the model's own modules:

| pipeline vs | max | mean |
| --- | --- | --- |
| `RoPE(RMSNorm(x))` | **0.000000** | **0.00000000** |
| `RMSNorm(RoPE(x))` | 0.046875 | 0.00113948 |

Exact zero. The pipeline provably applies the norm first, as you said.

One methodology note worth recording: this only shows up if you measure in **bfloat16**. Computing the references in fp32 while the pipeline runs bf16 injects ~0.002 of dtype noise, which makes both candidates look equidistant and understates the ordering error by ~20×. I got that wrong on my first pass. bf16 amplifies the ordering difference rather than damping it.

**Magnitude**, both measured identically (bf16, real captured prefill tensor, layer 0, vs the `mlx_lm` reference):

| | max abs | mean abs | relative |
| --- | --- | --- | --- |
| `main` today (norm skipped) | 6.3838 | 0.931429 | **97.19%** |
| this PR (ordering residual) | 0.0469 | 0.001139 | **0.1189%** |

So the PR takes it from 97% to 0.12% — a real improvement, and still not correct.

**The ordering error is big enough to matter in isolation.** Swapping only the order inside `mlx_lm` on a low-margin prompt gives a max logit delta of 0.805 against a top-1/top-2 margin of 0.188, and the argmax flips. Controls: stock-vs-stock and a correct-order re-implementation both give max delta exactly 0.00000, so the harness is exact.

**But I could not reproduce an output difference in vllm-metal.** Ten constructed cases — the 5 golden prompts, a 3,089-token long-context prompt, and 4 deliberately low-margin prompts including the one that flips in isolated `mlx_lm` — all produced tokens identical to the reference. I also implemented the corrected order directly in `prepare_sdpa_qkv` and re-ran: output unchanged. Why that gap exists is unproven; I'm not going to guess at it in the PR.

**On the fix itself** — I looked at keying the order off the attribute name and it doesn't hold:

| Model | Spelling | Order |
| --- | --- | --- |
| qwen3, gemma3_text | `q_norm` | norm → RoPE |
| stablelm | `q_layernorm` | **norm → RoPE** |
| lfm2 | `q_layernorm` | **norm → RoPE** |
| hunyuan_v1_dense | `query_layernorm` | **RoPE → norm** |

StableLM and LFM2 use the non-standard *name* with the standard *order*, so a name heuristic would silently mis-order them once they reach this path. It needs an explicit per-architecture signal, which is the larger change you flagged. Filed as #660 with the reproduction method and the open question.

Also added the `docs/supported_models.md` row per your other comment — marked 🔵 rather than ✅, since "tested and passing" here means short-context greedy parity and the ordering question is still open. Happy to move it to ✅ if you'd rather.

If you'd rather the ordering fix landed in this PR, I can attempt it — but I should flag that I currently have no way to verify it works. The same ten cases pass with and without the corrected order, so I'd be reshaping a hot path with no test that distinguishes the two states. I'd rather resolve the open question in #660 first, unless you have a way to exercise the difference that I've missed.

WindChimeRan (approved):
* removed the tests because they cannot guarantee correctness. 
* added inline TODO issue for follow up  fixes.

WindChimeRan:
Thanks for trying vllm-metal and fixing the bug! @ritunjaym

PR #662, opened 2026-08-30
[Release] Retain only the latest dev release

WindChimeRan:
## Problem

Automated development prereleases accumulate on the GitHub Releases page even though only the newest wheel is installed.

## Change

- Serialize release publication per Git ref so overlapping `main` runs cannot race during cleanup.
- Remember the current `.dev` release before publishing its replacement.
- Delete the previous Release record only after the replacement release and wheel are published successfully, while retaining its Git tag.
- Leave the stable release path and installer unchanged.

## Manual cleanup on the release page

- I plan to clean the Releases page manually and retain one representative release for each previous version line: `v0.1.0`, `v0.2.0`, and `v0.3.0`.

I'll do it now if you agree @LxYuan0420

## Validation

- `./scripts/lint.sh`
- The live release selector resolves the current development prerelease.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,9 @@ on:
         required: true
         type: string
 
+concurrency:
+  group: release-${{ github.ref }}
+
 jobs:
   build-and-release:
     runs-on: macos-15
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -141,6 +141,13 @@ main() {
   fi
 
   if [ "$channel" = "dev" ]; then
+    local main_sha
+    main_sha=$(git ls-remote --exit-code origin refs/heads/main | cut -f1)
+    if [ "$(git rev-parse HEAD)" != "$main_sha" ]; then
+      echo "Skipping stale dev release: checkout is no longer the tip of main."
+      exit 0
+    fi
+
     previous_dev_tag=$(gh release list --limit 30 --json tagName \
       --jq '[.[] | select(.tagName | contains(".dev"))][0].tagName // ""')
   fi
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -123,7 +123,7 @@ main() {
   fi
   verify_wheel_artifacts "${wheels[0]}"
 
-  local tag
+  local tag previous_dev_tag=""
   tag="v${version}"
   echo "Generated tag: $tag"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -10,6 +10,9 @@ on:
         required: true
         type: string
 
+concurrency:
+  group: release-${{ github.ref }}
```
LxYuan0420, line 14: Add a stale-rerun check before publishing dev releases. Serialized jobs still allow an old `main` rerun to publish stale code as the newest dev build.

scripts/release.sh
```diff
@@ -139,7 +139,17 @@ main() {
   if [ "$prerelease" -eq 1 ]; then
     release_args+=(--prerelease)
   fi
+
+  if [ "$channel" = "dev" ]; then
+    previous_dev_tag=$(gh release list --limit 30 --json tagName \
+      --jq '[.[] | select(.tagName | contains(".dev"))][0].tagName // ""')
```
LxYuan0420, line 152: I understand the one-dev-release cleanup model, but I prefer automatic retention. After a successful publish, keep the latest 30 dev releases and delete older dev release records.

LxYuan0420 (commented):
I prefer a rolling dev window over manual cleanup plus one current dev release. It keeps recent dev builds available for debugging, avoids manual maintenance, and still keeps the Release page bounded. Latest 30 is preferred; latest 10 is acceptable if we want it tighter.

WindChimeRan:
@LxYuan0420 

keeping the latest 10 will overflow the release page and the stable release will still be in the page 2, 
image

If we use latest 1 for dev, then it's 

* dev latest (user will install this one)
* stable 0.4.0 (with major version release log)
* stable 0.3.0 (same)
* ...

LxYuan0420:
> If we use latest 1 for dev, then it's
> 
> * dev latest (user will install this one)
> * stable 0.4.0 (with major version release log)
> * stable 0.3.0 (same)
> * ...

Ahh, I see your point now. . I’m ok with this model once we add the stale-rerun guard before dev publish

```
Example:

Before:
1. D1 = v0.4.0.dev20260831001336
2. S4 = v0.4.0
3. S3 = v0.3.0

Publish:
N = v0.4.0.dev20260831120000

Delete:
D1 release record

After:
1. N  = latest dev
2. S4 = v0.4.0
3. S3 = v0.3.0
```

LxYuan0420 (approved):
Approving first to unblock and thanks for working on this!

.github/workflows/release.yml
WindChimeRan: added!

PR #663, opened 2026-08-31
[Docs] Add an architecture figure to the README

WindChimeRan:
Adds a four-layer stack diagram to the README showing that upstream vLLM supplies the server, scheduler, and paged block manager, mlx_lm supplies the token-wise model layers, and vllm-metal owns the attention path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,9 @@ vLLM Metal is a plugin that enables vLLM to run on Apple Silicon Macs using MLX 
 
 ## Architecture
 
-![vllm-metal in the vLLM stack](docs/assets/architecture.svg)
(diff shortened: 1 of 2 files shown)


LxYuan0420 (approved):
Looks good overall. The figure is a useful README addition. A few minor tweaks:

- Boundary label: I think this is more accurate as `SchedulerOutput · scheduled-token counts · KV block IDs`. vllm-metal packs token IDs and builds `cu_seqlens`; Q is created later by `q_proj`.
- Ownership wording: `built on MLX-LM primitives` and `vllm-metal runtime orchestration` are more accurate than `reused unchanged` and `custom Metal`, since we own some loading/wrapping paths and MTP draft/verify is orchestration.
- PyPI rendering: I’m not fully sure, but I think the relative image path may break when README is rendered as package metadata. A stable absolute asset URL plus a one-sentence text summary would be safer.

WindChimeRan:
Thanks. accepted all the changes.

PR #664, opened 2026-08-31
[Bugfix] Make QK norm placement architecture-aware

### LINKED ISSUE #660: [Bug] prepare_sdpa_qkv hardcodes norm→RoPE ordering; Hunyuan needs RoPE→norm, and attribute spelling does not predict order
## Summary

`prepare_sdpa_qkv` applies the per-head Q/K norms **before** RoPE. That matches Qwen3 and Gemma3, but `hunyuan_v1_dense` applies them **after**. The pipeline hardcodes one order for every architecture, so Hunyuan is computed as `RoPE(RMSNorm(x))` where its reference is `RMSNorm(RoPE(x))`.

Identified by @WindChimeRan in review of #659. That PR fixes the separate, far larger bug where the norms were skipped entirely (97.19% → 0.1189% error); this issue tracks the ~0.12% ordering residual it leaves behind.

A correct fix needs an **explicit per-architecture ordering signal** — a name-based heuristic will not work, see below.

## The trap: attribute spelling does not predict ordering

The tempting fix is "`query_layernorm` spelling implies post-RoPE". That is false:

| Model | Spelling | Order |
| --- | --- | --- |
| `qwen3`, `gemma3_text` | `q_norm` / `k_norm` | norm → RoPE |
| `stablelm` | `q_layernorm` / `k_layernorm` | **norm → RoPE** |
| `lfm2`, `lfm2_moe` | `q_layernorm` / `k_layernorm` | **norm → RoPE** |
| `hunyuan_v1_dense` | `query_layernorm` / `key_layernorm` | **RoPE → norm** |

`stablelm` and `lfm2` use a non-standard *name* with the standard *order*. Hunyuan is the outlier on both axes, independently. Anything keying order off the name will silently mis-order StableLM and LFM2 the moment they reach this path.

(For reference, `stablelm` is currently blocked earlier by the missing `sm_scale` attribute — gap 1 of #656 — and `lfm2` fails `is_sdpa` because it uses `out_proj` rather than `o_proj`.)

## Confirming the ordering, bit-exactly

Rather than reading source, capture the real tensors from a live paged run and compare against both candidate orders recomputed offline.

**Method** (reproducible):

1. Instrument `prepare_sdpa_qkv` to dump `queries` immediately **before** the norm block, and again immediately **after** `apply_attention_rope`, as `float32` `.npz`.
2. Serve `mlx-community/Hunyuan-1.8B-Instruct-4bit` @ `5a121aef6911` with `VLLM_METAL_USE_PAGED_ATTENTION=1`, one short prompt, `max_tokens=1`.
3. Offline, load the pre-norm tensor, cast to **bfloat16** (the real compute dtype), transpose to `(B, H, L, D)`, and compute both candidates with the model's own `query_layernorm` and `rope` modules at `offset=0`.

**Result** — pipeline vs each candidate:

| pipeline vs | max | mean |
| --- | --- | --- |
| `RoPE(RMSNorm(x))` | **0.000000** | **0.00000000** |
| `RMSNorm(RoPE(x))` | 0.046875 | 0.00113948 |

Exact zero against `RoPE(RMSNorm(x))`. The pipeline is unambiguously applying the norm first.

**dtype matters, and gets this backwards if you use fp32.** Computing the references in `float32` while the pipeline runs `bfloat16` injects ~0.002 of dtype noise, which makes both candidates look equidistant and *understates* the ordering error ~20×. bf16 amplifies the ordering difference rather than damping it. Measure in bf16.

**Magnitude** (bf16, real captured prefill tensor, layer 0, vs the `mlx_lm` reference):

| | max abs | mean abs | relative |
| --- | --- | --- | --- |
| `main` before #659 (norm skipped) | 6.3838 | 0.931429 | 97.19% |
| with #659 (ordering residual) | 0.0469 | 0.001139 | **0.1189%** |

## It is large enough to change output — in isolated mlx_lm

Monkeypatching `hunyuan_v1_dense.Attention.__call__` to swap only the order, on the prompt `"In summary, the animal mentioned above is the"`:

| Comparison | max logit delta | argmax same |
| --- | --- | --- |
| stock vs stock (determinism control) | 0.00000 | yes |
| stock vs correct-order re-implementation | 0.00000 | yes |
| stock vs wrong order | **0.80469** | **no** |

Top-1/top-2 margin is 0.188, so a 0.805 perturbation flips it. The two zero-deltas confirm the harness is exact.

## Open question — why this does not surface in the paged pipeline

Despite the above, **I could not produce a vllm-metal output difference.** Ten constructed cases, all greedy, all comparing the paged path against the `mlx_lm` reference:

- the 5 golden prompts from `tests/test_hunyuan_golden.py` — identical
- a 3,089-token long-context prompt — identical, common prefix 10/10
- 4 deliberately low-margin prompts, including the one that flips in isolated `mlx_lm` — identical
- **the corrected order implemented directly in `prepare_sdpa_qkv`** — output unchanged from the wrong order

So the paged path computes the wrong order (proven bit-exactly) yet produces the reference's tokens.

**I do not have an explanation and am not going to guess at one.** A plausible line of investigation is whether the paged Metal attention kernel's numerics differ from `mlx_lm`'s dense SDPA enough that the two pipelines are not comparable at a 0.19-logit margin — but I have not tested that, and it should not be assumed.

Resolving this is the prerequisite for everything else here: until it is understood, there is no way to write a regression test that fails on the bug, and no way to verify that a fix actually fixed anything end-to-end.

## Why there is no test for this yet

A golden-token test cannot currently express this bug. All ten cases above pass with the defect present, so such a test would be green today and `xfail` would XPASS and break CI.

**Recommendation: do not add more golden-token cases for this until the open question above is resolved.** I burned ten constructed cases finding that out. The likely shape of a real regression test is white-box — asserting the *operation order* directly, or asserting `prepare_sdpa_qkv`'s output against a reference computed in bf16 — rather than comparing sampled tokens.

## Environment

Apple M2 / 8 GB, macOS 26.6.2, Python 3.12.2 (arm64), vllm 0.28.0+cpu, mlx 0.32.0, mlx-lm 0.31.3, vllm-metal at `main` + #659. Checkpoint `mlx-community/Hunyuan-1.8B-Instruct-4bit` @ `5a121aef6911ff1c99a3b3f1f7b028414cdfa618`.

WindChimeRan:
## Summary

Fixes #660.

Hunyuan dense applies its per-head Q/K norms after RoPE, while the paged SDPA path hard code every architecture's norms before RoPE. This PR makes that placement an explicit attention contract and registers Hunyuan's non-default order.

```
# hunyuan
Before:        RoPE(RMSNorm(x))
After:         RMSNorm(RoPE(x))
```

Before:

```
mlx-lm transformer layer
  → SDPAPagedAttentionWrapper
  → sdpa_forward
  → prepare_sdpa_qkv
      project Q/K/V
      normalize / apply RoPE
  → write K/V into paged cache
  → Metal attention kernel
  → output projection
```
After
```
mlx-lm transformer layer
  → SDPAPagedAttentionWrapper
  → sdpa_forward
  → prepare_sdpa_qkv
      explicit attention contract and registers Hunyuan's non-default order.
  → write K/V into paged cache
  → Metal attention kernel
  → output projection
```

## What changed

- Add a small immutable paged-attention contract with `BEFORE_ROPE` as the default.
- Resolve the contract once when an attention module is wrapped, then pass it to Q/K preparation.
- Register `mlx_lm.models.hunyuan_v1_dense` with `AFTER_ROPE`.
- Add a deterministic BF16 numeric regression using MLX-LM's actual Hunyuan `Attention` class.

The existing norm-name resolver is unchanged. Unregistered architectures keep the exact previous norm-before-RoPE path.

## Regression coverage

The numeric test uses non-uniform learned norm weights so the two orders cannot commute. Before the Hunyuan registration it fails with a Q max-absolute error of `1.05859375`; with the registration, both prepared Q and K match `RMSNorm(RoPE(x))` exactly (`0.0` max-absolute error).

## Validation

- `pytest tests/test_attention_sdpa.py tests/test_hunyuan_qk_norm.py -q` — 27 passed
- `mlx-community/Hunyuan-1.8B-Instruct-4bit` at `5a121aef6911ff1c99a3b3f1f7b028414cdfa618`: six greedy prompts, 10 tokens each, paged output token-identical to the non-paged MLX reference


LxYuan0420 (approved):
Nice fix. I’m not a big fan of carrying our own model-behavior registry, but since vllm-metal decomposes `mlx-lm` attention, we need an explicit contract to preserve each model’s math. We should revisit this if `mlx-lm` exposes the norm/RoPE order directly later.

vllm_metal/attention/model_patches/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    AttentionContract,
+    QKNormPlacement,
+    register_attention_contract,
+    resolve_attention_contract,
+)
+from vllm_metal.attention.model_patches.hunyuan_v1_dense import (
+    register as _register_hunyuan_v1_dense,
+)
+
+_register_hunyuan_v1_dense()
```
LxYuan0420: I thought about this more. This would be easier to read and extend as one small `attention_contracts.py` table instead of a registration package. A frozen `AttentionContract`, `Literal["before_rope", "after_rope"]`, and `_CUSTOM_ATTENTION_CONTRACT_MAPPINGS` keeps the contract explicit, avoids import-time registration side effects, and still lets us add future model-specific attention behavior without touching `sdpa.py`.

Optional suggested shape:

```python
QKNormPlacement = Literal["before_rope", "after_rope"]

@dataclass(frozen=True, slots=True)
class AttentionContract:
    qk_norm_placement: QKNormPlacement = "before_rope"


_DEFAULT_ATTENTION_CONTRACT = AttentionContract()

_CUSTOM_ATTENTION_CONTRACT_MAPPINGS = {
    "mlx_lm.models.hunyuan_v1_dense": AttentionContract(
        qk_norm_placement="after_rope",
    ),
}


def attention_contract_for(module: object) -> AttentionContract:
    return _CUSTOM_ATTENTION_CONTRACT_MAPPINGS.get(
        type(module).__module__,
        _DEFAULT_ATTENTION_CONTRACT,
    )
```

WindChimeRan: Thanks, this is much cleaner! I applied the suggested change, except for: I prefer Enum > Literal

---

Will merge the release PR first, then use this PR to test auto release refresh mechanism.

Issue #644, opened 2026-08-26
Add Nemotron-H (Mamba-2 + MoE hybrid) paged attention support on Apple Silicon


ricky-chaoju:
vllm-metal cannot serve Nemotron-H models on the default configuration. `use_paged_attention` defaults to True, and on that path a nemotron_h checkpoint patches zero of its layers, logs LOADED, accepts a request, then dies inside mlx-lm on the first forward. Turning the default off forces `VLLM_METAL_MEMORY_FRACTION=auto` and gives up every paged-only feature.

This is not a repeat of #57 or #31. Those shipped GDN for Qwen3-Next. Mamba-2 is a different recurrence, and Nemotron-H mixes three kinds of layer rather than two: 23 Mamba-2, 23 MoE, 6 attention out of 52. Our `num_layers - num_sdpa_layers` idiom reports 46 recurrent layers for it when only 23 are.

mlx-lm ships a complete `nemotron_h.py` whose `ssm_update` is already a Metal kernel, and vLLM 0.27.1 registers `NemotronHForCausalLM`. I checked the three things that could have killed this on `mlx-community/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-4bit`: vLLM produces a usable MAMBA2 spec, our `GDNPagedStateCache` already allocates the right shapes at Mamba-2 dimensions, and the unmodified mlx-lm mixer runs bit-identically against externally owned per-slot state. Numbers go in the PRs.

## PR Stack

- [x] PR 1: Own hybrid geometry behind a family owner. Typed layer plan and state geometry, zero behavior change. https://github.com/vllm-project/vllm-metal/pull/668
- [x] PR 2: Derive hybrid routing from vLLM model config, plus the state-family table and its rejection. https://github.com/vllm-project/vllm-metal/pull/696
- [x] PR 3: Run Mamba-2 mixers on paged recurrent state. Wrapper only, not wired in yet. #710
- [x] PR 4: Serve Nemotron-H hybrids through the paged runtime. Family module, docs row, parity evidence. #735

## Design

- The state family should be a loader table inside the existing `is_hybrid` arm, shaped like `_ENCODER_BACKEND_LOADERS`, not a fourth branch in `build_paged_attention_runtime`. Choosing between MHA, MLA and hybrid stays with #450.
- Layer topology should be a typed plan that answers which kind a layer is, not a boolean that leaves the rest to a default branch.
[...]

PR #607, opened 2026-08-14
Add paged MLA/KDA serving for Ling-3.0 (Bailing V3)

FENP:
## Summary

- Add paged serving support for Ling-3.0 Bailing V3 hybrid models.
- Serve MLA layers with the paged latent cache and KDA layers with scheduler-managed recurrent state.
- Reuse the generic hybrid runtime for continuous batching and align-mode prefix caching.
- Register KDA as a state family and restrict model construction to `BailingMoeV3ForCausalLM` with the supported KDA configuration.
- Preserve `architectures` from the Hugging Face config when MLX-LM's `ModelArgs` omits it.
- Document the supported Ling-3.0 checkpoint formats and dependencies.

## Dependencies

- vLLM-Metal #723 updated the MLX and MLX-LM pins and contains the repository-wide compatibility changes required by this PR. This PR is rebased on the merged change.
- vLLM Bailing V3 model support was merged in https://github.com/vllm-project/vllm/pull/51045 and released in [vLLM 0.28.0](https://github.com/vllm-project/vllm/releases/tag/v0.28.0), which is the version currently pinned by vLLM-Metal.
- MLX-LM Bailing V3 support was merged in https://github.com/ml-explore/mlx-lm/pull/1711. PR #723 pins the merged commit `9e6acca691e64d6d8bb808c328fcdea459099cca` until a suitable MLX-LM release includes it.

## Validation

Tested on an Apple M3 Pro with 18 GB unified memory, macOS 14.4.1, and Python 3.12.12.

Full non-slow test suite:

```bash
VLLM_METAL_BUILD_FROM_SOURCE=1 \
  .venv/bin/python -m pytest -p no:cacheprovider \
  -m "not slow" tests -q
```

Result: `2008 passed, 18 skipped, 40 deselected`.

After the latest rebase, the five affected test suites passed with `140 passed, 1 skipped`.

`ruff check`, `ruff format --check`, `mypy`, and `git diff --check` also passed.

Full-model serving used the 24-layer Ling-3.0 Tiny checkpoint converted from the official BF16 checkpoint to MLX-native MXFP8. The server was launched with paged attention and thinking disabled:

```bash
VLLM_METAL_BUILD_FROM_SOURCE=1 \
VLLM_METAL_USE_PAGED_ATTENTION=1 \
  .venv/bin/vllm serve /path/to/ling_v3_tiny_mxfp8_selective \
  --host 127.0.0.1 \
  --port 8000 \
  --trust-remote-code \
[...]

PR #598, opened 2026-08-11
LFM2/LFM2.5 paged serving with prefix caching

WindChimeRan:
## Summary

(I've been working on this for a very long time, because I wanna settle the shared state management & align mode prefix caching from Qwen3.5+ first. Now they are all in a usable shape, so I pick it up. LFM is a good model for mac inference! )

Adds paged text generation for LFM2 and LFM2.5 dense and MoE models, with prefix caching. 

* **supported**: Prefix caching defaults to `align`(same with upstream vllm). 
* **unsupported**: `all` mode and speculative decoding remain unsupported.

* commit 1 is the full LFM support (conv layer, prefix caching .etc)
* commit 2 is pure renaming, because now we have two kinds of state now. I checked the upstream vllm naming, and use state for shared worker machinery, and retain GDN / ShortConv for the concrete implementations.

## Performance

Serving output throughput is generated output tokens divided by total benchmark wall time, including prefill and scheduling. These runs do not isolate decode-kernel speed.

Sonnet, nominal **1024 input + 128 output tokens**, **100 requests/run**, rate **10/s**, concurrency **32**. The figure shows medians of two fresh-server runs on an M5 Pro (64 GiB); all 1,600 requests succeeded. Memory fraction: 0.3 for the small pair, 0.5 for the 8B pair.

benchmark_comparison

<details>
<summary>Reproduce the performance comparison</summary>

Run from revision `761e583`, with the vllm-metal environment active. Tested with macOS 26.6.2, vLLM 0.28.0+cpu, MLX 0.32.0, MLX-LM `254d153`, mlx-vlm 0.6.4, and Transformers 5.12.1. Native checkpoint dtypes are preserved.

Select one row below, then run with `APC=off` and `APC=on`, twice each. Restart the server for every run; reverse model order in the second repetition. No client warmup or discarded measured runs.

| `MODEL` | `REVISION` | `FRACTION` |
|---|---|---:|
| `Qwen/Qwen3.5-0.8B` | `2fc06364715b967f1860aea9cf38778875588b17` | 0.3 |
| `LiquidAI/LFM2.5-1.2B-Instruct` | `df58c174f05ff733f83f8cae10ea9298224c8006` | 0.3 |
| `Qwen/Qwen3-8B` | `b968826d9c46dd6066d109eabc6255188de91218` | 0.5 |
| `LiquidAI/LFM2.5-8B-A1B` | `b9aebfcbe28b6cb374042f495d733037550ab146` | 0.5 |

```bash
[...]

PR #668, opened 2026-09-01
Own hybrid geometry behind a family owner

ricky-chaoju:
Groundwork for the Nemotron-H stack in #644, PR 1 of 4. The hybrid runtime's geometry is currently ten undeclared attributes that lifecycle mutates onto the runner, and `cache_policy` sizes the linear side as `num_layers` minus the SDPA count. That arithmetic breaks as soon as a model has a third kind of layer. On the 52-layer Nemotron checkpoint it reports 46 recurrent layers when only 23 are.

Because of that, this PR moves the geometry behind one owner, a typed layer plan built by `families/gdn.py`, in the same shape as the encoder pooling loaders from #608. Lifecycle shrinks to a four-line route. The loader table comes with the next PR, so lifecycle imports the GDN builder directly for now.

For evidence, the main one is an interleaved A/B on `Qwen/Qwen3.5-0.8B`. Greedy token ids are bit-identical and `per_block_bytes` is identical. `num_blocks` wobbles a few blocks on both arms since overhead is measured at profile time. The full suite gives 1921 passed with 15 new tests, and error messages on malformed configs got stricter, each pinned by a full-string test.

#598 and #607 both touch these hunks, and #607 passes constructor kwargs this PR deletes, so whichever lands second rebases. Happy to be the one who does. The `MambaSpec` shape and the attention-side wiring stay put until the table lands.

### DIFF
--- a/tests/attention/test_hybrid_runtime_plan.py
+++ b/tests/attention/test_hybrid_runtime_plan.py
@@ -62,8 +62,8 @@ class _FakeModel(nn.Module):
     def __init__(self, roles: str) -> None:
         super().__init__()
         self.layers = [
-            _Layer(_FakeGDN() if k == "s" else _FakeSDPA(), linear=(k == "s"))
-            for k in roles
+            _Layer(_FakeGDN() if role == "s" else _FakeSDPA(), linear=(role == "s"))
+            for role in roles
         ]
 
 
@@ -119,15 +119,6 @@ class TestGdnPlanDecision:
         assert plan.geometry.num_v_heads == 4
         assert plan.geometry.value_head_dim == 16
         assert plan.geometry.key_head_dim == 32
-
-    def test_family_policy_matches_the_gdn_runtime_contract(self) -> None:
-        plan = build_gdn_hybrid_plan(GDN_ARGS, 8)
-
-        assert plan.family.label == "gdn"
-        assert plan.family.wrapper_cls is GDNPagedAttentionWrapper
-        assert plan.family.mamba_type is MambaAttentionBackendEnum.GDN_ATTN
-        assert plan.family.recurrent_dtype is torch.float32
-        assert plan.family.supported_cache_modes == ("none", "align")
 
 
 class TestGdnPlanRejection:
--- a/tests/attention/test_hybrid_runtime_plan.py
+++ b/tests/attention/test_hybrid_runtime_plan.py
@@ -82,7 +82,7 @@ def _make_tiny_plan() -> HybridRuntimePlan:
 
 def _make_runtime() -> HybridPagedAttentionRuntime:
     return HybridPagedAttentionRuntime(
-        plan=_make_tiny_plan(),
+        hybrid_plan=_make_tiny_plan(),
         max_num_seqs=2,
         num_kv_heads=1,
         head_dim=4,
@@ -93,10 +93,8 @@ def _make_runtime() -> HybridPagedAttentionRuntime:
 
 class TestGdnPlanDecision:
     def test_topology_follows_the_interval_rule(self) -> None:
-        # Arrange / Act
         plan = build_gdn_hybrid_plan(GDN_ARGS, 8)
 
-        # Assert
         assert plan.layers.attention_indices == (3, 7)
         assert plan.layers.state_indices == (0, 1, 2, 4, 5, 6)
         assert plan.layers.num_attention == 2
@@ -113,10 +111,9 @@ class TestGdnPlanDecision:
         )
 
     def test_geometry_packs_qk_and_v_into_the_conv_stream(self) -> None:
-        # Arrange / Act
         plan = build_gdn_hybrid_plan(GDN_ARGS, 8)
 
-        # Assert — conv_dim = 2*32*2 + 4*16 = 192, hand-written.
+        # conv_dim = 2*32*2 + 4*16 = 192, hand-written.
         assert plan.geometry.conv_kernel_dim == 3
         assert plan.geometry.conv_dim == 192
         assert plan.geometry.num_v_heads == 4
@@ -124,10 +121,8 @@ class TestGdnPlanDecision:
         assert plan.geometry.key_head_dim == 32
 
     def test_family_policy_matches_the_gdn_runtime_contract(self) -> None:
-        # Arrange / Act
         plan = build_gdn_hybrid_plan(GDN_ARGS, 8)
 
-        # Assert
         assert plan.family.label == "gdn"
         assert plan.family.wrapper_cls is GDNPagedAttentionWrapper
         assert plan.family.mamba_type is MambaAttentionBackendEnum.GDN_ATTN
@@ -140,53 +135,34 @@ class TestGdnPlanRejection:
         "missing", ["linear_num_key_heads", "linear_conv_kernel_dim"]
     )
     def test_omitted_linear_dim_rejects_with_the_key_name(self, missing: str) -> None:
-        # Arrange
         args = {k: v for k, v in GDN_ARGS.items() if k != missing}
         expected = f"GDN hybrid model args are missing a usable {missing!r}: got None."
 
-        # Act / Assert
         with pytest.raises(ValueError) as excinfo:
             build_gdn_hybrid_plan(args, 8)
         assert str(excinfo.value) == expected
 
     def test_non_positive_layer_count_rejects(self) -> None:
-        # Arrange
-        expected = "GDN hybrid requires a positive layer count, got 0."
-
-        # Act / Assert
-        with pytest.raises(ValueError) as excinfo:
+        with pytest.raises(ValueError, match="positive layer count"):
             build_gdn_hybrid_plan(GDN_ARGS, 0)
-        assert str(excinfo.value) == expected
 
 
 class TestLayerPlanDispatch:
-    def test_kind_of_rejects_out_of_range_layers(self) -> None:
-        # Arrange
+    def test_layer_kind_rejects_out_of_range_layers(self) -> None:
         plan = build_gdn_hybrid_plan(GDN_ARGS, 8)
-        expected = "hybrid layer plan covers layers 0..7, got layer 8"
 
-        # Act / Assert
-        with pytest.raises(ValueError) as excinfo:
-            plan.layers.kind_of(8)
-        assert str(excinfo.value) == expected
+        with pytest.raises(ValueError, match="got layer 8"):
+            plan.layers.layer_kind(8)
 
     def test_cache_index_rejects_the_wrong_kind(self) -> None:
-        # Arrange — layer 0 is a state layer.
         plan = build_gdn_hybrid_plan(GDN_ARGS, 8)
-        expected = (
-            "hybrid layer plan has no attention cache index for layer 0; "
-            "attention layers are (3, 7)"
-        )
 
-        # Act / Assert
-        with pytest.raises(ValueError) as excinfo:
+        with pytest.raises(ValueError, match="no attention cache index for layer 0"):
             plan.layers.attention_cache_index(0)
-        assert str(excinfo.value) == expected
 
 
 class TestStateCacheSpec:
     def test_spec_threads_geometry_family_and_mode(self) -> None:
-        # Arrange
         plan = make_gdn_hybrid_plan(
             4,
             [1, 3],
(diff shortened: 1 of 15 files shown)


vllm_metal/attention/runtime/families/gdn.py
```diff
@@ -0,0 +1,83 @@
...
+    RecurrentStateGeometry,
+    StateFamilySpec,
+)
+
+_INTERVAL_KEY = "full_attention_interval"
+# Scheduler-side mamba caching strategies the GDN state path implements.
+_SUPPORTED_CACHE_MODES = ("none", "align")
+
+
+def build_gdn_hybrid_plan(
```
LxYuan0420, line 111: Move the family config and typed model-args parsing above the public builder. `model_args` is the third-party boundary, so validation belongs here, but parse it once into a small `GDNHybridConfig` dataclass instead of repeatedly calling `_require_positive_int()` across the file.

vllm_metal/attention/runtime/hybrid_plan.py
```diff
@@ -0,0 +1,158 @@
...
+        cache_idx: int,
+        state_cache: GDNPagedStateCache,
+    ) -> None: ...
+
+    def rebind_state_cache(
+        self, state_cache: GDNPagedStateCache, *, cache_idx: int
+    ) -> None: ...
+
+
+@dataclass(frozen=True, slots=True)
```
LxYuan0420, line 44: `HybridLayerPlan` should have one source of truth. Today `kinds`, `attention_indices`, and `state_indices` can disagree. Accept layer kinds and derive the index tuples internally, or validate them in `__post_init__`; also use clearer names like `from_layer_kinds()` and `layer_kind()`.

vllm_metal/attention/runtime/hybrid.py
```diff
@@ -79,62 +43,46 @@ class HybridPagedAttentionRuntime(PagedAttentionRuntimeBase):
     def __init__(
```
LxYuan0420, line 43: Use descriptive ownership names here: `hybrid_plan`, `state_family`, and `state_geometry`. After assigning `self._hybrid_plan`, use that consistently instead of switching between the constructor argument and instance state.

vllm_metal/v1/cache_policy.py
```diff
@@ -67,6 +69,17 @@ def _align_state_pool_count(num_linear_layers: int, num_sdpa_layers: int) -> int
     return num_linear_layers
 
 
+def _require_runner_hybrid_plan(runner: MetalModelRunner) -> HybridRuntimePlan:
```
LxYuan0420: This top-level require helper plus `_require_hybrid_plan()` wrapper is defensive plumbing. Keep the invariant check in one owner, preferably a private `ModelCachePolicy._hybrid_plan()` method, and make the cache path read as: get plan, inspect layer role, emit spec.

vllm_metal/v1/cache_policy.py
```diff
@@ -365,27 +381,33 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
             num_spec_layers, _ = self._runner._yoco_cache_mapping
         specs: dict[str, KVCacheSpec] = {}
         use_deferred_mha_layout = self._uses_deferred_mha_layout()
+        hybrid_plan = self._require_hybrid_plan() if self._runner.is_hybrid else None
         for layer_idx in range(num_spec_layers):
```
LxYuan0420: Compute the layer role once and dispatch clearly. The repeated `kind_of()` calls and impossible fallback make this harder to read. Handle the roles PR 1 supports now; add `stateless` in PR 2 when Nemotron-H has a real consumer.

vllm_metal/v1/model_runner.py
```diff
@@ -459,6 +460,11 @@ def __init__(self, vllm_config: VllmConfig):
         # present: it is consulted on the KV-sizing path, which runs for every
         # model, not only the YOCO ones that install a real mapping.
         self._yoco_cache_mapping: tuple[int, dict[int, int]] | None = None
+        # Hybrid layer split and state family, resolved once by ModelLifecycle
```
LxYuan0420: This comment is doing too much. The field name should carry most of the meaning; keep one short sentence and avoid adding heavy architecture notes to the runner.

tests/attention/test_hybrid_runtime_plan.py
```diff
@@ -0,0 +1,324 @@
...
+        plan=_make_tiny_plan(),
+        max_num_seqs=2,
+        num_kv_heads=1,
+        head_dim=4,
+        block_size=4,
+        dtype=mx.float32,
+    )
+
+
+class TestGdnPlanDecision:
```
LxYuan0420, line 94: Trim the test scaffolding. Remove `Arrange / Act / Assert` comments and avoid pinning full internal error strings unless the wording is user-facing. These tests should prove topology, spec sizing, and wrapper placement.

LxYuan0420 (commented):
This is the right direction, but I would tighten the contract before merging. Moving hybrid geometry into a typed owner is better than lifecycle mutating many runner fields, but the current shape still has inconsistent topology construction, defensive helper plumbing, and naming that makes the flow harder to read.

ricky-chaoju:
Done. I ended up deleting from_kinds rather than renaming it; the constructor takes kinds directly now, so the factory had nothing left to do. Kept the full-string asserts only where a user can actually hit the message; the internal ones match fragments now.

vllm_metal/v1/model_runner.py
ricky-chaoju: ok,will watch it.

vllm_metal/attention/runtime/families/gdn.py
```diff
@@ -0,0 +1,99 @@
...
+    def _require_positive_int(model_args: Mapping[str, Any], key: str) -> int:
+        value = model_args.get(key)
+        if not isinstance(value, int) or isinstance(value, bool) or value <= 0:
+            raise ValueError(
+                f"GDN hybrid model args are missing a usable {key!r}: got {value!r}."
+            )
+        return int(value)
+
+
+def build_gdn_hybrid_plan(
```
LxYuan0420, line 111: I noticed this still accepts `full_attention_interval=1` and `full_attention_interval > num_layers`, which creates “hybrid” plans with zero state layers or zero SDPA layers instead of failing at the GDN config boundary.

vllm_metal/attention/runtime/hybrid_plan.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def attention_cache_index(self, layer_idx: int) -> int:
+        """Return the compact KV cache ordinal owned by an attention layer."""
+        return self._cache_index(self.attention_indices, layer_idx, ATTENTION_LAYER)
+
+    def state_cache_index(self, layer_idx: int) -> int:
+        """Return the compact state ordinal owned by a state layer."""
+        return self._cache_index(self.state_indices, layer_idx, STATE_LAYER)
+
+    @staticmethod
```
LxYuan0420: `_cache_index()` feels like private helper plumbing inside a small value object. Keep this direct and readable, or expose it as `cache_index()` if it is part of the object contract.

vllm_metal/attention/runtime/families/gdn.py
```diff
@@ -0,0 +1,99 @@
...
+    model_args: Mapping[str, Any], num_layers: int
+) -> HybridRuntimePlan:
+    """Resolve GDN layer topology and recurrent geometry from model args."""
+    config = GDNHybridConfig.from_model_args(model_args)
```
LxYuan0420: Use more descriptive names here: `gdn_config` and `layer_roles` read clearer than `config` and `kinds`. This is the family-owned topology builder, so the variable names should explain the model layout directly.

vllm_metal/attention/runtime/families/gdn.py
```diff
@@ -0,0 +1,99 @@
...
+    """GDN dims parsed once from the third-party model args boundary."""
+
+    full_attention_interval: int
+    num_key_heads: int
+    num_value_heads: int
+    key_head_dim: int
+    value_head_dim: int
+    conv_kernel_dim: int
+
+    @classmethod
```
LxYuan0420, line 39: This parsing path is still noisier than it needs to be. `model_args` is the boundary, so validation belongs here, but the private static helper makes the dataclass harder to read. Prefer a direct `GDNHybridConfig.from_model_args()` flow, or one small module-level reader if we really need shared parsing.

vllm_metal/attention/runtime/hybrid_plan.py
```diff
@@ -0,0 +1,157 @@
...
+from typing import Any, Literal, Protocol, TypeAlias
+
+import mlx.nn as nn
+import torch
+from vllm.v1.attention.backends.registry import MambaAttentionBackendEnum
+from vllm.v1.kv_cache_interface import MambaSpec
+
+from vllm_metal.attention.caches.gdn_cache import GDNPagedStateCache
+
+LayerKind: TypeAlias = Literal["attention", "state"]
```
LxYuan0420: `LayerKind` / `kinds` is still ambiguous. I would call this `LayerRole` with a `layer_roles` field, because consumers are dispatching cache and runtime ownership by layer role.

vllm_metal/v1/cache_policy.py
```diff
@@ -365,25 +376,23 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
             num_spec_layers, _ = self._runner._yoco_cache_mapping
         specs: dict[str, KVCacheSpec] = {}
         use_deferred_mha_layout = self._uses_deferred_mha_layout()
+        hybrid_plan = self._hybrid_plan() if self._runner.is_hybrid else None
```
LxYuan0420: This loop still mixes generic cache policy with hybrid dispatch details. Since `HybridRuntimePlan` owns layer roles, make this read as: get plan, get layer role, emit state spec or attention spec. Avoid repeated `hybrid_plan is not None` checks inside the loop.

vllm_metal/v1/cache_policy.py
```diff
@@ -1292,9 +1292,14 @@ def _hybrid_align_state_bytes_per_block(self) -> int:
             return 0
         if runner.cache_config.mamba_cache_mode != "align":
             return 0
-        num_linear = runner.num_layers - len(runner.sdpa_layer_indices)
-        pools = _align_state_pool_count(num_linear, len(runner.sdpa_layer_indices))
-        return runner.linear_cache_bytes_per_slot() * pools // num_linear
+        # linear_cache_bytes_per_slot routes through ModelCachePolicy, whose
+        # _hybrid_plan() enforces the plan-resolved invariant.
```
LxYuan0420: The comment plus `assert plan is not None` is a smell. If `linear_cache_bytes_per_slot()` already proves the hybrid-plan invariant, keep using the same resolved plan path instead of asserting a second time on `runner.hybrid_runtime_plan`.

tests/stub_runner.py
```diff
@@ -202,3 +212,48 @@ def make_gemma4_mixed_mha_runner(
         head_dim_per_layer=head_dim_per_layer,
         sliding_window_per_layer=sliding_windows,
     )
+
+
+# The production family spec, resolved once through the public builder (the
```
LxYuan0420: This test helper currently builds a GDN family spec from a degenerate no-state plan. Once the GDN builder rejects invalid hybrid layouts, this should use a valid tiny hybrid config instead of relying on a shape production should not accept.

tests/attention/test_hybrid_runtime_plan.py
```diff
@@ -0,0 +1,278 @@
...
+        plan = build_gdn_hybrid_plan(GDN_ARGS, 8)
+
+        assert plan.family.label == "gdn"
+        assert plan.family.wrapper_cls is GDNPagedAttentionWrapper
+        assert plan.family.mamba_type is MambaAttentionBackendEnum.GDN_ATTN
+        assert plan.family.recurrent_dtype is torch.float32
+        assert plan.family.supported_cache_modes == ("none", "align")
+
+
+class TestGdnPlanRejection:
```
LxYuan0420, line 124: Add focused rejection coverage for `full_attention_interval=1` and `full_attention_interval > num_layers`. Those are the real malformed hybrid cases that still pass today.

vllm_metal/v1/cache_policy.py
```diff
@@ -606,7 +615,7 @@ def _adopt_hybrid_scheduler_group(
             kv_cache_config,
             tuple(
                 f"layers.{layer_idx}.self_attn"
-                for layer_idx in sorted(self._runner.sdpa_layer_indices)
+                for layer_idx in self._hybrid_plan().layers.attention_indices
```
LxYuan0420: `_adopt_hybrid_scheduler_group()` calls `_hybrid_plan()` several times and rebuilds names from the plan each time. Fetch `hybrid_plan` and `layer_plan` once at the top, then use those locals. It keeps the scheduler-layout path easier to audit.

LxYuan0420 (approved):
Hey @ricky-chaoju, I pushed a few commits to tighten this up before merge. Please review it as well when you have a chance.

ricky-chaoju:
All good. Thanks @LxYuan0420 !

PR #669, opened 2026-09-01
[Docs] Document speculative ingest chunk setting

POWEHI2023:
## Summary

- document `VLLM_METAL_SPEC_INGEST_CHUNK`
- explain its default, purpose, and how to restore single-forward ingest

## Testing

- `git diff --check`
- verified all registered `VLLM_METAL_*` and `VLLM_MLX_*` variables appear in the configuration reference


LxYuan0420 (approved):
Thanks, this docs update is valid. I missed out the PR that added this env var, so I will review that separately. Approving this PR for now.

PR #672, opened 2026-09-02
Reserve null block in non-paged scheduler capacity

TheVerdman:
While testing #644, I found that a max-length request could remain in the waiting queue indefinitely on the non-paged path. `single_sequence_estimate` reports exactly enough block-aligned cache memory for the request, but vLLM's `BlockPool` reserves one physical block as its `null_block`, leaving the scheduler one block short.

This adds one cache block to the reported capacity so the original one-sequence capacity remains available after the null block is reserved. Paged-attention reporting is unchanged. This complements #513, which handled the same reservation during `max_model_len` auto-fit.

I added a focused regression test and ran the worker/platform tests plus the full non-slow suite locally (`1848 passed, 18 skipped, 39 deselected`).

### DIFF
--- a/tests/test_v1_worker.py
+++ b/tests/test_v1_worker.py
@@ -88,7 +88,7 @@ class TestWorkerRunnerBoundaryDelegation:
         worker.get_cache_block_size_bytes.assert_called_once_with()
 
     def test_determine_available_memory_single_sequence_mode(self) -> None:
-        """Test MLX path returns one max-length sequence estimate (PR #229)."""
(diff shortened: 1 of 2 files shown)


LxYuan0420 (approved):
Thanks!

 Note: Once `vllm-project/vllm#48724` lands in our pinned vLLM, we should revisit the null-block compat code. `#513` may become removable, while `#672` should be rechecked against upstream’s final BlockPool behavior.

Issue #655, opened 2026-08-30
[Bug] Falcon-H1 crashes EngineCore on first request: 'OffsetCache' object is not subscriptable (hybrid models need a nested CacheList)


ritunjaym:
## Summary

Serving a Falcon-H1 model (`FalconH1ForCausalLM`) — e.g. `mlx-community/Falcon-H1-1.5B-Instruct-4bit` — loads successfully and reaches `Application startup complete`, but the **first completion request crashes EngineCore** with `TypeError: 'OffsetCache' object is not subscriptable`.

`vllm-metal` passes a flat list of `OffsetCache` objects as the per-layer cache, but Falcon-H1 is a Mamba-2 + attention hybrid whose forward pass expects each layer's cache to be a **two-element `CacheList`** (SSM state + KV cache) and indexes it as `cache[0][0]` / `cache[0][1]`.

Related to #654 (LFM2), which shares the same underlying cause — `OffsetCache` is an attention-only stand-in handed to every layer — but fails differently: #654 is a mask **type** mismatch, this is a cache **shape** mismatch. Falcon-H1 crashes strictly earlier and never reaches the mask path in #654.

## Environment

- **vllm-metal** 0.3.0.dev20260829020049 (prebuilt wheel via `install.sh`); repo at `a9184aa`
- **vllm** 0.28.0+cpu
- **mlx** 0.32.0, **mlx-lm** 0.31.3, **mlx-metal** 0.32.0, **transformers** 5.12.1
- macOS 26.6.2 (25G83), Apple M2 / 8 GB unified memory
- Python 3.12.2 (arm64)
- `VLLM_METAL_USE_PAGED_ATTENTION=1`, `VLLM_METAL_MEMORY_FRACTION=0.8`

Model: `mlx-community/Falcon-H1-1.5B-Instruct-4bit` @ `6f5e4f6879b43846b7e960a5e68425f3d5d8c801`
(`model_type: falcon_h1`, 24 layers, 4-bit affine group 64)

## Reproduction

```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 \
  VLLM_METAL_MEMORY_FRACTION=0.8 \
  vllm serve mlx-community/Falcon-H1-1.5B-Instruct-4bit \
    --revision 6f5e4f6879b43846b7e960a5e68425f3d5d8c801 \
    --max-model-len 512 --max-num-batched-tokens 64
```

```bash
curl -s -X POST http://localhost:8000/v1/completions \
[...]

PR #674, opened 2026-09-02
[Bugfix] Batch nested CacheList caches on the non-paged path

### LINKED ISSUE #679: [Bug] Non-paged path: first batched decode step crashes EngineCore for CacheList models (Falcon-H1 & friends): Unsupported cache type for batching
## Summary

On the non-paged MLX cache path (`VLLM_METAL_USE_PAGED_ATTENTION=0`), a model whose `make_cache` returns a `CacheList` per layer serves a single request correctly, but the first step that batches two or more decode requests kills EngineCore:

```
File "vllm_metal/v1/model_runner.py", line 1117, in _batched_decode
TypeError: Unsupported cache type for batching: CacheList
```

Every request after that gets `EngineDeadError`. Affected mlx_lm families (a layer owns several stateful modules and indexes `cache[0]` / `cache[1]`): `falcon_h1`, `deepseek_v32`, `longcat_flash`, `baichuan_m1`. Since the non-paged path is the documented fallback for these hybrids while paged support lands (#655, #644), this is the difference between the fallback being usable and it being single-request-only.

## Reproduction

```bash
VLLM_METAL_USE_PAGED_ATTENTION=0 VLLM_METAL_MEMORY_FRACTION=auto \
  vllm serve mlx-community/Falcon-H1-1.5B-Instruct-4bit --revision 6f5e4f68 --max-model-len 8192
```

One `/v1/completions` request at a time works. Send two concurrently (so one decode step batches both) and the engine dies with the traceback above; both requests return 500. `--max-model-len` must be large enough that two requests are admitted at once — at 512 the second request instead waits forever, which is #672.

## Root cause

`mlx_lm.models.falcon_h1.Model.make_cache` returns `CacheList(ArraysCache(size=2), KVCache())` per layer (Mamba-2 mixer and attention run in parallel). The non-paged merge/extract helpers in `vllm_metal/v1/contiguous_cache.py` dispatch on `ArraysCache`, `RotatingKVCache` and `KVCache` only, so the container type raises when requests are merged into a batch.

## Environment

Apple M1 Pro, 16 GiB, macOS 27.0, Python 3.12.14, vllm 0.28.0+cpu, mlx 0.32.0, mlx_lm 0.31.3, vllm-metal source install at 07f1987 (`VLLM_METAL_BUILD_FROM_SOURCE=1`), model `mlx-community/Falcon-H1-1.5B-Instruct-4bit@6f5e4f68`.

Patch: #674.

rishabhsinha17:
## Summary

Fixes #679.

Falcon-H1 (`mlx-community/Falcon-H1-1.5B-Instruct-4bit`) on the non-paged MLX path (`VLLM_METAL_USE_PAGED_ATTENTION=0`) serves a single request correctly, but the first step that batches two or more decode requests kills EngineCore:

```
File "vllm_metal/v1/model_runner.py", line 1117, in _batched_decode
TypeError: Unsupported cache type for batching: CacheList
```

This teaches the non-paged merge/extract helpers about mlx_lm's `CacheList`, so models whose layer owns several stateful modules can take batched decode. Related: #655 (the paged-path crash for the same model stays open; that needs the Mamba-2 paged state work tracked in #644).

## Root cause

`mlx_lm.models.falcon_h1.Model.make_cache` returns `CacheList(ArraysCache(size=2), KVCache())` per layer, because every Falcon-H1 layer runs a Mamba-2 mixer and attention in parallel and indexes `cache[0]` / `cache[1]`. `vllm_metal/v1/contiguous_cache.py` dispatched on `ArraysCache`, `RotatingKVCache` and `KVCache` only, so `_merge_kv_caches` raised for the container type. Other mlx_lm models with the same shape: `deepseek_v32`, `longcat_flash`, `baichuan_m1`.

## Changes

- `_merge_layer_caches` / `_extract_layer_cache`: per-layer dispatch factored out of the loops; a `CacheList` recurses member-wise (width-checked, type-checked) so each nested cache keeps its own batching path (`ArraysCache` densification, `BatchKVCache` left padding) and extraction rebuilds the per-request nesting.
- `AnyCache` / `BatchedCache` aliases include `CacheList`.
- Tests: round-trip through a `CacheList(ArraysCache, KVCache)` layer with unequal sequence lengths, plus rejection of width mismatch and mixed layer types.

## Validation

Apple M1 Pro, 16 GiB, macOS 27.0, Python 3.12.14, vllm 0.28.0+cpu, mlx 0.32.0, mlx_lm 0.31.3, source install with `VLLM_METAL_BUILD_FROM_SOURCE=1`.

```
$ pytest tests/test_prefix_cache.py -q
12 passed in 2.48s
$ pytest -m "not slow" tests/ -q
1952 passed, 15 skipped, 52 deselected in 30.23s
Success: no issues found in 126 source files
```

Serving `mlx-community/Falcon-H1-1.5B-Instruct-4bit@6f5e4f68` with `VLLM_METAL_USE_PAGED_ATTENTION=0 VLLM_METAL_MEMORY_FRACTION=auto vllm serve ... --max-model-len 8192` (large enough that two requests fit; with `--max-model-len 512` the hybrid block alignment leaves one block and the request waits forever, which is #672):

- Before: two concurrent `/v1/completions` requests -> HTTP 500, EngineCore dead with the traceback above.
- After: both return 200; with `logprobs=2` and greedy sampling the per-token logprobs of the batched run are identical to the solo run for all three prompts (40 tokens each, `/metrics` shows `num_requests_running` reaching 2, no preemptions).
- The solo output matches a plain `mlx_lm` greedy loop (`model(...)` with `make_prompt_cache`) token for token on the same prompt ids.
- Teacher-forced in-process check mirroring `_batched_decode` (merge, one batched forward, extract, every step) with requests joining and leaving the batch: bit-identical logprobs to sequential decode for the unpadded request; a left-padded request can pick up fp16-level differences from the masked SDPA kernel (first seen as a 1e-3 deviation at one layer with bit-identical q/k/v), the same behaviour dense models and `mlx_lm.batch_generate` show today, so not introduced here.

### DIFF
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -261,3 +261,54 @@ class TestHybridCacheMergeExtract:
         kv_cache = contiguous_cache.KVCache()
         with pytest.raises(TypeError, match="Mixed cache types in a single layer"):
             contiguous_cache._merge_kv_caches([[arrays_cache], [kv_cache]])
+
+    def _make_cache_list(
+        self, *, seq_len: int, kv_value: float, v0: float, v1: float
+    ) -> contiguous_cache.CacheList:
+        # Falcon-H1's per-layer shape: CacheList(ArraysCache(size=2), KVCache()).
+        return contiguous_cache.CacheList(
+            self._make_arrays_cache(v0, v1), self._make_kv_cache(seq_len, kv_value)
+        )
+
+    def test_cache_list_merge_extract_roundtrip(self) -> None:
+        """Nested ``CacheList`` layers merge member-wise and extract back.
+
+        mlx_lm gives a layer that owns several stateful modules (Falcon-H1:
+        parallel Mamba-2 mixer + attention) one ``CacheList`` per layer.  Merging
+        must recurse into each member so batched decode sees a batched cache in
+        every slot, and extraction must rebuild the per-request nesting.
+        """
+        req0 = self._make_cache_list(seq_len=2, kv_value=1.0, v0=3.0, v1=33.0)
+        req1 = self._make_cache_list(seq_len=4, kv_value=2.0, v0=4.0, v1=44.0)
+
+        merged = contiguous_cache._merge_kv_caches([[req0], [req1]])
+
+        assert isinstance(merged[0], contiguous_cache.CacheList)
+        assert isinstance(merged[0][0], contiguous_cache.ArraysCache)
+        assert isinstance(merged[0][1], contiguous_cache.BatchKVCache)
+
+        for idx, src in enumerate((req0, req1)):
+            out = contiguous_cache._extract_kv_cache(merged, idx)[0]
+            assert isinstance(out, contiguous_cache.CacheList)
+            arrays_out, kv_out = out.caches
+            assert isinstance(arrays_out, contiguous_cache.ArraysCache)
+            assert isinstance(kv_out, contiguous_cache.KVCache)
+            assert bool(mx.allclose(arrays_out.state[0], src[0].state[0]))
+            assert bool(mx.allclose(arrays_out.state[1], src[0].state[1]))
+            assert kv_out.offset == src[1].offset
+            assert bool(mx.allclose(kv_out.keys, src[1].keys))
+            assert bool(mx.allclose(kv_out.values, src[1].values))
+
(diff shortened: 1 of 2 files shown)


WindChimeRan:
we no longer maintain the non-paged path.

Issue #654, opened 2026-08-30
[Bug] LFM2 crashes EngineCore on prefill: cache.make_mask() returns the string "causal" into ShortConv


ritunjaym:
## Summary

Serving an LFM2 model (`Lfm2ForCausalLM`) — e.g. `mlx-community/LFM2.5-1.2B-Instruct-4bit` — loads successfully and reaches `Application startup complete`, but the **first completion request crashes EngineCore** with `TypeError: string indices must be integers, not 'tuple'`.

The cause is a mask-contract mismatch at the `vllm-metal` ↔ `mlx-lm` boundary: `PagedAttentionContext.make_mask()` returns the **string** `"causal"`, which is valid for attention layers but is also handed to LFM2's `ShortConv` layers, which index the mask as an array.

Only prefill is affected — `make_mask` returns `None` when `N == 1`, so single-token decode steps would be fine. Any prompt longer than one token hits it.

## Environment

- **vllm-metal** 0.3.0.dev20260829020049 (prebuilt wheel via `install.sh`); repo at `a9184aa`
- **vllm** 0.28.0+cpu
- **mlx** 0.32.0, **mlx-lm** 0.31.3, **mlx-metal** 0.32.0, **transformers** 5.12.1
- macOS 26.6.2 (25G83), Apple M2 / 8 GB unified memory
- Python 3.12.2 (arm64)
- `VLLM_METAL_USE_PAGED_ATTENTION=1`, `VLLM_METAL_MEMORY_FRACTION=0.8`

Verified the crash reproduces with exactly the two env vars above — nothing else is required.

Model: `mlx-community/LFM2.5-1.2B-Instruct-4bit` @ `dee2f8a2786e6648bb644a7ca40652842490034b`
(`model_type: lfm2`, 16 layers — `layer_types` = 10 × `conv` + 6 × `full_attention`, 4-bit affine group 64)

## Reproduction

```bash
VLLM_METAL_USE_PAGED_ATTENTION=1 \
  VLLM_METAL_MEMORY_FRACTION=0.8 \
  vllm serve mlx-community/LFM2.5-1.2B-Instruct-4bit \
    --revision dee2f8a2786e6648bb644a7ca40652842490034b \
    --max-model-len 512 --max-num-batched-tokens 64
```

```bash
curl -s -X POST http://localhost:8000/v1/completions \
  -H "Content-Type: application/json" \
[...]

PR #675, opened 2026-09-02
Refuse paged attention for layers with unmanaged per-request state

rishabhsinha17:
## Summary

Serving a model whose layers keep per-request state the paged runtimes do not manage currently loads, warms up, reports healthy, and then kills EngineCore on the first request (#655). This refuses such models at `validate_paged_attention_support` time with a message that names the offending native cache slot and the non-paged fallback, so the failure moves from "500 on first completion" to "engine does not start".

Related: #655, #654, #644. Since this PR was opened, #598 and #710 landed paged serving for LFM2 and Nemotron-H, and #668/#696 moved hybrid geometry behind the state-family table. This revision validates against the table's layer plans and keeps the fail-loud for everything the table does not cover, such as Falcon-H1's per-layer cache stacks, per-request state in models vLLM does not flag as hybrid, and a declaration that drifts from a registered family's plan.

## Approach (revised per review)

Instead of walking model attributes for conv-carrying modules, validate mlx-lm's own cache declaration, as suggested in review: `make_prompt_cache()` is the authoritative statement of what per-request state each layer keeps.

`ModelCachePolicy.validate_paged_attention_support` (`vllm_metal/v1/cache_policy.py:296`) now calls `_require_managed_native_caches` (`cache_policy.py:314`):

- Non-hybrid: the declared slot count must equal `num_kv_cache_layers` (YOCO models declare exactly their unique slots, so Gemma4-style sharing passes), and every slot must be a single KV-style cache. An `ArraysCache` (Mamba/Mamba-2/ShortConv state) or `CacheList` (Falcon-H1's per-layer stack) slot raises `NotImplementedError` naming the slot, the type, and `VLLM_METAL_USE_PAGED_ATTENTION=0`.
- Hybrid: `_validate_hybrid_native_caches` (`cache_policy.py:355`) checks the declared split against `hybrid_runtime_plan.layers`, which serves all three registered families: one KV-style slot per attention layer, one `ArraysCache` per state layer, and no slot for stateless layers (Nemotron-H MLP/MoE blocks, which mlx-lm skips in `make_cache`).
- Cache classes from both `mlx_lm.models.cache` and `mlx_vlm.models.cache` are accepted (`cache_policy.py:84`): the two packages declare parallel hierarchies and Qwen3.5 text serving loads through mlx-vlm, where an `isinstance` against only mlx-lm's classes would misclassify every slot. Both packages are hard dependencies, so the previous `importlib` resolver is now two module imports and two class tuples.

`docs/supported_models.md` points at the state-family table and describes the validation backstop.

## Tests

`tests/test_attention_dispatch.py`, real mlx_lm modules with tiny configs against stub runners:

- Falcon-H1 (`CacheList` at slot 0) and a model declaring fewer slots than the KV plan raise.
- Qwen3 (all KV slots) and Qwen3.5 (matching GDN plan) pass; a hybrid whose declared split disagrees with the layer plan raises.
- LFM2 and Nemotron-H, which the previous revision rejected, now pass through their real family plans built by the family table; the Nemotron-H case uses an `M*-` pattern to cover the stateless-slot skip.
- Hybrid stubs carry `hybrid_runtime_plan` via `make_gdn_hybrid_plan` / `make_nemotron_hybrid_plan` instead of the removed runner attributes.
- `tests/test_vlm_forward_model.py` stub gains real layers so the adapter-delegation test passes the new validation.

## Validation

Apple M1 Pro, 16 GiB, macOS 27.0, Python 3.12.14, vllm 0.28.0+cpu, mlx 0.32.1, mlx_lm 0.32.0, mlx_vlm 0.6.17, source install with `VLLM_METAL_BUILD_FROM_SOURCE=1`.

```
$ pytest tests/ -q -m "not slow"
2156 passed, 15 skipped, 53 deselected in 54.42s
$ pytest tests/test_attention_dispatch.py::test_qwen35_paged_attention_hybrid -q
1 passed in 35.62s
Success: no issues found in 137 source files
```

The slow Qwen3.5 E2E exercises the hybrid validation against the real `mlx_vlm` loading path, the same path the `test (macos-15)` smoke failed on after #668 removed `runner.sdpa_layer_indices`.

### DIFF
--- a/docs/supported_models.md
+++ b/docs/supported_models.md
@@ -60,6 +60,18 @@ load, since mlx-lm does not read `attention_head_dim` from the
 `nvidia/Nemotron-H-*` configs yet. These values describe default engine
 behavior, not exhaustive per-model benchmarking on Metal.
 
+Hybrid checkpoints route through the state-family table in
+`vllm_metal/attention/runtime/families/`; a hybrid model type with no
+registered family (Falcon-H1, Jamba, PLaMo-2) is refused at startup with a
+clear error ([#655](https://github.com/vllm-project/vllm-metal/issues/655)).
+Paged-attention setup also validates every native cache slot mlx-lm declares
+(`ModelCachePolicy.validate_paged_attention_support`): dense models must
+declare one KV-style cache per layer, and a hybrid family's declaration must
+match its layer plan, so per-request state nothing manages fails loud at
+setup instead of corrupting requests. Supporting a new family means
+registering an owner in the table (the GDN, Nemotron-H, and ShortConv owners
+are the examples).
+
 HF AWQ checkpoints load through mlx-lm's `_transform_awq_weights` repack, with an
 entry-point preflight that normalizes AutoAWQ aliases (`w_bit`, `q_group_size`,
 uppercase `"GEMM"`) and rejects unsupported variants (`gemv`, `bits != 4`,
--- a/tests/test_attention_dispatch.py
+++ b/tests/test_attention_dispatch.py
@@ -295,3 +295,181 @@ def test_mixer_is_probed_only_when_a_family_opts_in() -> None:
 
     assert find_attn_attr(block) is None
     assert find_attn_attr(block, (*DEFAULT_ATTN_ATTR_NAMES, "mixer")) == "mixer"
+
+
+# ---------------------------------------------------------------------------
+# validate_paged_attention_support refuses unmanaged native cache topologies
+# ---------------------------------------------------------------------------
+
+
+def _make_policy_runner(model, *, num_layers: int, hybrid_plan=None, **attrs):
+    """Stub runner around a real mlx_lm model for cache-policy validation."""
+    from tests.stub_runner import make_stub_runner
+
+    return make_stub_runner(
+        model=model,
+        num_layers=num_layers,
+        num_kv_cache_layers=num_layers,
+        num_kv_heads=2,
+        is_hybrid=hybrid_plan is not None,
+        hybrid_runtime_plan=hybrid_plan,
+        **attrs,
+    )
+
+
+def _tiny_gdn_plan(num_layers, attention_indices):
+    """GDN plan with explicit topology; geometry is irrelevant to validation."""
+    from tests.stub_runner import make_gdn_hybrid_plan
+
+    return make_gdn_hybrid_plan(
+        num_layers,
+        attention_indices,
+        conv_kernel_dim=2,
+        conv_dim=4,
+        num_v_heads=1,
+        value_head_dim=4,
+        key_head_dim=32,
+    )
+
+
+def test_validate_rejects_falcon_h1_parallel_mamba():
+    """Falcon-H1 runs a Mamba-2 mixer next to self_attn in every layer (#655).
+
+    mlx-lm declares that topology as a ``CacheList`` per layer; no state
+    family owns it, and the dense paged runtimes manage one KV-style cache
+    per slot, so setup must refuse at validation time rather than crash on
+    the first request.
+    """
+    from mlx_lm.models.falcon_h1 import Model, ModelArgs
+
+    args = ModelArgs(
+        hidden_size=64,
+        intermediate_size=128,
+        num_hidden_layers=2,
+        num_attention_heads=4,
+        num_key_value_heads=2,
+        head_dim=16,
+        mamba_n_heads=4,
+        mamba_d_head=16,
+        mamba_d_ssm=64,
+        mamba_d_state=16,
+        vocab_size=100,
+    )
+    runner = _make_policy_runner(Model(args), num_layers=2)
+
+    with pytest.raises(NotImplementedError) as excinfo:
+        runner.validate_paged_attention_support()
+
+    message = str(excinfo.value)
+    assert "CacheList" in message
+    assert "native cache slot 0" in message
+    assert "VLLM_METAL_USE_PAGED_ATTENTION=0" in message
+
+
+def test_validate_rejects_native_cache_count_mismatch():
+    """A model declaring fewer native slots than KV layers must refuse."""
+    from types import SimpleNamespace
+
+    from mlx_lm.models.cache import KVCache
+
+    model = SimpleNamespace(make_cache=lambda: [KVCache()])
+    runner = _make_policy_runner(model, num_layers=2)
+
+    with pytest.raises(
+        NotImplementedError, match=r"expected 2 native cache slots.*declares 1"
+    ):
+        runner.validate_paged_attention_support()
+
+
+def test_validate_accepts_dense_and_gdn_hybrid_models():
+    """Qwen3 (all KV slots) and Qwen3.5 (matching GDN plan) both pass."""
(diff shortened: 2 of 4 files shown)


vllm_metal/attention/patching.py
```diff
@@ -49,6 +51,65 @@ def find_attn_attr(layer: Any) -> str | None:
     return None
 
 
+# Attribute names under which known mlx_lm layers keep a second stateful
+# module next to (or instead of) the one ``find_attn_attr`` selects: Mamba /
+# Mamba-2 mixers (Falcon-H1 ``mamba``, Jamba ``mamba``, Nemotron-H / PLaMo-2
+# ``mixer``) and LFM2's ShortConv ``conv``.
+_STATE_ATTR_NAMES = ("mamba", "mixer", "conv", "ssm", "linear_attn", "self_attn")
```
WindChimeRan: nit: rename to `_KNOWN_STATE_ATTR_NAMES`

WindChimeRan (commented):
I think we need to update the doc on supported model page. 

For new model attention support it will need to register here.

vllm_metal/attention/patching.py
```diff
@@ -49,6 +51,65 @@ def find_attn_attr(layer: Any) -> str | None:
...
+def _require_managed_state(
+    layer_idx: int, layer: Any, managed_attr: str | None
+) -> None:
+    unmanaged = find_unmanaged_state_attrs(layer, managed_attr)
+    if not unmanaged:
+        return
+    described = ", ".join(
+        f"'{name}' ({type(getattr(layer, name)).__name__})" for name in unmanaged
+    )
+    raise NotImplementedError(
```
WindChimeRan: I'd revise this error message ...

WindChimeRan (commented):
https://github.com/vllm-project/vllm-metal/blob/4d7228a7203c886462d6d1612ecaaad3b62e3300/vllm_metal/v1/cache_policy.py#L279-L294

I think the guard should be here: `validate_paged_attention_support`, maybe in this way (I didn't verify):
```
from mlx_lm.models.cache import (
    ArraysCache,
    CacheList,
    make_prompt_cache,
)


def validate_paged_attention_support(self) -> None:
    self._require_supported_per_layer_shapes()

    runner = self._runner
    native_caches = make_prompt_cache(runner._forward_model)

    # Registered hybrid runtimes must validate against their own layer plan.
    if runner.is_hybrid:
        self._validate_hybrid_native_caches(native_caches)
        return

    expected = runner.num_kv_cache_layers
    if len(native_caches) != expected:
        raise NotImplementedError(
            f"Paged attention expected {expected} native cache slots, but "
            f"mlx-lm declares {len(native_caches)}. This model uses a cache "
            "topology the selected paged runtime does not manage; set "
            "VLLM_METAL_USE_PAGED_ATTENTION=0."
        )

    for slot, cache in enumerate(native_caches):
        if isinstance(cache, (ArraysCache, CacheList)):
            raise NotImplementedError(
                f"Paged attention cannot serve this model: mlx-lm declares "
                f"{type(cache).__name__} at native cache slot {slot}, but the "
                "selected runtime manages only one KV-style cache per slot. "
                "Set VLLM_METAL_USE_PAGED_ATTENTION=0."
            )
```

rishabhsinha17:
Moved as suggested, thanks. The guard now lives in `validate_paged_attention_support` (vllm_metal/v1/cache_policy.py:303, helper at :321, hybrid plan check at :364) and validates `make_prompt_cache()`'s declaration: slot count against `num_kv_cache_layers` (YOCO models declare exactly their unique slots, so Gemma4 sharing passes), `ArraysCache`/`CacheList` slots refused, and registered hybrids checked against their own sdpa/linear layer plan. One adaptation to your sketch: the isinstance checks resolve cache classes across both `mlx_lm.models.cache` and `mlx_vlm.models.cache` (cache_policy.py:75), because Qwen3.5 text serving loads through mlx-vlm, whose parallel cache hierarchy fails an mlx-lm-only isinstance (caught by the slow Qwen3.5 E2E).

`walk_and_wrap` is back to upstream, which also moots both inline nits. Added the supported-models doc note you asked for.

Tests relocated to the new raise site and green locally: `pytest tests/ -q -m "not slow"` 1955 passed / 15 skipped, plus `test_qwen35_paged_attention_hybrid` (real hybrid via mlx-vlm) passing; ruff and mypy clean. New head: a421548.

WindChimeRan (commented):
* #668 has removed `runner.sdpa_layer_indices` but this PR rely on the dead attribute. 
* Please simplify the code & cleanup the tests.

rishabhsinha17:
Rebased on main and reworked for the state-family table. The hybrid check now validates mlx-lm's declared split against `hybrid_runtime_plan.layers` (cache_policy.py:355) instead of the removed `sdpa_layer_indices`, which also makes it serve all three registered families, including the stateless Nemotron-H blocks mlx-lm skips when declaring caches. Since both cache modules are hard dependencies now, the importlib resolver shrank to two imports and two class tuples (cache_policy.py:84).

Tests cleaned up to match. Rejecting LFM2 and Nemotron-H became obsolete once #598 and #710 landed paged serving for those families. Those two tests turned into acceptance coverage through their real family plans (the Nemotron-H case uses an `M*-` pattern to cover the stateless-slot skip), and hybrid stubs now build plans with `make_gdn_hybrid_plan` / `make_nemotron_hybrid_plan` instead of setting dead attributes. Doc note updated to point at the family table.

Green locally on M1 Pro with current pins (mlx 0.32.1, mlx-lm 0.32.0, mlx-vlm 0.6.17). `pytest -m "not slow" tests/` gives 2156 passed / 15 skipped, ruff and mypy are clean, and the slow `test_qwen35_paged_attention_hybrid` E2E passes, which is the path the `test (macos-15)` smoke died on. New head: d9965fa, PR description refreshed to match.

LxYuan0420 (approved):

WindChimeRan:
The original crash addressed by #675 was superseded by #696, which routes hybrid models through a family registry and rejects unsupported families at startup before they can crash on the first request.

rishabhsinha17:
Agreed on the original crash. #696 rejects unsupported families at startup, and the September 3 version of this branch went after that same gap. After #696 merged I rebuilt this on top of its family table, and the final head checks a different thing, one the registry cannot see from config. Registry routing picks a family plan and rejects models that have none. This guard then validates what mlx-lm declares against the chosen plan, checking the `make_prompt_cache` slot split against `hybrid_runtime_plan.layers` for every registered family, including the stateless Nemotron-H slots mlx-lm skips when declaring caches. It also refuses `ArraysCache` and `CacheList` slots on models vLLM does not treat as hybrid, which never reach the registry at all. Drift between an mlx-lm release and the family table never shows up in config. It shows up in the declared caches, and the guard reads those. On timing, @LxYuan0420 refreshed the branch against current main and approved the final head 840ddfd a few hours before the close, with CI green.

If the declaration backstop is not something the project wants to carry, or #764's direction ends up covering it, leaving this closed is fine with me. Happy to rebase and reopen if that would be useful. Either way I'll leave a review on #764.

PR #681, opened 2026-09-03
Add KV cache offloading for the Metal backend

RobbieJ:
Adds vLLM's KV cache offloading to Apple Silicon: a host pool plus an `fs`
disk tier, driven by the same `--kv-offloading-size` and
`--kv-transfer-config` flags as CUDA. Core vLLM's `OffloadingConnector` and
`TieringOffloadingSpec` do the work; only the copy engine, the shared
region and the disk I/O are Metal-specific. On a 128 GB Mac a 32B model
restores an evicted prefix 48x faster than it recomputes it. Off by
default, and nothing changes when it is off.

Supersedes #530. On `main` at vLLM 0.28.0. Measured on 128 GB M-series
machines, macOS 15 and 26.

## The problem

On Apple Silicon the KV cache is small and does not survive a restart.
Metal memory is shared with the OS, other applications and the model
weights, and what is left for KV shrinks as the model grows. On a 128 GB
machine at memory fraction 0.45, Qwen2.5-32B-4bit has room for 124,464
tokens of KV and Llama-3.3-70B-4bit for 34,448: about one long document.
A 1.5B model holds 1.76M and never runs out.

Prefix caching only holds what fits. An evicted prefix pays a full prefill
on its next request, and a restart (development loops, redeploys,
sleep/wake) discards everything. Prefill is the slow part here: on the 70B
a 2,320-token prefix takes 42.74 s to prefill and 6.65 s to restore.

## Results

Offloading against no offloading, same model, same machine:

| Case | Model | TTFT | End-to-end | n |
|---|---|---|---|---|
| Restart, same traffic | Llama-3.3-70B | **10.50x** | 2.47x | 3 |
| Restart, same traffic | Qwen2.5-32B | **9.50x** | 2.46x | 3 |
| Evicted prefix, requested again | Llama-3.3-70B | **11.11x** | 5.19x | 3 |
| Evicted prefix, requested again | Qwen2.5-32B | **29.09x** | 7.78x | 2 |
| Restart, memory fraction 0.8 | Qwen2.5-32B | **17.06x** | 2.20x | 3 |
| Evicted prefix, memory fraction 0.8 | Qwen2.5-32B | **48.09x** | 5.56x | 3 |

Rows at 0.45 ran on a host also in interactive use. Rows at 0.8, the
documented setting, ran on a dedicated M4 Max with IQR under 1% except
3.2% on the restart offload arm. Restart ranges from 2.16x to 10.50x over
12 model configurations at 0.45 and grows with model size. Some baseline
IQRs are wide, up to 97% on the 70B restart, and the 32B evicted-prefix
row at 0.45 is two repetitions. Per-row IQR, n and the raw JSON accompany
the benchmark tool, which is a separate PR.

**Where offloading cannot help, it delays admission.** With no reuse every
secondary-tier lookup misses and the scheduler holds the request one step.
Median TTFT over 11 no-reuse controls at 0.45 is 0.91x, about 10% added,
worst 0.79x. At 0.8 the two controls read 1.00x and 0.98x on TTFT, and
1.00x and 0.98x end to end.

**The speedup is restores, not better prefix caching.** One evicted-prefix
run was repeated with server logs kept. The in-GPU prefix cache hit rate is
within 0.3 points across both arms (52.2% against 51.9%). The offload arm
served 72.4% of GPU cache misses from the external tier and prefilled
about 87k prompt tokens against the baseline's 312k, 72.1% less.

**Restored output is the same output.** An end-to-end harness, not in this
PR, asserts restored text is byte-identical to the no-offload control
across 9 models and 33 scenarios, with `external_kv_transfer > 0` and
`gpu_cache_hit_tokens == 0` so the tokens came from the tier.

## Design

Core vLLM's offloading, with the platform-bound pieces replaced.

```bash
vllm serve <model> --kv-offloading-size 32
```

That is the host pool. The disk tier is what survives a restart. It is
configured as on CUDA:

```bash
PYTHONHASHSEED=0 vllm serve <model> --kv-offloading-size 32 \
  --kv-transfer-config '{"kv_connector_extra_config":
    {"secondary_tiers": [{"type": "fs", "root_dir": "/path/to/store"}]}}'
```

`PYTHONHASHSEED` is upstream's requirement for the `fs` tier, not a Metal
one: vLLM seeds `NONE_HASH` from it, and from `os.urandom` when it is
unset, whatever `--prefix-caching-hash-algo` says. Without it a restarted
server cannot find what the previous one wrote. The plugin warns at
startup when a persistent tier is configured and the seed is unset.

One spec, `MetalTieringOffloadingSpec(TieringOffloadingSpec)`, with or
without a disk tier. On unified memory the host pool and the Metal cache
are the same RAM, so a CPU-only spec would add code without adding a tier.
The connector overrides only worker-side cache registration. Block files
are upstream's format byte for byte, so a store written here is readable
by an upstream `fs` tier and the reverse, and the tier emits the same
`BlockStored` events.

| Tier | Status |
|---|---|
| Host pool | Supported |
| `fs` (disk) | Supported |
| `p2p` | Rejected at config time. No NVLink on Apple Silicon. |
| `obj` | Rejected at config time. NIXL has no macOS build. |

### What is Metal-specific

**The worker.** Upstream's copy engine is CUDA-only (streams, pinned host
buffers, a Triton kernel). The layout is not the blocker; upstream's
canonical form can express split K/V. The blocker is that the Metal cache
rebinds `key_caches[L]` to a fresh MLX array on every layer call, so a
torch view taken once at registration goes stale. `MetalKVOffloadWorker`
moves blocks between the live MLX arrays and the host pool, reading each
layer's current list entry at store time so the read is ordered after
pending writes. A completeness guard fails registration if the cache
grows a per-block array that is not in the offload inventory, so a new
model family cannot be silently half-offloaded. TurboQuant's scale and
zero arrays are in the inventory.

**The shared region.** Upstream's is `/dev/shm` plus Linux `madvise`.
Here it is a process-private buffer in a registry keyed by engine id,
which works because scheduler and worker share one process. That is why
offloading requires the single-process executor. A pool above half of
physical RAM is refused at startup, as upstream refuses a pool `/dev/shm`
cannot hold.

**The `fs` tier.** `fcntl(F_NOCACHE)` on block file descriptors, because
macOS has no `O_DIRECT` and upstream's probe falls back to buffered I/O,
which would push multi-GB KV churn through the Unified Buffer Cache. The
flag has to be set on the descriptor before the first write, so the tier
runs the Python store and load loop rather than upstream's C batch path.
Block files are 0o600 under a 0o700 root, where upstream creates 0o644;
KV blocks are conversation-derived and with a pinned hash seed the
filenames let anyone who can list the directory test for known prompts.
Upstream could take the same change. Blocks live under `blocks.noindex`
so Spotlight ignores the churn, and TurboQuant stores get a path component
of their own, since upstream's path hashes none of the fields TurboQuant
changes.

### What the platform hook does

`check_and_update_config` translates `--kv-offloading-size` itself,
because vLLM's own translation runs after the hook and would select a
connector this platform cannot serve. It clears
`cache_config.kv_offloading_size` to disarm that translation, routes the
connector and spec to the Metal classes by module path, forces
`kv_role=kv_both` as upstream does, and rejects at config time: any other
connector, a non-`native` backend, `VLLM_USE_SIMPLE_KV_OFFLOAD`, pipeline
parallelism, a multi-process executor, and any tier type but `fs`. When
KV cache events are enabled globally it sets `enable_kv_events` on each
tier; the reverse combination is refused, because that tier would publish
nothing and a KV-aware router would route as if the instance held no
prefixes. A transfer config with no connector is inert upstream and passes
through untouched.

### What changes on the inference path

Only with a connector configured. The runner opens upstream's connector
step at the top of `execute_model`, holds it across the
`execute_model` / `sample_tokens` split, and closes it when the deferred
decode is submitted, not when it resolves. The close queues the step's
stores, and the next step's `handle_preemptions` submits them before their
GPU blocks are reallocated. Closing later would copy blocks the
intervening forward had overwritten, a silent wrong-KV write into the
pool. Steps with no forward still progress KV transfers. Every new call
site in the runner, the worker and the decode pipeline is behind
`has_kv_transfer_group()`, and the existing runner tests pass unchanged.

## Sizing

The host pool should be larger than the Metal KV cache. If it is smaller,
the pool's own LRU drops blocks before the cache would have re-requested
them, so restores mostly miss and the tier costs more than it saves. The
plugin warns at startup when that is the shape and no disk tier is
configured. At memory fraction 0.8 there is no room for a larger pool, so
the 0.8 rows use an 8 GiB pool behind the disk tier, and the disk tier
carries the restores. The same 8 GiB pool with no disk tier gives 1.00x on
the evicted prefix and 0.99x on restart, three repetitions each. At a high
memory fraction the disk tier is the only tier that adds capacity, which
is why there is no separate CPU-only path.

## Not supported

Rejected at engine start with a specific message.

- Hybrid and sliding-window models. Single uniform full-attention KV cache
  group only.
- Draft-model speculative decoding. vllm-metal#630 registers the draft KV
  in the same cache group, and draft and target dims usually differ.
- Multi-rank. The host pool is shared inside one process. The
  single-process executor is the default for pipeline_parallel_size 1.

Nothing evicts block files, as with upstream's `fs` tier. The store grows
to whatever the workload touches; a no-reuse run on the 32B wrote 84 GB.
The user page says so and names the directory to delete.

## Relationship to vllm-project/vllm#49328

#530 was closed pending that fix, which shipped in vLLM 0.28.0. The
livelock guard #530 carried is deleted; the `fs` tier populates upstream's
`_load_job_keys` and the existing livelock test passes against upstream's
guard.

## Reviewing this

Two commits, feature and docs. About 3,900 lines: 1,600 in the package,
2,200 in tests, 130 in docs. Suggested order:

1. `platform.py`, `spec.py` and `shared_region.py` together. The executor
   guard exists because of the region.
2. `model_runner.py` and `decode_pipeline.py`, the step lifecycle above.
   The only files that touch request flow.
3. `kv_offload/worker.py`. The copy engine and the read-ordering rule.
4. `fs_tier.py`.
5. Tests and docs.

```bash
pytest tests/test_kv_offload_*.py     # 76 tests, about a minute
scripts/lint.sh                       # ruff, format, mypy, shellcheck
```

Full suite 2032 passing on macOS with vLLM 0.28.0. On a clean machine
`install.sh` bootstrapped uv and Python 3.12 and the suite passed
unchanged.

The tests worth a reviewer's time:

- `test_kv_offload_connector_step_order.py` pins the store fence: the
  step's close is ordered before the next step's `handle_preemptions`.
- `test_kv_offload_cuda_parity.py` checks block placement against
  upstream's own pointer arithmetic, on the real region.
- `test_kv_offload_tiering.py` drives `get_manager()` through upstream's
  body and asserts the Metal classes come out. The `fs` tier is selected
  by upstream's `module_path` hook; the region by rebinding one module
  global around the call. If either stopped working, production would
  get buffered I/O and 0o644 files with nothing logged.
- `test_kv_offload_fs_tier.py` drives upstream's failed-load negative
  cache through this subclass, and covers the empty-job path.

## Benchmark method

Four workloads per model: no reuse and reuse-inside-cache as controls,
evicted prefix, and restart. Each runs with offloading off and on against
a freshly started server, order rotated per repetition,
`--max-concurrency 4` so TTFT measures prefill rather than queueing,
`--max-model-len 8192`.

The evicted-prefix workload runs two passes against one server. The first
overflows the cache and is untimed; the second is measured, after the
early prefixes have been evicted. vLLM's `prefix_repetition` dataset emits
every request for a prefix consecutively, so a single pass never revisits
a prefix after eviction.

The offload arm holds more KV bytes than the baseline, so there is a
control with matched capacity: the baseline gets a larger Metal cache
instead of a host pool. That cuts the offload arm's memory advantage from
+131.6% to +4.5% and the result from 29.09x to 22.19x. The rest is the
disk tier.

The benchmark tool is a separate PR. It probes the KV cache, sizes the
host pool at 1.3x it and the evicted-prefix workload at 1.4x it, and exits
non-zero if either control shows offloading ahead. The command behind the
0.8 rows:

```bash
python tools/benchmark/kv_offload_benchmark.py \
    --model mlx-community/Qwen2.5-32B-Instruct-4bit \
    --memory-fraction 0.8 --offload-size-gib 8 \
    --max-model-len 8192 --prefix-len 3072 --suffix-len 512 --repeats 3 \
    --out kv-offload-benchmark.json
```

Caveats: the 70B rows and the 12-model matrix are at 0.45 on a host also
in interactive use; load average stayed near 2.5 on both hosts, so
absolute latencies include that load while the ratios, from interleaved
and rotated runs, do not depend on it.

## Docs

`docs/kv_offloading.md` is the user page, under Features in the nav. The
read-ordering rule is documented in `kv_offload/worker.py`, where it
applies.

### DIFF
--- a/docs/kv_offloading.md
+++ b/docs/kv_offloading.md
@@ -1,0 +1,139 @@
+# KV Cache Offloading
+
+vllm-metal supports vLLM's KV cache offloading. Blocks evicted from the wired
+Metal KV cache spill to host memory and, optionally, to disk, so a prefix that
+no longer fits can be restored instead of recomputed. The flags are vLLM's own,
+so a command that works on CUDA works here.
+
+Two things prefix caching cannot do and offloading can: serve a prefix after it
+has been evicted, and serve one after the server restarts.
+
+## Quick Start
+
+```bash
+VLLM_METAL_USE_PAGED_ATTENTION=1 vllm serve mlx-community/Qwen2.5-32B-Instruct-4bit \
+  --kv-offloading-size 32
+```
+
+That gives the host-memory tier. On Apple Silicon it is the same physical RAM
+as the wired cache, so it buys capacity beyond the wired cap rather than a
+faster medium.
+
+For persistence, which is where the benefit is largest, add a disk tier:
+
+```bash
+PYTHONHASHSEED=0 \
+VLLM_METAL_USE_PAGED_ATTENTION=1 vllm serve mlx-community/Qwen2.5-32B-Instruct-4bit \
+  --kv-offloading-size 32 \
+  --kv-transfer-config '{"kv_connector_extra_config":
+    {"secondary_tiers": [{"type": "fs", "root_dir": "/path/to/kv-store"}]}}'
+```
+
+Paged attention is required. Offloading is off by default.
+
+## Configuration
+
+| Flag | Description |
+|---|---|
+| `--kv-offloading-size N` | Host pool size in GiB. Enables offloading. |
+| `--kv-offloading-backend` | Must be `native`. |
+| `--kv-transfer-config` | Secondary tiers, as JSON. Only `fs` is supported. |
+
+Tier keys:
+
+| Key | Default | Description |
+|---|---|---|
+| `type` | required | `fs`. `p2p` needs NVLink, `obj` needs NIXL, which has no macOS build. |
+| `root_dir` | required | Where block files live. Nothing removes them; see Disk usage. |
+| `enable_kv_events` | auto | Set for you when KV events are on globally. |
+
+### `PYTHONHASHSEED` is required for reuse across restarts
+
+Block filenames are content hashes, and Python seeds its hash per process.
+Without `PYTHONHASHSEED=0` a restarted server cannot find what the previous one
+wrote, so the disk tier silently gives you nothing on the case it exists for.
+The server warns at startup when a persistent tier is configured and the seed
+is unset.
+
+## Sizing
+
+**The host pool should be larger than the wired KV cache.** If it is smaller,
+the pool's own LRU drops blocks before the wired cache would have re-requested
+them, so restores mostly miss and the tier costs more than it saves. The
+server logs a warning when this happens and no disk tier is configured.
+Behind a disk tier a small pool is the expected shape.
+
+**Offloading only helps once the working set overflows the wired cache.**
+Below that point prefix caching already serves every request and there is
+nothing to restore. Anyone validating the feature has to size the memory
+fraction so the traffic actually evicts, or offloading and no offloading
+read the same.
+
+The wired cache size is `VLLM_METAL_MEMORY_FRACTION` of the recommended working
+set, minus the model weights. It is printed at startup as
+`max_tokens_cached=`. Note the tension: at a high memory fraction there is no
+room left for a pool larger than the cache, and the disk tier is then the only
+secondary tier that adds real capacity.
+
+The host pool is pageable memory and comes out of the same RAM as the weights
+and the wired cache, so it is taken out of the `VLLM_METAL_MEMORY_FRACTION`
+budget: a larger pool means a smaller wired cache, and the fraction stays the
+bound on total resident memory. The startup log shows it as
+`kv_offload_pool=` in the memory breakdown. A pool above half of physical RAM
+is refused outright.
+
+## Disk usage
+
+**Nothing evicts block files.** The store grows to whatever the workload
+touches and stays there until you delete `root_dir` yourself. A workload with
+no prefix reuse writes every evicted block and reads none of them back; on a
+32B model that reached 84 GB in one benchmark run.
+
+Blocks live under a `blocks.noindex` subdirectory so Spotlight does not index
+them, and files are `0600` under a `0700` root because prompt content is
+recoverable from them. The store is not excluded from Time Machine for you.
+Run `tmutil addexclusion /path/to/kv-store` if it is on a backed-up volume.
+
+## KV-aware routing
+
+A KV-aware router places requests by knowing which instance already holds
+a prefix. It learns that from the `BlockStored` events the offload tier
+publishes, so the events have to be turned on:
+
+```bash
+PYTHONHASHSEED=0 \
+VLLM_METAL_USE_PAGED_ATTENTION=1 vllm serve <model> \
+  --kv-offloading-size 32 \
+  --kv-events-config '{"enable_kv_cache_events": true, "publisher": "zmq",
+    "endpoint": "tcp://*:5557"}' \
+  --kv-transfer-config '{"kv_connector_extra_config":
+    {"secondary_tiers": [{"type": "fs", "root_dir": "/path/to/kv-store"}]}}'
+```
+
+`enable_kv_events` on the tier is set automatically when
+`--kv-events-config` enables events globally. Setting it on the tier *without*
+enabling events globally is refused at startup, because the tier would publish
+nothing and the router would silently route as if this instance held no
+prefixes.
+
+## Not supported
+
+Rejected at startup with a specific message, rather than failing later.
+
+- Hybrid and sliding-window models. A single uniform full-attention KV cache
+  group only.
+- Draft-model speculative decoding, which puts the draft KV in the same cache
+  group.
+- Multiple workers. Offloading needs the single-process executor, because the
+  host pool is shared between the scheduler and the worker in one process.
+- The `p2p` and `obj` tiers.
+
+## Known limitations
+
+Block files carry no checksum, matching upstream's on-disk format so a store
+written here is readable by an upstream `fs` tier and the reverse. A torn write
+or bit rot therefore produces a right-sized file whose contents are wrong, and
+it is restored without complaint. Only truncation is caught.
+
+Deleting block files does not immediately return disk space, because APFS local
+snapshots are volume-wide and have no per-path exclusion.
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -73,6 +73,7 @@ nav:
     - Text Pooling: text_embedding_pooling.md
     - Speech-to-Text: stt.md
     - Turboquant: turboquant.md
+    - KV Cache Offloading: kv_offloading.md
     - GPU Profiling: profiling.md
     - Distributed (Ray): distributed.md
   - Contributing: CONTRIBUTING.md
--- a/tests/kv_offload_helpers.py
+++ b/tests/kv_offload_helpers.py
@@ -1,0 +1,122 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Shared helpers for the KV offloading tests.
+
+The snapshot/zero helpers mirror the cache kernels' write-then-rebind idiom
+(see vllm_metal/attention/caches/kv_cache.py): mutate through the current
+list entry, then rebind it, so later readers get provenance.
+"""
+
+import mmap
+import uuid
+
+import mlx.core as mx
+import numpy as np
+from vllm.v1.kv_offload.base import GPULoadStoreSpec
+
+from vllm_metal.attention.caches.kv_cache import MetalPagedKVCache
+from vllm_metal.v1.kv_offload.shared_region import MetalSharedOffloadRegion
+from vllm_metal.v1.kv_offload.worker import CACHE_ATTRS, MetalKVOffloadWorker
+
+
+def make_cache(dtype: mx.Dtype = mx.float16, **kwargs) -> MetalPagedKVCache:
+    kwargs.setdefault("num_layers", 2)
+    kwargs.setdefault("num_kv_heads", 2)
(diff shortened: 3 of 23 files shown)


now-ing:
Independently reproduced the headline evicted-prefix result, with an A/B on the same machine and cache size — since the benchmark tool is a separate PR, I wrote a minimal scenario myself (ten ~4.5k-token synthetic prompts: request A, nine fillers sized to force eviction at ~29k cached tokens, then A again) rather than trying to match your harness exactly.

**Setup**: M4 Max 128 GB, macOS 26, vLLM 0.28.0 wheel + this branch, `mlx-community/Qwen2.5-32B-4bit` (base, not Instruct — cache behavior is prompt-agnostic), `--gpu-memory-utilization 0.24` → `max_tokens_cached=29408`, `--kv-offloading-size 24` + fs tier, `PYTHONHASHSEED=0`, greedy.

| | A re-request TTFT | first/re-request | output |
|---|---|---|---|
| **offloading (host + fs tier)** | **1.21 s** | 54.1x | identical |
| **no offloading (same 0.24 frac)** | **74.65 s** | 0.97x (full recompute) | identical |

So the A/B gap on this box is ~62x, in the same regime as the 48x row in the PR table (mine is a fully-evicted single-prompt case, which is the most favorable shape). The disk tier does its job end to end: 900+ content-hash block files under `blocks.noindex/`, and restores produce byte-identical output.

One reproduction note worth having in the thread: at a comfortable memory fraction the wired cache simply doesn't evict — my first attempt at 0.65 showed an identical-fast re-request on the *baseline* too, because ~60 GB of wired KV held all ten prompts. Anyone validating this needs to size the fraction so the fillers actually overflow the wired cache before the offload-vs-baseline difference exists at all.

RobbieJ:
Thanks for running it, and for doing it on a shape I had not measured. A single fully evicted prompt is the most favourable case, so 62x landing in the same regime as the 48x row (108 prefixes, two passes) is the right relationship between the two.

Your fraction note is the one to keep. Below the point where the working set overflows the wired cache, prefix caching already serves everything and offloading has nothing to restore. The docs page says this under Sizing, but your wording is clearer and I will lift it.

The benchmark tool that produced the table follows as its own PR once this one has had a look.

RobbieJ:
@LxYuan0420 This is the revival of #530, rebased onto main at 0.28.0 with the benchmarks that one lacked. I know it is large for this repo. If it would help to review the benchmark tool first, or to take the host pool and the fs tier as two PRs, say which and I will restructure.

WindChimeRan (commented):
Personally, I think SSD KV cache offloading is an important feature, similar to oMLX.

But Apple Silicon’s UMA introduces a few design considerations:

1. Offloading KV to a separate CPU/RAM cache is not very meaningful. It doesn’t add physical memory capacity and may duplicate existing KV.
2. If a RAM intermediary is necessary for upstream’s plumbing, it needs to be included in the initial memory budget, and maybe just fake it as SSD intermediary? (I'm not sure if this is feasible)
3. SSD usage should also be bounded, with a capacity limit and eviction policy.

DGX Spark faces similar UMA challenges, so upstream’s design and code could be a useful reference.

I’d prefer SSD-backed prefix caching with an explicitly budgeted, bounded RAM intermediary.

RobbieJ:
Agreed on all three, and the disk tier is the point of the PR. On unified memory the host pool adds no capacity; measured on the 32B at fraction 0.8, the pool alone gives 1.00x on an evicted prefix while the same pool with the fs tier behind it gives 48x.

The pool exists because upstream's tiering is built around a primary RAM tier, and it is write-through: a block goes to disk as soon as its store to the pool completes, so the pool is a staging buffer, not a cache. It can be small; the 0.8 rows used 8 GiB against a 276k-token cache. Making it an mmap of a file would write the same bytes to SSD twice, since the primary tier is slot-addressed and the fs tier is content-addressed, and only the second survives a restart. I also measured an mmap store and read path against the F_NOCACHE one while tuning the tier: 5.4x slower, and it grows the page cache by every byte moved.

Two changes follow from your review, and I will make both:

1. Budget the pool. Today it sits outside VLLM_METAL_MEMORY_FRACTION. I will subtract it from the Metal budget when offloading is on, so the fraction stays the bound on total resident memory.
2. Bound the disk. Upstream's fs tier has no cap on CUDA either. I will add a max_size_gib tier key with oldest-first eviction, and a default. If you would rather have that as a follow-up so this PR stays reviewable, say so; otherwise it goes in here.

On the Spark: we have one, and I could not find unified-memory handling in upstream's offload code to borrow from. If you have a pointer I will read it.

LxYuan0420:
I agree with Ranran’s direction here. The use case is real, and the benchmark evidence is useful, but I do not want to merge this design as-is.

Before we split this into implementation PRs, we should first agree on the Apple Silicon resource model. The design likely needs to be SSD-backed, with a small bounded RAM staging layer, clear disk limits, narrow model scope, and serving/benchmark proof.

I’m closing this PR for now. We can revisit with a smaller design proposal based on those constraints.

RobbieJ:
Design proposal for the resource model, as asked: #740.

PR #683, opened 2026-09-03
Fix VLLM_USE_MODELSCOPE=1 silently falling back to HuggingFace for weight download

### LINKED ISSUE #682: [Bug] VLLM_USE_MODELSCOPE=1 silently falls back to HuggingFace for weight download (only "true" is accepted)
## Summary

`vllm_metal.utils.get_model_download_path()` only treats `VLLM_USE_MODELSCOPE` as enabled when its value is exactly the string `"true"`:

```python
# vllm_metal/utils.py:33
if os.environ.get("VLLM_USE_MODELSCOPE", "False").lower() == "true":
```

vLLM core accepts both `"1"` and `"true"`:

```python
# vllm/envs.py:727-728
"VLLM_USE_MODELSCOPE": lambda: (
    os.environ.get("VLLM_USE_MODELSCOPE", "False").strip().lower() in ("1", "true")
),
```

So with `VLLM_USE_MODELSCOPE=1`, vLLM core downloads configs/tokenizer via ModelScope, but the plugin's weight loader falls through to the HuggingFace fallback in `get_model_download_path()`.

## Impact

This is silent and looks half-working: the startup log shows ModelScope config downloads succeeding, then weight fetching crawls from HF Hub (with unauthenticated rate-limit warnings). In regions where HF is slow or blocked this effectively hangs the server. It cost us a while to diagnose because both paths mention the same env var.

## Repro

```bash
VLLM_USE_MODELSCOPE=1 vllm serve mlx-community/Qwen3.8-27B-4bit
# log: "Downloading: 13 files ... " via ModelScope  (configs, OK)
# then: "Warning: You are sending unauthenticated requests to the HF Hub" (weights, wrong)
```

With `=true` instead of `=1`, the log correctly shows `Downloading model ... from ModelScope...`.

## Environment

- vllm-metal 0.28.0 (source checkout of main) + vLLM core 0.28.0 wheel, macOS 26 / M5 Pro
- modelscope 1.39.1

## Suggested fix

Accept the same value set as vLLM core, or better, reuse the core parser so the two can't drift:

```python
from vllm import envs
if envs.VLLM_USE_MODELSCOPE:
    ...
```

suntp:
Fixes #682

## Problem

`get_model_download_path()` enabled the ModelScope path only when the env var was exactly `"true"`:

```python
if os.environ.get("VLLM_USE_MODELSCOPE", "False").lower() == "true":
```

vLLM core accepts both `"1"` and `"true"` (`vllm/envs.py`), so `VLLM_USE_MODELSCOPE=1` produced a half-working setup: configs/tokenizer downloaded via ModelScope, then weights silently crawled from HuggingFace (rate-limited or unreachable in some regions), with no hint in the logs that the env var had been ignored.

## Fix

Reuse vLLM core's parsed flag instead of re-parsing the string, so the plugin and core cannot drift apart:

```python
from vllm.envs import VLLM_USE_MODELSCOPE

if VLLM_USE_MODELSCOPE:
```

(`from vllm.envs import ...` is already the established pattern elsewhere in the plugin, e.g. `vllm_metal/__init__.py`.)

## Verification

- `VLLM_USE_MODELSCOPE=1 vllm serve mlx-community/Qwen3.8-27B-4bit` now logs `Downloading model ... from ModelScope...` and weights come from modelscope.cn (previously fell through to HF).
- `VLLM_USE_MODELSCOPE=true` unchanged.
- `import vllm_metal.utils` OK.

Tested on vllm-metal 0.28.0 source checkout + vLLM core 0.28.0 macOS arm64 wheel.

### DIFF
--- a/vllm_metal/utils.py
+++ b/vllm_metal/utils.py
@@ -30,7 +30,11 @@ def get_model_download_path(model_repo_name: str) -> str:
     if Path(model_repo_name).exists():
         return model_repo_name
 
-    if os.environ.get("VLLM_USE_MODELSCOPE", "False").lower() == "true":
+    # Reuse vLLM core's own env parsing (accepts "1"/"true") so the plugin
+    # and core cannot drift apart on which spellings enable ModelScope.
+    from vllm.envs import VLLM_USE_MODELSCOPE
+
+    if VLLM_USE_MODELSCOPE:
         try:
             from modelscope.hub.snapshot_download import snapshot_download
 
--- a/vllm_metal/utils.py
+++ b/vllm_metal/utils.py
@@ -2,7 +2,6 @@
 """Metal utility functions for vLLM Metal plugin."""
 
 import logging
-import os
 from pathlib import Path
 
 logger = logging.getLogger(__name__)


LxYuan0420 (approved):
Pretty straightforward. Thanks

PR #688, opened 2026-09-04
Cache SDPA metadata across layers

LxYuan0420:
This PR is:

- To cache `max_seq_len` with the existing SDPA kernel metadata
- To reuse it across attention layers in the same forward pass
- To reuse TurboQuant V centroids inside one SDPA forward

Before this PR, every attention layer recomputed:

```python
max(ctx.context_lens)
```

That value does not change during one forward pass, so we can compute it once when building kernel metadata and reuse it.

A focused microbenchmark of this Python setup path improved from 566 µs to 394 µs per forward, about 1.44x faster for this metadata work only. This is not an end-to-end inference speedup claim.


ricky-chaoju (approved):
LGTM

PR #692, opened 2026-09-04
[Bugfix] Fix OLMo 3 full-projection Q/K norms

### LINKED ISSUE #691: [Bug]: Paged attention applies OLMo 3 Q/K norms after splitting heads
## Description

The paged SDPA path crashes on OLMo 3 because it treats `q_norm` and `k_norm` as per-head norms.

In mlx-lm, `Olmo3Attention` [defines the Q/K RMSNorm widths over the full projections](https://github.com/ml-explore/mlx-lm/blob/254d153fdeb6f150edd4fc5a54f9828638481fa8/mlx_lm/models/olmo3.py#L75-L80) and [applies them before reshaping the projections into heads](https://github.com/ml-explore/mlx-lm/blob/254d153fdeb6f150edd4fc5a54f9828638481fa8/mlx_lm/models/olmo3.py#L99-L107):

```python
queries = self.q_norm(self.q_proj(x))
keys = self.k_norm(self.k_proj(x))

queries = queries.reshape(B, L, self.num_attention_heads, -1)
keys = keys.reshape(B, L, self.num_key_value_heads, -1)
```

The vLLM Metal paged path currently reshapes Q/K into `(B, L, heads, head_dim)` first and then applies the norms. For `allenai/Olmo-3-7B-Instruct`, the norm weight has 4096 elements while the reshaped tensor's last dimension is 128, so the first paged forward fails with:

```text
ValueError: [rms_norm] (*weight) must have the same size as the
last dimension of x but has 4096 elements.
```

The non-paged mlx-lm path succeeds with the same checkpoint.

## Reproduction

```bash
VLLM_ENABLE_V1_MULTIPROCESSING=0 \
VLLM_METAL_MEMORY_FRACTION=0.90 \
python - <<'PY'

llm = LLM(
    model="allenai/Olmo-3-7B-Instruct",
    max_model_len=64,
    max_num_seqs=1,
    trust_remote_code=True,
)
llm.generate("Hello", SamplingParams(temperature=0, max_tokens=1))
PY
```

Tested with:

- vLLM Metal base: `bcfc199`
- vLLM: `0.28.0+cpu`
- mlx-lm: `254d153fdeb6f150edd4fc5a54f9828638481fa8`
- Model snapshot: `allenai/Olmo-3-7B-Instruct@6e5971d9eba42665f5bd5a0fcf047f299ce1dccc`
- Apple M4, 24 GiB unified memory

## Expected behavior

The paged path should preserve mlx-lm's OLMo 3 preprocessing order:

```text
projection -> full-projection Q/K norm -> split heads -> RoPE
```

## Suggested fix and validation

See **https://github.com/vllm-project/vllm-metal/pull/692**. The proposed change makes the expected Q/K norm layout architecture-aware, applies OLMo 3's norms before head splitting, and adds focused regression coverage plus an official-checkpoint red-to-green validation.

HanningLin:
## Summary

Fix compatibility with mlx-lm's `Olmo3Attention` full-projection Q/K normalization.

- Add `QKNormPlacement.BEFORE_HEAD_SPLIT` for full-projection normalization.
- Keep the existing per-head `BEFORE_ROPE` behavior as the default.
- Preserve Hunyuan's per-head `AFTER_ROPE` behavior.
- Add numeric regression coverage using the real mlx-lm `Olmo3Attention` class.

Fixes #691

## Why

mlx-lm's `Olmo3Attention` [constructs `q_norm` and `k_norm` over the complete Q/K projection widths](https://github.com/ml-explore/mlx-lm/blob/254d153fdeb6f150edd4fc5a54f9828638481fa8/mlx_lm/models/olmo3.py#L75-L80). Its forward path then [normalizes each projection before reshaping it into attention heads](https://github.com/ml-explore/mlx-lm/blob/254d153fdeb6f150edd4fc5a54f9828638481fa8/mlx_lm/models/olmo3.py#L99-L107):

```python
queries = self.q_norm(self.q_proj(x))
keys = self.k_norm(self.k_proj(x))

queries = queries.reshape(B, L, self.num_attention_heads, -1)
keys = keys.reshape(B, L, self.num_key_value_heads, -1)
```

The paged SDPA preparation path assumed Q/K norms were per-head and applied them after reshaping. On the official 7B checkpoint, that passes a tensor whose last dimension is `head_dim=128` to an RMSNorm whose weight width is 4096, causing the first paged forward to fail.

This change represents the three supported normalization points with one enum:

- `BEFORE_ROPE`: default per-head normalization.
- `AFTER_ROPE`: Hunyuan per-head normalization after RoPE.
- `BEFORE_HEAD_SPLIT`: OLMo 3 full-projection normalization before reshape.

This does not change the attention kernel, RMSNorm math, model weights, or the default behavior of other architectures.

### Focused regression

The new test instantiates the real mlx-lm `Olmo3Attention` with a small deterministic configuration and verifies the mlx-lm ordering:

```text
projection -> full-projection norm -> reshape -> RoPE
```

```bash
pytest -q \
  tests/test_olmo3_qk_norm.py \
  tests/test_hunyuan_qk_norm.py \
  tests/test_attention_sdpa.py
```

```text
35 passed
```

The companion default-placement test verifies that existing Qwen-style norms still run per head. The existing Hunyuan tests verify that post-RoPE norm placement is unchanged.

### Official checkpoint

Tested `allenai/Olmo-3-7B-Instruct@6e5971d9eba42665f5bd5a0fcf047f299ce1dccc` on Apple Silicon with paged attention, greedy sampling, and:

```text
prompt="Hello", max_model_len=512, max_tokens=4
```

Pre-fix behavior:

```text
ValueError: [rms_norm] (*weight) must have the same size as the
last dimension of x but has 4096 elements.
```

Post-fix behavior on the rebased vLLM 0.29.0 branch:

```text
text: 'OOOOOOOO'
ids:  [20066, 20066, 20066, 20066]
```

The output text itself is not a quality evaluation because this uses a raw completion prompt with an instruct checkpoint. The relevant result is that the official checkpoint changes from a deterministic first-forward shape failure to successful paged prefill and decode.

### Regression and static checks

```text
pytest -m "not slow" tests/ -q
2162 passed, 15 skipped, 53 deselected

ruff check:        passed
ruff format check: passed
mypy:              passed
git diff --check:  passed
```


WindChimeRan (commented):
This PR seems to be over-complicated. Could you please try this design? 
(Also please resolve the merge conflicts. )
```
class QKNormPlacement(Enum):
    BEFORE_ROPE = auto()        # Default: per-head normalization
    AFTER_ROPE = auto()         # Hunyuan
    BEFORE_HEAD_SPLIT = auto()  # OLMo 3: full-projection normalization
```

```
q_proj_out = inner.q_proj(x)

if (
    norm_placement is QKNormPlacement.BEFORE_HEAD_SPLIT
    and q_norm is not None
):
    q_proj_out = q_norm(q_proj_out)

queries = q_proj_out.reshape(B, L, n_heads, -1)
```

now-ing:
The core fix looks right to me — OLMo 3's per-checkpoint full-width norms are exactly what `QKNormLayout` describes, and the real-class tests are the right shape. Two data points that support @WindChimeRan's "fold into one enum" direction, from reading the diff against the current registry:

1. **`FULL_PROJECTION + AFTER_ROPE` has no reachable user.** `_apply_full_projection_qk_norms` at the two AFTER_ROPE call sites (both the shared-KV path and the standard path) exists only for that combination, but the only FULL_PROJECTION registration (olmo3) is BEFORE_ROPE — and a full-projection-width norm after RoPE has already interleaved head dimensions doesn't correspond to anything in mlx_lm today. That's a whole helper plus two branches of zero-coverage surface area.
2. **The gated-`q_proj` branch silently skips full-projection norms.** In `prepare_sdpa_qkv`, the `q_full_head == 2 * head_dim` (Qwen3.5-style gate split) arm applies no `q_norm` at all; only the else arm does `q_norm(q_proj_out)`. No current model combines gating with FULL_PROJECTION, but if one appears the norm would drop silently rather than fail loudly — with the single-placement-enum design this hole doesn't need to exist.

Neither blocks the fix; they're the kind of surface that disappears if the layout becomes a third placement value, which would also resolve the naming wrinkle where the packed-qkv branch calls the per-head-named `_apply_qk_norms` on full-width (pre-reshape) tensors.

HanningLin:
Thanks @WindChimeRan and @now-ing. I rebased onto the latest `main`, resolved the conflicts, and simplified the implementation to the suggested single-enum design in `92c5078`:

- added `QKNormPlacement.BEFORE_HEAD_SPLIT` for OLMo 3;
- retained `BEFORE_ROPE` as the default per-head behavior and `AFTER_ROPE` for Hunyuan;
- removed `QKNormLayout`, the unreachable `FULL_PROJECTION + AFTER_ROPE` paths, and the full-projection-after-reshape helper;
- apply `BEFORE_HEAD_SPLIT` directly to Q/K projection outputs before reshaping. Q normalization now also occurs before the gated-Q split, so this placement is no longer silently skipped by that branch.

Validation on the rebased branch:

```text
focused OLMo 3/Hunyuan/SDPA tests: 35 passed
full non-slow suite: 2162 passed, 15 skipped, 53 deselected
ruff check / format / mypy / diff check: passed
official Olmo-3-7B-Instruct paged run: passed
```

WindChimeRan (approved):
LGTM. Tested e2e locally against mlx-lm with `tools/check_parity.py` 

(`mlx-community/Olmo-3-7B-Instruct-4bit`: all 40 prompts, 20 output tokens, top-k 5.)

| --- | ---: | ---: | ---: |
| 1 | 32 | 8 | 0 |
| 2 | 26 | 14 | 0 |

Both runs passed. 

I also added OLMo 3 to the supported-models table, and removed low-value tests. 


---

doc updates & e2e parity tests are required for new model supports. I'll update contributing guide as followup

PR #693, opened 2026-09-05
perf: GQA-shared flash-decode kernel inside the paged primitive (long-context decode 2-3x)

suntp:
## What

Reworked per the review direction above: the GQA flash-decode kernel now lives **inside the paged primitive** — `paged_attention_gqa_decode` dispatched from `paged_attention_primitive` in `paged_ops.cpp`, merged through the unchanged `v2_reduce`. No additional decode path, no Python-side routing; native SDPA is dropped from production and remains only as a test-only numeric ceiling (`sdpa.py` net −275 lines, `VLLM_METAL_NATIVE_SDPA_DECODE` env removed).

**Kernel design** — one threadgroup per (512-token partition, kv head, sequence); each simdgroup owns one query head of the GQA group, so every K/V row load is shared across the group (the per-token kernel re-reads each row once per query head from separate threadgroups); online softmax stays entirely in registers (log2 domain, `exp2`); partials follow the existing ps512 contract so the standard reduce consumes them unchanged. Instantiated for head sizes {64, 96, 128, 256} × block sizes {8, 16, 32} × half/bf16 in the v2 metallib; a missing specialization fails loudly.

**Conservative gate**: pure-decode batches, `window_seqlen_q <= 1`, true decode batch (`num_decode_requests == num_seqs`; callers that omit it keep the legacy single-seq gate), same-dtype fp16/bf16, no TurboQuant/sinks/softcap/sliding-window, GQA group ∈ [1, 8], aggregate KV ≥ 16384 (measured occupancy crossover on M5 Pro — below it the established split-KV kernel wins). Spec-decode verify windows are excluded so `test_spec_window_parity.py` stays bitwise.

## Numbers

**Kernel-level A/B** (same process, same tensors, gate on/off; Qwen3.8-like 24q/4kv/hs256 bf16):

| context | conc | per-token kernel | GQA kernel | speedup |
|---:|---:|---:|---:|---:|
| 16k | 1 | 63 GB/s | 133 GB/s | 2.1× |
| 64k | 1 | 44 GB/s | 176 GB/s | 4.0× |
| 64k | 2 | 44 GB/s | 174 GB/s | 4.0× |
| 256k agg | 4 | 46 GB/s | 226 GB/s | 4.9× |

Below the gate the dispatch falls back to the established kernels (GQA is 0.68× at 4k×1 — that is what the threshold is for). Small-GQA models (16q/8kv/hs128) see 1.0–1.8×.

**End-to-end, server topology, 100k context, single stream** (3 runs each, variance < 0.5%; methodology and measurement hazards documented in #713):

| configuration | decode tok/s |
|---|---:|
| pre-PR baseline (this PR's branch, gate forced off) | 4.25 |
| previous draft's fast path (native SDPA + python kernel) | 9.19 |
| **this PR (GQA merged into the paged primitive)** | **9.44** (quiet desktop: 11.2) |

**200k**: 8.70 tok/s sustained (250 GB/s effective), ~3× the pre-PR baseline. Honest disclosure: the previous draft's native-SDPA contiguous branch reached 12.1 in ideal conditions at 200k (~350 GB/s on the KV scan); the single-path GQA kernel currently scans at ~200–226 GB/s, so the longest contexts give up ~14–28% vs that draft's peak while gaining the unified path, multi-sequence support and simpler dispatch. Kernel headroom toward 300+ GB/s is identified as follow-up work.

The paged kernel's long-context collapse that motivates this was independently corroborated on M3 Max and M4 Max in this thread (10–33 GB/s across three chip generations while both alternative paths reach 45–79 GB/s on the same tensors).

## Tests

- Attention/decode suites **246/246** including the 13 new `test_gqa_paged_decode.py` cases (16k boundary, half-block tails, interleaved + contiguous tables, fp16/bf16, multi-seq) and **103/103** `test_spec_window_parity` (bitwise).
- Full tree: 1988 pass; the 21 failures + 6 errors are environment-only (Hub-unreachable, unrelated GGUF oracle, subprocess e2e).
- `ruff check` / `ruff format` clean.
- The same-process e2e A/B harness used for the validation is in-tree: `tools/e2e_qwen38.py`.

## Related

- #695 — the backing analysis (single-sequence decode bandwidth collapse and the dispatch diagnosis).
- #713 — measurement methodology used for the e2e numbers above (in-process vs server topology, desktop GPU co-tenancy, probe pitfalls).

### DIFF
--- a/tests/test_attention_sdpa.py
+++ b/tests/test_attention_sdpa.py
@@ -618,8 +618,9 @@ class _PagedRoutingOpsSpy:
         _out: mx.array,
         window_seqlen_q: int = 1,
         sinks: mx.array | None = None,
+        **_kwargs,
     ) -> None:
-        del window_seqlen_q, sinks
+        del window_seqlen_q, sinks, _kwargs
         self.calls[-1].block_tables = block_tables.tolist()
         self.calls[-1].block_size = block_size
 
--- a/tests/test_gqa_paged_decode.py
+++ b/tests/test_gqa_paged_decode.py
@@ -1,13 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 """GQA-shared flash-decode pass inside ``paged_attention_primitive``.
 
-Eligible long single-sequence decode (no TQ/sinks/softcap/window, head size
-in {64,96,128,256}, GQA group <= 8, context >= ``GQA_DECODE_MIN_SEQ_LEN``)
-is dispatched to ``paged_attention_gqa_decode`` and merged by the existing
-``paged_attention_v2_reduce``.  Shorter contexts, multi-sequence batches,
-and spec-decode verify windows stay on the established per-token / split-KV
-family — a different kernel family would break
-``tests/test_spec_window_parity.py`` bitwise identity.
+Eligible pure-decode batches (no TQ/sinks/softcap/window, head size in
+{64,96,128,256}, GQA group <= 8, ``window_seqlen_q <= 1``, aggregate KV
+``num_seqs * max_seq_len >= GQA_DECODE_MIN_SEQ_LEN``) dispatch to
+``paged_attention_gqa_decode`` and merge through ``paged_attention_v2_reduce``.
+Spec-decode verify windows pass ``window_seqlen_q = K+1`` and stay on the
+established family so ``tests/test_spec_window_parity.py`` stays bitwise.
 
 These tests drive the shipped primitive (not a reimplementation) against
 ``ref_paged_attn``.
@@ -61,6 +60,7 @@ def _run_primitive(
     seed: int,
     window_seqlen_q: int = 1,
     query_lens: list[int] | None = None,
+    num_decode_requests: int = -1,
 ) -> tuple[mx.array, mx.array]:
     mx.random.seed(seed)
     num_seqs = len(kv_lens)
@@ -72,11 +72,13 @@ def _run_primitive(
     n_blocks_needed = (max_kv_len + BLOCK_SIZE - 1) // BLOCK_SIZE
     tables = []
     max_blk = 0
-    for _ in range(num_seqs):
+    # Offset each sequence's pages so multi-seq batches do not alias.
+    page_stride = n_blocks_needed * 3
+    for s in range(num_seqs):
         if interleaved:
-            table = _interleaved_table(n_blocks_needed)
+            table = [b + s * page_stride for b in _interleaved_table(n_blocks_needed)]
         else:
-            table = list(range(n_blocks_needed))
+            table = list(range(s * n_blocks_needed, (s + 1) * n_blocks_needed))
         tables.append(table)
         max_blk = max(max_blk, max(table))
     num_cache_blocks = max_blk + 4
@@ -114,6 +116,7 @@ def _run_primitive(
         -1,
         out,
         window_seqlen_q=window_seqlen_q,
+        num_decode_requests=num_decode_requests,
     )
     mx.eval(out)
     ref = ref_paged_attn(
@@ -160,10 +163,7 @@ def test_gqa_decode_matches_reference(
 
 
 def test_below_crossover_stays_on_split_kv() -> None:
-    """Contexts under GQA_DECODE_MIN_SEQ_LEN must not take the new pass.
-
-    They still match the reference via the established split-KV kernel.
-    """
+    """A single sequence under GQA_DECODE_MIN_SEQ_LEN stays on split-KV."""
     ops = get_ops()
     kv_len = ops.GQA_DECODE_MIN_SEQ_LEN - 1
     assert kv_len > ops.PARTITION_SIZE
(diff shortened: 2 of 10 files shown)


WindChimeRan (commented):
I love the diagnosis on "At long contexts that kernel's effective KV-scan bandwidth collapses to **\~40 GB/s** on M5 Pro (physical peak \~320 GB/s)" but I think we should not maintain one more additional path using contiguous path. We should do it in a more principled way. 

I'd prefer to use native SDPA as a performance reference, and fix the paged decode implementation itself (e.g., maybe look into the upstream vllm source code as implementation reference)

What do you think?

suntp:
Thanks @WindChimeRan — I think we're actually pointing at the same target, and I'm happy to reshape the PR. My read of the data first, then two concrete options.

**The kernel is the fix you're describing.** `_gqa_decode_kernels()` is a ground-up fix of the paged decode path itself — split-KV two-pass flash-decode over the block table, same structure as vLLM's FlashInfer/flash-decode decode path, adapted to the paged layout. On the non-contiguous hybrid-interleave tables (Qwen3-Next-style 48 GDN : 16 attention, runs of 2) it lifts the KV-scan from ~39 GB/s to **168–188 GB/s** measured at 100k–200k contexts, which is where the whole 3–6× e2e decode win (9.8/7.1/4.8/2.9 → 15.7/15.3/12.1/12.1 tok/s at 30k/50k/100k/200k) comes from. That part has no extra dispatch surface: it *is* the paged implementation.

**Where I agree on the contiguous fast path:** it's a second dispatch case, and honestly it's the smaller half of the win (~201 GB/s zero-copy native SDPA vs 168–188 GB/s from the kernel — roughly 10–15% on top). I proposed it because for the common single-request decode case the blocks usually are physically contiguous, so native SDPA gives a free ceiling probe — which is exactly the "performance reference" role you describe.

So, two shapes I can take this to, your call:

1. **Kernel-only**: drop the contiguous/native-SDPA dispatch entirely from the PR. One code path, all layouts, ~4–5× on hybrid-interleave models. I'd keep the native-SDPA numbers in the PR description purely as the reference target.
2. **Kernel + reference path env-gated off by default** (`VLLM_METAL_NATIVE_SDPA_DECODE=0` ships), so the default behavior is exactly option 1, and the native path stays visible as a benchmark/reference tool for future kernel work.

I'd lean toward **1** for maintainability — the 10–15% delta isn't worth a second path, and if we later close the gap in the kernel itself (there's headroom: the merge pass and table upload are still untuned) the question disappears. Say the word and I'll push the slimmed version to this branch.

LxYuan0420 (commented):
Lets keep native SDPA as a reference and focus on improving the existing paged-attention implementation instead of adding another decode path.

now-ing:
Ran an independent single-sequence decode bandwidth probe on two more Apple Silicon generations, to help ground the "fix the paged kernel itself" direction discussed above.

**Setup**: direct `paged_attention_primitive` call vs this PR's two fast paths (`_native_sdpa_decode_fast_path` with a contiguous table → native SDPA, and with the run-of-2/skip-1 interleaved table → the flash-decode kernel), single sequence GQA 24Q/4KV, head_dim 256, bf16, block 32, synthetic caches, effective KV-scan GB/s = KV bytes / kernel wall time (median of 12 after warmup). MLX 0.32.0, your PR head `aa20536`.

**M3 Max 96 GB** | 30k: paged 33.4 / native 72.2 / flash-decode 15.5 | 100k: 12.7 / 49.8 / 29.2 | 200k: 16.7 / 62.9 / 45.4

**M4 Max 128 GB** | 30k: paged 15.1 / native 71.5 / flash-decode 76.3 | 100k: 22.2 / 10.3* / 63.5 | 200k: 9.7 / 24.6 / 79.3

(* the 100k native-SDPA point on M4 Max hit an allocation outlier — 39.7 ms vs 1.7/13 ms for its neighbors; treat that cell as noisy.)

Two observations:

1. **The paged kernel's bandwidth collapse is not M5-specific.** It sits at 10-33 GB/s across three chip generations while both alternative paths reach 45-79 GB/s on the same tensors, which supports the position that the paged decode kernel has a structural problem worth fixing in place (the diagnosis in #695 looks general, not memory-subsystem-specific).
2. **The flash-decode kernel in this PR is genuinely good** — on M4 Max it is the fastest path at every context length tested (76/64/79 GB/s), including the interleaved-table case this PR targets. As a reference implementation for "what the fixed paged kernel should achieve" it is directly useful; the native-SDPA numbers likewise make a clean perf reference for contiguous runs.

Two smaller notes on the diff while I was in there:

- the bare `except Exception` around the fast-path dispatch falls back with no log at all (only `_NATIVE_FP_STATS` counts) — a rate-limited `logger.warning` would make production regressions visible as more than "just slower";
- the merge kernel's `pm <= -1e29f` guard assumes every `(tile, kv)` threadgroup wrote its partial; if grid/threadgroup divisibility ever fails, partials read as uninitialized memory rather than sentinel values. A static assert on divisibility (or an explicit out-of-range `continue` keyed on `nt`) would pin that down.

Happy to share the microbench script (it stubs `vllm.logger` so it runs without the full vLLM wheel) if useful.

suntp:
Direction confirmed by both reviews (@WindChimeRan @LxYuan0420) — reshaping this PR as the merged-into-paged version:

- The GQA flash-decode kernel moves **into the paged primitive** (`paged_attention_gqa_decode` dispatched from `paged_attention_primitive`, merged through the existing `v2_reduce`), so there is no additional decode path and no Python-side dispatch.
- Native SDPA is **dropped from production** and kept only as a test-only numeric ceiling.
- Gate stays conservative: pure-decode batches, no TurboQuant/sinks/softcap/sliding-window, GQA group ≤ 8, head sizes {64,96,128,256}, aggregate KV ≥ 16k (the measured occupancy crossover), spec-decode verify windows excluded so `test_spec_window_parity` stays bitwise.
- Multi-request decode batches engage the same kernel (one simdgroup per GQA q-head, KV loads shared across the group).

Local status on M5 Pro: attention/decode suites 246/246 (incl. 103 spec-window parity), kernel-level A/B 2.1-4.9× over the per-token/split-KV kernel at ≥16k, and a same-process e2e A/B on Qwen3.8-27B-4bit shows single-stream decode +23%/+40%/+43% at 32k/64k/100k (machine in a thermally-throttled state, so these are the conservative floor; both states will be labeled in the final description). @now-ing's cross-generation probe (M3 Max / M4 Max) was very helpful for grounding this — thanks.

Will push the reworked branch here once the remaining bits are committed (fresh-state re-measure + description rewrite).

suntp:
Reworked branch pushed (fast-forward from aa20536): the GQA flash-decode kernel is now dispatched from inside `paged_attention_primitive` and merged through the existing `v2_reduce`, native SDPA is test-only, and multi-request decode batches take the same kernel. Description rewritten with the server-protocol numbers (4.25 → 9.44 tok/s at 100k, 8.70 at 200k, quiet-desktop 11.2 at 100k) plus the kernel-level A/B and the spec-window parity guarantee. The known trade at 200k vs the previous draft's native-SDPA peak is called out explicitly in the description.

suntp:
Superseded by #715 — the same reworked implementation rebased as a single clean commit on current main (this PR carried the withdrawn native-SDPA draft in its history and was 8 commits behind main, incl. #696's sdpa.py changes). All discussion and the independent M3 Max / M4 Max validation above remain the motivating context; the numbers live on in #715's description with a corrected attribution note for the 200k figure. The benchmarking-hazard follow-up (#713) and its env-knob PR #714 will be rebased onto #715.

PR #694, opened 2026-09-05
[Bugfix] Populate prompt_logprobs_dict on both runner paths

### LINKED ISSUE #680: [Bug] echo+logprobs returns 500 (KeyError) and prompt_logprobs silently returns [None]: prompt_logprobs_dict never populated
## Summary

The Metal runners never populate `prompt_logprobs_dict`, so the two OpenAI-API surfaces that need prompt logprobs both misbehave (any model; observed on the paged path with `mlx-community/Qwen3-0.6B-4bit`):

1. `echo=true` + `logprobs=N` -> **HTTP 500** from deep inside vLLM's completion serving; the engine survives (per-request failure):

```
File ".../vllm/entrypoints/openai/completion/serving.py", line 690, in _create_completion_logprobs
    step_token = step_top_logprobs[token_id]
KeyError: 2326
```

2. Bare `prompt_logprobs=1` -> **HTTP 200 with `"prompt_logprobs": [None]`** — a single placeholder instead of one entry per prompt token. Silent empty data is arguably worse than the 500: clients get a well-formed-looking response with nothing in it.

## Reproduction

```bash
vllm serve mlx-community/Qwen3-0.6B-4bit --max-model-len 4096
# 500:
curl -s localhost:8000/v1/completions -H 'Content-Type: application/json' \
  -d '{"model":"mlx-community/Qwen3-0.6B-4bit","prompt":"The three most important properties of a distributed cache are","max_tokens":8,"echo":true,"logprobs":1}'
# -> {"error":{"message":"2326","type":"InternalServerError","code":500}}

# silent [None]:
curl -s localhost:8000/v1/completions -H 'Content-Type: application/json' \
  -d '{"model":"mlx-community/Qwen3-0.6B-4bit","prompt":"The three most important properties of a distributed cache are","max_tokens":4,"prompt_logprobs":1}'
# -> 200, "prompt_logprobs":[None]
```

## Root cause

- `vllm_metal/v1/model_runner.py` `_build_output` (line 2560 at 07f1987) hardcodes `prompt_logprobs_dict={}` in every `ModelRunnerOutput`; `stt_model_runner.py` does the same (lines 219, 228). The vLLM v1 contract (one `LogprobsTensors` per request covering positions `1..prompt_len-1`, delivered on the prompt-finishing step, sliced under chunked prefill) is never fulfilled.
- The decode-pipeline gate is aware of the flag but only uses it to keep the step eager: `SamplingShape.has_prompt_logprobs` -> `block_reason() == "prompt logprobs requested"` (`vllm_metal/v1/decode_pipeline.py:143`). Nothing downstream computes the values.
- With the dict empty, vLLM's API layer builds the echo response assuming well-formed prompt logprobs and indexes a generated-token top-logprobs dict with a prompt token id -> `KeyError`.

## Expected

Either populate prompt logprobs (needs per-position accounting through chunked prefill on both runner paths), or reject `prompt_logprobs` / `echo`+`logprobs` with a clean 400 until it is supported — anything but a 500 on one surface and silent `[None]` on the other.

I can take the full implementation; an interim clean 400 may be worth landing first if maintainers prefer.

## Environment

Apple M1 Pro, 16 GiB, macOS 27.0, Python 3.12.14, vllm 0.28.0+cpu, mlx 0.32.0, mlx_lm 0.31.3, vllm-metal source install at 07f1987 (`VLLM_METAL_BUILD_FROM_SOURCE=1`), `VLLM_METAL_USE_PAGED_ATTENTION=1` (default), `VLLM_METAL_MEMORY_FRACTION=auto`.

rishabhsinha17:
## Summary

Fixes #680.

The Metal runners hardcoded `prompt_logprobs_dict={}` in every `ModelRunnerOutput`, so the two OpenAI-API surfaces that need prompt logprobs both misbehaved on any model: `echo=true` + `logprobs=N` returned HTTP 500 (vLLM's completion layer indexes a generated-token top-logprobs dict with a prompt token id and hits `KeyError`), and bare `prompt_logprobs=N` silently returned `[None]` instead of one entry per prompt token. This fulfills the vLLM v1 contract instead: one `LogprobsTensors` per requesting prompt covering positions `1..prompt_len-1`, delivered on the step that finishes the prompt, sliced per chunk under chunked prefill.

## Changes

- `vllm_metal/v1/prompt_logprobs.py` (new): the accounting and tensor construction shared by both runner paths.
  - `prompt_logprobs_window` mirrors upstream `_get_prompt_logprobs_dict` chunk accounting: an intermediate chunk scores all its rows, the `==` boundary chunk defers delivery to the next step, and the completing chunk drops the tail row that predicts the first sampled token.
  - `PromptLogprobsAccumulator` fills one CPU `LogprobsTensors` chunk by chunk; `gather_prompt_logprobs` goes through vLLM's own `Sampler.compute_logprobs` / `gather_logprobs`, so column 0 is the target token, alternatives are the top-k, and ranks are 1-based — the same layout as sampled-token logprobs.
  - `PromptLogprobsTracker` owns in-progress state per request: kept across preemption (a resumed request re-runs its chunks over the same positions), dropped when the engine finishes the request.
- Paged path (`vllm_metal/v1/model_runner.py`): steps containing a requesting prefill skip both head-pruning fast paths — the projection-free intermediate forward (which produces no logits at all) and the selective-logits layout (which keeps only each prefill segment's last row) — so every prompt position has a logits row. `_sample_paged_batch` feeds each requesting chunk's rows to the tracker on both the intermediate-only and the sampling branch, and the completed tensors land in the step's output. The prefill pack now carries the full prompt for requesting chunks: each chunk's last row scores the first token of the *next* chunk, so targets always reach one past the chunk's own tokens.
- Non-paged path: `_prefill_single` forwards the whole prompt in one chunk and gathers one-shot (`full_prompt_logprobs`).
- `_build_output` passes the batch's completed tensors through instead of `{}`. Desyncs raise: a requesting chunk whose segment rows were pruned, or whose pack lacks the full prompt, is a `RuntimeError`, not silent bad data.

STT keeps `{}`: transcription requests have no prompt-logprobs surface.

## Tests

`tests/test_prompt_logprobs.py`:

- Window accounting against upstream's chunk cases (intermediate, `==` boundary, completing, one-token prompt).
- Gather layout (column 0 target, descending alternatives, 1-based ranks) and accumulator slicing across chunks.
- Tracker: delivery only on the completing chunk, values identical to a one-shot gather over the same rows, empty tensors for a one-token prompt, discard, row-count mismatch.
- Runner wiring regressions for both symptoms: `_build_output` carries the dict (was hardcoded `{}`), `_prefill_single` populates tensors for the non-paged path, `_gather_prefill_prompt_logprobs` delivers on the completing chunk and raises on pruned segments / missing full prompt.
- Slow E2E (`test_prompt_logprobs_end_to_end_paged`): `Qwen/Qwen3-0.6B` with `max_num_batched_tokens=16` forcing chunked prefill; asserts one real entry per prompt token with the prompt token present in each (the exact data the `echo`+`logprobs` path indexes, and the silent-`[None]` repro inverted).

## Validation

Apple M1 Pro, 16 GiB, macOS 27.0, Python 3.12.14, vllm 0.28.0+cpu, mlx 0.32.0, mlx_lm 0.31.3, source install with `VLLM_METAL_BUILD_FROM_SOURCE=1`.

```
$ pytest tests/ -q -m "not slow"
1973 passed, 15 skipped, 53 deselected in 36.66s
$ pytest tests/test_prompt_logprobs.py -q
25 passed in 15.59s  # includes the slow chunked-prefill E2E
Success: no issues found in 127 source files
```

Manual check of the issue's repro shape via the offline engine (same code path the server drives): `prompt_logprobs=1` now returns `len(prompt_token_ids)` entries, `[0] is None`, every other position carries the prompt token's logprob.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -25,6 +25,7 @@ from vllm_metal.v1.model_adapter import DefaultModelAdapter
 from vllm_metal.v1.pooling.backends.decoder.factory import (
     build_decoder_pooling_backend,
 )
+from vllm_metal.v1.prompt_logprobs import PromptLogprobsTracker
 from vllm_metal.v1.spec_decode import SpeculativeDecodeController
 from vllm_metal.v1.structured_output import MetalStructuredOutputApplier
 
@@ -98,6 +99,7 @@ def make_stub_runner(
         "_sampler": None,
         "_native_sample_key": None,
         "_structured_output_applier": MetalStructuredOutputApplier(),
+        "_prompt_logprobs_tracker": PromptLogprobsTracker(),
         "_lora": MetalLoRARuntime(),
         "_yoco_cache_mapping": None,
         "model_args": _model_args,
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -130,7 +130,7 @@ def make_stub_runner(
     runner._cache_policy = ModelCachePolicy(runner, runner._model_adapter)
     if "_decode_pipeline" not in attrs:
         runner._decode_pipeline = DecodePipeline(
-            build_output=runner._build_output,
+            build_output=mr._ExecutionBatch.to_model_runner_output,
             validate=runner._validate_scheduled_outputs,
         )
 
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -63,6 +63,7 @@ def make_stub_runner(
         "model_config": SimpleNamespace(
             runner_type="generate",
             get_head_size=lambda: 128,
+            logprobs_mode="raw_logprobs",
             max_model_len=2048,
             is_hybrid=is_hybrid,
         ),
--- a/tests/test_prompt_logprobs.py
+++ b/tests/test_prompt_logprobs.py
@@ -1,0 +1,476 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Prompt logprobs accounting and tensor construction."""
+
+from __future__ import annotations
+
+import mlx.core as mx
+import numpy as np
+import pytest
+import torch
+
+from vllm_metal.v1.prompt_logprobs import (
+    PromptLogprobsAccumulator,
+    PromptLogprobsWindow,
+    gather_prompt_logprobs,
+    prompt_logprobs_window,
+)
+
+
+class TestPromptLogprobsWindow:
+    def test_single_chunk_scores_every_prompt_token_but_the_first(self) -> None:
+        window = prompt_logprobs_window(start_pos=0, num_tokens=5, prompt_len=5)
+
+        assert window == PromptLogprobsWindow(0, 4, completes=True)
+        assert window.first_target == 1
+
+    def test_intermediate_chunk_scores_all_its_rows(self) -> None:
+        window = prompt_logprobs_window(start_pos=0, num_tokens=3, prompt_len=8)
+
+        assert window == PromptLogprobsWindow(0, 3, completes=False)
+
+    def test_final_chunk_drops_the_row_that_predicts_the_sampled_token(self) -> None:
+        window = prompt_logprobs_window(start_pos=3, num_tokens=5, prompt_len=8)
+
+        # rows for positions 3..7 -> targets 4..7 (four prompt tokens)
+        assert window == PromptLogprobsWindow(3, 4, completes=True)
+
+    def test_chunk_ending_one_before_the_prompt_end_is_not_yet_complete(self) -> None:
+        # Positions 0..6 forwarded for an 8-token prompt: the == case in vLLM.
+        window = prompt_logprobs_window(start_pos=0, num_tokens=7, prompt_len=8)
+
+        assert window == PromptLogprobsWindow(0, 7, completes=False)
+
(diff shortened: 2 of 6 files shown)


now-ing:
Ran the full suite as an independent check on a different machine generation (M4 Max 128 GB, macOS 26, vLLM 0.28.0 wheel, MLX 0.32.0, source-built artifacts): **25/25 passed**, including `test_prompt_logprobs_end_to_end_paged` with the real `Qwen/Qwen3-0.6B` through the Metal paged path (chunked prefill at `max_num_batched_tokens=16`, both the echo/logprobs and plain `prompt_logprobs` assertions).

For anyone reproducing behind a proxy: the end-to-end test needs `HF_HUB_DISABLE_XET=1` (or a clean cache) — the xet/CAS transfer path fails with "File reconstruction error: CAS Client Error" through an HTTP proxy, which initially looks like a test failure but is purely a download-layer issue.

LxYuan0420 (approved):
I pushed a few commits to fix the remaining prompt-logprobs lifecycle and valid-mode cases, then trimmed the runner/test glue.

PR #696, opened 2026-09-05
Route hybrid models through a state-family table

ricky-chaoju:
PR 2 of the Nemotron-H stack in #644, on top of #668. The platform already reads vLLM's typed `ModelConfig.is_hybrid`, but the runner still sniffs `full_attention_interval` out of the mlx args, so a Nemotron-H checkpoint is hybrid to the engine and dense to the worker. On main that is 0/52 layers patched, 38 GB of KV budgeted for six attention layers, and a 500 on the first request.

The runner now reads the typed field too, and "which family" comes from a table in `attention/runtime/factory.py` copied from the encoder pooling loaders in #608. It has one row, GDN, and every supported model goes through it, so it is load-bearing today: the typed field is also true for LFM2 and Jamba and the table is what keeps them out of the GDN parser. Lifecycle imports no family anymore. No stateless role yet; per your note on #668 it waits for the Mamba-2 row in PR 3. If you would rather the table land with that row, I can fold this into PR 3.

The first commit is a runtime fix this needed: transformers' Qwen3.5 text config drops `full_attention_interval` when it derives `layer_types`, and mlx-lm resolves it on the built text model, so `_extract_model_args` now reads omitted keys off `language_model.args`. Same bug class as #449, with the omitted-key test.

### DIFF
--- a/vllm_metal/attention/runtime/factory.py
+++ b/vllm_metal/attention/runtime/factory.py
@@ -1,0 +1,43 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Factory for hybrid state-family runtime plans."""
+
+from __future__ import annotations
+
+from collections.abc import Callable, Mapping
+from dataclasses import dataclass
+from typing import Any
+
+from vllm_metal.attention.runtime.families.gdn import (
+    build_gdn_hybrid_plan,
+    supports_gdn_hybrid,
+)
+from vllm_metal.attention.runtime.hybrid_plan import HybridRuntimePlan
+
+
+@dataclass(frozen=True, slots=True)
+class StateFamilyPlanBuilder:
+    supports: Callable[[Mapping[str, Any]], bool]
+    build: Callable[[Mapping[str, Any], int], HybridRuntimePlan]
+
+
+_STATE_FAMILY_PLAN_BUILDERS = (
+    # ``ModelConfig.is_hybrid`` only says a model mixes attention and state
+    # layers; the family that owns its topology and geometry is resolved here.
+    StateFamilyPlanBuilder(
+        supports=supports_gdn_hybrid,
+        build=build_gdn_hybrid_plan,
+    ),
+)
+
+
+def build_hybrid_runtime_plan(
+    model_args: Mapping[str, Any], num_layers: int
+) -> HybridRuntimePlan:
+    for builder in _STATE_FAMILY_PLAN_BUILDERS:
+        if builder.supports(model_args):
+            return builder.build(model_args, num_layers)
+
+    raise NotImplementedError(
+        f"Metal hybrid runtime has no state family for "
+        f"model_type={model_args.get('model_type')!r}."
+    )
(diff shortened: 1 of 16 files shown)


vllm_metal/attention/runtime/families/gdn.py
```diff
@@ -108,6 +108,11 @@ def state_geometry(self) -> RecurrentStateGeometry:
 )
 
 
+def supports_gdn_hybrid(model_args: Mapping[str, Any]) -> bool:
+    """Whether these model args describe a GDN linear-attention hybrid."""
+    return "full_attention_interval" in model_args
```
LxYuan0420, line 113: Non-blocking follow-up: this is fine for the current GDN-only table, but when the Mamba-2/Nemotron row lands, GDN matching should become more explicit than only full_attention_interval. I would match known Qwen GDN model types plus required linear_* geometry fields so unsupported hybrids keep failing closed.

vllm_metal/attention/runtime/factory.py
```diff
@@ -0,0 +1,43 @@
...
+from vllm_metal.attention.runtime.hybrid_plan import HybridRuntimePlan
+
+
+@dataclass(frozen=True, slots=True)
+class StateFamilyPlanBuilder:
+    supports: Callable[[Mapping[str, Any]], bool]
+    build: Callable[[Mapping[str, Any], int], HybridRuntimePlan]
+
+
+_STATE_FAMILY_PLAN_BUILDERS = (
```
LxYuan0420, line 23: Non-blocking follow-up for the next family row: keep this table as the only hybrid-family selector. ModelConfig.is_hybrid should answer hybrid or not; this table should answer which supported state family owns the topology.

vllm_metal/attention/runtime/factory.py
```diff
@@ -0,0 +1,43 @@
...
+    # ``ModelConfig.is_hybrid`` only says a model mixes attention and state
+    # layers; the family that owns its topology and geometry is resolved here.
+    StateFamilyPlanBuilder(
+        supports=supports_gdn_hybrid,
+        build=build_gdn_hybrid_plan,
+    ),
+)
+
+
+def build_hybrid_runtime_plan(
```
LxYuan0420, line 33: Non-blocking follow-up: once there is more than one family, please add a negative test where an unsupported hybrid model has some overlapping fields but still rejects instead of routing to the wrong family.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -494,7 +494,7 @@ def _reject_pipeline_parallel_with_per_layer_metadata(self) -> None:
     def _install_hybrid_state_plan(self, args: dict[str, Any]) -> None:
         """Resolve the state family that owns this model's hybrid layers."""
         if self._runner.is_hybrid:
-            self._runner.hybrid_runtime_plan = build_gdn_hybrid_plan(
+            self._runner.hybrid_runtime_plan = build_hybrid_runtime_plan(
```
LxYuan0420, line 497: Non-blocking follow-up: I like that lifecycle stays as one call into the family builder. As more hybrid families land, keep model-specific topology parsing inside attention/runtime/families/*, not in lifecycle or runner.

LxYuan0420 (approved):
Great work!

PR #697, opened 2026-09-05
[CI] Run lint only on macOS

WindChimeRan:
CI currently runs the same lint script on macOS, Ubuntu ARM, and Ubuntu x86. This PR removes the two Ubuntu matrix entries and retain the `lint (macos-15)` one. 

test on ubuntu is not very meaningful to vllm-metal.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       contents: read
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


LxYuan0420 (approved):
Those are not really the bottleneck; would rather see us tackle the slow macOS test job, which is the actual CI bottleneck.

PR #699, opened 2026-09-05
Honor configured LoRA dtype

LxYuan0420:
This PR is:

- To honor vLLM’s resolved `LoRAConfig.lora_dtype` when creating Metal LoRA adapter buffers
- To run LoRA and QLoRA adapter inputs/deltas in the configured adapter dtype
- To preserve the base layer output dtype when adding the LoRA delta

Previously, Metal LoRA used `kv_cache_dtype or mx.float16` for adapter buffers, so `--lora-dtype` could be silently ignored.
This keeps dtype ownership inside `MetalLoRARuntime` and keeps `model_runner.py` as orchestration only.

### DIFF
--- a/tests/test_lora.py
+++ b/tests/test_lora.py
@@ -9,6 +9,7 @@ from types import SimpleNamespace
 
 import numpy as np
 import pytest
+import torch
 from vllm.lora.layers import LoRAMapping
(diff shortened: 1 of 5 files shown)


WindChimeRan (approved):
LGTM!

PR #701, opened 2026-09-05
[Test][M5] Use CPU references for GGUF matmul parity

WindChimeRan:
* **cause**: On M5, MLX can use reduced-precision TF32 for the dense FP32 matmul used as the GGUF test reference. 
* **effect**: M5 only:  nine parity cases to fail even though the quantized results agree closely with an independent CPU reference.

This PR: Compute the four dense matmul references on CPU and explain the precision requirement in a code comment. The GGUF operations still run on GPU with the existing tolerances.

Validation on Apple M5 Pro, macOS 26.6.2, MLX 0.32.0:

- Native GGUF and wrapper suites, excluding slow tests: before the change, **9 failed / 48 passed** with `MLX_ENABLE_TF32=1`; after the change, **57 passed** with either `MLX_ENABLE_TF32=1` or `0`.
- Five slow cases were deselected in each run; the optional real-file case was not exercised.
- Ruff checks, formatting, and `git diff --check` passed.

### DIFF
--- a/tests/test_gguf_mlx_native.py
+++ b/tests/test_gguf_mlx_native.py
@@ -72,12 +72,11 @@ def test_matmul_matches_dense_oracle_f32(tmp_path, qtype):
     x = mx.random.normal((3, qt.in_features)).astype(mx.float32)
 
     out = qt.matmul(x)
-    expected = x @ mx.array(oracle).T
+    # M5 GPU matmul may use TF32; keep the reference at full FP32 precision on CPU.
+    expected = mx.matmul(x, mx.array(oracle).T, stream=mx.cpu)
     mx.eval(out, expected)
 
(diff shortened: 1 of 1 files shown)


LxYuan0420 (approved):

now-ing:
Ran the GGUF suite on an M4 Max (macOS 26, MLX 0.32.0) to check the CPU-reference approach on another GPU generation: `pytest -m "not slow" -k gguf` → **37 passed, 4 skipped**, including all the matmul-parity cases that motivated the CPU references. No flakiness at the current `atol`, so the M5-Pro-observed TF32 drift doesn't reproduce on M4 with references on CPU — consistent with the fix being about where the reference lives rather than GPU-family-specific numerics.

PR #703, opened 2026-09-06
install.sh: fall back to prebuilt artifacts or JIT when Metal toolchain is unavailable

### LINKED ISSUE #684: [FR] install.sh: graceful degradation when the Metal toolchain is unavailable (warn + JIT fallback hint, or fetch prebuilt artifacts from the matching release wheel)
## Background

A source-checkout `./install.sh` runs `ensure_metal_toolchain` + `build_native_artifacts`, which need the `metal` compiler to produce the `.metallib` files. On newer toolchains the Metal compiler ships as a separate component that only full Xcode can fetch (`xcodebuild -downloadComponent MetalToolchain`), so on a machine with only **Xcode Command Line Tools** the script dies at the very end:

```
Metal toolchain not available, downloading via xcodebuild...
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory
'/Library/Developer/CommandLineTools' is a command line tools instance
```

Everything before that step (venv, vLLM core wheel, editable install, Rust build) has already succeeded, and the CONTRIBUTING docs explicitly describe a supported flow that needs only CLT: run with `VLLM_METAL_BUILD_FROM_SOURCE=1`, where MLX compiles the `.metal` shaders in-process at runtime.

So the failure is (a) at the end of a long install, (b) recoverable without full Xcode, and (c) not connected in the error message to the documented CLT-only workflow.

## Proposal

Two options (not mutually exclusive):

**1. Degrade to a warning + actionable hint (small change)**

When the Metal toolchain probe fails and `xcodebuild` is unavailable, print a warning instead of exiting, skip `build_native_artifacts`, and finish the install with a prominent post-install note:

> Metal toolchain not found. Either install full Xcode and rerun `./install.sh`,
> or run inference with `VLLM_METAL_BUILD_FROM_SOURCE=1` (MLX will JIT-compile
> the shaders in-process; only Command Line Tools required).

**2. Fetch prebuilt artifacts from the matching release wheel (nicer UX)**

Since the source checkout pins its vLLM core to a release tag, the script could, on a missing Metal toolchain, download the same-version `vllm_metal-<version>-cp312-...whl` from GitHub Releases and unpack `_paged_ops*.so` + `*.metallib` into `vllm_metal/metal/`. The runtime already loads wheel artifacts as-is (no hash stamps), and the Python side stays editable. I verified this manually on M5 Pro / v0.28.0 source checkout + v0.28.0 wheel artifacts and Qwen3.8-27B-4bit serves correctly.

Happy to turn option 1 (and optionally 2) into a PR if the approach sounds good.

now-ing:
## Summary

Implements the graceful-degradation approach proposed by @suntp in #684 (thanks for the detailed write-up and the manual verification of the wheel-unpack path!).

On a machine with only Xcode Command Line Tools, \`./install.sh\` used to die at the very end: the Metal toolchain probe fails, and \`xcodebuild -downloadComponent MetalToolchain\` itself errors out because xcodebuild requires full Xcode. Everything before that step (venv, vLLM core wheel, editable install, Rust build) had already succeeded.

The source-checkout path now degrades instead of aborting:

1. **\`ensure_metal_toolchain\` tolerates a missing/failing xcodebuild** — it returns 1 ("toolchain unavailable") instead of letting the non-zero status abort the script under \`set -e\`.
2. **New \`fetch_prebuilt_native_artifacts\`** — downloads the newest release wheel whose tag matches this checkout's pyproject version (\`vX.Y.Z\` or \`vX.Y.Z.devN\`, so the ABI stays in family) and unpacks the \`_paged_ops*.so\` + the four \`.metallib\` files into \`vllm_metal/metal/\`. The runtime loads these as-is, so an editable Python install plus the unpacked files is fully functional (verified on the release artifacts below).
3. **Final fallback** — if no matching wheel exists either, the install finishes with a pointer to the documented \`VLLM_METAL_BUILD_FROM_SOURCE=1\` JIT path (MLX compiles the shaders in-process; CLT is sufficient).

## Issue

Fixes #684 (option 1 + option 2).

## Validation

- \`shellcheck\`, \`bash -n\` clean on the touched files (no new warnings; the SC2250 style notes pre-date this change and come from shellcheck 0.11 — CI runs an older shellcheck).
- End-to-end wheel unpack verified against the real \`v0.28.0\` stable wheel: all 5 expected artifacts (\`_paged_ops.cpython-312-darwin.so\`, \`paged_attention_v2_kern.metallib\`, \`gdn_kern.metallib\`, \`paged_mla_kern.metallib\`, \`paged_attention_nax_kern.metallib\`) extract to \`vllm_metal/metal/\` with the expected names resolved from \`build.py\`.
- The fallback chain is exercised only when \`ensure_metal_toolchain\` fails; machines with a working toolchain see no behavior change.

## Type of Change

- [x] Bug fix / robustness improvement
- [x] Feature (install UX)

## Area

- [x] \`install.sh\` / \`scripts/lib.sh\`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -266,8 +266,23 @@ EOF
 
     # Source checkouts build native artifacts; release installs use the wheel.
     uv pip install -e .
-    ensure_metal_toolchain
-    build_native_artifacts
+    if ensure_metal_toolchain; then
+      build_native_artifacts
+    else
+      # No Metal toolchain (e.g. Xcode Command Line Tools only): fall back to
+      # the matching release wheel's prebuilt artifacts, and failing that,
+      # finish the install with a pointer to the JIT runtime path.
+      if fetch_prebuilt_native_artifacts; then
+        echo ""
+        echo "Note: serving will use the prebuilt .metallib files unpacked above."
+        echo "To build them locally instead, install full Xcode and rerun ./install.sh."
+      else
+        warning "Continuing without prebuilt Metal artifacts."
+        echo "To compile them locally, install full Xcode and rerun ./install.sh." >&2
+        echo "Or run inference with VLLM_METAL_BUILD_FROM_SOURCE=1: MLX JIT-compiles" >&2
+        echo "the shaders in-process, and Command Line Tools are sufficient for that." >&2
+      fi
+    fi
   else
     local release_data selected release_tag wheel_url vllm_release_tag
     release_data=$(fetch_release "$repo_owner" "$repo_name" "$channel")
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -181,8 +181,11 @@ build_native_artifacts() {
 fetch_prebuilt_native_artifacts() {
   section "Fetching prebuilt native artifacts from release wheel"
 
+  # The venv python is already active here; reading pyproject through it
+  # avoids the noisy `uv run` VIRTUAL_ENV-mismatch warning.
   local version
-  if ! version=$(get_version); then
+  if ! version=$(python -c \
+      "import tomllib; print(tomllib.load(open('pyproject.toml','rb'))['project']['version'])"); then
     error "Failed to read the project version from pyproject.toml."
     return 1
   fi
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -6,6 +6,11 @@ error() {
   echo -e "Error: $*" >&2
 }
 
+# Print a warning message
+warning() {
+  echo -e "Warning: $*" >&2
+}
+
 # Print a success message
 success() {
   echo -e "✓ $*"
@@ -127,7 +132,14 @@ ensure_metal_toolchain() {
   fi
 
   echo "Metal toolchain not available, downloading via xcodebuild..."
-  xcodebuild -downloadComponent MetalToolchain
+  # A Command Line Tools-only machine has no xcodebuild at all; treat that as
+  # "toolchain unavailable" rather than letting the non-zero status abort the
+  # caller (install.sh falls back to prebuilt or JIT artifacts).
+  if ! xcodebuild -downloadComponent MetalToolchain; then
+    warning "Could not download the Metal toolchain (xcodebuild failed or unavailable)."
+    rm -rf "${tmpdir}"
+    return 1
+  fi
 
   if ! xcrun -sdk macosx metal -o "${metal_lib}" "${metal_src}" &> /dev/null; then
     error "Metal toolchain still unavailable after download; cannot compile .metallib."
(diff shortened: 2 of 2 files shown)


now-ing:
Validated the fallback chain end-to-end on a CLT-only setup (M4 Max, macOS 26, `DEVELOPER_DIR=/Library/Developer/CommandLineTools` on a machine that does have full Xcode — which reproduces the exact failure mode from the issue, since xcodebuild then reports CLT as the active developer directory).

```
=== Ensuring Metal toolchain ===
Metal toolchain not available, downloading via xcodebuild...
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory
'/Library/Developer/CommandLineTools' is a command line tools instance
Warning: Could not download the Metal toolchain (xcodebuild failed or unavailable).
=== Fetching prebuilt native artifacts from release wheel ===
Downloading vllm_metal-0.28.0.dev20260905205506-cp312-cp312-macosx_15_0_arm64.whl...
unpacked 5 artifacts
✓ Prebuilt native artifacts installed into vllm_metal/metal/

Note: serving will use the prebuilt .metallib files unpacked above.
To build them locally instead, install full Xcode and rerun ./install.sh.
✓ Installation complete!   (exit 0)
```

On unchanged main this run dies at the `xcodebuild -downloadComponent MetalToolchain` line; with this PR the whole install completes and `vllm_metal/metal/` ends up with the extension plus all four `.metallib` files (version-matched dev wheel: this checkout is 0.28.0, the resolver picked the newest `v0.28.0.dev*` release).

Two minor notes from the run:

- `get_version` shells out to `uv run python`, which prints a one-line `VIRTUAL_ENV ... does not match the project environment` warning under the already-activated venv. Harmless, but reading the version via the venv's own `python -c "import tomllib; ..." ` would keep the log clean — happy to switch if preferred.
- with `UV_INDEX_URL` unset the vLLM wheel's dependency resolution pulls everything from PyPI directly, which can be glacial on constrained networks; not this PR's problem, just noting that the fallback path keeps working behind a mirror.

WindChimeRan (commented):
I'm not very clear on what issue this PR is trying to solve. 

We currently have two paths
* users & python developer can use one-line installation from the readme
* kernel developers need to install many dependencies themselves, and compile by themselves. 

---

Maybe a doc clarification would just work?

suntp:
@WindChimeRan @now-ing Thanks for the review and for implementing and testing the fallback. After looking more closely at the source-development case, I agree that this can be scoped down.

My earlier validation was of a specific v0.28.0 source/wheel combination. In the current patch, `fetch_prebuilt_native_artifacts` selects the newest release within a version family such as `0.28.0.dev*`. That does not establish compatibility with the checkout's native interfaces, kernel implementation, or pinned MLX version. Copying those artifacts into a source checkout can also leave local Python changes running against a different native implementation. For unreleased-kernel validation such as #715, that is particularly important. The absence of wheel build stamps is not a compatibility check.

**Suggested revision:**

1. Prioritize the installation/development documentation: provide a complete CLT-only source/JIT setup and clearly separate it from wheel inference and ahead-of-time native builds.
2. If installer changes are retained, keep them small and explicit: preflight the chosen mode before expensive work; handle toolchain discovery/download failures with a clear next step. If an explicit JIT mode is added, skip ahead-of-time Metal compilation and make the required runtime setting clear. Avoid an unconditional “Installation complete” that leaves the default serve command unable to load native artifacts.
3. Remove the automatic “newest same-version-family wheel” extraction from this PR. If it is revisited later, it should have an explicit compatibility/provenance contract rather than relying on the version prefix alone.

For any retained installer behavior, validate the working-toolchain path, a CLT-only explicit JIT path, and early failure with actionable guidance when the selected build path lacks its prerequisites. A docs-only revision is also fine if it supplies a verified, complete setup procedure.

One independent merge blocker: lint and tests are green, but DCO is still `ACTION_REQUIRED` because the first commit, `d4cf9f6`, lacks a `Signed-off-by` line. @now-ing will need to add their sign-off to that commit and update the branch.

This would address #684 while keeping the initial change focused on setup clarity and predictable behavior.

PR #704, opened 2026-09-06
[Bugfix] Route Phi-1/1.5 through paged attention (dense output projection + derived scale)

now-ing:
## Summary

`mlx_lm.models.phi` (phi-1 / phi-1.5 / phi-2 checkpoints) currently cannot serve on the paged path, for two independent reasons:

1. **Output projection name**: `PhiAttention` spells its output projection `dense`, not `o_proj`, so `is_sdpa`'s split-projection contract rejects it (and the forward path would crash on `inner.o_proj` even if dispatched).
2. **Scale attribute**: like StableLM before #687, Phi computes the standard softmax scale inline in its forward (`scale = math.sqrt(1 / queries.shape[-1])`) and exposes neither `scale` nor `sm_scale`, so `sdpa_forward` raises `AttributeError` before dispatch.

Fixes:

- `_output_projection(module)` probes `o_proj` then `dense`; both `is_sdpa`'s split-projection branch and the `sdpa_forward` output call go through it, keeping dispatch and forward aligned on the same alias set.
- Registers `"mlx_lm.models.phi"` with `derive_scale_from_query=True` — the exact pattern #687 established for StableLM. The derived formula matches `PhiAttention`'s own forward verbatim. Unregistered scale-less modules keep failing loudly before dispatch (unchanged).
- Adds `docs/supported_models.md` row for Phi.

## Issue

Follow-up to the scale discussion in #656 (whose author measured mlx-lm having exactly two inline-scale SDPA families — StableLM, now fixed by #687, and Phi, addressed here).

## Validation

- New regression test drives the **real pinned** `mlx_lm.models.phi.PhiAttention` through `sdpa_forward` with a fake ops layer: asserts `is_sdpa(inner)` is True and the kernel receives `head_dim ** -0.5`.
- `pytest tests/test_attention_sdpa.py`: **30/30 passed** (including the StableLM and GPT-OSS scale regressions unchanged); `tests/test_hunyuan_qk_norm.py` passed.
- `ruff check` + `ruff format --check` clean; `mypy` clean for the touched modules.
- Note: `PhiAttention`'s partial RoPE (`partial_rotary_factor`) is encoded in its own `nn.RoPE` dimensions, which `prepare_sdpa_qkv` already applies as-is — same conclusion #687 reached for StableLM's RoPE.

## Type of Change

- [x] Bug fix

## Area

- [x] Paged attention SDPA dispatch

### DIFF
--- a/docs/supported_models.md
+++ b/docs/supported_models.md
@@ -95,6 +95,7 @@ Llama-3.2-1B-Instruct, and Mistral-7B-Instruct-v0.3 Q8_0
 | Mistral-7B | ✅ | GQA (paged) | ✅ | `mlx-community/Mistral-7B-Instruct-v0.3-4bit` |
 | Mistral-Small-24B | 🔵 | GQA (paged) | ✅ | `mlx-community/Mistral-Small-24B-Instruct-2501-4bit` |
 | StableLM 2 | ✅ | MHA + partial RoPE (paged) | ✅ | `mlx-community/stablelm-2-zephyr-1_6b-4bit` |
+| Phi 1.5 / Phi 2 | ✅ | MHA/MQA + partial RoPE (paged) | ✅ | `mlx-community/phi-2-4bit` |
 | GPT-OSS | 🔵 | Sink attention (paged) | ✅ | `openai/gpt-oss-20b` |
 | GLM-4.5 | 🟡 | MLA (paged latent cache, MLX SDPA — no Metal kernel) | 🟡 | — |
 | MiniCPM3-4B | ✅ | MLA (paged latent cache) | ✅ | `mlx-community/MiniCPM3-4B-4bit` |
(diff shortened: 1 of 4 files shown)


LxYuan0420 (approved):
Good fix. I pushed one small docs commit to replace the Phi example checkpoint with one that actually routes through mlx_lm.models.phi; the previous example used model_type="phi2", which does not match the pinned mlx-lm module path. With that mismatch fixed, this looks good to merge once CI passes.

LxYuan0420 (approved):
Good fix. I pushed one small docs commit to replace the Phi example checkpoint with one that actually routes through mlx_lm.models.phi; the previous example used model_type="phi2", which does not match the pinned mlx-lm module path. With that mismatch fixed, this looks good to merge once CI passes.

PR #706, opened 2026-09-06
[CI] Move model parity to nightly and /ci parity runs

WindChimeRan:
## Motivation

Our current CI is running slow e2e smoke test on every commit / PR. 

This PR aims to 
* make per commit / PR CI **faster**, by removing the slow e2e test.
* move the slow e2e tests to nightly CI, and it can be manually triggered by `/ci parity` with write access. In the future we can improve the test coverage, it's ok to be **slower** (e.g., add speculative decoding. It should have been catch a recent correctness regression). 

## Summary

Adds a daily run at 07:17 UTC on `main` and `/ci parity` for PRs, restricted to users with repository write access. 

Each run tests the captured commit against native `mlx-lm`: Qwen3-0.6B and Qwen3.5-0.8B, macOS 15/26, 40 shared prompts, 20 output tokens, top-K 5, and HTTP request batch sizes 1/2.

CI and local use invoke the same `tools/check_parity.py` command. For each model/OS, it generates one native MLX reference, exits that process, starts one `vllm serve` instance, checks `/health`, compares prompts through `/v1/completions` first individually, then in pairs within one request, and shuts the server down. The offline Metal execution path is removed.

The `Parity` check reports the tested SHA and whether it matches the current PR head at report time. It links to per-model summaries and artifacts containing the native reference, server log, per-batch results, and environment versions, retained for 14 days. Regular PR CI retains wheel validation, Metal platform checks, and non-slow tests.

Depends on #705. The comment and scheduled triggers activate once the workflow is on the default branch.

## Validation

- The canonical CLI passed all 160 local comparisons across both models, 40 prompts, 20 output tokens, and HTTP request batch sizes 1/2: 124 EXACT, 36 TOP_K_MATCH, 0 FAIL. Each model used one server for 40 single-prompt and 20 paired requests, followed by verified process cleanup. Local validation used `VLLM_METAL_MEMORY_FRACTION=0.10`; CI retains `0.65`.
- All 18 focused comparison, request-batching, orchestration, and reporting tests passed, including response ordering, partial final batches, and missing/duplicate choices. Ruff and workflow validation passed.
- The full hosted matrix has not been completed for this version.

## Read the result

click `Actions` tab in github, scroll down and find the run. 

image

## Lesson learnt

1. VLLM_METAL_MEMORY_FRACTION
- **No**: `VLLM_METAL_MEMORY_FRACTION=0.80`: hit the 45-minute timeout
- **Yes**: `0.65`: the same job passed in 9m33s Keep this setting for the small GHA runners.

2. Different Metal compilation behavior
- Metal Shading Language 3.2 on macOS 15 
- Metal Shading Language 4.0 on macOS 26
- So this CI keep both

## Acknowledgement

Learnt a lot from this post: https://vllm.ai/blog/2026-07-16-keeping-vllm-production-quality

### DIFF
--- a/.github/scripts/parity.py
+++ b/.github/scripts/parity.py
@@ -96,6 +96,28 @@ def start() -> dict | None:
 
 def finish(check_id: str, result: str) -> None:
     conclusion = result if result in {"success", "cancelled"} else "failure"
+    tested_sha = os.environ["PARITY_TESTED_SHA"]
+    summary = f"Tested commit: `{tested_sha}`.\n\n"
+    if os.environ["GITHUB_EVENT_NAME"] == "issue_comment":
+        event = json.loads(Path(os.environ["GITHUB_EVENT_PATH"]).read_text())
+        number = int(event["issue"]["number"])
+        try:
+            head_sha = github_api(f"pulls/{number}")["head"]["sha"]
+            freshness = (
+                "Up to date at report time."
+                if head_sha == tested_sha
+                else "Outdated — the current PR head was not tested by this run. "
+                "Comment `/ci parity` to test the current head."
+            )
+            summary += f"Current PR head: `{head_sha}`.\n\n{freshness}\n\n"
+        except (OSError, ValueError, KeyError) as error:
+            print(f"::warning::Could not read current PR head: {error}")
+            summary += "Freshness unknown: could not read the current PR head.\n\n"
+    summary += (
+        f"[View workflow run]({workflow_url()}) for per-model "
+        "summaries of EXACT, TOP_K_MATCH, and FAIL. Download the parity "
+        "artifacts for per-batch logs and environment versions."
+    )
     github_api(
         f"check-runs/{int(check_id)}",
         method="PATCH",
@@ -104,12 +126,12 @@ def finish(check_id: str, result: str) -> None:
             "conclusion": conclusion,
             "output": {
                 "title": f"Parity matrix: {conclusion}",
-                "summary": f"[View workflow run]({workflow_url()}) for per-model "
-                "summaries of EXACT, TOP_K_MATCH, and FAIL. Download the parity "
-                "artifacts for per-batch logs and environment versions.",
+                "summary": summary,
             },
         },
     )
+    with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as output:
+        print(f"### Parity matrix: {conclusion}\n\n{summary}", file=output)
 
 
 def main() -> None:
--- a/.github/scripts/parity.py
+++ b/.github/scripts/parity.py
@@ -1,0 +1,132 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Authorize parity runs and report checks; never execute PR code here."""
+
+import argparse
+import json
+import os
+import urllib.error
+import urllib.parse
+import urllib.request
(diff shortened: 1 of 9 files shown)


WindChimeRan:
Hope this will make reviewer's life easier! @LxYuan0420 @ricky-chaoju

.github/scripts/parity.py
```diff
@@ -0,0 +1,132 @@
...
+        raise
+    if permission not in {"admin", "maintain", "write"}:
+        return None
+    number = int(event["issue"]["number"])
+    pr = github_api(f"pulls/{number}")
+    if pr["state"] != "open" or not pr["head"]["repo"]:
+        return None
+    return {
+        "repository": pr["head"]["repo"]["full_name"],
+        "sha": pr["head"]["sha"],
```
LxYuan0420, line 70: /ci parity should be tied to the commit I reviewed. Please make the command include the full SHA and reject it if the PR head changed before the job starts.

.github/workflows/parity.yml
```diff
@@ -0,0 +1,159 @@
...
+          path: ~/.cache/huggingface/hub
+          key: ${{ steps.models.outputs.cache-primary-key }}
+      - name: Compare all batch sizes
+        run: |
+          source .venv-vllm-metal/bin/activate
+          printf '### %s\n\nCommit: %s\n\n| Batch | EXACT | TOP_K_MATCH | FAIL | Exit |\n|---|---:|---:|---:|---:|\n' "$MODEL_ID" "$TESTED_SHA" >> "$GITHUB_STEP_SUMMARY"
+          failed=0
+          for batch in 1 2; do
+            log="$RUNNER_TEMP/parity/batch-${batch}.log"
+            if python tools/check_parity.py --model "$MODEL_PATH" --top-k 5 --max-tokens 20 --batch-size "$batch" 2>&1 | tee "$log"; then
```
LxYuan0420: This replaces token parity, but it does not cover the old serving smoke. Please keep one small vllm serve check with /health and /v1/completions so API startup stays covered.

.github/scripts/parity.py
WindChimeRan: new changes: After the CI run finishes, log the commit SHA in the final report and compare it with the new HEAD to determine if it's up to date.

I think /ci parity SHA is overkill (you'll need to copy & paste the SHA each time). We'll add more commands in the future, e.g., weekly ci, spec decoding ci ...

LxYuan0420: that make sense; yes lets keep it simple for now

LxYuan0420 (approved):
Thanks for working on this!

.github/workflows/parity.yml
WindChimeRan: refactored the tool & CI around the http api

```
For each model / OS:
1. Generate the native MLX reference once, then exit.
2. Start one vllm server with max-num-seqs=2.
3. Check /health.
4. Run parity requests one at a time.
5. Run parity requests two at a time.
6. Stop the server.
```
This removes most of the duplicated model loading & reference gen & server waiting time. 

Reduce the CI nightly time from 40min to 20min (tested in GHA). https://github.com/vllm-project/vllm-metal/actions/runs/34509882546

nightly total time is not very stable, as it depends on GHA's machine availability.

PR #709, opened 2026-09-06
[feat] Experimental prefill/decode disaggregation via file-based KV connector

now-ing:
## Summary

This PR adds the first prefill/decode (PD) disaggregation path for vllm-metal, and to our knowledge the first on a pure Apple-Silicon cluster (the only prior public Mac PD work mixes a DGX Spark with Macs).

Upstream has never shipped a KV connector: the earlier drafts #474 / #497 / #499 were all closed unmerged with the reasoning that the plugin should wait *"until it is ready and demonstrates meaningful improvement."* We read that bar as two things: a complete, correctly-scoped implementation, and honest evidence of what it does. This PR is the bring-up attempt at the first half — the full v1 connector interface on vLLM 0.28.0, zero changes to the vllm core repo, and end-to-end correctness proof: **token-identical greedy decoding through a producer→consumer handoff, both on one Mac and across two Macs.** It is deliberately **correctness-first**: the transfer plane is a shared directory of per-layer safetensors files, which keeps the diff small, the format debuggable, and the transport swappable. Performance is explicitly not claimed and is mapped as follow-up work.

## Design

- **`vllm_metal/kv_connector.py` (new)** — `MetalFileConnector`, one class playing both roles via `kv_role`. Implements the full `KVConnectorBase_V1` surface: `get_num_new_matched_tokens`, `update_state_after_alloc`, `build_connector_meta`, `start_load_kv`, `save_finished_requests`.
  - Transfer medium: a shared directory (local FS / NFS / sshfs). One safetensors file per layer holding `mx.array` block-pool slices `{"key","value"}` with shape `[n_blocks, block_size, kv_heads, head_dim]`, plus a `done` marker written only after all layers are complete — a consumer can never read a half-written prefix.
  - Lookup key: sha256 of the block-aligned prompt prefix.
- **`vllm_metal/v1/model_runner.py`** — the worker-role connector is initialized lazily after `initialize_kv_cache`, loaded externally through `KVConnectorFactory` + `kv_connector_module_path` — **zero intrusion into the vllm core repo**. Bulk load before the forward; bulk store after the async forward materializes (in the `sample_tokens` path). In PD mode the decode pipeline gate is forced ineligible, preserving per-step synchronous sampling.
- **`vllm_metal/v1/worker.py`** — `get_kv_connector_handshake_metadata` returns `None`: a file-based connector needs no out-of-band handshake, and this keeps EngineCore's startup handshake-collection path working.
- **Scope guards** — single-KV-group models only (dense MHA/GQA); hybrid linear-attention raises `NotImplementedError`; the connector does not declare `SupportsHMA`, hence `--disable-hybrid-kv-cache-manager` is required for PD runs.
- **`docs/pd_disaggregation.md` (new)** — full documentation: architecture, quick start (single Mac + cross-Mac), verification method, benchmark, limitations, roadmap.

## Validation

- **Unit tests** (`tests/test_kv_connector.py`, 4/4 green; `ruff check .`, `ruff format --check .`, and `mypy vllm_metal` all clean): store→load block-level roundtrip into *different* block ids; load without a `done` marker raises `FileNotFoundError`; scheduler miss→hit after a store; `update_state_after_alloc` + `build_connector_meta` state cleanup.
- **End-to-end, single Mac, two processes** (Qwen3-0.6B @ `c1899de`, `block_size=16`, 28 KV layers, 190-token prompt): the producer stored 11 blocks × 28 layers (19 MB); the consumer reported the external hit, loaded the blocks, and its greedy output matched the non-disaggregated ground truth **28/28 tokens, identical**.
- **End-to-end, two Macs**: prefill on an M4 Max 128 GB → Wi-Fi LAN → decode on an M3 Max 96 GB. Same 11-block hit and load; output again token-identical to ground truth.
- **Benchmark** (3,400-token prompt, 48 generated tokens): monolithic on the decode Mac = 1.60 s; PD = 1.17 s prefill (M4 Max) + 5.12 s rsync ship (371 MB, demo transport) + **0.59 s decode (M3 Max, KV loaded)** — the decode node's work drops **2.7×**. End-to-end is currently dominated by the demo transport, not the connector; the numbers characterize the file transfer plane, not the ceiling of PD on Apple Silicon.
- **Why token identity is the headline check**: under greedy decoding, any corruption, dtype mismatch, or block-layout error in the transferred KV pool would diverge the output stream immediately — matching every token is the strongest practical proof the handoff is lossless.

## Limitations & Next steps

Honest scope of this PR:

- **Synchronous file transfer** — store/load sit on the forward path; performance was not the goal of this bring-up.
- **Demo-grade cross-machine transport** — the validated cross-Mac run used a two-phase rsync-over-ssh watcher (~72 MB/s observed over Wi-Fi); NFS/sshfs shares also work.
- **Short prompts** (< `block_size + 1` tokens) are not transferred.
- **Chunked prefill**: only requests scheduled as a single full chunk are stored (PoC limitation).
- **Hybrid / MLA models not supported** (hybrid linear-attention explicitly raises `NotImplementedError`).

Next steps, in rough order:

1. **Async, layered streaming** of store/load as the forward progresses.
2. **A TCP transfer connector** (NIXL or similar) — the interface stays, only the transport swaps; that is where any "meaningful improvement" claim would come from.
3. **Hybrid-model support**, lifting the single-KV-group restriction.
4. **Heterogeneous clusters** — the reported 2.8× for a DGX Spark + Mac mix is the shape where PD pays off around Macs today.

We're aware of the history on #474 / #497 / #499 and the "ready + meaningful improvement" bar. This PR is explicitly the *readiness* half — complete interface, zero core-repo intrusion, lossless-transfer proof, limits stated upfront — with the performance half deliberately left as the follow-up it needs to be. Happy to rescope, rename, or split anything the maintainers would rather see shaped differently.

### DIFF
--- a/docs/pd_disaggregation.md
+++ b/docs/pd_disaggregation.md
@@ -91,6 +91,7 @@ The decode instance's work drops **2.7×** (1.60 s → 0.59 s) — that is the d
 - **Synchronous file transfer.** Store/load sit on the forward path. Performance is explicitly not the goal of this bring-up; correctness is.
 - **Demo-grade cross-machine transport.** The validated cross-Mac setup used a two-phase rsync-over-ssh watcher (~72 MB/s observed over Wi-Fi). NFS/sshfs shares work; nothing here is fast yet.
 - **Short prompts don't transfer.** Prefixes shorter than `block_size + 1` tokens are not stored/loaded.
+- **All-or-nothing hits.** The transfer key is the whole block-aligned prompt, so there is no prefix sharing across different-length prompts: a shorter prompt does not hit a longer one's store. Fine for v1; prefix-granular keys are future work. `cache_salt` is folded into the key; LoRA adapters are not yet (same prompt under different adapters must not share stores today — use separate storage paths per adapter).
 - **Chunked prefill.** Only requests scheduled as a single full chunk are stored (PoC limitation).
 - **Single KV cache group only.** Dense MHA/GQA models. Hybrid linear-attention models raise `NotImplementedError`; MLA is not supported.
 - **HMA disabled.** The connector does not declare `SupportsHMA`; PD runs require `--disable-hybrid-kv-cache-manager`.
--- a/docs/pd_disaggregation.md
+++ b/docs/pd_disaggregation.md
@@ -80,6 +80,12 @@ Token identity is the core of the check: under greedy decoding, any corruption, 
 
 The decode instance's work drops **2.7×** (1.60 s → 0.59 s) — that is the disaggregation effect: the consumer no longer prefills. End-to-end time is currently dominated by the demo rsync transport, not by the connector; the numbers characterize the *file* transfer plane, not the ceiling of PD on Apple Silicon.
 
+## Design notes (anticipated review questions)
+
+- **Why store only the block-aligned prefix (the `len - 1` alignment)?** The consumer must compute at least the final prompt token itself, so the transferable prefix excludes the last (possibly partial) block. This mirrors `ExampleConnector`'s key semantics; the consumer's forward covers the unaligned tail, and greedy output was verified token-identical against monolithic serving.
+- **Why bulk `save_finished_requests` instead of per-layer `save_kv_layer`?** The Metal path has no per-attention-op hookpoint (MLX attention wrappers do not expose the layer boundary the CUDA custom-op path does), so the plugin's model runner performs one bulk store after the async forward materializes. This is a deliberate runner-side integration, trading the async overlap of per-layer saves for a minimal, auditable diff; the roadmap's layered streaming restores the overlap.
+- **Integrity.** Every store carries a `manifest.json` (layer count, block geometry, dtype); loads validate it and fail fast on mismatch — a wrong-shaped or stale store can never be scattered into the paged cache. `max_stored_prefixes` (extra config) optionally prunes oldest stores.
+
 ## Limitations
 
 - **Synchronous file transfer.** Store/load sit on the forward path. Performance is explicitly not the goal of this bring-up; correctness is.
--- a/docs/pd_disaggregation.md
+++ b/docs/pd_disaggregation.md
@@ -1,0 +1,97 @@
+# Prefill/Decode Disaggregation (Experimental)
+
+!!! note
+    This feature is **experimental** — a correctness-first bring-up of prefill/decode (PD) disaggregation on Apple Silicon, not a performance feature. It is validated end-to-end on Qwen3-0.6B, single Mac and across two Macs over Wi-Fi, with **token-identical** greedy output against non-disaggregated serving. Transfer is synchronous file I/O over a shared directory; see [Limitations](#limitations) and [Roadmap](#roadmap) before relying on it.
+
+vllm-metal can run **disaggregated prefill**: one instance (the *producer*, `kv_role=kv_producer`) computes prompt KV and writes it out; a second instance (the *consumer*, `kv_role=kv_consumer`) serving the same model detects the stored prefix, loads the blocks into its own KV cache, and decodes from there without re-prefilling. Both roles are played by a single connector class, `MetalFileConnector`, plugged in through vLLM's standard v1 KV-transfer config. To our knowledge this is the first PD-disaggregated serving path on a pure Apple-Silicon cluster (prior public Mac PD work mixes a DGX Spark with Macs, not Macs only).
+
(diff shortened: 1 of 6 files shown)


now-ing:
Did a careful pass over `kv_connector.py` (contract semantics against `KVConnectorBase_V1`, plus an independent model review on the side). The overall shape is right: the `-1` in `_align_to_block_size(len - 1, bs)` matches the ExampleConnector precedent, returning `load_kv_async=False` makes the synchronous bulk `start_load_kv` contract-compliant, and because vLLM v1 allocates blocks for externally-matched tokens, `block_ids[:num_blocks]` is well-defined under chunked prefill. The done-marker-last ordering is also the right basic idea. Below is what I'd want addressed before this lands, ordered by severity.

**1. save-vs-free race (correctness of the shared store).** `save_finished_requests` reads the request's blocks synchronously, but nothing here implements the `request_finished(...) -> delay_free_blocks` half of the connector contract. If the scheduler frees the request's blocks and a new request reuses them before the save's gather runs, the shared directory gets permanently poisoned with another request's KV — every future consumer of that prompt hash reads wrong data. Either prove the ordering (save strictly within the same step, before any free/reuse) in a comment, or implement `request_finished` returning `delay_free_blocks=True` + a `get_finished` handshake.

**2. No cross-instance validation of cache shape/identity.** The hash covers tokens + mm hashes only. Two engines with different `block_size`, different kv-heads/head-dims, or two different *models* sharing the directory silently scatter garbage. A tiny `meta.json` (model identity, block_size, kv shape, dtype) written next to `done` and asserted on load would make every mismatch loud. Also worth folding vLLM's `cache_salt` / LoRA id into the hash — same prompt under a different adapter must not hit.

**3. Non-atomic publication and the third state.** Layer files are written in place; a producer crash after `done`... can't happen (done last), but a consumer can see `done` while layer bytes are still in page cache on NFS (no fsync), and a crash mid-directory leaves "dir exists, no done" — which today means *neither* store nor load, so that prompt is never cached again (the producer branch skips because the dir exists; the consumer branch skips because done is absent). Writing to `tmp_<pid>/` + `os.rename` + fsync of the dir resolves publication atomicity and the third state in one move.

**4. `assert total_need_load == len(self._requests_need_load)`.** Request cancellation between `update_state_after_alloc` and `build_connector_meta` turns this into an EngineCore crash. The standard answer is to drain state in `request_finished` rather than assert.

**5. Minor.** (a) `get_num_new_matched_tokens` can return a negative number when internal prefix-cache hits already exceed the aligned external length — worth clamping to 0; (b) the uint32 fancy-index scatter into the pools deserves a read-back unit test on the MLX side (index-dtype support and lazy-materialization timing); (c) a design note worth stating in the docs: because the key is the whole block-aligned prompt, hits are all-or-nothing — no prefix sharing across different-length prompts, which is fine for v1 but should be an explicit known limitation.

Also left one drive-by on the branch: `ruff format` was failing on `kv_connector.py` (CI lint gate) — pushed a format-only commit (`d4efc98`) on top.

now-ing:
Follow-up: I also reproduced the end-to-end handoff independently on a fresh checkout of this branch (M4 Max 128 GB, macOS 26, vLLM 0.28.0 wheel, source-built artifacts, Qwen3-0.6B @ `c1899de`, ~190-token prompt).

- Monolithic ground truth (no `--kv-transfer-config`): 48 greedy tokens recorded.
- Producer on :8100 stores `12 KV blocks x 28 layers`; a **cold-restarted** consumer on :8200 (restarted specifically so its internal prefix cache cannot mask the transfer — worth doing in any future validation run, since a warm consumer hits its own vLLM prefix cache and matches trivially) then logs `External KV cache hit` + `Loaded 12 KV blocks x 28 layers` and produces **48/48 token-identical** output.

One operational note for anyone reproducing: the proxy pattern (forward to producer with `max_tokens=1`, then to the consumer) means a stale server already bound to the producer port silently breaks the chain — the producer phase hits a non-connector instance and the consumer quietly prefills locally. A startup assertion that the connector actually attached (or logging the resolved role loudly on first request) would make misconfiguration visible. Repro assets (stdlib demo proxy + three-instance driver script): I can upstream them into `tools/` or a docs snippet if wanted.

WindChimeRan (commented):
What's the motivation of PD-disaggregation on apple silicon? 

I checked your docs/pd_disaggregation.md, but It doesn’t establish a convincing motivation for vllm-metal.

PR #614, opened 2026-08-16
[1/5] Plan scheduler-owned speculative GDN state

PhilipJohnBasile:
## Summary

Phase 1 of the vLLM Metal #610 stack.

Reserve scheduler-owned Mamba/GDN speculative blocks, expose a per-request state-chain contract, and keep multi-token prefill on the ordinary path.

This is intentionally dependency-scoped and does not include later Qwen MTP phases.

## Validation

Validated on the upstream-matching `macos-15-arm64` runner against `vllm-project/vllm-metal@083f581f048b1b4460f166b31ec18dc6ec000e17`.

Focused gate:
- `tests/attention/test_align_gdn_state_manager.py`
- `tests/test_turboquant_hybrid_sizing.py`
- `tests/test_platform.py`
- 86 passed, 10 skipped

The complete five-phase stack also passed the full non-slow repository gate with 1,804 combined passing tests, plus Ruff, formatting, mypy, native Metal artifact build, and MLX↔PyTorch MPS bridge preflights.

Full validation run: https://github.com/PhilipJohnBasile/mlx-lm/actions/runs/31932330097

## Review order

- `vllm_metal/attention/context.py`
- `vllm_metal/attention/state/align.py`
- `vllm_metal/attention/runtime/hybrid.py`
- `vllm_metal/v1/cache_policy.py`

Part of #610.

Signed-off-by: Philip John Basile <PBasile@Basilecom.com>

PR #710, opened 2026-09-06
Run Mamba-2 mixers on paged state and serve Nemotron-H

ricky-chaoju:
PR 3 of the Nemotron-H stack in #644, on top of #696, with the wrapper and the family row in one PR since a wrapper nothing wires in is the scaffolding shape from #614. On main the 30B Nemotron 4bit checkpoint fails at load with no state family. With this it serves: 23 Mamba-2 mixers on paged state, 6 attention layers on the SDPA kernel, 23 MoE blocks left alone.

The wrapper does not re-implement the mixer. It hands mlx-lm's own mixer a small `ArraysCache` stand-in whose reads and writes land on rows of `GDNPagedStateCache`, which fits Mamba-2 unchanged. Prefill runs one request per call because a packed batch contaminates every request after the first, decode is one gathered batch, and the wrapper owns the mask, which was the `TypeError` in #644. The family row parses `hybrid_override_pattern` into a three-role plan, and the new stateless role keeps the MoE blocks out of the spec loop and `patch_model`. The GDN row matches on model type as you asked on #696, with the overlapping-fields test.
Prefill runs one request per mixer call only when several prompts prefill in the same step; a single prompt is one call per layer, and batching equal-length prefills inside the mixer's constraint is the follow-up.


now-ing:
Read through the wrapper and the family plumbing — the `ArraysCache`-as-pool-rows design is a really clean way to keep mlx_lm owning the SSD/SSM math, and the fail-loud posture throughout (pattern token whitelist, positive-int field validation, the `M`/`*` coexistence check, and especially the `ssm_state_size % 32` guard, which would otherwise be a silent unwritten-columns bug) is exactly right for a first serving path.

Three smaller things while reading:

1. **`_SlotArraysCache.lengths = None` is an implicit contract with the pinned mlx_lm.** "Slices arrive unpadded so the mixer's right-padding gather stays off" is a statement about mlx_lm's internal branch on `cache.lengths`; the repo pins mlx-lm by git SHA, so this is safe today, but when that pin next moves it's the kind of behavioral coupling that breaks silently. Worth one line in the class docstring naming it as a pin-sensitive invariant (or a startup smoke assert that a `lengths=None` cache with an unpadded slice skips the gather).
2. **Pattern length vs `num_layers` has no explicit check.** `layer_roles` is consumed by index (`is_attention_layer(layer_idx)` etc.); if a checkpoint's `hybrid_override_pattern` length disagrees with the layer count, the failure is a bare IndexError somewhere deep in cache planning rather than a named error next to the pattern parsing. `NemotronHHybridConfig.__post_init__` already validates tokens — adding a length-vs-num-layers check there (or wherever the plan meets the real layer count) would keep misconfigured checkpoints failing at load with a message that names the pattern.
3. **Serial per-request prefill** across 23 Mamba-2 mixers is the right correctness-first call (the SSD path's per-batch-state limitation is real), just flagging that the Python-level loop per layer per request is the dominant prefill cost on long prompts — if there's a roadmap item for chunking or batching within the constraint, worth stating in the PR body so nobody reads the single-request loop as an oversight.

vllm_metal/attention/patching.py
```diff
@@ -37,8 +37,9 @@ def find_layers(model: Any) -> list[Any]:
         )
 
 
-# Attribute names to probe on each layer, in priority order.
-_ATTN_ATTR_NAMES = ("self_attn", "linear_attn", "attention")
+# Attribute names to probe on each layer, in priority order; Nemotron-H names
+# every block's module ``mixer``.
+_ATTN_ATTR_NAMES = ("self_attn", "linear_attn", "attention", "mixer")
```
LxYuan0420: Keep `mixer` out of the global patcher. `walk_and_wrap()` is shared by MHA, MLA, and hybrid runtimes, and the plain paths do not prove the discovered module is SDPA before wrapping it. Please make `mixer` a Nemotron/hybrid-only probe so future mlx-lm models with non-attention `mixer` modules do not get patched accidentally.

Suggested shape:

```python
_DEFAULT_ATTN_ATTR_NAMES = ("self_attn", "linear_attn", "attention")

def find_attn_attr(
    layer: Any,
    attr_names: tuple[str, ...] = _DEFAULT_ATTN_ATTR_NAMES,
) -> str | None:
    for name in attr_names:
        if hasattr(layer, name):
            return name
    return None

def walk_and_wrap(
    model: Any,
    wrap_layer: Any,
    *,
    only_layers: list[int] | None = None,
    attr_names: tuple[str, ...] = _DEFAULT_ATTN_ATTR_NAMES,
) -> int:
    ...
    attn_attr = find_attn_attr(layer, attr_names)
    ...
```

Then the hybrid family can opt into `("mixer",)` only where needed.

vllm_metal/platform.py
```diff
@@ -530,6 +530,24 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
                     "'align' for models without SupportsMambaPrefixCaching). "
                     "Use align mode: --enable-prefix-caching resolves to it."
                 )
+            if cache_config.enable_prefix_caching and config.use_paged_attention:
+                from vllm_metal.attention.runtime.factory import (
+                    state_family_for_model_type,
+                )
+
+                state_family = state_family_for_model_type(
+                    model_config.hf_config.model_type
```
LxYuan0420: Use `model_config.hf_text_config.model_type` here. vLLM exposes `hf_text_config` as the text-model contract, and this platform hook runs before lifecycle flattens text args. Looking at outer `hf_config.model_type` can choose the wrong hybrid family for multimodal models.

Suggested shape:

```python
state_family = state_family_for_model_type(
    model_config.hf_text_config.model_type
)
```

A focused test can set different outer and text model types and assert the platform uses the text model type.

LxYuan0420 (approved):

Issue #685, opened 2026-09-03
[Verification + RFC] qwen3_5 vision works with mlx-vlm 0.6.8 + multimodal-native — auto-mode text-backbone override could key on the installed mlx-vlm version


suntp:
## Summary

The `auto` multimodal mode forces qwen3_5-family MLX-quantized checkpoints onto the text backbone because mlx_vlm "runs them with unset mRoPE state and produces garbled output" (`_matches_auto_text_backbone_override`). I verified on real hardware that the garbling is a **version artifact of the pinned mlx-vlm**, not inherent to the architecture: with mlx-vlm **0.6.8** (the version mlx-community used to convert these checkpoints, still within the plugin's `>=0.6.2,<0.7` range) and `VLLM_METAL_MULTIMODAL_MODE=multimodal-native`, **Qwen3.8-27B-4bit vision works correctly on the Metal path**.

## Verified setup

- M5 Pro / 64 GB, macOS 26.x, vllm-metal 0.28.0 source checkout + vLLM core 0.28.0 wheel
- mlx-vlm 0.6.8 (bumped from the resolver's 0.6.4) + transformers 5.15.1 (see caveat below)
- `VLLM_METAL_MULTIMODAL_MODE=multimodal-native vllm serve <local snapshot>`
- Sent a synthetic image (red triangle + blue square + the number 42); the model described every element, position, and the overlap/occlusion between the shapes accurately, in both `reasoning` and `content`. No garbling.

Two unrelated bugs had to be worked around to reach that result — both filed upstream:

1. vLLM core's ModelScope path conversion builds a stale cache layout so multimodal processor loading crashes ([vllm-project/vllm#55189](https://github.com/vllm-project/vllm/issues/55189)); serving the local snapshot path directly avoids it.
2. transformers 5.14+ auto-mapping for qwen3_vl/qwen3_5 references image processor classes whose module no longer exists, so `AutoProcessor` fails even on official checkpoints ([huggingface/transformers#48511](https://github.com/huggingface/transformers/issues/48511)); a small compat patch resolving the dangling names to `Qwen2VLImageProcessor` fixes it.

## RFC: refine the auto-mode override

[...]

PR #711, opened 2026-09-06
Gate the qwen3_5/3_6 auto-mode text-backbone override on the installed mlx-vlm

now-ing:
## Summary

Implements the version-scoped override proposed in #685.

The MLX-affine arm of `_matches_auto_text_backbone_override` currently forces every MLX-quantized Qwen3.5/3.6 conditional-generation wrapper onto the text backbone, because the pinned mlx-vlm (0.6.4) drives them with unset mRoPE state and garbles output. #685 established — and I independently A/B-verified at a second checkpoint size (comment on the issue) — that mlx-vlm **0.6.8**, the revision these checkpoints were converted with and still inside the plugin's `>=0.6.2,<0.7` pin, serves them correctly on the multimodal path:

- mlx-vlm 0.6.4 + `multimodal-native` + Qwen3.5-4B-MLX-4bit: unconditional multi-language token soup.
- mlx-vlm **0.6.8** + same model, same image, nothing else changed: accurate structured description (shapes, colors, positions, the number 42, and the overlap/occlusion).

This PR adds a tiny probe (`_mlx_vlm_garbles_qwen35_mrope`, version tuple < (0, 6, 8)) and gates only that arm on it, so `auto` mode keeps native vision once a fixed mlx-vlm is installed while known-bad versions keep today's safe behavior. The probe fails closed (returns True → keep override) when mlx-vlm cannot be imported or its version is unparseable. FP8 handling, the Gemma4 branch, and `multimodal-native` bypass behavior are all unchanged.

Note the transformers `<5.13` cap still gates actually *installing* 0.6.8 in one environment with vllm-metal (0.6.8 wants transformers>=5.14); that cap decision stays with the maintainers — this change only makes the override honest about which mlx-vlm is actually loaded, which also matters for anyone testing 0.6.8 in an isolated venv today (as #685's author did).

## Issue

Implements the RFC half of #685.

## Validation

- `pytest tests/test_model_adapter.py`: **77 passed** (existing override test now pins the probe to the known-bad side so it stays environment-independent; new test covers the fixed-version side asserting the override does not fire).
- `ruff check` + `ruff format --check` clean; `mypy` output identical to unchanged main (4 pre-existing gguf-enum notes in this local env, zero new).
- The runtime A/B evidence is in the #685 comment (4B) and the original issue (27B, M5 Pro).

## Type of Change

- [x] Bug fix / behavioral refinement

## Area

- [x] Multimodal model adaptation

### DIFF
--- a/tests/test_model_adapter.py
+++ b/tests/test_model_adapter.py
@@ -7,6 +7,7 @@ import mlx.core as mx
 import pytest
 
 import vllm_metal.envs as envs
+import vllm_metal.v1.model_adapter
 from vllm_metal.config import reset_config
 from vllm_metal.multimodal.paddleocr_vl import PaddleOCRVLMultimodalAdapter
 from vllm_metal.multimodal.qwen3_vl import Qwen3VLMultimodalAdapter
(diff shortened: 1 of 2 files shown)


ricky-chaoju:
Thanks for this. transformers 5.13 no longer breaks Metal on the runners, so #712 raises the mlx-vlm floor to 0.6.8 and removes the override instead of gating it, which makes this unnecessary.

PR #712, opened 2026-09-06
[Bugfix] Unpin transformers and serve MLX-quant Qwen3.5/3.6 natively

ricky-chaoju:
The `transformers<5.13` cap came from #471, added when 5.13.0 was A/B'd on the CI runner as the reason MLX Metal stopped resolving inside the vLLM process, and it has sat behind a TODO ever since. It is now also what blocks mlx-vlm 0.6.8, the release that fixes the unset-mRoPE garbling on MLX-quantized Qwen3.5/3.6 wrappers reported in #685, so auto mode still routes those checkpoints to the text backbone and they serve without vision.

Four transformers minors later the premise no longer holds, so this drops the cap rather than working around it. The mlx-vlm floor moves to 0.6.8, capped below 0.6.16 which drops `target_verify` from `Qwen3_5MLP.__call__` and trips the `CompiledMLPBlocks` guard. With a fixed mlx-vlm pinned, the MLX-affine arm of the auto-mode override is dead weight, so the second commit deletes it together with `_has_mlx_quantized_weights`. The Gemma4 and FP8 arms stay; they are separate unfixed problems.

This makes #711 unnecessary and is mutually exclusive with it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,16 +46,14 @@ dependencies = [
     # import tvm-ffi; this bounds the resolver only.
     # TODO: remove once xgrammar bounds its own apache-tvm-ffi range.
     "apache-tvm-ffi>=0.1.9,<0.1.13; platform_system == 'Darwin' and platform_machine == 'arm64'",
-    # mlx-vlm 0.6.2 includes PaddleOCR-VL multi-image M-RoPE fixes.
-    "mlx-vlm>=0.6.2,<0.7.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
+    # mlx-vlm 0.6.8 drives MLX-quantized Qwen3.5/3.6 wrappers with correct
+    # mRoPE state; 0.6.4 leaves it unset and garbles their output (#685).
+    # Capped below 0.6.16, which drops `target_verify` from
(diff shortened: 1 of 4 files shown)


reneleonhardt:
Thank you so much for unpinning desperately needed updates! ❤️

vllm-metal v0.28.0 declares requires-python = ">=3.12,<3.14" while not providing Python 3.13 wheels, and Python 3.15 around the corner.

LxYuan0420 (approved):
Good fix. Before merge, can you help me double-check the MoE path? `qwen3_5_moe` now stays on the native multimodal path, but I don’t see `qwen3_5_moe` / `Qwen3_5MoeForConditionalGeneration` in `build_multimodal_adapter()`. If that still returns `None`, an image request may hit the runner’s missing-adapter guard.

ricky-chaoju:
You're right. `build_multimodal_adapter` only knows `qwen3_5` and `Qwen3_5ForConditionalGeneration`, so the MoE and 3.6 wrappers would hit the runner guard with a None adapter. I only checked the dense checkpoint.

Pushed a fix that ties the override to whether an adapter actually exists, sharing one predicate with the dispatch so the two lists can't drift. Only the dense wrapper leaves the override now. Thanks @LxYuan0420 !

Issue #713, opened 2026-09-06
Benchmarking pitfalls on macOS: in-process LLM() decode penalty (1.5-2.1x), desktop GPU co-tenancy, non-diagnostic short probes


suntp:
## Summary

While validating the decode work in #715 (successor of #693) on an M5 Pro (64GB, macOS 26.6,
Qwen3.8-27B-4bit) we hit - and initially mis-attributed - three distinct
measurement hazards that can silently distort performance numbers on
macOS. All three reproduced with hard data; none is a vllm-metal bug per
se, but together they argue for a short "Benchmarking on macOS" doc and
two tiny quality-of-life additions. Cross-references: #695.

## Pitfall 1: in-process `LLM(...)` (VLLM_ENABLE_V1_MULTIPROCESSING=0)
understates decode 1.5-2.1x vs the server topology

Single sequence, 100k context, 119-token generations, 3 runs each,
variance < 0.5%:

| Configuration | server (`vllm serve`) | in-process | penalty |
|---|---|---|---|
| pre-PR paged kernel | 4.25 tok/s | 2.82 tok/s | 1.51x |
| GQA flash-decode | 9.44 tok/s | 4.41 tok/s | 2.14x |

The penalty **grows with the speed of the path measured**, so it also
compresses ratios: a true 2.22x A/B reads as 1.56x in-process. Relative
comparisons look plausible while both the absolute values and the ratio
are wrong. Likely mechanism: the engine core thread shares one
interpreter/GIL with the driver's output processing; per-step CPU-side
scheduling stretches and the GPU idles between steps (direct kernel
measurements inside the same process show full bandwidth, so the GPU is
fine). In-process mode is a debug mode by design - the hazard is using
it for performance claims.

## Pitfall 2: desktop UI co-tenancy silently eats GPU time

With vLLM fully idle and all engine processes killed, the GPU can still
sit at **57-59% device utilization** driven by WindowServer (~49% CPU)
and app renderer processes (an Electron-style chat client in our case):

```bash
ioreg -r -c AGXAccelerator -d "1" | grep -oE '"Device Utilization %"=[0-9]+'
```

Measured impact: a quiet desktop vs an active one moved real decode at
100k from **11.2 tok/s to 9.44 tok/s (-16%)** and a short GEMM probe by
**~5x**. It looks exactly like a machine regression or thermal
throttling (it is neither: `pmset -g therm` clean, AC, high-power mode),
[...]

PR #714, opened 2026-09-06
perf: VLLM_METAL_DISABLE_GQA_DECODE escape hatch + macOS benchmarking guide

suntp:
> **Stacked on an older #715 commit.** The disable flag, `sdpa_forward` plumbing, `docs/configuration.md` row, and routing tests now live **in #715 itself** (`b66217c`, per review). Please do **not** merge this before #715.
>
> After #715 merges, this PR should be rebased onto main as **docs-only**: the remaining unique file is `docs/benchmarking-macos.md` (#713). If that doc is folded into #715 instead, this PR can close.

## What remains here

**`docs/benchmarking-macos.md`** — three measurement hazards from #713 with detection recipes:

- server topology for absolute claims (in-process `LLM()` understates decode 1.5–2.1× and compresses ratios)
- AGX `Device Utilization %` gate for desktop GPU co-tenancy (measured −16% on real decode)
- why short-burst GEMM probes are not a fast/slow oracle

## What moved to #715

`VLLM_METAL_DISABLE_GQA_DECODE`, both `paged_attention_primitive` call sites, primitive equality, the dispatch kill switch, and the tests that assert the env reaches the primitive. Reviewing those diffs on this PR will duplicate (and, until rebase, **conflict with**) #715.

Addresses the documentation half of #713.

### DIFF
--- a/docs/benchmarking-macos.md
+++ b/docs/benchmarking-macos.md
@@ -2,7 +2,7 @@
 
 Three measurement hazards documented in [#713] can silently distort
 performance numbers on Apple Silicon. All three were hit while
-validating the decode work in #693 on an M5 Pro; each cost hours of
+validating the decode work in [#715](https://github.com/vllm-project/vllm-metal/pull/715) (successor of #693) on an M5 Pro; each cost hours of
 misattribution before being isolated.
 
 ## 1. Absolute numbers: use the server topology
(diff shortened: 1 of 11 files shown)


suntp:
Rebased onto #715 (the clean single-commit successor of the closed #693) — diff vs #715 is now just this PR's own commit.

suntp:
Per the review on #715, the disable flag + routing tests now live in #715 itself (`b66217c`). This PR's remaining unique content is `docs/benchmarking-macos.md` (#713). After #715 merges I'll rebase this onto main as docs-only (or close it if that doc is folded in). Please don't merge this before #715.

PR #715, opened 2026-09-06
perf: scoped GQA paged decode for single-request long contexts

suntp:
This PR adds a GQA decode kernel inside the paged attention primitive, with automatic routing restricted to the measured single-request, long-context cases requested in review. `VLLM_METAL_DISABLE_GQA_DECODE=1` keeps eligible calls on the established path, and tests inspect the actual C++ dispatch.

## Scope

Automatic routing requires one pure-decode request, matching FP16/BF16 Q/K/V, kernel block size 16, and one of these geometries and inclusive KV-length ranges:

| Query heads | KV heads | Head dimension | KV tokens |
|---:|---:|---:|---:|
| 32 | 8 | 128 | 32,768–131,072 |
| 24 | 4 | 256 | 32,768–131,072 |
| 16 | 2 | 128 | 65,536–131,072 |

All rows also require `ceil(KV tokens / 512) × KV heads >= 3 × detected GPU cores`; unknown core counts fall back. This occupancy guard and the table are empirical scope limits, not a guarantee for every GPU or competing workload. There is no model-name or device-name allowlist. The 16/2/128 geometry deliberately remains on the established path at 32K.

Multi-request batches, speculative verification windows, lengths outside the table, TurboQuant, sinks, soft-capping, and sliding-window attention retain the established family. `num_decode_requests` must be 1 or omitted, and the verification window must be at most 1. The disable switch cannot broaden eligibility.

The kernel groups query heads using the same KV head in a threadgroup and feeds 512-token partials into the existing reduction. Each simdgroup still issues its own K/V loads; grouping can improve locality, without an explicit cross-simdgroup broadcast or a fixed DRAM-byte saving. Startup calibration, cached thresholds, and mutable gate parameters have been removed. No runtime permission framework is added.

## Validation

Local revision: `efa0e31cda60fb2bb018d5a2588af3617e29ac63`; native artifacts rebuilt and worker-side stale-artifact checks clear. GitHub lint, test, and DCO checks pass for this head ([CI run](https://github.com/vllm-project/vllm-metal/actions/runs/34749135355)).

- Targeted routing, numerical, SDPA, and verification tests: **224 passed, 1 skipped**.
- Non-slow repository suite: **1,986 passed, 19 skipped, 39 deselected**. The local runner redirects only the Whisper tokenizer fixture to ModelScope-sourced files and blocks Hugging Face HTTP; this fixture-source redirection is outside the repository and does not alter assertions. These are local results, not remote CI results.
- Routing tests evaluate the primitive before checking `last_paged_dispatch()`, covering GQA selection, length bounds, geometry/block exclusions, multi-request fallback, and the disable flag. Numerical checks include reference attention, native SDPA, and relative-error bounds for small long-context outputs.

Serving environment: M5 Max (40 GPU cores, 48 GiB), macOS 26.6.2, Python 3.12.9, vLLM 0.28.0+cpu, MLX 0.32.0, mlx-lm 0.31.3, Transformers 5.12.1. Separate `vllm serve` arms, one request, cached prefixes, prefill chunk 2,048, at least 30 seconds warmup per length, 3 × 128 generated tokens; the table reports median decode tok/s, computed as `(completion_tokens - 1) / (last_token_time - first_token_time)` with returned token IDs and usage verified. Compiled MLP and native sampling are off. `max_model_len` / memory fraction: Qwen3-4B 40,960 / 0.4; Qwen3.8-27B-4bit 33,024 / 0.85; MiniCPM5-2B 131,072 / 0.4.

| Model | Prompt tokens | Default route | Default tok/s | Disabled tok/s | Ratio |
|---|---:|---|---:|---:|---:|
| Qwen3-4B | 32,768 | `gqa_decode` | 40.32 | 29.89 | 1.35× |
| Qwen3.8-27B-4bit | 32,768 | `gqa_decode` | 24.38 | 21.14 | 1.15× |
| MiniCPM5-2B | 16,384, reverse-order control | `per_token_ps512` | 74.82 | 75.00 | 1.00× |
| MiniCPM5-2B | 32,768 | `per_token_ps512` | 51.40 | 51.85 | 0.99× |
| MiniCPM5-2B | 65,536 | `gqa_decode` | 43.59 | 29.27 | 1.49× |
| MiniCPM5-2B | 120,000 | `gqa_decode` | 29.67 | 14.03 | 2.11× |

Worker checks after every measured request confirm the listed route; every disabled arm uses `per_token_ps512`. Positive GQA cells have within-arm CV ≤ 1.1%. Existing background tasks were retained. Recorded thermal states included nominal and fair, with no new swapouts during measured phases. GPU-utilization readings include the benchmark itself and do not establish identical resource conditions between arms.

The initial default-first MiniCPM5 16K run was 65.57 vs 73.64 tok/s (0.89×), despite both arms selecting the baseline. The difference did not reproduce with reversed arm order (74.82 vs 75.00, 0.998×); both results are retained. The first discrepancy is not attributable to GQA kernel execution.

MiniCPM5's 64K and 120K free-running generations first diverge at output token 36. Follow-up fixed-history checks at three positions per length use two-token requests: the first token comes from prefill, and the second from an actual decode forward. All six pairs have identical decode histories and choose the same second token; Top-10 overlap is 9–10/10. Full-vocabulary BF16 logits are not identical (relative L2 0.97–2.75%, maximum absolute difference 0.4375). Re-prefilling the preceding history changes its KV computation; these sampled checks do not reproduce the original accumulated decode trajectory or establish full-generation equivalence. Qwen3-4B, Qwen3.8-27B-4bit, and the shorter MiniCPM5 runs had matching output tokens.

Separate single-evaluation primitive probes cover each geometry's lower and upper bounds in FP16/BF16: 12 cells, six alternating AB/BA pairs each, all observed pairs favor GQA, with cell median ratios 1.51–3.84×. Five cells exceed 3% CV (maximum 3.72%); these are noisier measurements, not 12 uniformly stable results. Primitive ratios are not HTTP throughput ratios.

Earlier M5 Pro figures in this discussion refer to `b66217c`, not this revision.

## Broader routing investigated

We tested an occupancy floor plus a scalar potential-KV-reread proxy, fitted at startup. Additional geometries, submission modes, and competing GPU work did not justify extrapolating one fitted threshold to every supported shape. Actual DRAM traffic depends on cache residency, and a deterministic selection rule does not make its performance independent of runtime load.

Further experiments used alternating paired timings, independent confirmation, numerical/dispatch checks, telemetry, and compute/bandwidth competition. An invocation-bound permit prototype added expiry, one-time use, revocation, and native validation before encoding. It handled known invalidation but did not establish performance for subsequent requests after resource conditions changed; always-fallback runs did not validate a positive optimization. That machinery remains outside this PR.

These findings support returning to explicit measured scope, not a conclusion that general optimization is impossible. Wider routing should come with its own numerical, dispatch, and real-serving evidence. See [routing documentation](https://github.com/suntp/vllm-metal/blob/perf/gqa-decode-paged/docs/gqa-decode.md) and the [macOS benchmarking discussion](https://github.com/vllm-project/vllm-metal/issues/713).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -8,6 +8,7 @@
 | `VLLM_MLX_DEVICE` | `gpu` | MLX device (`gpu` or `cpu`) |
 | `VLLM_METAL_USE_PAGED_ATTENTION` | `1` | Enable experimental paged KV cache |
 | `VLLM_METAL_DISABLE_NAX` | `0` | Emergency override for automatic M5 NAX prefill attention. Set to `1` to force the non-NAX fallback. |
+| `VLLM_METAL_DISABLE_GQA_DECODE` | `0` | Force-close the GQA flash-decode dispatch gate ([#715](https://github.com/vllm-project/vllm-metal/pull/715) / [#713](https://github.com/vllm-project/vllm-metal/issues/713)). Set to `1` so long-context single-request decode stays on the established per-token / split-KV kernels — the server-topology A/B hatch (mirrors `VLLM_METAL_DISABLE_NAX`). Off by default. The GQA kernel only auto-routes a single decode request with KV ≥ 16384; multi-request batches never take it. |
 | `VLLM_METAL_MULTIMODAL_MODE` | `auto` | Multimodal serve mode: `auto` uses the compatibility allowlist; `multimodal-native` disables overrides |
 | `VLLM_USE_MODELSCOPE` | `False` | Set True to change model registry to <https://www.modelscope.cn/> |
 | `VLLM_METAL_MODELSCOPE_CACHE` | None | Specify the absolute path of the local model |
--- a/docs/gqa-decode.md
+++ b/docs/gqa-decode.md
@@ -1,0 +1,94 @@
+# GQA decode routing
+
+The paged attention primitive can use `paged_attention_gqa_decode` for a
+limited set of single-request, long-context decode calls. Automatic routing
+is intentionally narrower than the shapes supported by the shader.
+
+## Automatic eligibility
+
+The following geometry and KV-length bounds are inclusive:
+
+| Query heads | KV heads | Head dimension | Minimum KV tokens | Maximum KV tokens |
+|---:|---:|---:|---:|---:|
+| 32 | 8 | 128 | 32,768 | 131,072 |
+| 24 | 4 | 256 | 32,768 | 131,072 |
+| 16 | 2 | 128 | 65,536 | 131,072 |
+
+Every row additionally requires:
+
+- One pure-decode request, with `num_decode_requests` equal to 1 or omitted.
+- A verification window of at most 1.
+- Matching FP16 or BF16 query, key-cache, and value-cache types.
+- Kernel block size 16. This is the block size passed to the primitive after
+  any hybrid-cache view conversion.
+- No TurboQuant, attention sinks, logit soft-capping, or sliding-window
+  attention.
+- Enough partition threadgroups for the conservative occupancy guard:
+  `ceil(KV tokens / 512) * KV heads >= 3 * detected GPU cores`.
+  An unknown GPU core count falls back.
+
+Calls outside these bounds use the established attention family. That
+includes multi-request batches, even if each request individually matches a
+row, and lengths above the maximum. The 16/2/128 geometry deliberately starts
+at 64K; small measured gains at 32K were not used to widen automatic routing.
+
+The gate checks geometry rather than model names and does not contain a
+device-name allowlist. Its bounds and occupancy guard are empirical choices,
+not a claim that GQA wins for every device or competing GPU workload.
+
+## Disable switch
+
+Set `VLLM_METAL_DISABLE_GQA_DECODE=1` before starting the server to keep
+eligible requests on the established path. This switch provides an A/B and
+operational fallback; it cannot enable an otherwise ineligible call.
+
+There is no startup calibration or disk-cached performance threshold for
+this gate. `VLLM_METAL_GQA_AUTOTUNE` and the former mutable gate-parameter
+APIs are no longer used.
+
+## Validation
+
+`tests/test_gqa_paged_decode.py` evaluates the primitive and checks
+`last_paged_dispatch()` alongside the numerical reference. Positive cases
+must actually report `gqa_decode`; boundary, multi-request, verification,
+feature, and disabled cases must report the appropriate established family.
+`tests/test_attention_sdpa.py` checks that the environment switch reaches
+the primitive, and `tests/test_native_sdpa_decode.py` provides a separate
+numerical comparison.
+
+`last_paged_dispatch()` records the last family selected in the process. It
+is suitable for serial tests and isolated worker checks; it is not
+per-request telemetry for concurrent serving. Evaluate the operation before
+reading it, because MLX builds graphs lazily.
+
+For performance validation, rebuild native artifacts from the tested
+revision and use the real `vllm serve` process topology. Record the model,
+runtime versions, hardware, warmup, KV lengths, repetitions, and background
+GPU activity; compare enabled and disabled arms with actual worker-side
+dispatch checks. Separate primitive timing from HTTP decode throughput,
+and keep noisy measurements visible. The
+[macOS benchmarking discussion](https://github.com/vllm-project/vllm-metal/issues/713)
+explains why in-process engine measurements and short probes are not
+interchangeable with serving results.
+
+## Why the broader gate was removed
+
+An earlier design combined a partition-occupancy floor with a scalar
+potential-KV-reread proxy and fitted a threshold at startup. Additional
+geometry, submission-mode, and load tests did not justify extending that
+single threshold to all shader-supported cases. The proxy is not measured
+DRAM traffic: each simdgroup still issues its own K/V loads, and the actual
+traffic depends on cache residency. Grouping related heads can improve
+locality without an explicit cross-simdgroup KV broadcast.
+
+A separate experiment added paired timing, confirmation runs, numerical
+checks, telemetry, and invocation-bound permits with expiry, revocation,
+and native validation before encoding. It rejected known invalidation but
+did not establish performance for later requests after resource conditions
+changed. Tests in which requests always fell back did not validate positive
+GQA performance. That experimental machinery is not part of this routing
+implementation.
+
+These limitations motivate the explicit scope above; they do not establish
+that a more general optimization is impossible. Expanding automatic routing
+requires its own dispatch, numerical, and real-serving benchmark evidence.
--- a/tests/test_attention_sdpa.py
+++ b/tests/test_attention_sdpa.py
@@ -701,6 +701,53 @@ class TestSDPAForward:
         assert captured["num_kv_heads"] == 2
         assert captured["scale"] == 0.5
 
+    def test_gqa_disable_env_reaches_primitive(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """``sdpa_forward`` must pass ``VLLM_METAL_DISABLE_GQA_DECODE`` through."""
+        captured: dict[str, bool | None] = {}
+
+        class _FakeOps:
+            def reshape_and_cache(
+                self,
+                _keys,
+                _values,
+                key_cache,
+                value_cache,
+                _slot_mapping,
+            ) -> tuple[mx.array, mx.array]:
+                return key_cache, value_cache
+
+            def paged_attention_primitive(self, *_args, **kwargs) -> None:
+                captured["gqa_disabled"] = kwargs.get("gqa_disabled")
+
+        inner = _make_inner()
+        inner.o_proj = lambda out: out
+        cache = MetalPagedKVCache(
+            num_layers=1,
+            num_kv_heads=_N_KV_HEADS,
+            head_dim=_HEAD_DIM,
+            num_blocks=1,
+            block_size=8,
+            dtype=mx.float16,
+        )
+        x = mx.ones((_BATCH, _SEQ_LEN, _HIDDEN), dtype=mx.float16)
+        zeros = mx.zeros((_BATCH, _SEQ_LEN, _N_HEADS * _HEAD_DIM), dtype=mx.float16)
+
+        def _run() -> None:
+            with (
+                patch.object(sdpa_mod, "get_ops", return_value=_FakeOps()),
+                patch.object(sdpa_mod, "truncate_padded_output", return_value=zeros),
+            ):
+                sdpa_forward(inner, x, _make_ctx(_SEQ_LEN), cache, layer_idx=0)
+
+        _run()
+        assert captured.get("gqa_disabled") is False
+
+        monkeypatch.setenv("VLLM_METAL_DISABLE_GQA_DECODE", "1")
+        _run()
+        assert captured.get("gqa_disabled") is True
+
     def test_mixed_batch_routes_slots_and_page_tables_by_layer_group(self) -> None:
         """Full and sliding layers consume their scheduler-group metadata."""
         layout = MHAKVCacheLayout(
--- a/tests/test_attention_sdpa.py
+++ b/tests/test_attention_sdpa.py
@@ -619,8 +619,9 @@ class _PagedRoutingOpsSpy:
         _out: mx.array,
         window_seqlen_q: int = 1,
         sinks: mx.array | None = None,
+        **_kwargs,
     ) -> None:
-        del window_seqlen_q, sinks
+        del window_seqlen_q, sinks, _kwargs
         self.calls[-1].block_tables = block_tables.tolist()
         self.calls[-1].block_size = block_size
 
--- a/tests/test_gqa_paged_decode.py
+++ b/tests/test_gqa_paged_decode.py
@@ -1,15 +1,20 @@
 # SPDX-License-Identifier: Apache-2.0
 """GQA-shared flash-decode pass inside ``paged_attention_primitive``.
 
-Eligible pure-decode batches (no TQ/sinks/softcap/window, head size in
-{64,96,128,256}, GQA group <= 8, ``window_seqlen_q <= 1``, aggregate KV
-``num_seqs * max_seq_len >= GQA_DECODE_MIN_SEQ_LEN``) dispatch to
-``paged_attention_gqa_decode`` and merge through ``paged_attention_v2_reduce``.
-Spec-decode verify windows pass ``window_seqlen_q = K+1`` and stay on the
-established family so ``tests/test_spec_window_parity.py`` stays bitwise.
+Eligible single-request, long-context decode (no TQ/sinks/softcap/window,
+head size in {64,96,128,256}, GQA group <= 8, ``window_seqlen_q <= 1``,
+``num_seqs == 1`` with the scheduler confirming one real decode request,
+KV ``max_seq_len >= GQA_DECODE_MIN_SEQ_LEN``) dispatches to
+``paged_attention_gqa_decode`` and merges through ``paged_attention_v2_reduce``.
+Multi-request batches, spec-decode verify windows, and everything under the
+crossover stay on the established per-token / split-KV family.
+``VLLM_METAL_DISABLE_GQA_DECODE`` (mirror kwarg ``gqa_disabled``) forces
+eligible batches back to the established kernels.
 
-These tests drive the shipped primitive (not a reimplementation) against
-``ref_paged_attn``.
+The routing tests assert the dispatch family via ``ops.last_paged_dispatch()``
+(the gate lives below the Python boundary, so the family name is the only
+faithful record of which kernel ran); parity is checked against
+``ref_paged_attn`` on every path.
 """
 
 from __future__ import annotations
@@ -52,6 +57,11 @@ def _assert_close(out: mx.array, ref: mx.array, dtype: mx.Dtype) -> None:
     )
 
 
+def _dispatch_family() -> str:
+    """Kernel family chosen by the most recent primitive eval."""
+    return get_ops().last_paged_dispatch()
+
+
 def _run_primitive(
     kv_lens: list[int],
     dtype: mx.Dtype,
@@ -61,11 +71,15 @@ def _run_primitive(
     window_seqlen_q: int = 1,
     query_lens: list[int] | None = None,
     num_decode_requests: int = -1,
+    gqa_disabled: bool = False,
+    num_query_heads: int = NUM_QUERY_HEADS,
+    num_kv_heads: int = NUM_KV_HEADS,
+    head_size: int = HEAD_SIZE,
 ) -> tuple[mx.array, mx.array]:
     mx.random.seed(seed)
     num_seqs = len(kv_lens)
     max_kv_len = max(kv_lens)
-    scale = HEAD_SIZE**-0.5
+    scale = head_size**-0.5
     if query_lens is None:
         query_lens = [1] * num_seqs
     total_q = sum(query_lens)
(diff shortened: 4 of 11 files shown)


LxYuan0420 (commented):
This should not merge with automatic routing for many cases yet. Please limit it to the single-request, long-context case you measured and tested. Include the disable flag and tests showing when the new GQA kernel is and is not used. Support for more shapes and batching can come later with more test and benchmark results.

suntp:
@LxYuan0420 thanks — agreed, the first cut should only auto-route the case with serve-topology numbers.

Pushed `b66217c` on this PR:

- **Gate is now `num_seqs == 1`** (and `num_decode_requests` is `1` or omitted). Two 16k decode rows that would each take GQA alone stay on split-KV (`test_two_individually_eligible_requests_stay_off_gqa`). The 4×64k 4.9× figure stays in the table as a kernel microbench, marked **not** on the default path.
- **`VLLM_METAL_DISABLE_GQA_DECODE`** is in this PR (env → `sdpa_forward` → `gqa_disabled` on the primitive), with tests that the eligible 32k shape falls back to `per_token_ps512` and matches the reference either way.
- **`last_paged_dispatch()`** so tests assert the family the C++ gate actually chose: `gqa_decode` / `per_token_ps512` / `window_*`. Includes the measured 24q/4kv/hs256 shape.

`test_gqa_paged_decode.py` is 19 cases; spec-window parity still 103/103 bitwise.

Multi-request GQA can come later with serve-topology numbers (and occupancy on `sum(seq_lens)` rather than `N * max_seq_len`).

suntp:
Local serve-protocol check of this head (`b66217c`) on Qwen3.8-27B-4bit (M5 Pro, `vllm serve`, APC on, compiled MLP + native sampling, temperature=0, ~20k prefix, `usage.completion_tokens` tok/s):

| | GQA on (default) | `VLLM_METAL_DISABLE_GQA_DECODE=1` | ratio |
|---|---:|---:|---:|
| single-stream decode (3×120 tok) | 14.44 tok/s | 12.63 tok/s | **1.14×** |
| two in-flight decodes (64 tok each) | 7.57 tok/s | 7.55 tok/s | **1.00×** |

No engine traceback/NaN. Dual matching disable is the review gate: a 2-seq decode batch does not take GQA. Single-stream 1.14× at 20k is smaller than the 100k 2.22× in the PR table because 16 GQA layers are a smaller fraction of hybrid e2e at 20k than at 100k (48 GDN + MLP still dominate).

suntp:
Follow-up serve check on the same head (`b66217c`), Qwen3.8-27B-4bit, **single request**, APC-warmed prefix, 3×120 generated tokens, `usage.completion_tokens` tok/s:

| prefix | GQA on | `DISABLE_GQA_DECODE=1` | ratio |
|---:|---:|---:|---:|
| 20k (earlier comment) | 14.44 | 12.63 | 1.14× |
| **50k** | **12.52** (12.52/12.51/12.52) | **7.55** (7.55×3) | **1.66×** |
| **100k** | **10.60** (10.61/10.60/10.60) | **4.80** (4.80/4.80/4.79) | **2.21×** |

100k matches the PR table (4.25 → 9.44, 2.22×) on this machine. No traceback/NaN. Prefill TTFT (cold) was ~171s at 50k and ~473s at 100k; decode TTFT after APC was ~3.3s / ~3.7s.

suntp:
## Validation on Apple M5 Max (48 GB, macOS 26.6.2)

Checked out `pr-715` (`3a98081` + `b66217c`), rebuilt native artifacts in-tree (`python -m vllm_metal.metal.build`, NAX SDK OK), and validated with the real Qwen3.8-27B model (mlx-4bit, 64 layers = 16 full-attention + 48 GDN). Env: Python 3.12.9, vllm 0.27.1, mlx 0.32.0.

### Unit tests — all pass

- `tests/test_gqa_paged_decode.py`: **19/19** ✅ (incl. `test_qwen38_gqa_shape_takes_kernel`, multi-request stays on split-KV, spec-window stays on window path, disable-flag parity)
- `tests/test_native_sdpa_decode.py` + `tests/test_attention_sdpa.py`: **34/34** ✅

### Kernel microbenchmark (M5 Max, this PR's shape: 24q / 4kv / hs256 / bf16 / block16, interleaved hybrid-GDN block table)

Single-request decode, 5 warmup + 50 timed iters, effective KV-scan bandwidth = `kv_len × 4 × 256 × 2 (K+V) × 2 B / t`. Dispatch family verified via `ops.last_paged_dispatch()` on every run; numerics match `ref_paged_attn` (max_abs_err ≤ 2.4e-4, identical across both paths).

| KV len | GQA decode (PR) | per_token_ps512 (established) | Speedup |
|---:|---:|---:|---:|
| 16,384 | 243.9 µs/tok · **275.1 GB/s** | 321.2 µs/tok · 208.9 GB/s | **1.32x** |
| 32,768 | 337.7 µs/tok · **397.5 GB/s** | 941.6 µs/tok · 142.5 GB/s | **2.79x** |
| 65,536 | 709.6 µs/tok · **378.3 GB/s** | 2,601.4 µs/tok · 103.2 GB/s | **3.67x** |
| 131,072 | 1,406.9 µs/tok · **381.6 GB/s** | 5,949.4 µs/tok · 90.2 GB/s | **4.23x** |

Same trend as the M5 Pro numbers in the PR (~40→~190 GB/s), shifted up with M5 Max's bandwidth: GQA path sustains ~380–400 GB/s while the established path degrades to ~90 GB/s at 128k. The 16k crossover is also a win here (1.32x) — no need to lower the threshold on this generation.

### End-to-end (qwen3.8-27b-mlx-4bit, greedy 64-token decode, prefix caching on, in-process engine)

A/B via `VLLM_METAL_DISABLE_GQA_DECODE`; `last_paged_dispatch()` confirmed `gqa_decode` vs `per_token_ps512` on every run; **greedy outputs byte-identical in all configs**.

| Context | GQA on | GQA off | Speedup |
|---:|---:|---:|---:|
| 32,768 | 15.21 tok/s | 14.02 tok/s | **1.09x** |
| 65,536 | 12.98 tok/s | 9.73 tok/s | **1.33x** |
| 88,064 | 14.52 tok/s | 8.73 tok/s | **1.66x** |
| 131,072 | 13.71 tok/s | 6.96 tok/s | **1.97x** |

E2E gain is bounded by weight reads (4-bit 27B ≈ 15 GB/token-step at ~400 GB/s) and converges toward the kernel-level ratio as context grows — exactly the expected behavior.

### Memory notes for 128k on 48 GB (KV = 16 full-attn layers × ~66 KB/token, bf16)

| `VLLM_METAL_MEMORY_FRACTION` | KV budget | Max context |
|---|---:|---:|
| 0.85 | 6.03 GiB | ~93.6k |
| 0.92 | 7.32 GiB | ~114k |
| 0.96 | 8.04 GiB | ~126k |
| **0.98** | **8.78 GiB (175 blocks)** | **132k — 131,072 works** ✅ |

256k (~17 GiB KV) is out of reach on 48 GB even at fraction 1.0 (15 GB weights + 17 GiB KV + runtime > total); 64 GB machines should be fine, consistent with prior experience.

### Conclusion

Correct and behaving as designed on M5 Max: 53/53 tests pass, routing gate verified in real serving (single-request ≥16k → `gqa_decode`; disable flag / fallback all correct), outputs identical, kernel-level 2.8–4.2x at ≥32k, E2E up to ~2x at 128k. LGTM from this machine.

suntp:
## Follow-up: benchmarking-pitfall audit of my own numbers above

Prompted by the BigRig pitfall list posted in this thread, I re-checked each trap against how the M5 Max data above was actually measured, and re-ran the e2e A/B interleaved. Honest scorecard:

**Where I was clean**
- *Peak-memory high-water inheritance across configs (trap 1):* N/A — no peak-memory figures published; A/B metric was throughput.
- *RSS blindness (trap 3, first half):* never used RSS. Memory math came from vLLM's KV-budget logs plus config-derived per-token arithmetic (16 full-attn layers × ~66 KB/token, which the logs confirm: 8.78 GiB → 132,082 tokens).
- *Baseline stated (trap 4, second half):* the baseline was the same process with `VLLM_METAL_DISABLE_GQA_DECODE=1` → `per_token_ps512`, dispatch verified per run.

**Where I fell in (caught before posting)**
- *MLX lazy-eval variant of per-op timing (pitfall 3):* my first kernel-bench version discarded the lazy outputs, dead-code elimination skipped the kernels entirely, and it "measured" 4,745 GB/s. Caught because that number is impossible on this hardware; fixed by enqueueing 50 calls and `mx.eval`-ing them together, which also avoids per-op GPU start/stop. Residual caveat: repeated same-input calls may leave KV slightly L2-warm at 16k (67 MB), so the small-context bandwidth is a bit flattering; at 128k (268 MB) the effect is negligible, and the e2e trend corroborates the kernel numbers.

**Where I was exposed (closed today)**
- *Sequential single-sample A/B (trap 4):* the original e2e was one 64-token sample per config, sequential. Noise was already visible in my own table (88k "on" at 14.52 tok/s > 64k at 12.98). Today's interleaved re-test at 65,536 ctx, same process, alternating on/off ×3 rounds, best-of-3:
  - gqa_on: 14.29 / 14.39 / 14.41 tok/s → best 14.41
  - gqa_off: 10.89 / 10.92 / 10.90 tok/s → best 10.92
  - **speedup 1.32x** (original sequential: 1.33x) — ratio stable.
  - Absolute throughput moved ~10% across days (12.98/9.73 → 14.41/10.92) as available memory went 34.1 → 29.3 GB with the machine in use — a live confirmation that anything else on the Mac inflates whichever round it lands in, and that ratios survive while point estimates drift.
- *stdin detach on background runs:* my earlier nohup runs didn't `< /dev/null`; nothing hung by luck. Fixed in the current scripts.

**Still open**
- *Compressor/swap under memory pressure (trap 3, second half):* during today's interleaved decode at fraction 0.90, vm_stat showed zero compressor growth and zero swapouts. The 128k run used fraction 0.98 and I did not sample vm_stat then, so swap contamination there can't be ruled out retroactively — bounded by the ~9 s decode window, and the on/off ratio was measured within one process, but stated for completeness.

Bottom line: the posted ratios (kernel 2.8–4.2x at ≥32k, e2e 1.09–1.97x) hold up under interleaving; treat absolute tok/s as ±10% machine-state noise, and the 128k point as the least hardened of the set.

suntp:
Additional local validation on 2026-09-09 of `b66217c` with **OpenBMB/MiniCPM5-2B**, original BF16 weights downloaded from ModelScope, on **M5 Pro / 64 GiB / macOS 26.6.2**. vLLM 0.28.0+cpu, MLX 0.32.0, Transformers 5.12.1.

This is a useful group-8 shape: **42 full-attention layers, 16 query heads / 2 KV heads / head dim 128**.

**Local tests:** added 22 MiniCPM5-shape cases (BF16/FP16, 16k boundary, partial/fragmented blocks, 32k/64k/120k, disable flag at 16k/32k/120k, two eligible requests staying off GQA, and speculative-window fallback). Together with existing local GQA/SDPA/spec-window tests: **190 passed**. Every shape case checks actual dispatch plus independent reference numerics. The 22 new cases also passed against the unmodified PR-head helper. These additional test files are currently local and are not part of this PR diff.

**Serve A/B:** separate `vllm serve` processes, single request, APC-warmed prefix, compiled MLP + native sampling, temperature 0, 3 × 120 tokens. Decode tok/s uses `(usage.completion_tokens − 1) / (last_token_time − first_token_time)`. Prompts are repeated technical prose followed by a counting task, with `ignore_eos=True` to keep each run at 120 tokens. Same exact token inputs and original BF16 weights in both arms; only `VLLM_METAL_DISABLE_GQA_DECODE` changes.

| Prompt tokens | GQA on tok/s | GQA off tok/s | Speedup |
|---:|---:|---:|---:|
| 4,096 | 37.21 | 37.47 | 0.99× |
| 20,000 | 29.34 | 24.79 | 1.18× |
| 50,000 | 21.28 | 12.44 | 1.71× |
| 100,000 | 15.07 | 6.56 | 2.30× |

**Additional 16k / 120k serve check** (same settings and model, separate run later the same day; compare the A/B pair within each row):

| Prompt tokens | GQA on tok/s | GQA off tok/s | Speedup |
|---:|---:|---:|---:|
| 16,384 | 38.28 | 36.94 | 1.04× |
| 120,000 | 17.33 | 6.17 | 2.81× |

16k here is the exact 16,384-token activation boundary. Both additional on rows confirm `gqa_decode`; both off rows confirm `per_token_ps512`. Each arm is 3 × 120 generated tokens; all texts match at each length, and neither server log contains traceback/NaN/errors. 120k repetitions: on 17.35 / 17.32 / 17.33, off 6.19 / 6.17 / 6.16 tok/s. This adds a modest 1.04× win at the gate and a 2.81× win at 120k.

A worker RPC read of `last_paged_dispatch()` after every measured request confirms `gqa_decode` for long-context on, `per_token_ps512` for off and for both 4k arms. **All generated texts are byte-identical across all six runs at each length**, and both arms pass a Chinese chat smoke test. No engine traceback/NaN/errors. The 100k on repetitions are 15.05 / 15.07 / 15.09 tok/s; off is 6.43 / 6.82 / 6.44 tok/s. The table uses means; 4k is effectively unchanged.

The separate synchronized primitive microbench at this shape gives **1.12× / 1.76× / 2.96× / 3.72×** at 16k / 32k / 64k / 128k, with maximum absolute reference error ≤ 2.45e-4 for eligible rows. These are wall-clock primitive timings, including dispatch/synchronization. Real-model serving was tested through 120k, and real-model concurrency was not tested in this run.

This supports the existing single-request long-context gate for MiniCPM5-2B on this machine.

suntp:
## Update: universal two-clause dispatch gate + startup autotune (M5 Max validated)

Two new commits on this branch replace the shape-specific dispatch conditions with a physically derived, cross-machine gate, and add a startup autotune that re-fits the gate's one hardware constant on the local GPU.

### What changed

**1. Universal two-clause gate** (`gqa_decode_shape_eligible`)

The previous gate accumulated per-shape conditions (hs64/96 never, hs128 needs ≥32k, hs256 needs group≥6, group<4 blocked) — every untested model was a guess, and the list would keep growing. It is replaced by two clauses over quantities available at dispatch time:

```
use_gqa = (partitions × num_kv_heads ≥ 3 × gpu_cores)        # parallelism floor
       AND ((gqa_group − 1) × kv_bytes ≥ waste_min_bytes)     # bandwidth-waste threshold
where kv_bytes = kv_len × num_kv_heads × head_size × 4B (K+V, bf16)
```

- Clause 1: `gqa_decode`'s grid is `partitions × kv_heads`, i.e. `group`× smaller than per-token's; below ~3× core count the GPU is under-filled. GPU cores are detected via IORegistry, no per-device table.
- Clause 2: `(group − 1) × kv_bytes` is exactly the extra DRAM traffic the per-token kernel re-reads per decode step; sharing only pays when that saving is large enough. `group == 1` (MHA) yields zero waste and is excluded by the formula with no special case.

Back-checked against 25 measured (shape × length) points from five models on M5 Max: **22 optimal, 3 conservative misses (≤1.27x upside lost, all in the noise band), 0 regressions**. It correctly separates same-shape collisions, e.g. Qwen3-0.6B@32k vs MiniCPM5@32k (134 MB vs 235 MB waste; measured 0.50x vs 1.34x).

**2. Startup autotune** (`gqa_autotune.py`, hooked into `warm_up_kernels`)

The waste threshold is hardware-dependent (bandwidth vs. core-count trade-off), so instead of hard-coding the M5 Max fit, the gate constants are runtime-tunable (`set_gqa_gate_params`) and re-fitted on the local GPU at startup:

- Times both decode kernels on three reference shapes (group 4/8 at hs128, group 4 at hs256) × five context lengths (16k–65k), forcing routing via permissive/restrictive gate params.
- Timing discipline (per the macOS benchmarking pitfalls discussed above): 50 batched launches under a single `mx.eval` per round, best of 3 rounds, process-wide warm-up of **both** kernels before measuring — an earlier linear-regression version of this fit 328 MB because the first probe ate a cold-start outlier; the warm-up + decision rule eliminated that.
- Per shape, crossover = interpolation at speedup 1.0 into the clear-win (≥1.15x) suffix; the most conservative crossover across shapes wins; result clamped to [100 MB, 400 MB] and cached per `(chip, gpu_cores)` in `~/.cache/vllm-metal/`. `VLLM_METAL_GQA_AUTOTUNE=0` keeps static defaults, `=force` recalibrates; any failure falls back to static defaults silently.
- At dispatch time the gate remains a pure arithmetic predicate — nanoseconds per request, deterministic, immune to runtime load. Calibration happens once at startup (~1.7 s first run, then cache-hit).

### M5 Max measurements (48 GB, 40 GPU cores)

Autotune fit on M5 Max:

```
(16q/4kv/hs128): 101MB:0.76  151MB:1.21  201MB:1.53  302MB:1.92  403MB:2.42 → crossover 128MB
(16q/2kv/hs128): 117MB:0.83  176MB:1.09  235MB:1.46  352MB:2.08  470MB:2.15 → crossover 176MB
(16q/4kv/hs256): 201MB:0.86  302MB:1.32  403MB:2.12  604MB:2.98  805MB:3.35 → crossover 232MB
fitted threshold = max(128, 176, 232) = 231 MB
```

**231 MB vs. the 220 MB hand-fit it replaces — within 5%**, two independent methods converging on the same boundary. On other chips (M5 Pro, M4 Max, …) the fit moves with the hardware; no per-device conditions needed.

End-to-end A/B with the autotuned gate (interleaved ×3, best-of-3, greedy 64 tokens, dispatch verified every round, outputs byte-identical):

| model @ ctx | waste | gate | speedup |
|---|---:|---|---:|
| Qwen3-0.6B @ 32k | 134 MB | blocked → per_token | 1.00x |
| Qwen3-0.6B @ 40.6k | 166 MB | blocked (conservative miss¹) | 1.00x |
| Qwen2.5-0.5B @ 28.7k | 88 MB | blocked → per_token | 1.00x |
| Qwen2.5-0.5B @ 64k² | 201 MB | blocked → per_token | 1.00x |
| Qwen3-4B @ 24.5k | 302 MB | **gqa_decode** | **1.13x** |
| Qwen3-4B @ 32k | 403 MB | **gqa_decode** | **1.25x** |
| MiniCPM5-2B @ 16k | 117 MB | blocked → per_token | 1.00x |
| MiniCPM5-2B @ 64k | 470 MB | **gqa_decode** | **1.53x** |
| Qwen3.8-27B-4bit @ 32k | 671 MB | **gqa_decode** | **1.10x** e2e |

¹ 0.6B@40k measured 1.22x at kernel level but sits below threshold — conservative direction, no mis-route.
² 0.5B caps at 32768 positions; 64k used `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1` for routing A/B only.

Kernel-level (27B full-attn shape, 24q/4kv/hs256): 2.79x @ 32k, 3.67x @ 64k, 4.23x @ 128k. E2E on the 27B hybrid model reaches **1.97x @ 128k** (attention share limited by ~15 GB weight reads per token).

### Tests

- New `tests/test_gqa_autotune.py` (6): disabled keeps static defaults; force-calibration fits and writes cache; cache hit skips calibration; calibration failure falls back; forced gate params flip real routing (`last_paged_dispatch` verified per_token_ps512 ↔ gqa_decode); crossover suffix rule unit tests.
- Truth tables in `test_gqa_paged_decode.py` / `test_native_sdpa_decode.py` updated to the universal formula, rows re-judged against measurements.
- **111/111 pass** (74 GQA decode + 31 SDPA + 6 autotune).

suntp:
Follow-up to [the startup-autotune proposal above](https://github.com/vllm-project/vllm-metal/pull/715#issuecomment-5645190995): **this approach did not achieve the intended general routing guarantee and has been replaced in the PR.** Further validation did not support extending a startup-fitted threshold to unmeasured shapes or changing GPU resource conditions. The earlier claims of exact extra DRAM traffic and immunity to runtime load were overstated.

The current revision, [efa0e31](https://github.com/vllm-project/vllm-metal/pull/715/commits/efa0e31cda60fb2bb018d5a2588af3617e29ac63), follows the requested narrower scope: single-request, long-context decode for three measured attention geometries, with explicit length bounds and fallback conditions. It retains `VLLM_METAL_DISABLE_GQA_DECODE=1` and tests that verify the actual native dispatch for enabled, disabled, and fallback cases. Startup calibration and mutable gate parameters have been removed; the later permit-lifecycle experiment is also outside this revision.

This supersedes the linked proposal. Please refer to the updated [PR description](https://github.com/vllm-project/vllm-metal/pull/715#issue-5365789316) for the current implementation, validation results, and limitations.

PR #716, opened 2026-09-06
[Bugfix] Keep draft KV in sync when dynamic K is zero

POWEHI2023:
## Summary

- Keep the draft model KV cache synchronized when dynamic speculative
  decoding selects zero speculative tokens for a batch.
- Perform cache ingestion before returning from `propose()` when `K == 0`.
- Preserve the existing behavior of skipping draft token generation for
  zero-token batches.

## Problem

Dynamic speculative decoding may assign different speculative-token counts
based on batch size. For example:

- batch size 1: K = 3
- batch size 2: K = 0

Previously, `DraftModelProposer.propose()` returned immediately when K was
zero. This skipped not only draft generation, but also the cache-ingestion
step for committed target tokens.

When the batch later returned to K > 0, the draft model resumed from a stale
KV-cache state. This reduced speculative-token acceptance even though the
target-model output remained correct.

## Fix

Move the K=0 return until after:

- scheduler plan collection;
- committed-token ingestion into the draft model;
- draft sequence-length updates.

Draft generation is still skipped when K is zero, so the change only keeps
the draft cache synchronized.

## Reproduction

Configuration:

- target model: Qwen/Qwen3-0.6B
- draft model: Qwen/Qwen3-0.6B
- two 384-token requests
- max batched tokens: 128
- batch size 1: K = 3
- batch size 2: K = 0

Before the change:

- K=0 proposer calls: 1
- cache-ingestion plans during the K=0 call: 0
- cold acceptance: 28/57 (49.12%)
- subsequent acceptance: 36/39, due to history masking the stale-cache issue

After the change:

- K=0 proposer calls: 1
- cache-ingestion plans during the K=0 call: 2
- acceptance: 36/39 across all three runs
- generated token IDs are unchanged

## Validation

- [x] `scripts/lint.sh`
- [x] Relevant proposer/cache tests: 167 passed
- [x] Full non-slow test suite: 1986 passed, 15 skipped
- [x] Dynamic-K reproduction: K=0 produced two cache-ingestion plans
- [x] Generated token IDs remained unchanged
- [x] Full `scripts/test.sh` smoke run: 1986 passed, 15 skipped

### DIFF
--- a/docs/speculative_decoding.md
+++ b/docs/speculative_decoding.md
@@ -75,6 +75,26 @@ VLLM_METAL_MEMORY_FRACTION=0.55 \
     --speculative-config '{"method":"draft_model","model":"Qwen/Qwen3-0.6B","num_speculative_tokens":3}'
 ```
 
+### Dynamic speculative decoding
+
+See the upstream [dynamic speculative decoding guide](https://docs.vllm.ai/en/latest/features/speculative_decoding/dynamic_speculative_decoding/)
+for configuration details. This example sets `K=3` for one scheduled request
+and `K=0` for two:
+
+```bash
+VLLM_METAL_MEMORY_FRACTION=0.55 \
+  vllm serve Qwen/Qwen3-8B \
+    --max-model-len 2048 \
+    --max-num-seqs 2 \
+    --no-async-scheduling \
+    --speculative-config '{
+      "method": "draft_model",
+      "model": "Qwen/Qwen3-0.6B",
+      "num_speculative_tokens": 3,
+      "num_speculative_tokens_per_batch_size": [[1, 1, 3], [2, 2, 0]]
+    }'
+```
+
 ## N-gram
 
 Follow the upstream [N-gram guide](https://docs.vllm.ai/en/latest/features/speculative_decoding/n_gram/)
--- a/vllm_metal/v1/draft_model_proposer.py
+++ b/vllm_metal/v1/draft_model_proposer.py
@@ -280,8 +280,6 @@ class DraftModelProposer:
 
     def propose(self, ctx: ProposeContext) -> DraftTokenIds | None:
         num_speculative_tokens = ctx.num_speculative_tokens
-        if num_speculative_tokens <= 0:
-            return None
         if self._committed_group_index is None:
             raise RuntimeError(
                 "DraftModelProposer.propose() called before "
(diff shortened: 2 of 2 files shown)


WindChimeRan (approved):
Thanks for the good catch! I'll add a short doc paragraph.

ricky-chaoju (commented):
The K=0 sync itself works, 48 of 48 zero-K calls ingest on this branch versus 0 on main, and the output tokens are identical.

I couldn't reproduce the acceptance part though. On main the first K=3 call after a K=0 run does one 172-token catch-up ingest before it drafts, so by draft time the cache is already repaired. The proposed draft tokens come out byte-identical on both sides (same sha over 23 proposals), not just the same acceptance rate. What does change is the peak ingest, 172 down to 127, at 11% more wall time here.

Your pr has one K=0 call and mine has twelve, so the shapes differ. Can you share the script? If the 28/57 case holds this should go in as is; if not, it might be worth restating the win as a lower peak ingest.
CC @WindChimeRan

WindChimeRan:
@ricky-chaoju 

I think there's a pre-existing draft-model speculative decoding bug: https://github.com/vllm-project/vllm-metal/pull/721

POWEHI2023:
@ricky-chaoju Thanks — I reproduced both behaviors. The difference is whether a valid draft-KV boundary exists before K becomes zero.

In my original case, the request is first seen during a K=0 prefill. Main neither ingests its first 123 tokens nor records a draft boundary. Later prefill starts from the next scheduler offset, so the missing prefix is forgotten and cannot be caught up. This reduces cold-run acceptance from 36/39 to 28/57.

The fixed branch eagerly ingests during every K=0 step. This lowers peak ingest from 172 to 127 tokens, but was about 10% slower because it performs draft-model work even for requests that may finish before drafting resumes.

A more efficient solution would record the earliest missing draft-KV position during K=0 without running the draft model, then lazily ingest the complete missing range before the next K>0 proposal.
[repro_dynamic_k_draft_kv.py](https://github.com/user-attachments/files/31903802/repro_dynamic_k_draft_kv.py)

ricky-chaoju:
You're right. Let's land #721 first, then this. I stacked them locally and it's clean. Worth noting the ~10% in the description though, and your lazy-ingest idea sounds like the right follow-up.

PR #718, opened 2026-09-06
Reuse vLLM hybrid state dtypes

WindChimeRan:
Follow-up to #598. Reuse vLLM's resolved hybrid state dtypes for cache sizing and allocation (inspired by #699), replacing local model-specific rules. GDN fallback rejects dtype combinations it cannot update in place.

### DIFF
--- a/vllm_metal/attention/impls/linear.py
+++ b/vllm_metal/attention/impls/linear.py
@@ -417,6 +417,13 @@ class GDNPagedAttentionWrapper(nn.Module):
         # Promote only as needed to preserve input, state, and output precision.
         recurrent_pool = self._gdn_state_cache.recurrent_states[self._gdn_cache_idx]
         kernel_dtype = mx.result_type(q, k, v, g, beta, recurrent_pool, state.x)
+        if recurrent_pool.dtype != kernel_dtype:
+            raise RuntimeError(
+                f"GDN fallback does not support {recurrent_pool.dtype} recurrent "
+                f"state with {kernel_dtype} kernel buffers. Set "
+                f"--mamba-ssm-cache-dtype {str(kernel_dtype).removeprefix('mlx.core.')} "
+                "to enable in-place state updates."
+            )
         q_flat = mx.contiguous(q.reshape(total_tokens, n_hk, d_k).astype(kernel_dtype))
         k_flat = mx.contiguous(k.reshape(total_tokens, n_hk, d_k).astype(kernel_dtype))
         v_flat = mx.contiguous(v.reshape(total_tokens, n_hv, d_v).astype(kernel_dtype))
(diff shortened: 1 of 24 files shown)


vllm_metal/attention/runtime/families/shortconv.py
```diff
@@ -38,7 +39,7 @@ def _create_shortconv_state_cache(
         conv_kernel_dim=geometry.conv_kernel_dim,
         conv_dim=geometry.conv_dim,
         initial_seqs=initial_seqs,
-        dtype=dtype,
+        dtype=dtypes[0],
```
LxYuan0420, line 42: I noticed BF16 input + FP32 ShortConv state returns FP32 output here. That leaks cache dtype into activation dtype. Please keep the ShortConv output in `x.dtype` and add a dtype assertion in the mixed-dtype test.

vllm_metal/attention/impls/linear.py
```diff
@@ -418,9 +420,17 @@ def _run_recurrent_fallback(
...
 
         # Flatten for kernel: remove batch dim.
-        # Use float32 for kernel dispatch to avoid float16 overflow in
-        # recurrent state accumulation.  Output is cast back after.
-        kernel_dtype = mx.float32
+        # The native kernel uses one buffer dtype and accumulates in float32.
+        # Promote only as needed to preserve input, state, and output precision.
+        recurrent_pool = self._gdn_state_cache.recurrent_states[self._gdn_cache_idx]
+        kernel_dtype = mx.result_type(q, k, v, g, beta, recurrent_pool, state.x)
+        if recurrent_pool.dtype != kernel_dtype:
```
LxYuan0420, line 420: This makes GDN mixed-dtype support depend on request shape. One-token prefill skips the lazy path and reaches this fallback, so the same config can work for normal prefill/decode but fail later. Please either restore typed fallback or reject this dtype combo at init.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -494,8 +495,14 @@ def _reject_pipeline_parallel_with_per_layer_metadata(self) -> None:
     def _install_hybrid_state_plan(self, args: dict[str, Any]) -> None:
         """Resolve the state family that owns this model's hybrid layers."""
         if self._runner.is_hybrid:
+            model_config = self._runner.model_config
+            model_cls, _ = ModelRegistry.resolve_model_cls(
```
LxYuan0420: Please use `model_config.registry` here instead of global `ModelRegistry`. That is vLLM’s path for applying `model_class_overrides`, and bypassing it can resolve the wrong model class/state dtype in workers.

LxYuan0420 (approved):
few minor gaps to fix; approving to unblock the progress

vllm_metal/attention/impls/linear.py
WindChimeRan: fix: reject this dtype combo at `vllm_metal/v1/model_lifecycle.py`

PR #719, opened 2026-09-07
[Bugfix] preserve model revisions across Metal loading paths

skundu42:
Forward requested revisions to text, multimodal, AWQ, draft-model, and ModelScope loaders. Keep AWQ config detection on the same revision as weight loading.

### DIFF
--- a/vllm_metal/v1/model_lifecycle.py
+++ b/vllm_metal/v1/model_lifecycle.py
@@ -112,7 +112,9 @@ class GenerationLoadRequest:
             model_name=(
                 gguf_source.weights_path
                 if gguf_source is not None
-                else get_model_download_path(model_config.model)
+                else get_model_download_path(
+                    model_config.model, revision=model_config.revision
+                )
             ),
             model_config=model_config,
             hf_config=hf_config,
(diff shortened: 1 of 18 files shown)


vllm_metal/v1/model_lifecycle.py
```diff
@@ -112,7 +112,9 @@ def from_runner(
             model_name=(
                 gguf_source.weights_path
                 if gguf_source is not None
-                else get_model_download_path(model_config.model)
+                else get_model_download_path(
+                    model_config.model, revision=model_config.revision
```
LxYuan0420, line 116: Please thread `model_config.revision` through the STT path too. Detection and loading still resolve the default revision, so `--revision` can select the wrong STT runner or load different weights than requested.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -276,7 +278,12 @@ def _load_generation_model(
                 self._runner.vllm_config.load_config,
```
LxYuan0420, line 278: Add `revision=None` to the slow AWQ test doubles. Real vLLM `ModelConfig` declares `revision`; using `getattr(..., None)` in production would hide an inaccurate test fixture instead of fixing the contract.

skundu42:
Thanks for the comments. I have addressed them in this new commit.

LxYuan0420 (approved):

PR #635, opened 2026-08-21
Support remote GGUF references

LxYuan0420:
This PR is:

- To add remote GGUF serving for `repo_id:quant` references.
- To extend the existing local-only GGUF path without changing local GGUF behavior.
- To support local or Hugging Face tokenizer/config sources for GGUF weights.
- To fail clearly for unsupported qtypes, sharded GGUF, ambiguous matches, and multimodal GGUF.

Before this PR, vllm-metal only supported local `.gguf` weight files. After this PR, users can serve either a local GGUF file or a remote GGUF reference:

```bash
vllm serve Qwen/Qwen3-0.6B-GGUF:Q8_0 \
  --tokenizer Qwen/Qwen3-0.6B
```

```bash
vllm serve /local/model.gguf \
  --tokenizer Qwen/Qwen3-0.6B
```

Remote GGUF support is intentionally limited to single-file `Q8_0`, `Q4_0`, and `Q4_1` checkpoints for now.

PR #720, opened 2026-09-07
[GGUF] Fail fast on local sharded GGUF files

ricky-chaoju:
Passing one shard of a split GGUF to vllm-metal today loads the whole file and then dies in the completeness check with a confusing "Incomplete GGUF load" error. The remote path already rejects sharded matches before download (#635); the local path had no guard.

This adds one: the loader now rejects shard-named paths like `model-00001-of-00002.gguf` up front, before opening the file. A `-00001-of-00001.gguf` file is one complete file and still loads. Remote selection behavior is unchanged.

### DIFF
--- a/tests/test_gguf_loader.py
+++ b/tests/test_gguf_loader.py
@@ -181,6 +181,12 @@ def _gguf_module_histogram(model: nn.Module) -> dict[str, int]:
     return counts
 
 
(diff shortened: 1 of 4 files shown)


LxYuan0420 (approved):
LGTM

PR #721, opened 2026-09-07
[Bugfix] [Spec Decoding] Fix the first draft window after prefill

WindChimeRan:
comparison

Fixes a regression from #630 that leaves the first draft window one token behind. Target prefill has already committed its first output token, but the drafter ingested only the prompt and predicted that token again. This fix ingests the sampled token before drafting, aligning the first proposal with the verifier's next position.

With **Qwen3-0.6B as both drafter and target**, greedy acceptance should be near 100%, apart from numerical differences. Measured acceptance rises from **65.3% to 98.1%** (40 prompts × 3 runs, batch 1, K=3, 10 output tokens).

### Why the first draft was rejected

For `The capital of France is`, target prefill emits ` Paris`. Verification now checks the token **after Paris**, which is `.` in this trace:

| | Drafter input | First proposal | Verifier expects |
|---|---|---|---|
| Before | Prompt only | ` Paris` | `.` |
| After | Prompt + ` Paris` | `.` | `.` |

Minimal same-model regression repro, from a configured checkout. It fails before the fix (`Paris` versus `.`, logit gap 8.125) and passes on this branch:

```bash
PYTHONPATH=. VLLM_METAL_MEMORY_FRACTION=0.108 python -m pytest \
  tools/test_spec_decode_draft_model_e2e.py -q -m slow -k capital
```

Validation: all 64 focused proposer/cache tests pass after updating the sampled-token fixtures and ingest expectations.

### NAX precision issue (out of scope)

below are the notes from GPT6-astral. NAX has some numeric bugs. The fix is non-trivial. (my fix result in performance degradation). So future work. 

> On M5, the expanded same-model corpus retains one later full-prefill rejection: the translation prompt's third verification window has a logit gap of **0.1875**, after two fully accepted windows. Replaying identical attention inputs isolates precision loss in NAX's `P × V` multiplication; draft prefill and target decode use different numerical paths. Disabling NAX removes this rejection. This precision loss also reproduces in standalone native MLX 0.32.0 attention. Attention-kernel changes are outside this PR; the **two-BF16-ULP rejection tolerance remains unchanged**.

### DIFF
--- a/tests/test_draft_model_proposer.py
+++ b/tests/test_draft_model_proposer.py
@@ -160,7 +160,7 @@ def _prefills_context(
         decode_segments=[],
         decode_token_ids=[],
         prefill_reqs=prefill_reqs,
-        prefill_token_ids=[],
+        prefill_token_ids=[42] * len(prefill_reqs),
         prefill_result_modes=["final"] * len(prefill_reqs),
         request_states={
             req_id: _request_state(committed_block_ids=[0]) for req_id, _ in prefills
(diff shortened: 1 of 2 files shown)


ricky-chaoju (approved):
LGTM. Thanks!

PR #707, opened 2026-09-06
[Cleanup] Remove the legacy non-paged serving path

WindChimeRan:
## Background

New contributors are confused about non-paged path e.g., https://github.com/vllm-project/vllm-metal/pull/674. And I think now it's the time to retire it. For parity test, we can just compare mlx_lm output e2e with the new tool & CI.

## Summary 

With independent `mlx-lm` parity in #705, remove the legacy non-paged serving path.

Serving always uses paged attention; `VLLM_METAL_USE_PAGED_ATTENTION` is removed. Retain existing golden and speculative tests, updating their legacy-path switches and references as needed.

Depends on #706. Test retirement is reviewed separately on top of this PR.

Model-runner deletions:

| Deleted code | Why it can be removed |
|---|---|
| `_prefill_single`, `_batched_decode`, `_sequential_decode`, `_run_non_paged_decode_batch` | Legacy cache execution; paged forward and sampling remain. |
| `estimate_one_sequence_kv_bytes` | Legacy memory estimate; paged cache sizing remains. |
| `_make_sampling_metadata` | Test-only wrapper; shared `SamplingBatch` metadata construction remains. |
| `generate()` | Standalone MLX helper with only test callers in the repository; unused by vLLM serving. |

PR #724, opened 2026-09-07
[Tests] Retire model golden tests and generators

WindChimeRan:
Replace model-specific frozen token goldens with live comparisons against native `mlx-lm` through `tools/check_parity.py`.

- Remove seven golden tools/generators and three model-golden test files; update dependent imports and contributor guidance.
- Retain specialized speculative-decoding, prefix-cache, sliding-window dispatch, and multimodal tests.
- Preserve the EXAONE config workaround in the native reference loader.

The shared corpus retains all raw prompts from the retired files. Comparisons use plain-text inputs. Other model checkpoints can be checked manually with `--model`; nightly CI continues to cover Qwen3-0.6B and Qwen3.5-0.8B.

Depends on #707.


tools/test_glm47_flash_golden.py
```diff
@@ -1,203 +0,0 @@
...
-    return results
-
-
-def run_paged() -> dict[str, list[int]]:
-    """Run vllm with paged + MLA kernel and return per-prompt token IDs."""
-    # Force the kernel path: the whole point of this test is to compare
-    # the Metal MLA kernel against standalone mlx_lm. A caller with
-    # VLLM_METAL_MLA_KERNEL=0 exported would silently fall back to MLX
-    # SDPA, which matches all goldens and produces a false pass.
-    os.environ["VLLM_METAL_MLA_KERNEL"] = "1"
```
LxYuan0420, line 125: I’m wondering if we should keep a thin GLM live parity wrapper here without the frozen token dict. This path needs the chat template, VLLM_METAL_MLA_KERNEL=1, and a backend-reach check, while the generic plain-text parity tool would not cover that.

tools/test_gpt_oss_smoke.py
```diff
@@ -1,91 +0,0 @@
...
-    "The largest planet in our solar system is": [200005, 35644, 200008, 976, 1825, 31064, 25, 392, 976, 10574, 17921, 306, 1039, 17624, 2420, 382, 4050, 3164, 6960, 1682, 290, 6052, 25, 79575, 13, 3164, 3572, 1682, 261, 18128, 13, 2632, 6052, 25, 79575, 13, 138743, 8633, 4275, 290, 10574, 13, 2632, 9570, 25, 79575, 13, 200007, 200006, 173781, 200005, 17196, 200008, 976, 10574, 17921, 306, 1039, 17624, 2420, 382, 6240, 41, 26451, 410, 13, 200002],
-    "Water boils at a temperature of": [200005, 35644, 200008, 976, 1825, 5003, 25, 392, 27874, 165683, 540, 261, 12088, 328, 4050, 3164, 6960, 1682, 290, 79667, 2438, 328, 3411, 13, 3072, 290, 4928, 382, 60592, 25, 392, 27874, 165683, 540, 261, 12088, 328, 4050, 3164, 3572, 1682, 290, 6052, 25, 220, 1353, 26557, 540, 220, 16, 83327, 11, 503, 220, 19584, 68854, 13, 3072, 10112, 1023, 1682, 290, 12088, 306, 181775, 25, 220, 33797, 13, 1055, 658, 13, 623, 1825, 3572, 413, 35885, 261, 52077, 6052, 13, 623, 4928, 382, 60592, 889, 6960, 1023, 1682, 290, 79667, 2438, 13, 2632, 6052, 25, 220, 1353, 26557, 350],
-}
-# fmt: on
-
-
-def _apply_chat_template(model_name, prompts):
```
LxYuan0420, line 49: On the GPT-OSS side, would deleting this leave us with no chat-formatted real-model check for sink attention? The generic parity tool is useful, but plain tokenizer.encode(prompt) does not exercise the same profile.

WindChimeRan: Chat-template correctness (GPT-OSS & GLM) is upstream vllm’s responsibility, and chat formatting is not required for a valid numerical parity test.

chat template and sink attention are not in the same abstraction layers ...

tools/test_glm47_flash_golden.py
WindChimeRan: added VLLM_METAL_MLA_KERNEL to the env var

WindChimeRan:
results here: 
https://github.com/vllm-project/vllm-metal/actions/runs/34551686878

WindChimeRan:
@LxYuan0420 In case you missed it 😉

LxYuan0420 (approved):
Nicely done! 
Thanks for working on this

PR #725, opened 2026-09-07
Support newer mlx-vlm Qwen3.5 MLP calls

LxYuan0420:
This PR is:

- To remove the temporary `mlx-vlm<0.6.16` cap added in #712
- To support both released Qwen3.5 MLP call contracts: `(x, target_verify=False)` and `(x)`
- To keep compiled MLP wrapper selection explicit for the installed `mlx-vlm` contract

PR #712 capped `mlx-vlm` below 0.6.16 because newer Qwen3.5 MLP blocks changed their call signature and broke compiled MLP wrapping. This PR handles both signatures: older blocks keep the `target_verify` wrapper, and newer unary blocks use the plain compiled wrapper.

Proof from local checks:

- Checked the `mlx-vlm 0.6.17` wheel: Qwen3.5 dense and MoE MLP blocks use unary `(x)` calls
- Served `Qwen/Qwen3.5-0.8B` with `mlx-vlm 0.6.17` and compiled MLP enabled; the model loaded, compiled MLP wrapped 24 blocks, and `/v1/completions` returned HTTP 200

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,9 +47,7 @@ dependencies = [
     "apache-tvm-ffi>=0.1.9,<0.1.13; platform_system == 'Darwin' and platform_machine == 'arm64'",
     # mlx-vlm 0.6.8 drives MLX-quantized Qwen3.5/3.6 wrappers with correct
     # mRoPE state; 0.6.4 leaves it unset and garbles their output (#685).
-    # Capped below 0.6.16, which drops `target_verify` from
-    # Qwen3_5MLP.__call__ and trips the CompiledMLPBlocks wrapper guard.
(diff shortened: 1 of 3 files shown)


LxYuan0420:
@ricky-chaoju Please help me double check this one since it is highly related to your previous PR.

ricky-chaoju (approved):
Test this on 0.6.15 and 0.6.17. 0.6.15 picks `CompiledTargetVerifyMLPBlock`, 0.6.17 picks `CompiledMLPBlock`, full suite 2097 on each. All three targets are unary at 0.6.17, not just the dense one.

Also poked at the fail-fast since that's what I'd worry about with this shape. A wrong default, keyword-only, or a different name all still raise. Good to have the 0.6.16 pin lifted.

Thanks @LxYuan0420 . LGTM

PR #726, opened 2026-09-07
Remove stale prompt-logprobs decode gate

LxYuan0420:
This PR is:

- To remove the decode-pipeline prompt-logprobs blocker made stale by #694
- To let pure decode steps use the existing decode pipeline after prompt logprobs are delivered during prefill
- To delete the unit test that constructed an unreachable active prompt-logprobs decode state

After #694, prompt logprobs are tracked per request and delivered on the completing prefill step. Pure decode steps no longer need prompt-logprob rows, and prefill steps are already blocked from the decode pipeline by `has_prefill_phase_requests`.


WindChimeRan (approved):
LGTM

PR #728, opened 2026-09-07
[Bugfix] Added min_p filtering natively in MLX ops

### LINKED ISSUE #727: Support min_p sampling parameter
Repro:                                                                                                                                                                                                   
                                                                                                                                                                                                           
  ## Terminal 1: start server with min_p                                                                                                                                                                    
  VLLM_METAL_BUILD_FROM_SOURCE=1 vllm serve mlx-community/Qwen3.5-9B-MLX-4bit \                                                                                                                         
    --override-generation-config '{"temperature": 1.0, "top_p": 1.0, "top_k": 40, "min_p": 0.2, "presence_penalty": 1.5}'                                                                                  
                                                                                                                                                                                                           
  ## Terminal 2: send a chat request                                                                                                                                                                        
  vllm chat                                                                                                                                                                                                
  > hi                                                                                                                                                                                                      
                                                                                                                                                                                                           
```
  Error:                                                                                                                                                                                                   
  openai.APIError: vllm-metal does not support sampling controls backed by vLLM logits processors (min_p). (parameter=min_p)        
```                                                                    
                                                                                                                            
  Root cause:  Upstream vLLM implements min_p as a MinPLogitsProcessor, but vllm-metal does not support vLLM logits processors. make_sampling_metadata passes an empty LogitsProcessors object, so min_p has no effect. validate_request rejects it to avoid silently ignoring it. Unlike logit_bias and min_tokens (which are sent per-request by clients), min_p is set once at server startup via                 
  --override-generation-config and applied as a default to every request, so the rejection breaks the server for all requests.

trackerrr:
Fix: Implement min_p filtering natively in MLX ops (softmax → max → threshold → mask), matching upstream vLLM's sampling order (after temperature, before top-k/top-p). Covers both the native MLX path  
  and the torch fallback path. Remove min_p from the rejection list in validate_request. 
Closes #727

### DIFF
--- a/tests/test_sampling.py
+++ b/tests/test_sampling.py
@@ -26,6 +26,8 @@ from vllm_metal.v1.model_runner import (
 from vllm_metal.v1.sampling_batch import (
     SamplingBatch,
     sample_decode_tokens,
+    _apply_min_p_mlx,
+    _apply_min_p_torch,
     sample_from_logits,
 )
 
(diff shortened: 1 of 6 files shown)


vllm_metal/v1/sampling_batch.py
```diff
@@ -557,6 +597,8 @@ def sample_from_logits(
     logits_torch = mlx_to_torch(
         logits_2d.astype(mx.float32), device=SamplingBatch.SAMPLER_DEVICE
     )
+    # min_p: applied before the vLLM sampler since logitsprocs are not wired up.
```
LxYuan0420: Do not pre-mask logits before `Sampler.forward()`. vLLM computes raw logprobs first, then applies temperature, then applies min_p. This path applies min_p to raw logits, so seeded/fallback requests with `temperature != 1` can sample from the wrong candidate set, and raw logprobs are no longer raw. Keep min_p inside the sampler/logits-processor order.

vllm_metal/v1/sampling_batch.py
```diff
@@ -519,6 +519,46 @@ def make_sampling_metadata(
         )
 
 
+# ---------------------------------------------------------------------------
```
LxYuan0420: Please remove the banner comment and avoid adding free private helper functions for sampling policy. `SamplingBatch` should own the sampling decision and processor setup. Otherwise every new sampling option will add another module-level helper and another test surface.

tests/test_sampling.py
```diff
@@ -25,7 +25,12 @@
     _create_request_generator,
     _ExecutionBatch,
 )
-from vllm_metal.v1.sampling_batch import SamplingBatch, sample_from_logits
+from vllm_metal.v1.sampling_batch import (
+    SamplingBatch,
+    _apply_min_p_mlx,
```
LxYuan0420: Please test this through the public sampling path instead of importing private helpers. The current tests mostly prove the helper formula, while the real bug is sampler ordering. Keep this focused: one native path test and one seeded fallback test where min_p leaves a known candidate set.

trackerrr:
Thank you @LxYuan0420. I've addressed the comments and applied min_p in vLLM's sampler order on both sampling paths. Could you take a look again?

The main change:
- MLX path: fold min_p into `_top_k_top_p_min_p_masked_logits`
- Torch path: register `BatchMinPLogitsProcessor` through
  `SamplingMetadata.logitsprocs` rather than masking ahead of
  `Sampler.forward`.

LxYuan0420 (approved):
Pushed a small cleanup on top: rebased to current main, trimmed the long comments, and kept the native mask helper name so the existing native-sampling parity tests still run. The sampler-order concern is fixed. Please wait for the test workflow before merging.

Issue #700, opened 2026-09-05
[vllm 0.30.0] Compatibility with upstream KV-cache refactors


WindChimeRan:
vllm-metal depends on vLLM’s scheduler and cache-planning interfaces. A recent upstream vllm Attention Backend Refactor introduced breaking change: 
* RFC: https://github.com/vllm-project/vllm/issues/42449. 
* design doc: https://docs.google.com/document/d/1-TnWZf8jI6nWgE-xaQkgsWy130CVUqqt1NYqY-mAOe4/edit?tab=t.2gn7lfqebggq#heading=h.yr1y11b4ihee

Upstream https://github.com/vllm-project/vllm/pull/51718 has merged (not in our pinned vLLM v0.28.0). There might be more upstream PRs.

Static source review identifies three incompatibilities:

- **GDN prefix caching:** upstream removed `KVCacheTensor.shared_by`, which [[our hybrid cache initialization](https://github.com/vllm-project/vllm-metal/blob/7bc6a0ac5d91685bf21be229bdc925adcdc65c12/vllm_metal/v1/cache_policy.py#L695-L715)](https://github.com/vllm-project/vllm-metal/blob/7bc6a0ac5d91685bf21be229bdc925adcdc65c12/vllm_metal/v1/cache_policy.py#L695-L715) uses to derive shared state pools.
- **Mixed full/sliding-window attention:** [[our layout translator](https://github.com/vllm-project/vllm-metal/blob/7bc6a0ac5d91685bf21be229bdc925adcdc65c12/vllm_metal/attention/caches/mha_layout.py#L131-L150)](https://github.com/vllm-project/vllm-metal/blob/7bc6a0ac5d91685bf21be229bdc925adcdc65c12/vllm_metal/attention/caches/mha_layout.py#L131-L150) rejects nonzero strides/offsets and assumes the old tensor-sharing representation.
- **Context auto-fit:** [[our compatibility shim](https://github.com/vllm-project/vllm-metal/blob/7bc6a0ac5d91685bf21be229bdc925adcdc65c12/vllm_metal/compat.py#L76-L84)](https://github.com/vllm-project/vllm-metal/blob/7bc6a0ac5d91685bf21be229bdc925adcdc65c12/vllm_metal/compat.py#L76-L84) calls `_pool_bytes_per_block` with two arguments; upstream now accepts one.

[...]

PR #731, opened 2026-09-07
[Experimental] Reuse upstream vLLM models on PyTorch MPS

lucamotz:
The current Metal runner loads model implementations through MLX, so supporting an upstream vLLM architecture still depends on a separate MLX implementation and adapter. This draft explores a smaller compatibility boundary: execute **upstream vLLM model classes and weight loaders on PyTorch MPS**, while reusing upstream request handling and sampling.

`VLLM_METAL_MODEL_BACKEND=torch` selects an experimental out-of-tree platform. The existing MLX path remains the default. The implementation adds four backend files (about 720 lines including types, comments, and whitespace), with no changes to vLLM core and no copied model implementations.

### Reuse and scope

- `get_model` / `ModelRegistry` own architecture resolution and weight loading.
- `CPUModelRunner` owns request updates, batching, slot mapping, KV allocation/binding, prompt logprobs, and sampling. A small wrapper transfers inputs to MPS and logits to CPU; parameters, hidden states, and KV storage stay on MPS.
- Attention consumes `CommonAttentionMetadata` and uses native PyTorch SDPA over vLLM's paged cache. It supports causal/chunked prefill, decode, GQA, prefix reuse, and sliding windows. The cache adapter accepts both the 0.28 split layout and main's packed per-layer layout without copying cache storage.
- Setup rejects features outside the initial dense, unquantized, single-device, eager text-generation scope. An explicit but unavailable MPS selection fails after plugin probing, where vLLM cannot silently swallow the error and fall back to CPU.
- KV storage defaults to at most 512 MiB, bounded by system/MPS headroom. An explicit `kv_cache_memory_bytes` is checked against that budget. This is a KV limit, not a total-memory guarantee.

This is **not full upstream model coverage or a performance claim**. MoE, SSM/hybrid models, quantization, multimodal/pooling, LoRA, speculative decoding, distributed execution, and transfer/offload paths need further backend work. CUDA-only operations in otherwise dense architectures may still fail. SDPA gathers visible cache tensors and expands GQA heads, so long-context scratch memory and throughput need work. The normal macOS vLLM CPU extension is required for the inherited slot-mapping operation; an `empty` build is insufficient.

The alternative loader-only approach would still have to translate upstream Torch forward contexts and KV caches into the MLX runner's model-specific contract. Keeping model execution behind the existing upstream runner avoids that additional model-porting layer. Details and usage are in `docs/pytorch_backend.md`.

### Validation

Tested on Apple M4, 24 GB unified RAM, macOS 26.6.2, Python 3.12.13, PyTorch 2.13.0, the official vLLM 0.28 macOS CPU wheel, and this repository's pinned MLX dependencies. Large-model and performance suites were not run given the machine's memory pressure.

```bash
python -m pytest tests/test_register.py tests/test_macos_defaults.py \
  tests/test_platform.py tests/test_pytorch_attention.py \
  tests/test_pytorch_model_runner.py tests/test_tensor_bridge.py -q
# 137 passed, 10 skipped

python -m pytest tests/test_pytorch_models.py -q
# 4 passed

ruff check .
ruff format --check .
mypy vllm_metal
# All pass; 299 files formatted, 141 source files type-checked.
```

The model tests create tiny Llama, Qwen2, and Qwen3 checkpoints locally. Greedy token IDs match both Hugging Face CPU and `mlx_lm` references, including chunked prefill and repeated prefix-cache use. They assert that actual upstream model classes and all parameters are on MPS, and exercise logit bias plus prompt/output logprobs. Attention tests include real MPS float32/float16/bfloat16 execution, physical block addressing, released sliding-window prefix blocks, and both cache layouts. Memory-budget tests allocate no large buffers.

Additional manual checks:

- **SmolLM2-135M**, revision `93efa2f097d58c2a74874c7e644dbc9b0cee75a2`: all **36/36 greedy generated token IDs** matched the HF CPU reference across three prompts, in float32 with a **16 MiB KV cache**. Post-generation MPS allocation was 557,486,848 bytes; this is not a peak-process-memory measurement.
- A normally spawned localhost `vllm serve` process returned HTTP 200 for `/health` and `/v1/completions`, with expected continuation text and output logprobs. SIGTERM cleanup logged a forced-worker shutdown/semaphore warning; clean server shutdown needs further validation.
- Against upstream vLLM Python sources at **`537af2c3a4ba7462ddc9bc94ec7a4ea496da6d2e`**, tiny Llama/Qwen2/Qwen3 greedy checks and Llama chunked-prefill/prefix-cache/sampling-control checks passed. These used the installed 0.28 macOS wheel's C extension through a source overlay, **not a fresh native build of main**. The plugin's vLLM version pin is unchanged.

### Draft review

Checked open vllm-metal and upstream vLLM proposals on 2026-09-07. No open PR implements this upstream-model/MPS runner path. Existing MLX sampling, model-adapter, and KV-offload proposals operate on the existing MLX path; this does not replace their work or the version-migration checklist in #700.

AI assistance was used for investigation, implementation, and tests. This is intentionally a draft for human review of the reuse boundary before expanding model coverage or optimizing kernels. Broader checkpoint/dtype evaluation and a fresh current-main macOS build remain follow-up validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@ vLLM Metal is a plugin that enables vLLM to run on Apple Silicon Macs using MLX 
 
 Upstream vLLM supplies the API server, scheduler, and paged block manager; mlx_lm supplies the token-wise model layers; vllm-metal owns the request-aware attention path — the paged varlen kernel, M5 NAX prefill, and speculative decoding.
 
+An opt-in [PyTorch/MPS compatibility backend](docs/pytorch_backend.md) executes
+upstream vLLM model implementations directly. It is experimental and currently
+targets unquantized dense text generation on one Mac.
+
 ![vllm-metal in the vLLM stack](https://raw.githubusercontent.com/vllm-project/vllm-metal/main/docs/assets/architecture.svg)
 
 ## Requirements
--- a/docs/pytorch_backend.md
+++ b/docs/pytorch_backend.md
@@ -1,0 +1,89 @@
+# Experimental PyTorch/MPS backend
+
+Set `VLLM_METAL_MODEL_BACKEND=torch` before starting vLLM to execute upstream
+vLLM model implementations on Apple Silicon through PyTorch MPS. The default
+remains `mlx`.
+
+This is a compatibility prototype for unquantized dense decoder models. It
+does not establish support for every model in vLLM's registry. Models still need
+PyTorch-native implementations of all their operations, and this path is not
+optimized for throughput.
+
+## Usage
+
+Use the normal vllm-metal development installation, including the macOS vLLM
+CPU wheel and its native extension. An `empty` vLLM build is insufficient:
+the upstream CPU runner uses its native slot-mapping operation.
+
+Start with a small original Hugging Face checkpoint:
+
+```bash
+VLLM_METAL_MODEL_BACKEND=torch vllm serve HuggingFaceTB/SmolLM2-135M \
+  --host 127.0.0.1 --dtype float32 --max-model-len 256 \
+  --max-num-seqs 2 --max-num-batched-tokens 64 \
+  --kv-cache-memory-bytes 16777216
+```
+
+The same environment variable selects this backend for `LLM(...)`. Eager
+execution and synchronous scheduling are configured by the platform. The model
+loader honors upstream checkpoint loading and architecture resolution; no MLX
+model conversion is involved.
+
+The default KV budget is capped at 512 MiB and reduced when system or MPS
+headroom is lower. `--kv-cache-memory-bytes` overrides that cap within the
+available budget. These limits cover KV storage, not total process memory:
+weights, activations, and attention scratch space also consume unified memory.
+Use small context and batch limits on memory-constrained Macs.
+
+## Reused upstream components
+
+The plugin selects a separate out-of-tree platform before applying the MLX
+compatibility patches. Its worker reuses `CPUModelRunner` for request updates,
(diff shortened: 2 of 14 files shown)


WindChimeRan (commented):
Quick feedback:

I also don't like relying on `mlx_lm` / `mlx_vlm` for model configs, because they tend to be pretty messy. For example, `mlx_lm` always trims the MTP head, and support for visual encoders mostly depends on `mlx_vlm`, which is maintained by a smaller community.

But switching to vLLM models is non-trivial because vLLM is deeply coupled with PyTorch. From my own benchmarks, PyTorch MPS is far slower than MLX, which makes it basically unusable on Apple Silicon.

So I'd still vote for maintaining only the MLX runtime backend, even if that means keeping a number of ugly patches. I'm not sure whether it's feasible to read vLLM's model configs and translate them cleanly into MLX, though...

LxYuan0420 (commented):
Thanks for exploring this, but I’m not sold on this direction for vllm-metal. A separate PyTorch MPS runtime would bypass our optimised Metal kernels, support fewer features, and significantly increase our maintenance burden :)

lucamotz:
@WindChimeRan @LxYuan0420 thanks for the quick feedback haha! Didn’t expect you guys to even look at a draft. I’ll just close it again then, taking it on board. We would have run some benchmarks at work over the next few weeks to see if we can get this to work, but if you say it’s truly not worth it, we’ll rely on other software for our Mac experiments. It’s a shame really!

PR #733, opened 2026-09-08
Use MLX 0.32 APIs in loader and TurboQuant paths

LxYuan0420:
This PR is:

- To use native `mx.searchsorted` in TurboQuant quantization now that MLX 0.32.1 is pinned
- To remove redundant `mx.load()` runtime type guards from GGUF, LoRA, and encoder loading paths
- To keep file-format validation at the loader boundary and trust MLX’s named-tensor load contract internally

The TurboQuant change removes a local `searchsorted` workaround. The loader cleanup keeps the same runtime behavior for valid files while reducing repeated defensive checks after the MLX dependency bump.


ricky-chaoju (approved):
LGTM

PR #734, opened 2026-09-08
[Spec Decode] Lazily catch up draft KV after zero-K steps

POWEHI2023:
## Motivation

This is a follow-up to #716.

#716 keeps the draft KV cache synchronized by running draft-model ingest even
when dynamic speculative decoding selects K=0. This fixes the missing-KV
problem, but adds draft-model work to every zero-K scheduler step.

During the review and validation of #716, a more efficient alternative was
suggested: record the earliest missing draft-KV position during K=0 without
running the draft model, then lazily ingest the complete missing range before
the next K>0 proposal.

This draft PR implements that approach.

## Summary

When `enable_prefix_caching=False`, K=0 now records the earliest valid draft-KV
boundary without running the draft model. The missing range is ingested once,
immediately before the next K>0 proposal. This preserves cache correctness
while avoiding work for requests that finish before drafting resumes.

`enable_prefix_caching=True` keeps the eager #716 behavior so scheduler-visible cache hits
cannot reference draft KV that has not been materialized.

## Implementation

- Preserve the physical KV boundary with `setdefault()` during K=0.
- Reconstruct deferred prefill ranges in `_make_prefill_plan()`
- Return unused speculative scratch blocks during zero-budget steps.
- Pass the prefix-caching setting `enable_prefix_caching ` from the runner to the proposer.

## Performance

Three alternating eager/lazy pairs were run in fresh processes using
`Qwen/Qwen3-0.6B` as both target and draft. Dynamic-K used K=3 for batch size 1
and K=0 for batch size 2.

| Metric | Eager (#716) | Lazy (this PR) |
|---|---:|---:|
| Median wall time | 0.8610 s | 0.7754 s |
| K=0 ingest tokens | 426 | 0 |
| Total ingest tokens | 439 | 184 |
| Peak per-request ingest | 128 | 172 |
| Accepted/proposed | 18/18 | 18/18 |

Median wall time decreased by **9.94%**, while total ingest decreased by
**58.09%**. Proposal hashes, target outputs, and acceptance counts were
identical across all six processes. Lazy catch-up increases the largest resume
ingest from 128 to 172 tokens; the result is specific to this workload.

The 255-token reduction comes from the companion request, which finishes while
K=0 and therefore never needs draft KV. Eager mode still ingests its 127- and
128-token chunks; lazy mode only records its boundary and later drops the
request. The survivor performs the same 172 tokens of catch-up work in both
modes: eager spreads it across `127 + 44 + 1` tokens, while lazy performs one
172-token ingest before drafting resumes. If every K=0 request later returns to
K>0, total ingest tokens may be similar; the remaining difference would be
forward scheduling and peak catch-up size.

## Test

```shell
pytest -m "not slow" tests/ -q --tb=short
2108 passed, 15 skipped, 53 deselected, 16 warnings
```

### DIFF
--- a/tests/test_draft_model_proposer.py
+++ b/tests/test_draft_model_proposer.py
@@ -141,37 +141,23 @@ def _prefills_context(
     prefills: list[tuple[str, list[int]]],
     *,
     start_pos: int = 0,
-    end_pos: int | None = None,
     result_mode: str = "final",
+    full_prompt_token_ids: list[int] | None = None,
     num_speculative_tokens: int = 1,
-    sampled_token_id: int = 42,
-    committed_block_ids: tuple[int, ...] = (0,),
 ) -> ProposeContext:
-    """Build a proposer context for one or more prefill requests."""
+    """Context whose requests are final-chunk prefills (greedy, drafting)."""
     prefill_reqs = [
         PrefillRequest(
             req_id=req_id,
-            token_ids=list(
-                full_prompt_token_ids[
-                    start_pos : (
-                        len(full_prompt_token_ids) if end_pos is None else end_pos
-                    )
-                ]
-            ),
+            token_ids=list(token_ids),
             sampling_params=SamplingParams(temperature=0.0),
-            block_ids=[list(committed_block_ids)],
+            block_ids=[[0]],
             generator=None,
-            prompt_len=(
-                None
-                if result_mode in {"intermediate", "final"}
-                else len(full_prompt_token_ids)
-            ),
+            prompt_len=None,
             start_pos=start_pos,
-            full_prompt_token_ids=(
-                list(full_prompt_token_ids) if start_pos > 0 else None
-            ),
+            full_prompt_token_ids=full_prompt_token_ids,
         )
-        for req_id, full_prompt_token_ids in prefills
+        for req_id, token_ids in prefills
     ]
     return ProposeContext(
         target_hidden_states=None,
@@ -179,14 +165,10 @@ def _prefills_context(
         decode_segments=[],
         decode_token_ids=[],
         prefill_reqs=prefill_reqs,
-        prefill_token_ids=[sampled_token_id] * len(prefill_reqs),
+        prefill_token_ids=[42] * len(prefill_reqs),
         prefill_result_modes=[result_mode] * len(prefill_reqs),
         request_states={
-            req_id: _request_state(
-                committed_block_ids=list(committed_block_ids),
-                token_ids=full_prompt_token_ids,
-            )
-            for req_id, full_prompt_token_ids in prefills
+            req_id: _request_state(committed_block_ids=[0]) for req_id, _ in prefills
         },
         cu_seqlens=[],
         num_decode_segments=0,
@@ -321,21 +303,16 @@ def test_intermediate_prefill_chunk_ingests_without_drafting() -> None:
     assert len(model.block_tables) == 1
 
 
-def test_zero_k_eagerly_ingests_when_prefix_caching_is_enabled() -> None:
+def test_eager_zero_k_still_ingests() -> None:
     """Prefix-cache mode must materialize scheduler-publishable draft KV."""
     model = _StubDraftModel()
     proposer = _proposer(model)
-    prompt = list(range(PROMPT_LEN))
 
     drafts = proposer.propose(
         _prefills_context(
-            [("r1", prompt)],
-            start_pos=0,
-            end_pos=8,
+            [("r1", list(range(8)))],
             result_mode="intermediate",
             num_speculative_tokens=0,
-            sampled_token_id=VOCAB_SIZE - 1,
-            committed_block_ids=(0, 1),
         )
     )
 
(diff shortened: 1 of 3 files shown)


vllm_metal/v1/draft_model_proposer.py
```diff
@@ -150,11 +152,16 @@ def __init__(
         controller: SpeculativeDecodeController,
         extract_logits: Callable[[Any], mx.array],
         merge_ingest_windows: bool = False,
+        enable_prefix_caching: bool = True,
```
LxYuan0420: I see what you did here, but the proposer does not need to know about prefix caching. Pass defer_zero_k_ingest directly, and let the runner decide that from cache_config.enable_prefix_caching.

vllm_metal/v1/draft_model_proposer.py
```diff
@@ -365,6 +378,25 @@ def release_requests(self, req_ids: set[str]) -> None:
 
     # -- internals -----------------------------------------------------------
 
+    def _record_deferred_ingest_boundaries(self, ctx: ProposeContext) -> None:
```
LxYuan0420: This can be simpler. We do not need to build a boundaries list just to loop once. Iterating decode rows and prefill rows directly would be easier to read.

vllm_metal/v1/draft_model_proposer.py
```diff
@@ -487,6 +519,44 @@ def _make_prefill_plan(
             return None
         req_id = prefill.req_id
         committed_len = prefill.start_pos + len(prefill.token_ids)
+        draft_seq_len = self._draft_seq_lens.setdefault(req_id, prefill.start_pos)
+        if draft_seq_len >= committed_len:
+            return None
+
+        if draft_seq_len < prefill.start_pos:
```
LxYuan0420, line 510: This block is doing a lot now: normal prefill planning, deferred prompt reconstruction, final-token stitching, and length validation. I would split the ingest-token selection into a small named method so _make_prefill_plan() stays easy to follow.

tests/test_draft_model_proposer.py
```diff
@@ -79,6 +79,7 @@ def _proposer(
     *,
     committed_num_blocks: int = COMMITTED_NUM_BLOCKS,
     scratch_reserve_blocks: int = SCRATCH_RESERVE_BLOCKS,
+    enable_prefix_caching: bool = True,
```
LxYuan0420: This default now shows up in the common test helper, but only the new lazy path needs it. I would keep old tests on the default path and set the defer policy only in the new zero-K lazy tests.

tests/test_draft_model_proposer.py
```diff
@@ -298,6 +321,182 @@ def test_intermediate_prefill_chunk_ingests_without_drafting() -> None:
     assert len(model.block_tables) == 1
 
 
+def test_zero_k_eagerly_ingests_when_prefix_caching_is_enabled() -> None:
```
LxYuan0420: These tests feel larger than the behavior we need to prove. I would trim them to the main cases: eager mode still ingests on K=0, lazy mode skips K=0 and catches up before drafting, and the preserved-boundary case if that comes from a real trace.

POWEHI2023:
@LxYuan0420  Thanks for reviewing this! I’ve kept the defer_zero_k_ingest decision in the runner, inlined _record_deferred_ingest_boundaries into the lazy path in propose(), and extracted _select_prefill_ingest_tokens() from _make_prefill_plan().
The test helper now defaults to eager ingest. Trimming the new tests is a bit tricky because they cover behavior that’s coupled with assumptions across the project. Could you help identify which cases you’d prefer to drop or consolidate?

tests/test_draft_model_proposer.py
POWEHI2023: Thanks, I trimmed the coverage to three focused cases:
- eager K=0 still ingests;
- lazy K=0 skips ingest and catches up before the next K>0 draft;
- lazy K>0 -> K=0 preserves the existing physical KV boundary.

I also simplified the prefill helper -> _prefills_context().

POWEHI2023:
@LxYuan0420 Thanks for the review. I’ve addressed the test-size concern in c1565db:

- kept the eager K=0 ingest case;
- consolidated the lazy coverage into K=0 -> K>0 catch-up before drafting;
- retained the K>0 -> K=0 boundary case because it comes from a real Dynamic-K trace;
- removed the larger consecutive-prefill test and simplified the test helper.

The targeted proposer tests pass, and the full non-slow suite passed.
Could you please take another look when you have time? If this addresses your concerns, would you mind approving the PR?

LxYuan0420 (approved):
LGTM. One minor naming nit: `defer_zero_k_ingest` reads like a generic performance toggle; `allow_deferred_zero_k_ingest` would make the prefix-cache safety boundary clearer.

POWEHI2023:
@LxYuan0420 Thanks for the approval and the naming suggestion. I renamed defer_zero_k_ingest to allow_deferred_zero_k_ingest throughout the runner, proposer, and tests. The name now makes the prefix-cache safety condition clearer.

PR #735, opened 2026-09-08
Run Nemotron-H through the decode pipeline

ricky-chaoju:
PR 4 of the Nemotron-H stack in #644, on top of #710, which kept Nemotron-H off the one-step-ahead decode pipeline until it had mileage. On the 30B 4bit checkpoint the pipeline takes greedy chat completions from 58.6 to 84.0 tok/s on one stream and from 143.2 to 170.1 on four, with identical completions over 256 tokens (M1 Ultra, `VLLM_METAL_DECODE_PIPELINE=0` vs `1`, same shape as #611). The wrapper's decode is one gathered mixer call with lazy pool writes, so nothing in the step needs the sync the pipeline defers.

With every family on the pipeline, the per-family opt-out has no reader left, so this deletes `supports_decode_pipeline` and the runner capability it fed; `VLLM_METAL_DECODE_PIPELINE=0` stays the kill switch.

The parity tool on this head gives the same output as #710 with the pipeline engaged. Compiled MLP dispatch for the MoE blocks and batched right-padded prefill were both measured and not shipped, no gain and not token-identical respectively.


LxYuan0420 (approved):
LGTM

PR #736, opened 2026-09-08
[Compat] Load official Qwen3-Embedding checkpoints (#730)

### LINKED ISSUE #730: [Model Support] Qwen3-Embedding: official Qwen checkpoints fail to load, flat keys need a `model.` prefix remap
## Environment

- vllm-metal 0.28.0.dev from main @ 207155c, installed with install.sh
- vLLM 0.28.0, transformers 5.16.1, mlx 0.32.0, mlx-lm pinned to git 254d153f
- macOS Version 26.6.2 (25G83), Apple Silicon M4 24GB

## Reproduction

```bash
vllm serve Qwen/Qwen3-Embedding-0.6B \
  --runner pooling \
  --dtype auto \
  --max-model-len 600 \
  --port 8100 \
  --trust-remote-code
```

The model load in EngineCore fails:

```
ValueError: Received 310 parameters not in model:
  embed_tokens.weight, layers.0.input_layernorm.weight, ..., norm.weight
```

## Root cause

The official Qwen3-Embedding checkpoints for the 0.6B, 4B and 8B models store backbone weights without the `model.` prefix:

```
embed_tokens.weight
layers.0.input_layernorm.weight
...
norm.weight
```

The `Qwen3` class in mlx-lm, `mlx_lm/models/qwen3.py`, wraps the backbone:

```python
class Model(nn.Module):
    def __init__(self, args):
        self.model = Qwen3Model(args)   # expects model.embed_tokens.*, model.layers.*, model.norm.*
```

so `load_weights(strict=True)` rejects all 310 tensors. The upstream `sanitize()` only drops a tied `lm_head.weight`. It never remaps keys.

This is the checkpoint Qwen ships. sentence-transformers and CUDA vLLM both load it as-is, but the vllm-metal docs only list `mlx-community/Qwen3-Embedding-0.6B-8bit` for this model. Anyone who wants the unquantized official weights has no working path today.

## Proposed fix

A small compat patch in `vllm_metal/compat.py`, using the same `_wrap_model_sanitize()` pattern as the existing Qwen3.5 FP8 and gemma4 patches:

```python
_QWEN3_BACKBONE_ROOTS = frozenset({"embed_tokens", "layers", "norm", "lm_head"})


def _patch_mlx_lm_qwen3_flat_weight_prefix() -> None:
    """Remap flat official Qwen3 checkpoints to mlx_lm's model.* layout."""
    from importlib import import_module
    from importlib.util import find_spec

    if find_spec("mlx_lm.models.qwen3") is None:
        return
    qwen3 = import_module("mlx_lm.models.qwen3")
    model_cls = getattr(qwen3, "Model", None)
    if model_cls is None:
        return
    if _wrap_model_sanitize(
        model_cls,
        "_vllm_metal_qwen3_flat_prefix_patch",
        _qwen3_flat_weight_prefix,
    ):
        logger.debug("Patched mlx_lm Qwen3 flat weight prefix compatibility")


def _qwen3_flat_weight_prefix(_self, weights):
    """Re-key flat official Qwen3 checkpoints to mlx_lm's ``model.*`` layout."""
    if any(str(key).startswith("model.") for key in weights):
        return weights  # already MLX layout, e.g. mlx-community repos
    roots = {str(key).split(".", 1)[0] for key in weights}
    if not roots or not roots <= _QWEN3_BACKBONE_ROOTS:
        return weights  # not a flat Qwen3 backbone, leave untouched
    if bool(getattr(getattr(_self, "args", None), "tie_word_embeddings", False)):
        weights = {k: v for k, v in weights.items() if k != "lm_head.weight"}
    return {"model." + str(k): v for k, v in weights.items()}
```

The patch gets registered in `apply_compat_patches()`.

Why it is safe:

- Checkpoints that already use the `model.*` layout pass through unchanged. That covers every mlx-community repo.
- The remap only fires when all top-level keys are `embed_tokens`, `layers`, `norm` or `lm_head`, so unrelated flat layouts are left alone.
- A tied `lm_head.weight` is dropped, matching what upstream `sanitize()` already does.
- The wrapper is idempotent and re-applies cleanly in the EngineCore process via model_lifecycle.py.

## Validation

I applied the patch to my local install and ran it against the original Qwen safetensors:

- The server starts. /v1/embeddings returns 1024-dim, L2-normalized vectors.
- I compared its output against my existing Qdrant collection. It holds 56,523 document chunks built earlier with the same model on vLLM/CUDA in bf16 on a rtx 3090. On 32 sampled chunks of 185 to 502 tokens, the new embeddings score 0.99983 median cosine against the stored vectors, 0.99972 in the worst case. Pre-built indices stay compatible.
- No other model path changes.

## Alternatives

- Fix it in mlx-lm's qwen3.sanitize. Also correct, but vllm-metal pins a git SHA of mlx-lm and already keeps equivalent shims in its compat layer. Fixing here works without waiting on an mlx-lm release.
- Point users at mlx-community/Qwen3-Embedding-0.6B-8bit. It loads, but I measured ~2e-3 cosine drift against the original weights, and it is a quantized re-pack, not Qwen's own artifact.

## AI assistance

AI assistance was used to reproduce the failure, trace the root cause, benchmark alternatives and draft the patch and this report. Harness: pi coding agent. Model: Qwen3.8-27B via local llama.cpp. The benchmark and index numbers above were produced on my machine by the assistant, so please reproduce them before any fix is merged. I reviewed the findings and the code.

mhdimo:
Fixes #730

## Summary

Official Qwen3-Embedding checkpoints (0.6B, 4B, 8B) fail to load on
vllm-metal with `ValueError: Received 310 parameters not in model`. The
root cause is two stacked defects at the `vllm_metal` ↔ `mlx_lm` boundary;
this PR fixes both with one narrow compat patch in `vllm_metal/compat.py`.

## Root cause

1. **Flat backbone keys.** Qwen ships these checkpoints with keys like
   `embed_tokens.weight`, `layers.0....`, `norm.weight`, but mlx_lm's
   `qwen3.Model` wraps the backbone as `self.model = Qwen3Model(args)` and
   expects `model.`-prefixed keys. Upstream `sanitize()` only drops a tied
   `lm_head.weight` — it never remaps — so strict load rejects every
   backbone tensor.
2. **Headless but untied (8B only).** `Qwen/Qwen3-Embedding-8B` sets
   `tie_word_embeddings: false` yet ships **no** `lm_head.weight` at all
   (398 keys, roots `{embed_tokens, layers, norm}`). mlx_lm constructs the
   head when untied, so even with correct prefixes, strict load fails with
   `Missing 1 parameters: lm_head.weight`. Neither the issue's proposed
   patch nor plain mlx_lm handles this.

## Fix

`_patch_mlx_lm_qwen3_flat_weight_prefix()` wraps `mlx_lm.models.qwen3`
`Model.sanitize` via the existing `_wrap_model_sanitize()` helper (same
pattern as the Qwen3.5 FP8 patch), with a transform that:

- passes through unchanged when **any** key already starts with `model.`
  (every mlx-community repo), or when top-level keys are not all Qwen3
  backbone roots;
- otherwise adds the `model.` prefix to backbone keys;
- keeps `lm_head.weight` top-level — upstream `sanitize` drops it when
  tied, untied checkpoints own it there;
- when an untied flat checkpoint ships no head tensor, deletes the
  `mlx.nn` module mlx_lm just constructed (`__delattr__` removes it from
  `parameters()`), so strict load stops demanding a tensor Qwen never
  wrote. Embed pooling reads body hidden states via `forward_packed` and
  never touches `lm_head`; reranker checkpoints always ship the head
  (their classify path needs it), so that path is unaffected.

The wrapper is idempotent via the shared sanitize-wrap sentinel.

## Testing

All runs on M5 / 32 GB, vLLM 0.28 pooling runner, `max_model_len=600`:

- Pre-fix: `Qwen/Qwen3-Embedding-0.6B` fails with the issue's exact error.
- Post-fix, real checkpoints load and embed (L2-normalized):

  | Checkpoint | Embedding dim | Result |
  |---|---|---|
  | `Qwen/Qwen3-Embedding-0.6B` | 1024 | pass |
  | `Qwen/Qwen3-Embedding-4B` | 2560 | pass |
  | `Qwen/Qwen3-Embedding-8B` | 4096 | pass |
  | `mlx-community/Qwen3-Embedding-0.6B-8bit` (regression) | 1024 | unchanged |

- 12 new unit tests in `tests/test_compat.py`: remap rules, tied/untied/
  headless `lm_head` cases, pass-through for prefixed and non-Qwen3
  layouts, idempotent wrap. `python -m pytest tests/test_compat.py` →
  42 passed, 3 pre-existing skips. `scripts/lint.sh` clean.

Reviewer repro:

```bash
vllm serve Qwen/Qwen3-Embedding-0.6B --runner pooling --max-model-len 600
```

## Notes / follow-ups

- `docs/supported_models.md` still lists only the mlx-community 8bit
  repo; updating the table is a deliberate non-goal here and can follow
  once this lands.
- The 8B headless case is per-checkpoint detection (flat layout + untied
  config + absent tensor), not a general "skip lm_head" loader change;
  a pooling-path fix could subsume it later if more headless checkpoints
  appear.

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -921,9 +921,27 @@ class TestQwen3FlatWeightPrefixTransform:
 
         assert out == weights
 
-    def test_untied_headless_checkpoint_drops_lm_head_module(self) -> None:
+    def test_untied_headless_checkpoint_drops_lm_head_module_in_pooling_scope(
+        self,
+    ) -> None:
         # The official 8B embedding checkpoint ships tie_word_embeddings=false
-        # but no lm_head tensor at all; strict load must not demand one.
+        # but no lm_head tensor at all; a pooling load must not demand one.
+        fake_self = _HeadfulQwen3Self(tie_word_embeddings=False)
+        assert fake_self.lm_head is not None
+
+        with compat.pooling_load_scope():
+            out = compat._qwen3_flat_weight_prefix(
+                fake_self, _flat_official_qwen3_weights()
+            )
+
+        assert "model.norm.weight" in out
+        assert not hasattr(fake_self, "lm_head")
+
+    def test_untied_headless_checkpoint_keeps_lm_head_module_outside_pooling(
+        self,
+    ) -> None:
+        # Generation and classify need the head. Outside the pooling scope the
+        # module stays, so strict load fails with the missing-tensor error.
         fake_self = _HeadfulQwen3Self(tie_word_embeddings=False)
         assert fake_self.lm_head is not None
 
@@ -932,7 +950,7 @@ class TestQwen3FlatWeightPrefixTransform:
         )
 
         assert "model.norm.weight" in out
-        assert not hasattr(fake_self, "lm_head")
+        assert fake_self.lm_head is not None
 
     def test_headful_checkpoint_keeps_lm_head_module(self) -> None:
         fake_self = _HeadfulQwen3Self(tie_word_embeddings=False)
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -810,35 +810,44 @@ class TestInstallMetalOverride:
                 delattr(RayWorkerProc, "_metal_patched")
 
 
-def _install_fake_qwen3_module(monkeypatch):
-    mlx_lm_pkg = ModuleType("mlx_lm")
-    mlx_lm_models = ModuleType("mlx_lm.models")
-    mlx_lm_pkg.models = mlx_lm_models
-    monkeypatch.setitem(sys.modules, "mlx_lm", mlx_lm_pkg)
-    monkeypatch.setitem(sys.modules, "mlx_lm.models", mlx_lm_models)
+def _unwrap_qwen3_sanitize(model_cls) -> bool:
+    """Restore upstream sanitize if this session already wrapped it."""
+    original = getattr(model_cls.sanitize, "_vllm_metal_original_sanitize", None)
+    if original is None:
+        return False
+    model_cls.sanitize = original
+    return True
 
-    qwen3_module = ModuleType("mlx_lm.models.qwen3")
 
-    class Model:
-        def __init__(self, tie_word_embeddings: bool = False) -> None:
-            self.args = SimpleNamespace(tie_word_embeddings=tie_word_embeddings)
+def _tiny_qwen3_args(tie_word_embeddings: bool):
+    from mlx_lm.models.qwen3 import ModelArgs
 
-        def sanitize(self, weights):
-            if self.args.tie_word_embeddings:
-                weights.pop("lm_head.weight", None)
-            return weights
+    return ModelArgs(
+        model_type="qwen3",
+        hidden_size=16,
+        num_hidden_layers=1,
+        intermediate_size=32,
+        num_attention_heads=2,
+        rms_norm_eps=1e-6,
+        vocab_size=32,
+        num_key_value_heads=2,
+        max_position_embeddings=64,
+        rope_theta=10000.0,
+        head_dim=8,
+        tie_word_embeddings=tie_word_embeddings,
+    )
 
-    qwen3_module.Model = Model
-    monkeypatch.setitem(sys.modules, "mlx_lm.models.qwen3", qwen3_module)
-    mlx_lm_models.qwen3 = qwen3_module
 
-    def _fake_find_spec(name: str):
-        if name == "mlx_lm.models.qwen3":
-            return object()
-        return None
+def _official_flat_weights(model) -> dict[str, object]:
+    """Backbone tensors in the official flat layout: no ``model.`` prefix."""
+    from mlx.nn.utils import tree_flatten
 
-    monkeypatch.setattr(importlib.util, "find_spec", _fake_find_spec)
-    return qwen3_module
+    flat = dict(tree_flatten(model.parameters()))
+    return {
+        key.removeprefix("model."): value
+        for key, value in flat.items()
+        if key.startswith("model.")
+    }
 
 
 def _flat_official_qwen3_weights() -> dict[str, object]:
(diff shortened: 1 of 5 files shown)


vllm_metal/compat.py
```diff
@@ -913,3 +915,78 @@ def _patched_sanitize(self, weights):
...
+        return weights
+    roots = {key.split(".", 1)[0] for key in keys}
+    if not roots or not roots <= _QWEN3_BACKBONE_ROOTS:
+        return weights
+    if (
+        "lm_head.weight" not in weights
+        and not model.args.tie_word_embeddings
+        and getattr(model, "lm_head", None) is not None
+    ):
+        delattr(model, "lm_head")
```
LxYuan0420, line 978: Do not remove `lm_head` in the global Qwen3 sanitize patch. Missing `lm_head` is only safe for embedding/pooling because that path never calls logits. Generation/classify should still fail on an untied headless checkpoint. Please scope this to pooling or make it an explicit pooling-only shim.

vllm_metal/compat.py
```diff
@@ -913,3 +915,78 @@ def _patched_sanitize(self, weights):
...
+
+
+def _patch_mlx_lm_qwen3_flat_weight_prefix() -> None:
+    """Teach mlx_lm's Qwen3 loader to accept flat official Qwen3 checkpoints.
+
+    Qwen ships the Qwen3-Embedding checkpoints (0.6B/4B/8B) with backbone keys
+    such as ``layers.0....`` while mlx_lm's ``qwen3`` wrapper expects
+    ``model.layers.0....``; upstream ``sanitize`` never remaps, so
+    ``load_weights(strict=True)`` rejects all tensors. Mirrors the Qwen3.5 FP8
+    patch: narrow to the affected model module, upstream control flow intact.
```
LxYuan0420, line 992: Please add the upstream mlx-lm issue/PR link and removal condition. This is an upstream key-layout compat shim, not permanent vllm-metal behavior.

tests/test_compat.py
```diff
@@ -803,3 +808,185 @@ def test_patch_ray_distributed_targets_real_worker(self) -> None:
                 delattr(RayWorkerProc, "get_node_and_physical_gpu_ids")
             if not had_sentinel and "_metal_patched" in RayWorkerProc.__dict__:
                 delattr(RayWorkerProc, "_metal_patched")
+
+
+def _install_fake_qwen3_module(monkeypatch):
```
LxYuan0420: Please trim the fake-module test matrix and add one test against the real `mlx_lm.models.qwen3.Model` strict-load shape. The test should prove the actual #730 failure path, not only our local transform.

mhdimo: Removed.

vllm_metal/compat.py
mhdimo: there is no mlx-lm issue/pr for this, i added a `TODO:` for tracking this issue in the future.

vllm_metal/compat.py
mhdimo: Fixed.

vllm_metal/v1/model_lifecycle.py
```diff
@@ -153,7 +153,11 @@ def load(self) -> None:
             return
 
         request = GenerationLoadRequest.from_runner(self._runner, self._model_adapter)
-        loaded_model = self._load_generation(request)
+        # Pooling-only compat shims (the Qwen3 headless lm_head drop) must stay
+        # off for generation loads: strict load has to fail there, not the
+        # first forward pass.
+        with pooling_load_scope(enabled=self._runner._is_pooling):
```
LxYuan0420: `_is_pooling` is too broad here because it also covers Qwen3 reranker `classify`, which needs `lm_head`. Please make this embedding-only and context-local, so missing heads still fail for classify and generation loads.

mhdimo: Fixed

mhdimo:
rebased over upstream

LxYuan0420 (approved):

PR #737, opened 2026-09-08
Add a KV offloading serving benchmark

RobbieJ:
The tool behind the numbers in #681. It is separate so the feature PR
stays the feature, and so anyone can re-run the measurement on their own
machine before reading 3,900 lines. Stdlib only, no dependency on #681's
code; it drives `vllm serve` and `vllm bench serve`.

## What it measures

Four workloads, each with offloading off and on against a freshly started
server, order rotated per repetition:

- **no reuse** and **reuse inside the cache**: controls. Offloading cannot
  help on either. If it shows ahead by more than noise the benchmark is
  measuring admission queueing rather than prefill, and the tool exits
  non-zero.
- **evicted prefix, requested again**: two passes against one server. The
  first overflows the cache untimed; the second is measured, after the
  early prefixes have been evicted. One pass cannot measure this, because
  vLLM's `prefix_repetition` dataset emits every request for a prefix
  consecutively.
- **restart, same traffic**: a second process. Nothing survives a restart
  without a disk tier.

It reads the KV cache size from the server's own startup line and sizes
the host pool at 1.3x it and the spilling workload at 1.4x it. It warns
when a configuration would swap, scales the spilling and no-reuse
workloads down to what the disk can hold, and exits non-zero when a
sample fails.

## Run

```bash
python tools/benchmark/kv_offload_benchmark.py \
    --model mlx-community/Qwen2.5-32B-Instruct-4bit \
    --memory-fraction 0.8 --offload-size-gib 8 \
    --max-model-len 8192 --prefix-len 3072 --suffix-len 512 --repeats 3 \
    --out kv-offload-benchmark.json
```

That command, on a 128 GB M4 Max with the #681 branch, gave TTFT ratios
of 1.00x and 0.98x on the controls, 48.09x on the evicted prefix and
17.06x on restart, three repetitions each with IQR under 1% except 3.2%
on one arm. `--no-disk-tier` on the same configuration gave 1.00x and
0.99x, which is the measurement behind #681 having no CPU-only path.

Without #681 the server rejects the offloading flags, so the tool has
nothing to measure until that lands. It reviews on its own.

Flags: `--memory-fraction`, `--offload-size-gib`, `--baseline-memory-fraction`
(give the baseline a larger cache instead of a pool, for an equal-bytes
control), `--no-disk-tier`, `--kv-cache-tokens`, `--prefix-len`,
`--suffix-len`, `--num-prompts`, `--repeats`, `--bench-timeout`,
`--keep-logs`. One file, 772 lines, `ruff` clean.

### DIFF
--- a/tools/benchmark/kv_offload_benchmark.py
+++ b/tools/benchmark/kv_offload_benchmark.py
@@ -188,11 +188,14 @@ class ServerProcess:
         offload: bool,
         store_dir: Path | None,
         log_path: Path,
+        fraction: float | None = None,
     ) -> None:
         self.args = args
         self.offload = offload
         self.store_dir = store_dir
         self.log_path = log_path
+        # Explicit memory fraction, overriding the arm-based choice in _env.
+        self.fraction = fraction
         self.port = free_port()
         self.proc: subprocess.Popen | None = None
         self.max_tokens_cached = 0
@@ -237,6 +240,8 @@ class ServerProcess:
         fraction = self.args.memory_fraction
         if not self.offload and self.args.baseline_memory_fraction:
             fraction = self.args.baseline_memory_fraction
+        if self.fraction is not None:
+            fraction = self.fraction
         env["VLLM_METAL_MEMORY_FRACTION"] = str(fraction)
         # Block filenames are content hashes and hash seeding is per process.
         # Without this a restarted server cannot find what the previous one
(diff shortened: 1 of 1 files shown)


WindChimeRan (commented):
Please change the status to "Draft". I think this PR is not ready for review.

PR #739, opened 2026-09-09
Accept remote plain-typed GGUF reference tags

ricky-chaoju:
A local F16/F32/BF16 GGUF file loads today, but the same checkpoint referenced as `org/repo:F16` is rejected. The tag grammar already accepts these; the only block is the qtype allowlist in `resolve()`. This lifts that allowlist to match what the loader already executes, one line. K-quant and IQ tags stay rejected, and the gate still sits before both the online and offline branches from #732.

Before: `Remote GGUF qtype 'F16' is not supported by vllm-metal; supported qtypes: Q4_0, Q4_1, Q8_0.` After, the reference resolves and serves. For evidence I served `unsloth/Qwen3-0.6B-GGUF:BF16` (revision 50968a44) end to end with `tools/gguf_serve_smoke.py` and the greedy continuation is correct. One note: `mx.load` reads BF16 tensors as float16, exact within f16 range, same as the local path already does. 83/83 in the touched test files with eleven new cases, full suite 2143, ruff and mypy clean.

### DIFF
--- a/docs/gguf.md
+++ b/docs/gguf.md
@@ -42,8 +42,7 @@ before model load.
 ## Current scope
 
 - Supported model families: Qwen2, Qwen3, Llama, and Mistral.
-- Supported qtypes: Q8_0, Q4_0, and Q4_1.
-- Remote references additionally accept `:F16`, `:F32`, and `:BF16` tags for
-  the plain-typed checkpoints the loader already executes.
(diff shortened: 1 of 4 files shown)


LxYuan0420 (approved):
nit: the supported-qtype wording should be consistent for local and remote. F16/F32/BF16 already work for local GGUF too, so the docs should not make them sound remote-only.

LxYuan0420 (approved):

PR #744, opened 2026-09-10
Bump vLLM to 0.29.0

ricky-chaoju:
0.29.0 changed the KV cache layout [(vllm#51718)](https://github.com/vllm-project/vllm/pull/51718). `KVCacheTensor` no longer says which layers share a slot; groups sit on one shared allocation and a layer's region starts at `offset + l * layer_stride`, so `MHAKVCacheLayout` now derives slots from those addresses and the hybrid state pools use the same helper. The engine core also asks workers which layouts they support, and the default probe would pick the CPU backend's LBHNC, so `MetalWorker` answers LBNHC and waits on MLX in `synchronize_device`.

Upstream reserves the null block itself now (vllm#47272), so our compat patch for that is deleted; it would have crashed on the new `_pool_bytes_per_block` signature anyway. Model Runner V2 is the default (vllm#53183) and Metal only stayed on V1 because macOS has no Triton, so the plugin sets `VLLM_USE_V2_MODEL_RUNNER=0`. `transformers` moves to 5.10.4, `huggingface_hub` to 1.28.0, and `pyproject` to 0.29.0 per `RELEASE.md`. Nothing else discharges.

`MetalPlatform` resolves after a clean `./install.sh`, lint passes, 2156 non-slow tests pass, both smokes return their goldens, the GGUF serve smoke passes, TurboQuant hybrid and Nemotron-H serves boot, and both spec decode e2e gates pass 7/7. The slow suite's failures are the known in-process memory issue; each passes rerun alone.

`vllm bench serve`, Qwen3-0.6B, sonnet, 100 prompts, `--request-rate inf --temperature 0 --ignore-eos --seed 0`, three sessions per arm, all at 100 requests, 54256 input and 15000 generated tokens. One 0.28.0 session ran slow; against its other two, 0.29.0 is +0.5%. No regression.

| Metric | vLLM 0.28.0 | vLLM 0.29.0 | Delta |
|---|---:|---:|---:|
| Output tok/s | 1040.71 | 1057.80 | +1.6% |
| Total tok/s | 4805.01 | 4883.93 | +1.6% |
| Mean TTFT (ms) | 4559.01 | 4374.38 | -4.0% |
| P99 TTFT (ms) | 7199.94 | 7000.75 | -2.8% |
| Mean TPOT (ms) | 64.65 | 64.39 | -0.4% |
| P99 TPOT (ms) | 79.12 | 78.94 | -0.2% |
| Duration (s) | 14.42 | 14.18 | -1.6% |


vllm_metal/__init__.py
```diff
@@ -127,6 +135,7 @@ def _register() -> str | None:
     _configure_logging()
     _apply_macos_defaults()
     _apply_mlx_buffer_defaults()
+    _apply_model_runner_default()
```
LxYuan0420: Could we move this after `MetalPlatform.is_available()`? I want to make sure we only set `VLLM_USE_V2_MODEL_RUNNER=0` when Metal is actually selected. Also, what should happen if a user sets `VLLM_USE_V2_MODEL_RUNNER=1`, since Metal still uses the V1 runner path?

LxYuan0420 (approved):
This looks close to me. Since this is a vLLM version bump with runtime behavior changes, I would like one more maintainer to take a look before we merge. This should help us avoid missing anything in the migration.

WindChimeRan (commented):
Image

I run benchmark on LFM & Qwen GDN model (they are different attentions). No regression on speed & correctness. 

Non blocking: But comparing prefix caching on versus off produced 16/80 mismatches for GDN and 24/80 for LFM, on both 0.28 and 0.29.  Seems like a pre-existing issue. I guess it's just a numeric noise. Need to rerun the test, after the top-k parity tool merged.

WindChimeRan (commented):
Non blocking: this PR does not fix the pre-existing GDN block double counting issue automatically.

vllm_metal/attention/caches/mha_layout.py
```diff
@@ -14,14 +14,30 @@
...
     SlidingWindowSpec,
 )
 
 NO_SLIDING_WINDOW = -1
+# vLLM's name for the page order Metal stores: [block, token, head, dim].
+KV_CACHE_LAYOUT = KVCacheLayout.LBNHC.name
 StandardMHASpec: TypeAlias = FullAttentionSpec | SlidingWindowSpec
 
 
+def layer_addresses(tensor: KVCacheTensor) -> list[tuple[str, int]]:
```
WindChimeRan, line 28: Non-blocking

def layer_addresses

I think this is a generic function, and should not be placed in a mha specific file ... maybe we should refactor this a bit later

WindChimeRan (approved):
LGTM!

ricky-chaoju:
@WindChimeRan Thanks for running both. The GDN mismatches match what I saw before opening this: the same four div_1B_* cells fail on 0.28.0 and 0.29.0, so I left it out of the bump. to rerun once #705 lands.

Agreed on the double counting, this PR does not touch Metal's accounting. And yes, layer_addresses belongs in a neutral module; I'll move it in a follow-up.

WindChimeRan:
@ricky-chaoju 

I think it's fine. top k new results on main: 

Comparison | Exact | Top-5 match | Failed
-- | -- | -- | --
Caching off vs on | 132 | 28 | 0
Both modes vs native MLX | 268 | 52 | 0

ricky-chaoju:
@WindChimeRan Thanks, good to see it all lands in top-5 with nothing failing. v0.29.0 is published: https://github.com/vllm-project/vllm-metal/releases/tag/v0.29.0

PR #746, opened 2026-09-11
Move KV placement helpers out of the MHA layout

ricky-chaoju:
pointed out on #744 that layer_addresses is about vLLM's KVCacheTensor placement, not MHA, yet it lived in mha_layout.py and the hybrid state-pool derivation imported it from there. The KV_CACHE_LAYOUT name the worker reports had the same problem.

This moves both into attention/caches/placement.py, a small module that only states vLLM's placement contract (the layout the worker reports and the byte address a tensor assigns to each layer). mha_layout.py keeps its translator and cache_policy imports the helper from the new owner.

### DIFF
--- a/tests/stub_runner.py
+++ b/tests/stub_runner.py
@@ -13,7 +13,7 @@ from vllm.config import CacheConfig
 from vllm.v1.attention.backends.utils import record_kv_cache_layout
 
 import vllm_metal.v1.model_runner as mr
-from vllm_metal.attention.caches.mha_layout import KV_CACHE_LAYOUT
+from vllm_metal.attention.caches.placement import KV_CACHE_LAYOUT
 from vllm_metal.attention.runtime.factory import build_hybrid_runtime_plan
 from vllm_metal.attention.runtime.hybrid_plan import (
     ATTENTION_LAYER,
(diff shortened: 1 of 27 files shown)


LxYuan0420 (approved):
LGTM; just need to rebase it first

WindChimeRan (changes requested):
I mean to do a more thorough refactoring. 

e.g., MHA name is misleading (it's also for GQA, MQA). and there is some stale naming from the old representation, TensorLayout --> slot? better doing a more thorough pass on refactoring or renaming, and check how upstream name these.

ricky-chaoju:
Rebased onto main and added the naming pass:

- mha_layout.py and its MHA* classes -> attention_layout.py / Attention* (the path serves MHA/GQA/MQA; upstream just says attention)
- MHATensorLayout / _tensor_layout / tensor_index -> SlotLayout / _slot_layout / slot_index, since slots come from region addresses now, not tensors

ricky-chaoju:
The runtime is now SDPAPagedAttentionRuntime / runtime/sdpa.py since that's what impls/ and the README already call this path. The cache_policy helpers and the mode string dropped the MHA prefix, and the leftover "tensor slot" wording and mixed_mha fixtures are renamed too.

WindChimeRan (approved):
LGTM

PR #747, opened 2026-09-11
Reject quantized --kv-cache-dtype values Metal does not store

hyunnnchoi:
## Purpose

`--kv-cache-dtype fp8` (and `fp8_*`, `int8_per_token_head`, `int4_per_token_head`, `nvfp4*`) is accepted, logs appear as below and starts serving. 

```
[cache.py:335] Using fp8 data type to store kv cache. It reduces the GPU memory footprint and boosts the performance. ...
```

We only reject the `turboquant_*` names, and only because those populate `kv_cache_dtype_skip_layers` and trip the guard at `vllm_metal/platform.py:470`. Everything else falls through and the paged KV cache is still allocated in the model dtype (`vllm_metal/v1/model_lifecycle.py:125`, stored as `runner.kv_cache_dtype` at `:559`). No saving, no warning.

Rejects those values in `check_and_update_config` next to the existing guard, pointing at `--kv-cache-dtype auto` or TurboQuant. `auto`, `float16` and `bfloat16` unchanged.

It's also admittable to make it a warning instead if breaking existing `fp8` invocations matters..  though that keeps the no-op, just louder. Didn't alias `fp8` to TurboQuant since they're different schemes.

## Test Plan

```bash
pytest tests/test_platform.py -q
ruff check . && ruff format --check . && mypy vllm_metal
```

Runtime check with the offline engine: `LLM("mlx-community/Qwen2.5-0.5B-Instruct-4bit", max_model_len=1024, gpu_memory_utilization=0.2, kv_cache_dtype=...)`.

## Test Result

| `kv_cache_dtype` | before | after |
|---|---|---|
| `auto` | 46,401 blocks × 196,608 B | same |
| `fp8` | 46,401 blocks × 196,608 B, plus the log line above | fails at startup |

```
NotImplementedError: vllm-metal does not support --kv-cache-dtype fp8: the paged KV cache is stored in the model dtype. Use --kv-cache-dtype auto, or enable TurboQuant with --additional-config '{"turboquant": true}'.
```

`test_check_and_update_config_rejects_quantized_kv_cache_dtype` (fp8, fp8_e5m2, int8_per_token_head, nvfp4): 4 failed on main, 4 passed here. Non-slow suite: 2160 passed, 15 skipped, 53 deselected. ruff and mypy clean.

Apple M1 Max 64 GB, macOS 15.7.1, vLLM 0.29.0, MLX 0.32.1.

### DIFF
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -501,6 +501,7 @@ class TestMetalPlatform:
             "max_model_len": 32768,
             "hf_config": SimpleNamespace(model_type="qwen3"),
             "is_hybrid": False,
+            "dtype": torch.float16,
         }
         model_fields.update(model or {})
         return self._platform_config(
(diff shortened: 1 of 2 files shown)


LxYuan0420 (approved):
made some edits and thanks for fixing it

Issue #743, opened 2026-09-10
Add Nemotron Nano VL/Omni support to paged serving


TheVerdman:
Following [Ricky’s guidance in #644](https://github.com/vllm-project/vllm-metal/issues/644#issuecomment-5597548115), I’d like to take on the Nemotron Nano VL/Omni adapter work in core vLLM-Metal.

This issue tracks multimodal input support with text output, split across smaller reviewable PRs. I’ll build on the Nemotron-H runtime from #710 and follow the Qwen3-VL adapter boundary over MLX-VLM’s `nemotron_h_nano_omni`.

Ricky suggested starting with Nano VL V2 image input, followed by the Parakeet audio encoder, so I’ll begin there. Video and mixed-modality support can follow once those paths are in place.

## Proposed PR stack

- [ ] **Nano VL V2 image serving**
  Add the model-owned adapter, vision encoding/projection, preprocessing, and embedding placement needed to serve image inputs through the paged text runtime.
- [ ] **Omni audio serving**
  Add the Parakeet audio encoder path, including feature extraction, projection, masks, and audio embedding placement.
- [ ] **Video serving**
  Add video input support once the intended checkpoint and MLX-VLM reference path are confirmed.
- [ ] **Mixed-modality support**
  Validate supported image, audio, and video combinations, including ordering and multi-turn behavior, and document the combinations that are known to work.

These boundaries are provisional. If shared processor or runner changes are needed along the way, I’ll split those into small prerequisite PRs so each change stays focused.

## Validation

Each modality PR will include focused regression coverage and reproducible greedy parity against a pinned direct MLX-VLM reference.

I’ll record the checkpoint revision, relevant runtime revisions, preprocessing settings, prompts, and hardware used for the comparison. I’ll require exact generated-token matches for the pinned parity case before claiming support.

As each path lands, I’ll also cover the relevant embedding accounting, unsupported-input handling, cancellation, and request-isolation cases.

I’ll link the implementation PRs here as they’re opened.

PR #749, opened 2026-09-11
[Multimodal] Adapter hooks for image-capable text backbones

Lavmee:
## Purpose

Three small hooks on the model-adapter contract that a text backbone needs
before it can serve image inputs, plus a `text-only` multimodal serve mode.
Nothing model-specific here — the Gemma 4 vision sidecar that consumes these
hooks follows in a separate PR.

## What changes

- **`supports_per_segment_positions`** — an adapter can opt out of per-segment
  positions when its text backbone wants plain sequential positions instead of
  the mrope-style per-segment layout.
- **`supports_selective_logits`** — an adapter can keep vLLM's selective-logits
  path on a VLM text backbone instead of materializing logits for every row.
- **`PlaceholderRange.is_embed`** is now honoured when splicing multimodal
  embeddings, so a placeholder span that mixes embed and non-embed rows (boi /
  eoi marker tokens around an image block) splices only the embed rows.
- **Encoder profiling** — when an adapter offers `profile_features`, the runner
  runs one encoder pass during profiling so the vision tower's peak allocation
  is part of the memory profile rather than a surprise on the first request.
- **`VLLM_METAL_MULTIMODAL_MODE=text-only`** forces the text-only backbone for
  every multimodal checkpoint. `auto` and `multimodal-native` are unchanged.

## Testing

`pytest -m "not slow" tests/ -q` — 2165 passed, 15 skipped. `ruff check`,
`ruff format --check` and `mypy vllm_metal` are clean. The new behaviour is
covered through stub adapters in `tests/v1/mm/test_paged_forward_mm.py`,
`tests/v1/mm/test_paged_forward_mm_gate.py` and
`tests/test_model_runner_profile.py`, including the empty-`profile_features`
branch. No existing adapter sets any of the new flags, so every current model
keeps its present behaviour.

## What follows

This is the first of four PRs bringing Gemma 4's image inputs to Metal. They
are stacked, so each is opened only once its parent has landed and it has been
rebased on `main` — the sizes below are the incremental ones:

1. **This PR** — the adapter hooks and the `text-only` serve mode (+418, 9
   files).
2. **Vision sidecar** — `vision_tower` and `embed_vision` loaded beside the
   mlx_lm text backbone rather than moving Gemma 4 onto an mlx-vlm composite,
   so MTP drafting, YOCO fast prefill and sliding-window dispatch keep running
   against the same text model (+2577, 20 files).
3. **Bidirectional attention inside image blocks** — HF's
   `create_masks_for_vision_model` semantics, reached by recomputing the image
   rows with MLX SDPA after the paged kernel (+2079, 27 files).
4. **The same mask inside the tiled Metal kernel** — vLLM's `mm_prefix`
   contract behind a function constant, with the recompute kept as a
   switchable reference path (+1269, 14 files).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -7,7 +7,7 @@
 | `VLLM_METAL_MEMORY_FRACTION` | `auto` | Metal memory budget mode; see [KV Cache Memory Settings](#kv-cache-memory-settings) |
 | `VLLM_MLX_DEVICE` | `gpu` | MLX device (`gpu` or `cpu`) |
 | `VLLM_METAL_DISABLE_NAX` | `0` | Emergency override for automatic M5 NAX prefill attention. Set to `1` to force the non-NAX fallback. |
-| `VLLM_METAL_MULTIMODAL_MODE` | `auto` | Multimodal serve mode: `auto` uses the compatibility allowlist; `multimodal-native` disables overrides |
+| `VLLM_METAL_MULTIMODAL_MODE` | `auto` | Multimodal serve mode: `auto` uses the compatibility allowlist (Gemma 4 gets the vision sidecar when its checkpoint allows); `multimodal-native` disables overrides; `text-only` forces the text-only path for every multimodal checkpoint |
 | `VLLM_USE_MODELSCOPE` | `False` | Set True to change model registry to <https://www.modelscope.cn/> |
 | `VLLM_METAL_MODELSCOPE_CACHE` | None | Specify the absolute path of the local model |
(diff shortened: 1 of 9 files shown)


TheVerdman:
I'm working on Nano VL V2 image serving in #743, which needs the `PlaceholderRange.is_embed` handling in this PR. My local branch has an equivalent fix, with regression coverage for marker-only chunks, internal text markers, and embedding slices across chunk boundaries.
I'd like to use this as the shared prerequisite for Nano. If the other changes need more time, would you be open to landing the placement fix separately? I can contribute the additional test cases.

PR #753, opened 2026-09-13
[Bugfix] Anchor install.sh source-checkout mode to the repo root

### LINKED ISSUE #752: [Bug] install.sh fails when run from outside the source checkout: paths resolve against cwd
## Summary

`install.sh` decides it is running from a source checkout by looking next to its own location (`$script_dir/scripts/lib.sh`), but the source-checkout branch then resolves every path against the caller's current directory instead of the repo. Running the script from any directory other than the repo root fails before installing anything.

## Steps to reproduce

1. Clone the repo and check out `main` (tested at `d7e7980`, v0.29.0).
2. Change to any directory that is not the repo root, for example a fresh temp directory:
   ```bash
   cd "$(mktemp -d)"
   ```
3. Run the installer by path:
   ```bash
   bash /path/to/vllm-metal/install.sh
   ```

## Observed

```
=== Creating virtual environment ===
Using CPython 3.12.3 interpreter at: /opt/homebrew/opt/python@3.12/bin/python3.12
Creating virtual environment with seed packages at: .venv-vllm-metal
 + pip==26.2.1
Activate with: source .venv-vllm-metal/bin/activate
Error: Missing vLLM release tag file: .github/vllm-release-tag.commit
```

A stray `.venv-vllm-metal` is left behind in the current directory and nothing is installed.

Running the same `bash install.sh` from inside the repo root gets past this step, so the failure is purely about where the script is invoked from.

## Expected

A source-checkout install behaves the same regardless of the directory it is launched from: it reads the tag file from the repo, installs the repo in editable mode, and creates the venv in a predictable location.

## Cause

In the source-checkout branch of `main()` in `install.sh`:

- `venv="$PWD/.venv-vllm-metal"` creates the venv in the caller's directory
- `read_vllm_release_tag` is called with its default relative path `.github/vllm-release-tag.commit`
- `uv pip install -e .` targets the caller's directory

All three use the caller's cwd, while `script_dir` (which already points at the repo) is only used to locate `scripts/lib.sh`. Even with the tag file check removed, `uv pip install -e .` fails next for the same reason.

## Environment

- macOS 15, Apple M1, 8 GB
- Homebrew Python 3.12.3 (arm64)
- vllm-metal `main` at `d7e7980`

abhijithneilabraham:
Fixes #752

## Problem

`install.sh` detects a source checkout by looking for `scripts/lib.sh` next to its own location, but the source-checkout branch of `main()` then resolves three things against the caller's current directory instead of the repo:

- the venv path (`$PWD/.venv-vllm-metal`)
- the vLLM release tag file (`.github/vllm-release-tag.commit`, read with a relative default)
- the editable install (`uv pip install -e .`)

So running `bash /path/to/vllm-metal/install.sh` from any other directory fails immediately with `Error: Missing vLLM release tag file`, after creating a stray venv in the caller's directory.

## Fix

Hoist `script_dir` out of the inner `if` so it stays in scope, and `cd` to it at the start of the source-checkout branch before creating the venv. Everything downstream (`read_vllm_release_tag`, `uv pip install -e .`, `ensure_metal_toolchain`, `build_native_artifacts`) already assumes the repo root as cwd, so no other changes are needed. The venv now lands at `<repo>/.venv-vllm-metal`, which is the same location as before when running from the repo root.

The `curl | bash` path is unaffected: `local_lib` is empty there and the branch is never taken.

## Verification

Steps from #752: `cd "$(mktemp -d)"`, then `bash /path/to/vllm-metal/install.sh`.

Before this change:

```
=== Creating virtual environment ===
Creating virtual environment with seed packages at: .venv-vllm-metal
Error: Missing vLLM release tag file: .github/vllm-release-tag.commit
```

After this change, from the same kind of directory: the venv is created under the repo, vLLM core installs, the editable install of the repo succeeds, and the script proceeds to `=== Ensuring Metal toolchain ===`. On a machine with Command Line Tools only (no Xcode) it stops there, which is the separate #684 problem and not related to this change.

`bash -n install.sh` passes. I do not have shellcheck installed locally, so I am relying on CI for that.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -216,8 +216,8 @@ EOF
   # Source shared library functions
   # Try local lib.sh first (when running ./install.sh), fall back to remote (when piped from curl)
   local local_lib=""
+  local script_dir=""
   if [[ -n "${BASH_SOURCE[0]:-}" ]]; then
-    local script_dir
     script_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" && pwd)"
     local_lib="$script_dir/scripts/lib.sh"
   fi
@@ -251,7 +251,11 @@ EOF
 
   local venv="$HOME/.venv-vllm-metal"
   if [[ -n "$local_lib" && -f "$local_lib" ]]; then
-    venv="$PWD/.venv-vllm-metal"
+    # Source checkouts read the release tag file, run the editable install,
+    # and build native artifacts relative to the repo root, so anchor the
+    # working directory there instead of the caller's cwd.
+    cd "$script_dir" || exit 1
+    venv="$script_dir/.venv-vllm-metal"
   fi
 
   ensure_venv "$venv"


abhijithneilabraham:
@StevenWang-CY @LxYuan0420

WindChimeRan:
please sign DCO @abhijithneilabraham. Then it's good to merge. 

---

This PR LGTM  as an idiot-agent-proof installation enhancement.

LxYuan0420 (approved):
Thanks for the fix.

Recommit it using `-s` and push with least force and we are good to go

abhijithneilabraham:
@LxYuan0420 DCO is signed, thanks for the guidance

PR #754, opened 2026-09-13
perf: compact physical slot mapping for align-mode GDN state

suntp:
## Symptom

With align-mode prefix caching, GDN state storage can grow to several GiB even when many allocated rows hold no state. This wastes memory during long-context requests and cache churn.

## Cause

GDN and attention KV blocks draw IDs from the same block pool. The old state arrays grow to the largest GDN block ID, so a few scattered IDs can materialize many unused rows. KV-only copy-on-write pairs can also raise that allocation high-water mark.

## Change

Map scheduler block IDs to compact physical state slots. Storage then grows with the number of mapped GDN blocks, rather than the span of their IDs.

- Preserve live state and reusable prefix checkpoints.
- Reclaim a slot when its block ID has been reassigned to a KV group. The block pool has already invalidated its old prefix hash at that point.
- Translate copy-on-write through the mapping, copy each shared state pool once, and fully overwrite a reused slot before reading it.

Startup memory budgeting and the scheduler-visible block limit stay unchanged. Assigning the saved reservation to more KV capacity is the separate, dependent draft [#764](https://github.com/vllm-project/vllm-metal/pull/764).

## Results and validation

On the earlier `a7e82c5`-based branch, a Qwen3.8-27B-4bit 128K workload on a 64 GB host showed:

| GDN state-array capacity | Before | After |
|---|---:|---:|
| Materialized rows | 287 | 96 |
| Capacity | 13.72 GiB | 4.59 GiB |

These are historical array-capacity measurements, not RSS/wired measurements or a rerun of the rebased head.

For the current head `a91b346` against its base `34c0adb`:

- All 40 complete 27B output comparisons match: 20 cases on each cache arm. Both revisions retain the same 16 cache-on/off divergences; this PR does not claim to fix them.
- The 0.8B quick gate passes. Unit coverage includes compact growth, slot reuse, checkpoint survival, shared-pool aliases, and pending writes before copy-on-write.
- Local non-slow tests have the same 11 failures on base and head. The shared environment did not meet all base dependency pins, so this is a controlled branch comparison, not clean-install qualification.

[Rebased output dumps, environment details, comparison and exact test failures](https://gist.github.com/suntp/f3344e2860c85b15aa0fc655b58367dc). The code is unchanged by this description update.


suntp:
Rebased onto `34c0adb` and updated the PR head to `a91b346`. All seven commits carry DCO sign-offs; DCO is green and GitHub now reports the branch as mergeable. The port follows the generalized `AlignStateManager` / `state_*` API, exposes one canonical layer per shared physical state pool, and moves the ShortConv CoW regression through the compact manager path.

Fresh validation:

- Qwen3.8-27B-4bit: final head vs `34c0adb` has identical complete strict-pass outputs for all 20 cases on each cache arm (40 full-output comparisons). Both standalone gates report 16 divergent / zero waived; the previous 2+2 result belongs to the old vLLM 0.28 stack.
- Qwen3.5-0.8B: the final head passes the vLLM 0.29 `--quick` gate, with 20 comparisons and zero divergence.
- Non-slow tests: head 2141 passed / 11 failed / 15 skipped; base 2126 passed / 11 failed / 15 skipped. The exact 11 failure IDs match. Both runs explicitly exclude the ngram module because local Numba rejects NumPy 2.5.2. Local ruff, format, mypy and shellcheck pass.

[New dumps, source/environment provenance, comparison and exact test failures](https://gist.github.com/suntp/f3344e2860c85b15aa0fc655b58367dc). The shared local venv does not yet meet all new main dependency requirements (notably MLX / mlx-lm / mlx-vlm), so this is a controlled branch comparison, not a clean-install qualification or proof of an upstream numerical regression. The PR body now separates this evidence from historical parity and memory measurements. Startup budgeting remains unchanged by design; the investigative APC issue will follow after this PR merges.

WindChimeRan (commented):
The problem seems plausible ... But your language is hard to read. 

Could you try to improve your PR description writing? What's the symptom, what's the root cause, what's the solution, what's the before/after result? Your agent was adding too many irrelevant and distracting details. 

I’m struggling to assess the scope and reasoning behind the change, which makes me hesitant to proceed with the code review.

(Also please mark followup PR as draft if it depends on this.)

suntp:
Thanks for pointing this out. I rewrote the description around the symptom, cause, mapping change, and before/after result, and moved the detailed validation context behind the evidence link. The historical memory measurement and the rebased output comparison are now distinguished explicitly.

No code changed. The dependent follow-up #764 remains a draft.

suntp:
@WindChimeRan I've rewritten the description around the symptom, root cause, solution and before/after results, with the detailed evidence linked separately. The code is unchanged, and the dependent #764 remains a draft. The updated description is ready for your review.

WindChimeRan (commented):
The issue is real. But the code is not very clean to merge. 

---

I need more days to review the design ...

suntp:
@WindChimeRan Thanks for confirming the issue. I agree the lifecycle and scope can be made easier to review.

The main design choice is to keep block-to-slot mapping in `AlignStateManager`, with caches operating on physical slots. Preserving checkpoints across request completion and detecting safe retirement from KV-group IDs requires ownership information across the runner/runtime boundary, which explains several interface changes. Retiring slots before CoW allocation avoids unnecessary growth; pending writes and shared-pool copies also constrain the ordering. That ordering can be expressed more clearly in one place.

The patch grew further because I included large-block parity-harness support and diagnostic tie classification. Those are separable from the mapping change. Looking back, leaving retirement preparation in two entry points and retaining ShortConv's old direct-ID copy helper also makes the final design less clear.

I propose this cleanup order:

1. Remove the legacy ShortConv scheduler-copy helper and stale ID/slot comments, preserving its shared-pool regression coverage.
2. Group each cache group's motion data together, preserving validation and capacity growth before state writes.
3. Consolidate state preparation into one path: pending updates → retirement → state CoW → copy-forward/zero-init, with call-order regressions.
4. Separate the parity-harness expansion and diagnostic tie classification, keeping the direct runtime regressions here and preserving reproducibility of the evidence.

When you've had time to review, could you point out the parts you find hardest to follow, and whether you prefer a different ownership or interface boundary? I can adjust this plan before restructuring the published patch.

vllm_metal/attention/runtime/hybrid.py
WindChimeRan, line 112: vllm-metal's initialize: 
```python
 def initialize(self, num_blocks: int) -> None:
        self._cache = MetalPagedKVCache( ... 
...
        self._state_cache = self._hybrid_plan.family.create_state_cache(
...
```

This is more like vllm-metal's previous tech debt (today we use two storage families, which is a bad design). I think this is the root cause, and we should follow upstream vllm's abstraction, manage kv cache in a unified way, then maybe we don't have to invent ad-hoc ways to solve this issue. 

WDYT?

suntp: @WindChimeRan I agree with this direction. After checking upstream's cache layout and running local feasibility tests with shared Metal storage, I think a unified KV/recurrent-state cache is the better approach. It would address the separate-storage design directly and let upstream block ownership govern both KV and state, removing the need for the extra compact-slot mapping in this PR.

I plan to propose this in a separate PR based on main. Does that approach match what you had in mind? Are there any abstraction boundaries or constraints you would like me to follow?

WindChimeRan: I opened tentative fix #776 #777 #778. 

I think the direction is better than #754, but I haven't check every detail deeply ... 

Maybe we can collaborate on this.

suntp: @WindChimeRan Thanks, I'd be happy to collaborate. I'll build on your #776–#778 stack and take on two areas:

- Shared-state correctness review and focused regression coverage: view aliasing, pending updates, copy/reset, and native fallback completion ordering.
- Qwen3.8-27B continuous-batching validation against main: mixed arrivals, prefill/decode cancellation, batch reordering, and preemption/recovery, checking complete token outputs and memory peaks.

I'll start from #778 at `77240dc` and share reproducible findings and focused patches for confirmed issues. Any temporary-memory budgeting or performance changes can follow the measurements as separate work.

WindChimeRan:
Close as superseded by #778. Thanks for looking into this issue!

PR #756, opened 2026-09-13
Remove VLLM_METAL_MEMORY_FRACTION in favor of --gpu-memory-utilization

### LINKED ISSUE #755: [Bug] KV cache OOM mitigation always names VLLM_METAL_MEMORY_FRACTION, even when --gpu-memory-utilization is the active knob
## Summary

When the paged KV cache does not fit, the startup error always says `increase VLLM_METAL_MEMORY_FRACTION`, even when that variable is unset and the fraction actually in use came from `--gpu-memory-utilization`. The log line printed just before the error already states which source is active, so the two messages contradict each other on the same screen.

## Steps to reproduce

1. Use a machine where a small model does not fit under a low memory fraction (tested on an 8 GB Apple M1).
2. Leave `VLLM_METAL_MEMORY_FRACTION` unset so the default `auto` mode applies.
3. Start a server with a deliberately low `--gpu-memory-utilization`:
   ```bash
   vllm serve mlx-community/Qwen3-0.6B-4bit \
     --gpu-memory-utilization 0.15 \
     --max-model-len 2048 \
     --max-num-seqs 1
   ```

## Observed

The cache planner logs where the fraction came from:

```
INFO [cache_policy.py:1307] Paged attention: VLLM_METAL_MEMORY_FRACTION=auto, using --gpu-memory-utilization=0.15
```

Then the error tells the user to change a different knob:

```
ValueError: Paged attention: not enough Metal memory for KV cache. metal_limit=5.73GB, fraction=0.15, usable_metal=0.86GB, model_memory=0.34GB, overhead=1.42GB, kv_budget=-0.89GB. Mitigations: increase VLLM_METAL_MEMORY_FRACTION; use a smaller or more quantized model.
```

Following the advice literally does work, because a numeric `VLLM_METAL_MEMORY_FRACTION` overrides the flag. But the user only ever typed `--gpu-memory-utilization`, that flag is what produced the number in the error, and the message never mentions it. The `fraction=0.15` in the breakdown gives no hint of its origin either.

## Expected

The mitigation names the knob that is actually in effect:

- in `auto` mode: `increase --gpu-memory-utilization (currently 0.15)`
- when the env var is set: `increase VLLM_METAL_MEMORY_FRACTION (currently 0.15)`

## Cause

In `vllm_metal/v1/cache_policy.py`, `WorkerCachePlanner._memory_fraction` resolves the value through `MetalConfig.effective_memory_fraction` and branches its log line on `metal_config.is_auto_memory`. That boolean is then dropped: `_PagedAttentionPlan` only stores the numeric `fraction`, and `_PagedAttentionPlan.format_mitigations` returns a hardcoded string.

## Environment

- macOS 15, Apple M1, 8 GB
- vllm-metal v0.29.0, main at `fbf41a6`

abhijithneilabraham:
Fixes #755

Per the discussion above, this removes `VLLM_METAL_MEMORY_FRACTION` rather than improving the message that mentioned it. `--gpu-memory-utilization` (`gpu_memory_utilization=` for `LLM()`) becomes the only way to size the Metal KV budget, matching upstream vLLM. Breaking change, no deprecation window, as agreed.

## What changes

- `envs.py`, `config.py`: the variable, the `AUTO_MEMORY_FRACTION` sentinel, `MetalConfig.memory_fraction`, `is_auto_memory` and `effective_memory_fraction` are removed.
- `platform.py`: `get_device_total_memory` / `get_device_available_memory` report raw unified memory (they previously scaled by the env var only when it was numeric). The MB-per-buffer default reads the flag directly.
- `cache_policy.py`: `_memory_fraction` reads `cache_config.gpu_memory_utilization` and logs it. The OOM mitigation reads `increase --gpu-memory-utilization (currently 0.15)`.
- `draft_model_proposer.py`: the scratch-pool exhaustion hint names the flag.
- `tools/check_parity.py` gains `--gpu-memory-utilization`, defaulting to 0.65 (the value the parity workflow has pinned since it was written), and `.github/workflows/parity.yml` passes it explicitly. The default matters for comment-triggered runs: `/ci parity` executes main's workflow file against the PR checkout, so until this merges the flag is not passed and the tool default is what the server sees.
- Tests that constructed `MetalConfig(memory_fraction=...)` drop the field; `docs/configuration.md` and `docs/profiling.md` document the flag.

## Verification

- `pytest -m "not slow" tests/`: 2066 passed, 18 skipped.
- `ruff check .`, `ruff format --check .`, `mypy vllm_metal`: clean.
- On an 8 GB M1 with `--gpu-memory-utilization 0.15` the OOM now reads:
  ```
  Paged attention: using --gpu-memory-utilization=0.15
  ValueError: ... Mitigations: increase --gpu-memory-utilization (currently 0.15); use a smaller or more quantized model.
  ```
  A normal launch at 0.4 serves and answers a chat request.

## Left for a follow-up

Kept this diff to what the removal strictly requires. A stale `VLLM_METAL_MEMORY_FRACTION` export is now ignored rather than rejected. Several tool scripts under `tools/`, the slow e2e tests, and example commands in a few other doc pages still set the variable; those become no-ops. The Gemma4 tools are left alone per #724.


abhijithneilabraham:
@LxYuan0420 @WindChimeRan

WindChimeRan (commented):
I prefer to remove `VLLM_METAL_MEMORY_FRACTION`. 

keeping two knobs that have identical meaning but in different paths is confusing. 

cc @LxYuan0420 , because you reviewed #514

abhijithneilabraham:
Do you think  we should to keep only `gpu-memory-utilization` and remove `VLLM_METAL_MEMORY_FRACTION` ? The only concern I have here is backwards compatibility.

@WindChimeRan

WindChimeRan:
@abhijithneilabraham yes. I think this is the right thing to do. 

* VLLM_METAL_MEMORY_FRACTION is a historic vllm-metal only knob. 
* gpu-memory-utilization aligns with vllm UX

I think breaking change is good. It will simplify the code. Less bugs. Less maintenance burden. More clarity in the doc.

abhijithneilabraham:
Reworked per the discussion above: this PR now removes `VLLM_METAL_MEMORY_FRACTION` entirely rather than fixing the message. The description is updated with the full scope and end-to-end verification. A stale export fails at startup with a message naming `--gpu-memory-utilization`, so nobody silently serves at a different budget.

WindChimeRan (approved):
nit: Need to update `docs/profiling.md` (still say `re-run with VLLM_METAL_MEMORY_FRACTION=0.1...` ) then it LGTM. 

(No need to update Gemma4 tools, We plan to remove those in #724 )

abhijithneilabraham:
@WindChimeRan I skip running some tests myself since I don't have enough hardware (would love to have it sometime). I run the tests in my apple macbook m1.  Hope you will run larger hardware tests in your end.

abhijithneilabraham:
@WindChimeRan

PR #758, opened 2026-09-13
[MLX] Use a zero-copy DLPack tensor bridge

WindChimeRan:
## Background

[DLPack](https://github.com/dmlc/dlpack) is a standard for sharing tensor storage between frameworks. It describes the tensor's dtype, shape, strides, and device, allowing another framework to use the same memory.

[June 22] [MLX #3531](https://github.com/ml-explore/mlx/pull/3531) added zero-copy Metal DLPack sharing between MLX arrays and PyTorch MPS tensors. This support lets us replace vllm-metal's NumPy/frombuffer bridge with direct tensor handoffs.

For vllm-metal, this bridge only needs the zero-copy path. Both directions use `copy=False`: share the existing allocation, or raise if sharing is unsupported. The bridge should not silently copy tensors to make an import or export succeed.

I also contributed [MLX #4497](https://github.com/ml-explore/mlx/pull/4497), which fixes a hang when copying Metal DLPack inputs by releasing the GIL while the copy completes. That fix applies to the copy path; vllm-metal's zero-copy bridge does not enter it.

## Summary

Replace the NumPy/frombuffer bridge with DLPack. Tensor handoffs share storage or raise when sharing is unsupported. The bridge performs no automatic data copies or device transfers.

## API

```python
from vllm_metal.pytorch_backend.tensor_bridge import (
    mlx_to_torch,
    torch_to_mlx,
)

mlx_array = torch_to_mlx(torch_tensor)
torch_cpu = mlx_to_torch(mlx_array, device="cpu")
```

Both directions request `copy=False`. CPU exports select CPU storage directly, avoiding an MPS import followed by `.cpu()`.

Current runtime exports send FP32 logits, embeddings, and scores to CPU. Imports cover encoder weights, image/audio features, integer image grids, and CPU grammar-mask rows.

## Sharing contract

- Preserve strides and storage offsets.
- Synchronize Torch MPS writes before MLX reads. MLX export evaluates pending work.
- Reject negative-stride exports before Torch can abort, and reject export devices other than CPU/MPS.
- Let callers request independent writable storage explicitly when their operation needs it.

Remove automatic contiguity conversions, unused synchronization helpers, and the copy-based MPS size threshold. Model loading now reads the dtype mapping directly instead of converting empty tensors.

## MLX compatibility

Validated with the pinned MLX 0.32.1 and PyTorch 2.13.0. No MLX version bump is required for this zero-copy bridge.

## Behavior change

* **Empty CPU tensors now raise** (e.g., torch.zeros(0, 4)).  our runtime paths do not require empty imports, so no compatibility handling is needed.
* **Float64 inputs now raise instead of silently downcasting to FP32.** if anyone need it, they should explicit cast the dtype. 

## Performance

No regression on the sampling path

3 interleaved runs per arm (main@916b97c vs PR@f1ea8b4), `--temperature 1.0 --top-p 0.9 --ignore-eos`:

| Metric           | main  | PR #758 | Δ     | run-to-run spread |
| ---------------- | ----: | ------: | ----: | ----------------- |
| Output tok/s     | 293.6 | 294.4   | +0.3% | ±0.9% / ±2.1%     |
| Mean TPOT (ms)   | 94.9  | 94.7    | −0.1% | ±0.2%             |
| Median TTFT (ms) | 449.1 | 452.7   | +0.8% | ±2.9%             |

Isolated microbenchmark says the bridge call itself got *cheaper* : 22.4 µs → 0.97 µs per export (main's `mx.contiguous` + eval dance dominated) — but that's ~0.01% of a step, hence invisible end to end.

Greedy control on main: **770 tok/s / 32.6 ms TPOT** vs 294 / 94.9 sampling. So the sampling path is still very slow.


LxYuan0420 (approved):
Great work!

PR #760, opened 2026-09-14
Add Homebrew installation and stable release updates

WindChimeRan:
Adds Homebrew installation for **stable releases**, with automatic formula-update PRs when a new stable release is published.

Users can run `vllm` without activating a Python environment:

```bash
brew tap vllm-project/vllm-metal https://github.com/vllm-project/vllm-metal
brew install vllm-project/vllm-metal/vllm-metal
```

The updater reads the latest stable release and its paired vLLM metadata, obtains the wheel checksums from GitHub, and creates or updates a formula-only PR. Dev releases are ignored; manual dispatch can retry the latest stable update. It opens the PR with a short-lived token from the `vllm-metal` GitHub App (see below) and leaves review/merge to maintainers; the bot cannot approve or merge.

## Maintainer UX

Our formula pins release URLs and SHA-256 checksums, so each stable release needs a formula update.

1. [Maintainer] Follow the stable release process: cut the release branch, build a draft release, then review and publish it. #652
2. **[Auto] The workflow fetches the wheel checksums from GitHub and opens or updates the Homebrew formula PR as `vllm-metal[bot]`. CI runs on it normally.**
3. **[Maintainer] A maintainer reviews the five-line diff and merges it.**

## Why a GitHub App

The organization's Actions policy does not let `GITHUB_TOKEN` open pull requests, and repository admins cannot override it. PRs opened by `GITHUB_TOKEN` would also never trigger CI. A GitHub App token has neither limitation.

- App: https://github.com/apps/vllm-metal, owned by @WindChimeRan (ownership can be transferred to the org later). Permissions: Contents and Pull requests, read & write; no organization permissions.
- Installed on this repository by @mgoin. The private key is the `HOMEBREW_APP_PRIVATE_KEY` repository secret; the App ID is public and hardcoded in the workflow.
- The token is minted per run by `actions/create-github-app-token` and expires after one hour. `GITHUB_TOKEN` is read-only in this workflow.

## Validation

Validated locally with Homebrew installation and native-kernel tests, Qwen3-0.6B serving, and updater dry runs. After switching to the App token, the updater script was re-run locally against the current stable release (v0.29.0) and produced no diff, as expected. The first end-to-end run happens on the next stable release; `workflow_dispatch` on `main` exercises the token step before then.


.github/workflows/update-homebrew.yml
```diff
@@ -0,0 +1,87 @@
...
+          import os
+          import re
+          import subprocess
+          from pathlib import Path
+
+          def api(path):
+              return json.loads(subprocess.check_output(["gh", "api", path]))
+
+          def wheel(release, name):
+              asset = next(a for a in release["assets"] if a["name"] == name)
```
LxYuan0420, line 41: Small robustness question: could we raise a clear error if the expected wheel asset is missing? `StopIteration` would make the updater failure harder to diagnose.

LxYuan0420 (approved):

PR #766, opened 2026-09-14
[Feature] Add DSpark speculative decoding

StevenWang-CY:
## Purpose

Add DSpark (DeepSeek's DeepSpec drafter) as a fourth speculative decoding method
on the Metal paged-attention path, next to MTP, draft model and N-gram.

DSpark proposes a whole block of tokens from the target's own hidden states: a
small parallel backbone cross-attends over the fused residuals of a few target
layers, predicts a `block_size`-token block (7 in the published checkpoints) in
one pass, and a rank-256 Markov head corrects each position on the token before
it. The target then verifies the block exactly as it does for the other Metal
methods, so greedy output is unchanged. A drafter is trained for one target, so
this PR ships the Qwen3 path with the matched standalone drafters from the
[official DeepSpec release](https://github.com/deepseek-ai/DeepSpec):

| Trained target | DSpark drafter |
|---|---|
| `Qwen/Qwen3-4B` | `deepseek-ai/dspark_qwen3_4b_block7` |
| `Qwen/Qwen3-8B` | `deepseek-ai/dspark_qwen3_8b_block7` |
| `Qwen/Qwen3-14B` | `deepseek-ai/dspark_qwen3_14b_block7` |

Upstream vLLM implements DSpark only in its GPU V2 model runner, which needs
Triton and so is unavailable on Metal; vllm-metal gets its own proposer on the
V1 paged-attention runner.

## This is PR 1 of a series

The complete implementation is not a reviewable single change, so it is being
sent as a series. **This PR is the first part, split purely for the convenience
of review.** The cut is the self-contained foundation: the drafter, the target
hidden-state capture, the per-request context, the greedy verification path,
and the memory accounting that keeps the drafter from competing with the target
KV cache. Each follow-up extends this foundation; none rewrites it.

The whole series is on
[`dspark-pr-final`](https://github.com/mhdimo/vllm-metal/commits/dspark-pr-final/),
rebased on current `main`. Its first commit is this PR; the three that follow
are the follow-up PRs, in order, each green on its own:

| # | Commit | Adds | Tests |
|---|---|---|---|
| 1 | [`fe15c42`](https://github.com/mhdimo/vllm-metal/commit/fe15c42986e092622284a004dd40a0435f5f362a) — **this PR** | Drafter, hidden-state capture, per-request context, greedy verification, memory plan | 2,357 |
| 2 | [`362f508`](https://github.com/mhdimo/vllm-metal/commit/362f508ec0d80711b2b58eb788efbe223fd821e8) | Sampled-request drafting by exact rejection sampling; admission caps | 2,408 |
| 3 | [`59d571a`](https://github.com/mhdimo/vllm-metal/commit/59d571adccf161cfa50acd65b9dcb8d98a4b5c6d) | Confidence calibration, measured cost model, adaptive draft length and bypass | 2,454 |
| 4 | [`5188220`](https://github.com/mhdimo/vllm-metal/commit/51882205455bae825b9c4c2cf34e8a6e87e304f7) | Context arena (per-row attention, batched ingest), async scheduling, load regime | 2,474 |

### Where the series ends up

Stating this here so the foundation reviewed below is judged against what it
carries. All figures are measured on an Apple M5 Max (48 GB) with
`mlx-community/Qwen3-4B-4bit` and `deepseek-ai/dspark_qwen3_4b_block7`.

**Serving, against a target-only server with identical flags.** Paired A/B: the
two servers alternate, five repeats per cell, 95% CI by bootstrap, and a pair is
discarded and retried if another process loads the machine while it runs.

| Prompt × output | Clients | Fixed `K=2` | Fixed `K=7` | **`adaptive`** |
|---|---|---|---|---|
| 128 × 128 | 1 | +41.1% | +5.7% | **+50.1%** [41.5, 57.0] |
| 128 × 128 | 8 | −6.6% | −11.8% | **−5.9%** [−28.0, 38.0] |
| 1024 × 128 | 1 | +19.2% | −18.1% | **+6.2%** [−27.3, 33.4] |
| 1024 × 128 | 8 | −14.1% | −30.4% | **−6.3%** [−20.6, 3.3] |

The point is the bounded downside. A fixed width has to be tuned per workload
and still loses 14–30% at eight clients; the adaptive planner needs no tuning
and its worst cell is −6.3%, because it stops drafting when verification will
not repay it — across the 128 × 128 cells at four, eight and sixteen clients it
issues 12, 26 and 28 draft tokens in total, against 24,063 for fixed `K=7`.
Where drafting does pay it keeps drafting: 128 × 512 is **+28.3%** at one client
and −1.3% at eight.

**Single stream, against the standalone runtime.** The same three prompts through
`mlx-dspark`, the reference standalone MLX implementation of this drafter, and
through this port — each cell is that runtime's own target-only rate → its DSpark
rate:

| Prompt | `mlx-dspark` | **This port** |
|---|---|---|
| chat | 181.3 → 212.4 tok/s (1.17×) | 142.8 → 217.6 tok/s (**1.52×**) |
| code | 182.8 → 205.9 tok/s (1.13×) | 142.3 → 311.2 tok/s (**2.19×**) |
| math | 182.1 → 209.4 tok/s (1.15×) | 138.5 → 317.0 tok/s (**2.29×**) |

The paged target starts slower than the standalone MLX one and ends faster.

**Accepted length per drafting round**, `T=1.0`, 64 prompts per set, against the
DSpark paper's Table 1 — the check that the port is faithful, not merely fast:

| Evaluation set | This port | Paper | Acceptance rate |
|---|---|---|---|
| gsm8k | 6.12 | 6.11 | 73.2% |
| math500 | 5.69 | 5.70 | 67.2% |
| humaneval | 5.11 | 5.38 | 58.8% |
| mbpp | 4.81 | 5.13 | 54.4% |
| mt-bench | 3.54 | 3.64 | 36.5% |
| alpaca | 3.40 | 3.54 | 34.3% |

This PR ships fixed-width greedy drafting only; the numbers above need commits
2–4. Its own measurements are under [Validation](#validation-on-this-commit).

## What changes

- `vllm_metal/v1/hidden_state_tap.py`: captures the requested Qwen3 layer
  residuals from the native mlx-lm body in the same forward that produces the
  target logits, without patching the live model — a shallow module copy owns
  the observer list, so a raising forward cannot leave the target instrumented.
  Wired through `target_forward` and the projection-free intermediate forward
  via `capture_layer_ids` (`proposer.py`, `model_adapter.py`, `model_runner.py`).
- `vllm_metal/v1/dspark/`: drafter config (validated `config.json` contract),
  MLX drafter model, streaming checkpoint loader (shard, header, shape, dtype
  and non-finite checks, one tensor at a time into the 4-bit conversion inside
  a memory budget, allocator limit restored on every path), the memory plan,
  and the supported-pair contract: a `Qwen3ForCausalLM` target with a
  `Qwen3DSparkModel` drafter of matching hidden size, vocabulary and layer
  count, `1 <= num_speculative_tokens <= block_size`, and explicit rejection of
  LoRA, tensor parallelism and every speculative option this proposer does not
  implement.
- `vllm_metal/v1/dspark_proposer.py`: the per-request drafter context. Every
  scheduled prefill and verification span is ingested at its absolute position,
  including prefill chunks that sample nothing; a rollback trims each physical
  layer back to the committed prefix; a scheduler recompute from zero rebuilds
  the context; the runner's existing invalidation path releases contexts on
  finish, cancellation and preemption before any request-id reuse. Eligible
  requests draft together in one batched backbone pass. A request whose target
  features are unavailable falls back to target-only generation rather than
  drafting from an incomplete context, and an allocation failure while drafting
  releases draft state and keeps the target's output.
- `vllm_metal/v1/model_runner.py`, `cache_policy.py`: the drafter is
  materialized before the target profiling forward, and its context, capture
  and workspace reservation is subtracted before the target KV cache is sized
  (`dspark_context_and_workspace` in the paged-attention plan line). A
  configuration that cannot fit fails at startup naming the limits to lower.
- `vllm_metal/platform.py`: vLLM 0.29 rejects `method="dspark"` on the V1
  runner inside `VllmConfig.__post_init__`, which runs after the platform hook.
  The hook validates the pair and then presents it to vLLM as the
  `draft_model` method; the runner and the cache planner recognise the drafter
  under either name through `is_dspark_config` and check it before the
  autoregressive draft-model path, so no scheduler KV group is created for it.
  `parallel_drafting` stays set, so the scheduler keeps booking
  `num_speculative_tokens` draft slots per request. No environment variable is
  needed: the plugin already pins the V1 runner.
- `tools/check_sd_lossless.py`: a greedy target-only versus speculative
  comparison usable by every Metal method. It classifies each first divergence
  against the base engine's top-K logprobs (`EXACT` / `TOP_K_MATCH` / `FAIL`,
  the vocabulary `check_parity.py` already uses), and can run a target-only
  control at a second sequence limit, so a divergence the target reproduces on
  its own is attributed to the target rather than to the verifier. It also
  gates its own verdict on the spec-decode draft counter: a run whose engine
  never drafted is reported `INCONCLUSIVE` (exit 2), not a pass, because a
  target-only fallback emits the target's own tokens and would otherwise match
  exactly without verification ever running.
- Docs: the DSpark section of `docs/speculative_decoding.md` and a
  `docs/tools.md` entry for the losslessness check.
- Tests: 8 new files (contracts, loader, memory plan, proposer lifecycle,
  target capture, hidden-state tap, audit regressions, losslessness tool)
  totalling 70 test functions and 179 cases, and 4 updated. No downloads: the
  capture-parity tests build a tiny seeded Qwen3 and compare native MLX against
  paged attention directly.

`config.py`, `model.py` and `loader.py` derive from the MIT-licensed
[ARahim3/mlx-dspark](https://github.com/ARahim3/mlx-dspark) and keep its header
alongside a `NOTICE` recording the provenance; everything else is Apache-2.0
like the rest of the repository. The drafter model keeps its upstream Gemma4
family branch so the vendored files stay close to their source, but only Qwen3
is enabled for serving.

## Usage

```bash
vllm serve mlx-community/Qwen3-4B-4bit \
  --revision 4dcb3d101c2a062e5c1d4bb173588c54ea6c4d25 \
  --host 127.0.0.1 --port 8000 \
  --gpu-memory-utilization 0.3 --max-model-len 4096 --max-num-seqs 4 \
  --generation-config vllm \
  --speculative-config '{"method":"dspark","model":"deepseek-ai/dspark_qwen3_4b_block7","revision":"3457dff1417cb84927f6098a5fcb7cee85c934b7","num_speculative_tokens":4}'
```

```bash
curl -sS http://127.0.0.1:8000/v1/completions \
  -d '{"model":"mlx-community/Qwen3-4B-4bit","prompt":"The capital of France is","temperature":0,"max_tokens":32,"return_token_ids":true}'
```

The drafter may be a Hugging Face repo id or a local path. Startup rejects an
unmatched pair, an unimplemented drafter option or insufficient capacity with
an explicit error; sampled requests, and requests whose target features are
unavailable, run target-only. The server log shows `DSpark drafter loaded for
speculative decoding: ... (block_size=7, target_layer_ids=[...])` and the
periodic `SpecDecoding metrics` line reports live acceptance.

## Validation on this commit

Apple M5 Max (48 GB), macOS 26.6, Python 3.12.12, vLLM 0.29.0, MLX 0.32.1,
mlx-lm 0.32.0, mlx-vlm 0.6.17, from a clean checkout, with
`mlx-community/Qwen3-4B-4bit` + `dspark_qwen3_4b_block7` at the revisions above.

### Lint, tests and docs

`pytest -m "not slow" tests/ .github/scripts/test_parity.py`: **2,357 passed**,
15 skipped, 53 slow tests deselected. `ruff check .`, `ruff format --check .`,
`mypy vllm_metal` (147 modules) and the strict documentation build all pass.
No shell script changes.

### Serving

The serve command above is healthy in 10 s. The drafter is materialized before
KV sizing (`dspark_context_and_workspace=1.94GB` in the paged-attention plan
line), the platform downgrades async scheduling as it does for the other
methods, and the curl returns `" Paris. The capital of Belgium is Brussels.
..."` (32 tokens) at a mean acceptance length of 3.44.

### Greedy losslessness

`tools/check_sd_lossless.py`, 12 prompts × 64 greedy tokens, each first
divergence classified against the base engine's top-2 logprobs:

| Run | EXACT | TOP_2_MATCH | Outside top-2 |
|---|---|---|---|
| DSpark K=4, one prompt at a time | 5 | 7 | 0 |
| DSpark K=7, one prompt at a time | 5 | 7 | 0 |
| DSpark K=2, batch of 12 | 10 | 2 | 0 |
| DSpark K=7, batch of 12 | 6 | 5 | 1 |
| N-gram K=7, batch of 12 (control) | 9 | 3 | 0 |

Every TOP_2_MATCH is a tie within one bfloat16 ULP in the target's own logits,
a gap of at most 0.125 nats. The control settles what those ties are: the
*target-only* engine served one prompt at a time differs from its own
12-prompt batch on the same 7 prompts, so the flips are the target's
batch-shape arithmetic rather than the verifier, and an unrelated drafter
(N-gram) on the same target shows the same class. The single K=7 outlier is
the target's runner-up by one ULP served single-row, and the target's *own*
argmax under 16-token prefill chunks. No divergence is attributable to the
drafter or to verification.

### Throughput

`vllm bench serve --temperature 0` against a target-only server with identical
flags, `--max-num-seqs 4`. Only greedy requests are drafted, so a sampled
benchmark would measure an idle drafter; the `/metrics` spec-decode counters
confirm every DSpark cell actually drafted.

Natural short prompts, 10 each from the DeepSpec gsm8k, humaneval, mbpp and
alpaca sets, chat template, up to 256 output tokens:

| Server | Clients | Output tok/s | Median TPOT | Accepted per round |
|---|---|---|---|---|
| target-only | 1 | 153.1 | 6.4 ms | – |
| DSpark K=4 | 1 | **178.7 (+17%)** | 5.1 ms | 1.79 of 4 |
| DSpark K=7 | 1 | 149.9 (−2%) | 6.5 ms | 2.05 of 7 |
| target-only | 4 | 463.3 | 8.5 ms | – |
| DSpark K=4 | 4 | 374.3 (−19%) | 9.7 ms | 1.39 of 4 |
| DSpark K=7 | 4 | 358.5 (−23%) | 9.6 ms | 1.53 of 7 |

The `sonnet` benchmark from CONTRIBUTING, ≈550-token prompts sharing a
200-token prefix and 150 output tokens, with prefix caching disabled on both
servers so the shared prefix does not turn the requests target-only:

| Server | Clients | Output tok/s | Median TPOT | Accepted per round |
|---|---|---|---|---|
| target-only | 1 | 138.1 | 6.5 ms | – |
| DSpark K=4 | 1 | **159.4 (+15%)** | 5.7 ms | 1.66 of 4 |
| target-only | 4 | 327.0 | 9.5 ms | – |
| DSpark K=4 | 4 | 248.5 (−24%) | 14.3 ms | 1.62 of 4 |

At this stage of the series, DSpark is a single-stream win at `K=4` and a loss
once several requests share a step: verifying `K+1` rows per request costs more
than the accepted tokens repay, and a wider block adds verification rows faster
than accepted tokens. Commit 3 of the series removes that trade-off by deciding
it per step. The documentation here says exactly what this commit does, the
method is opt-in, and no existing path changes.

## Limitations

- Greedy requests only; sampled requests run target-only, as for every Metal
  speculative method. Synchronous scheduling, single GPU, no LoRA.
- Matched target/drafter pairs only, and only the 4B pair is measured. A
  quantized target uses the drafter trained for its bf16 original; the drafter
  itself is converted to the MLX 4-bit recipe at load. Other quantized
  combinations need their own qualification.
- A request whose prompt hits the target prefix cache is never drafted, because
  its hidden states were never captured; pass `--no-enable-prefix-caching` for
  workloads with long shared prefixes. With caching left on, the sonnet
  workload above runs 11–13% slower than target-only, the idle drafter's cost.
- Greedy output is exact up to the target's own numerical stability, as
  measured above. Degenerate repetitive prompts can flip between near-tied
  tokens on any execution path, speculative or not.

Co-authored-by: Mihal Dimo <mihal@kakao.com>
Signed-off-by: Mihal Dimo <mihal@kakao.com>
Signed-off-by: Chuyue Wang <stevenwang0805@outlook.com>


StevenWang-CY:
@LxYuan0420 would you have time to look at this one?

It adds DSpark as a fourth speculative decoding method on the Metal V1 paged
path, alongside MTP, draft model and N-gram. The method is opt-in behind
`--speculative-config` and no existing path changes.

The full implementation was too large to review in one piece, so it is split
into four commits. This PR is the first: the self-contained foundation — the
drafter, the target hidden-state capture, the per-request context, greedy
verification, and the memory accounting that keeps the drafter from competing
with the target KV cache. The other three are finished and each green on its
own; they are on
[`dspark-pr-final`](https://github.com/mhdimo/vllm-metal/commits/dspark-pr-final/)
(first commit there is this one) if it helps to see where the series lands.

The description has the measurements for both: what this commit does on its
own, and what the finished series reaches. CI is green here — lint, test and
DCO.

Happy to move the boundary if you would prefer a different cut, or to answer
anything in more detail.

tools/check_sd_lossless.py
```diff
@@ -0,0 +1,302 @@
...
+            ("base", base_result),
+            ("sd", sd_result),
+            ("control", control),
+        ):
+            if payload is not None:
+                with open(os.path.join(args.dump, f"{name}.json"), "w") as f:
+                    json.dump({"args": vars(args), **payload}, f)
+    base, sd = base_result["tokens"], sd_result["tokens"]
+    print(f"[stats] base: {json.dumps(base_result['stats'])}")
+    print(f"[stats] sd:   {json.dumps(sd_result['stats'])}")
```
HanningLin, line 260: Could we check that the SD run generated at least one draft token before reporting PASS? Currently, a target-only fallback can produce identical outputs and pass > without exercising speculative verification. I reproduced this with mocked results containing identical tokens and zero draft-token counters. Missing metrics or zero drafting should be reported as inconclusive.

vllm_metal/v1/dspark/model.py
```diff
@@ -0,0 +1,470 @@
...
+    ) -> tuple[mx.array, mx.array]:
+        """Temperature draft for speculative *sampling*: sample each block position from
+        its (temperature-scaled, optionally top-p/top-k truncated) distribution and return
+        ``(tokens [k], probs [k, V])``. ``probs[i]`` is the draft distribution q_i that token
+        i was sampled from — the verifier needs it for the accept test ``min(1, p_i/q_i)`` and
+        residual resampling. Truncating q here (same top-p/top-k as the target) keeps
+        acceptance from collapsing when a client asks for nucleus sampling; losslessness comes
+        from the *target* side (see ``_spec_sample_accept``). Sequential because the Markov
+        bias for position i depends on the token sampled at i-1."""
+        from .sampling import sample_probs, truncate_probs
```
HanningLin: Could we remove sample_block_probs() from this PR? Sampled drafting is outside the supported scope, this method has no callers, and the .sampling module it imports is not included. We can add it alongside the sampled-drafting implementation.

tools/check_sd_lossless.py
StevenWang-CY: Confirmed, and fixed in fe15c42. You were right about the mechanism: the verdict only looked at `failed`, so a target-only fallback gave `N exact, 0 failed` and a PASS that said nothing about verification.

It now reads the draft counter from the SD engine's own metrics and reports `INCONCLUSIVE` (exit 2) when that is zero or missing. `tests/test_check_sd_lossless.py` covers your case and the no-metrics one; both fail against the unfixed tool.

Worth flagging: my first cut matched the wrong metric key and would have reported INCONCLUSIVE on every real run — the mocked test passed it happily. Running the tool for real caught it:

```
PASS: 3 prompts x 24 greedy tokens, method=dspark, K=4, max_num_seqs=1: 3 exact, 0 top-2 near-tie, 0 failed, 116 draft tokens
```

Chasing the same class of problem across the rest of the diff turned up one more in this tool: `--method dflash` was offered and documented, but vllm-metal has no dflash proposer, so that documented invocation could only fail at engine start. Dropped from the choices and the docs.

vllm_metal/v1/dspark/model.py
StevenWang-CY: Agreed — removed in fe15c42. All three points check out, and calling it here would have raised `ModuleNotFoundError`, since `.sampling` only arrives with the sampled-drafting commit.

It also cleared up churn I had missed: the later commits already drop the method, so its deletion was showing up in the next commit's diff. `model.py` is now unchanged between commits 1 and 2.

Your point applies to `sample_block()` just above it as well, so I removed that too. It is uncalled across the whole series, and the proposer already runs the same Markov-biased greedy block loop batched across requests (`markov_head.step_bias` + `argmax`), so the single-sequence copy was redundant rather than merely unused. I checked the removal against dynamic reach (no dispatch table, registry, or subclass; `mlx.nn.Module` enumerates the instance dict, so class methods are invisible to `parameters()`/`save_weights()`) and confirmed the suites are unchanged with it gone, while deleting `compute_logits` under the same probe breaks 41 tests — so the check discriminates.

`confidence_logits()` next to them stays: it is unused in this commit but the adaptive planner calls it two commits later, and the `ConfidenceHead` it drives is an `nn.Module` member whose weights load from the checkpoint.

WindChimeRan (commented):
I would not start with DSpark.

1. Adaptive verification is super difficult to get right, and I don’t think it’s settled upstream yet. Remove the Markov and confidence heads, and it’s basically DFlash.
2. DSpark is heavy (a 5-layer drafter). I’m not convinced the gain justifies that cost. Ideally, we should train a 3-layer DFlash with Speculators first.
3. Your e2e gain is narrow: 15–17% faster with one client, but 19–24% slower with four. upstream vllm's dspark is a universal win across batch sizes.

StevenWang-CY:
Thanks. Taking the points in turn.

**Point 3, the e2e gain.** The table you quoted is PR 1: fixed width, which is all this commit does. The completed series adds a planner that decides per step whether drafting pays and stops when it does not. At one client that is +50% on short prompts; from four clients up it holds parity instead of losing 19-24%. Numbers and method in the next comment.

I would not claim a universal win. Upstream's DSpark runs on the GPU V2 runner with Triton; on the Metal V1 paged runner the extra verify rows cost relatively more, which is why bypassing earns its place here.

**Adaptive verification.** Agreed it is unsettled, and this does not implement it: `enable_adaptive_verification` sits in the contract's rejected-options list and startup fails if it is set. What adapts here is draft length per request plus bypass, driven by the drafter's calibrated confidence and a measured cost model. Both artifacts are absent by default, so out of the box it serves at a plain fixed width.

**Markov and confidence heads.** Fair that without them the shape is close to DFlash. The Markov head is what reaches the published acceptance: on the DeepSpec sets at `T=1.0` this port matches the paper's Table 1 (gsm8k 6.12 vs 6.11, math500 5.69 vs 5.70, humaneval 5.11 vs 5.38). The confidence head is what the planner reads.

**Heaviness.** You are right: 5 layers, 2.8 GB bf16, converted to 4-bit at load, reserving 1.94 GB before the target KV cache is sized so the cost appears in the plan line. Whether that trade is worth it for a 4B target is a fair call to make.

**Starting with DFlash.** Your call on ordering. Most of this PR is not DSpark-specific: the hidden-state tap, the per-request context and the memory accounting are what any target-hidden-state drafter needs. If you would rather land that foundation with a 3-layer DFlash drafter on top, I can reshape it that way; the drafter-specific code is confined to `vllm_metal/v1/dspark/`.

StevenWang-CY:
Measurements for the completed series, since the description carries only PR 1's.

Adaptive against target-only, paired: servers alternate, five repetitions per cell, bootstrap intervals, pairs discarded under outside load. Two values where the cell was measured in two independent campaigns.

| Prompt × output | 1 client | 4 | 8 | 16 |
|---|---|---|---|---|
| 128 × 128 | **+50.1%** / +51.5% | +5.5% / +0.9% | −0.8% / +0.2% | −1.8% / +1.1% |
| 128 × 512 | **+28.3%** / +23.3% | +1.2% / +2.9% | −1.8% / +3.1% | +0.7% / −0.5% |
| 1024 × 128 | +6.2% / +9.3% | −3.4% / −4.7% | −14.5% / −13.4% | −12.4% / −12.3% |

From four clients up the planner bypasses (0 accepted of 0-2 drafted), so those cells are target-only work plus bookkeeping and every interval straddles zero. The 1,024-token rows at 8 and 16 clients are still down 12-14%; the record flags that eight-client cell as order-dependent and open.

Against the asynchronous target-only server, which is the production default and about 13% faster on its own, 128 × 128 at one client is +31.4% [29.5, 34.7]. A fixed width goes higher still there: +72.5% [61.8, 74.4] at K=4. The planner trades part of that for the higher-concurrency behaviour.

Four commits because it took four steps:

| Stage | 4-16 clients |
|---|---|
| Fixed width (this PR) | −4% to −28% |
| + calibrated planner and bypass | −6% to −20% |
| + context arena, target KV budget fix | −1% to −9% |
| + asynchronous scheduling | −2.5% to −4.3% |
| + load regime | +0.9% to +5.5% on 128 × 128 |

Correctness and stability:

- Greedy output is token-identical up to the target's own numerical stability; every divergence is a tie within one bfloat16 ULP that the target-only engine reproduces against itself.
- Accepted length matches the paper's Table 1 on a 4-bit target: gsm8k 6.12 vs 6.11, math500 5.69 vs 5.70, mt-bench 3.54 vs 3.64.
- Serving gates pass in fixed, adaptive and bypass modes, plus the stochastic gate.
- Ten-minute soak at eight clients: 1,507 requests, zero errors, resident set flat.

Single stream against `mlx-dspark` on the same three prompts: 1.52× / 2.19× / 2.29× on chat / code / math, where that runtime reaches 1.13-1.17× against its own baseline.

WindChimeRan:
@StevenWang-CY 

An agent built runnable prototype can still leave architectural questions unresolved. I think we should first make sure we understand the core challenges, break them down, and have each PR address one challenge. This should make vllm-metal easier to maintain and support more model × drafter combinations.

For example:
1. How are you handling drafter attention? Are you using MLX’s attention or vllm-metal’s attention?
2. How are you managing the drafter’s KV cache? You manage a separate per-request drafter context cache, this is not an ideal design.  
3. How are you extracting hidden states from the target model when mlx-lm doesn’t expose an API for this? Should we contribute to upstream mlx_lm for this? 

I saw many code agent reward hacking in the design. Coding agent tends to make things look easy.

WindChimeRan (commented):
@StevenWang-CY 

For exp: 

* **correctness**: on/off speculative decoding, under greedy sampling, output should be the same or under top-k
* **e2e speedup** DSpark/DFlash should be show e2e speedup than 1. target model only 2. draft-model speculative decoding (Qwen3-8B + 0.6B) 3. under both batch 1 and batch N (because these are the paper's claim, and empirically verified by upstream vllm)

If it's slower than draft-model or even slower than non-speculative decoding, then it's a strong indicator of a flawed design. 

---

Above is just for exp result criteria. Achieving this goal doesn't mean I'll give a green light to the design.

PR #768, opened 2026-09-15
[Bugfix] Fix OLMo 2 full-projection Q/K norms

### LINKED ISSUE #767: [Bug] OLMo 2 paged attention applies full-projection Q/K norms after splitting heads
## Problem

`allenai/OLMo-2-0425-1B-Instruct` loads and the server becomes healthy, but the first paged completion returns HTTP 500 and EngineCore fails with:

```text
ValueError: [rms_norm] (*weight) must have the same size as the last dimension of x but has 2048 elements.
```

The checkpoint has 16 attention heads with `head_dim=128`. Its Q/K RMSNorm weights each cover the complete 2048-dimensional projection. The default paged SDPA path applies those norms after splitting the projections into heads, when the last dimension is only 128.

## Reproduction

On unmodified vllm-metal `bc9bd16`, in the project's development environment:

```bash
MODEL_DIR="$(hf download allenai/OLMo-2-0425-1B-Instruct \
  --revision 48d788eca847d4d7548f375ad03d3c9312f6139e)"

VLLM_METAL_BUILD_FROM_SOURCE=1 python tools/check_parity.py \
  --model "$MODEL_DIR" \
  --prompt Hello \
  --max-tokens 4 \
  --batch-size 1 \
  --top-k 5 \
  --gpu-memory-utilization 0.5 \
  --output-dir /tmp/olmo2-before
```

Native mlx-lm reference generation succeeds. The paged server starts, then fails on the first completion. The tool exits with an HTTP 500 error; `serve.log` contains the RMSNorm exception above.

Environment: Apple M4, 24 GiB unified memory, macOS 26.6.2, Python 3.12.13, vLLM 0.29.0+cpu, MLX 0.32.1, and mlx-lm 0.32.0 pinned to `9e6acca691e64d6d8bb808c328fcdea459099cca`.

## Expected behavior

Paged attention should follow mlx-lm's OLMo 2 preparation order:

```text
projection → full-projection Q/K norm → split heads → RoPE
```

mlx-lm [constructs full-projection norms](https://github.com/ml-explore/mlx-lm/blob/9e6acca691e64d6d8bb808c328fcdea459099cca/mlx_lm/models/olmo2.py#L67-L68) and [applies them before splitting heads](https://github.com/ml-explore/mlx-lm/blob/9e6acca691e64d6d8bb808c328fcdea459099cca/mlx_lm/models/olmo2.py#L78-L85).

## Cause and proposed fix

`attention_contract_for()` falls back to `BEFORE_ROPE` for `mlx_lm.models.olmo2`. Register this module with the existing `QKNormPlacement.BEFORE_HEAD_SPLIT`, as already used for OLMo 3 in #692.

This report concerns `model_type=olmo2` and the checkpoint above; it does not claim support for OLMoE or other model types.

HanningLin:
Fixes #767

## Summary

OLMo 2 currently starts serving but crashes on its first paged completion: a 2048-element Q norm is applied to a tensor whose last dimension is `head_dim=128`. Register `mlx_lm.models.olmo2` with the existing `BEFORE_HEAD_SPLIT` placement so Q/K normalization follows mlx-lm's full-projection ordering.

- Reuse the existing preparation path; no attention-kernel change or new configuration is needed.
- Extend the existing OLMo 3 numeric regression into one parametrized OLMo 2/3 test. Remove the redundant enum-value assertion; the numeric test exercises registration through `attention_contract_for()` and fails with the original shape error if registration is missing.
- Add the verified OLMo 2 checkpoint to the supported-models table.

Scope is `model_type=olmo2`. OLMoE and MiniMax are separate follow-ups.

## Validation

Tested on Apple M4 with 24 GiB unified memory, macOS 26.6.2, vLLM 0.29.0+cpu and MLX 0.32.1. mlx-lm is pinned to `9e6acca691e64d6d8bb808c328fcdea459099cca`.

Checkpoint: `allenai/OLMo-2-0425-1B-Instruct@48d788eca847d4d7548f375ad03d3c9312f6139e` (BF16).

**Before the fix:** the native reference succeeds, but the first paged `Hello` completion fails with the 2048-element RMSNorm exception. The parametrized regression also fails for OLMo 2 while OLMo 3 passes.

**After the fix:** the existing parity tool passes all 40 built-in prompts with greedy sampling, 20 generated tokens, and request batch sizes 1 and 2:

```bash
MODEL_DIR="$(hf download allenai/OLMo-2-0425-1B-Instruct \
  --revision 48d788eca847d4d7548f375ad03d3c9312f6139e)"

VLLM_METAL_BUILD_FROM_SOURCE=1 python tools/check_parity.py \
  --model "$MODEL_DIR" \
  --max-tokens 20 \
  --batch-size 1 2 \
  --top-k 5 \
  --gpu-memory-utilization 0.5 \
  --output-dir /tmp/olmo2-after
```

| --- | ---: | ---: | ---: |
| 1 | 34 | 6 | 0 |
| 2 | 31 | 9 | 0 |

`TOP_K_MATCH` means mutual top-5 membership at the first divergent token, as defined by the existing tool; the remaining divergent continuation is not compared. These results do not claim exact token equality for every prompt. The standard parity run disables prefix caching.

**Prefix cache:** a separate HTTP check enables prefix caching and repeats the corpus's longest prompt (279 input tokens, 20 output tokens). The repeated request records 272 additional prefix-cache-hit tokens and exactly matches the uncached request's generated token IDs; both requests pass comparison against the native reference.

**Regression and static checks:**

```text
2179 passed, 15 skipped, 53 deselected

ruff check .                 passed
ruff format --check .        passed (300 files)
mypy vllm_metal              passed (139 source files)
git diff --check             passed
```

This extends the existing numeric regression instead of adding another copy or a frozen-token golden file. The OLMo 2 case protects the new registry entry in normal CI; the OLMo 3 case retains its prior coverage.

Validation covers this checkpoint and the short corpus above; it does not establish every OLMo 2 checkpoint or long-context configuration.


WindChimeRan (approved):

WindChimeRan (commented):
OLMo3 sliding-window attention (SWA) is not fully optimized. I marked it as Experimental

PR #769, opened 2026-09-15
[Bugfix] Reject non-greedy transcription at request admission

ricky-chaoju:
Temperature is a first-class field of upstream's transcription protocol, so any client can send one. `STTModelRunner._execute_stt` rejected it by raising `ValueError`, and that raise runs inside `execute_model`, which vLLM treats as a fatal EngineCore error. One such request returns 500, logs `EngineDeadError`, and leaves `/health` at 503 for everything after it.

Move the rejection to `MetalPlatform.validate_request`, the admission hook that already turns down `logit_bias` and `min_tokens`. The STT rule lives in `vllm_metal/stt/policy.py` next to `apply_stt_scheduler_policy`, which already owns config-time STT policy. The runner keeps no sampling check.

Before, on `openai/whisper-tiny` with `-F temperature=0.7`:

{"error":{"message":"EngineCore encountered an issue...","code":500}}
ValueError: STT models only support greedy decoding (temperature=0). Got temperature=0.7
vllm.v1.engine.exceptions.EngineDeadError
GET /health -> 503

After, the same request:

{"error":{"message":"vllm-metal transcribes with greedy decoding, so speech-to-text requests require temperature=0. (parameter=temperature, value=0.7)","param":"temperature","code":400}}
 GET /health -> 200


LxYuan0420 (approved):
This is the right safety fix. It would be better to spend further effort implementing proper seeded STT sampling across Whisper and Qwen3-ASR than polishing the greedy-only rejection path.

PR #770, opened 2026-09-16
[Bugfix] Honour request sampling params when transcribing

ricky-chaoju:
The one-shot STT decode picked every token with `mx.argmax`, so a request's `SamplingParams` never reached it. #769 fixed the half you could see, a sampled request killing the engine. The other half stayed. vLLM runs Whisper language detection as its own request with `allowed_token_ids` set to the language tokens, and an unconstrained argmax answers with something else. Silence, noise and a pure tone each returned 500 on the default call.

Rather than add a second sampler, the decode now uses the one we already have. `sample_from_logits` stays on `mx.argmax` for a greedy batch and otherwise hands off to vLLM's own `Sampler`, so every field gets upstream's semantics and greedy output does not move. An `STTSampling` carrier holds the params, the seeded generator and the shared sampler. `create_request_generator` moves into `sampling_batch.py` as its own commit. #769's guard then goes away.

`decode_budget` caps Whisper's loop at the request's `max_tokens`, and it has to, because `allowed_token_ids` masks the stop token out of language detection, 1.2s against 0.06s. Qwen3-ASR does not get it, since its raw stream is an envelope the adapter unwraps.

`python tools/stt_sampling_serve_smoke.py --audio speech.wav`, on this branch:
```
[200] greedy: ' the quick brown fox jumps over the lazy dog, Apple Silicon runs this model locally.'
[400] temperature=1.0 seed=7: 'speech-to-text requests require temperature=0.'
[500] non-speech, language auto-detected: '{"error":{"message":"","code":500}}'

[200] greedy: ' the quick brown fox jumps over the lazy dog, Apple Silicon runs this model locally.'
[200] temperature=1.0 seed=7: ' the quick brown fox jumps over the lazy dog ; Apple silicone runs this model locally.'
[200] temperature=1.0 seed=7 (repeat): ' the quick brown fox jumps over the lazy dog ; Apple silicone runs this model locally.'
[200] temperature=1.0 seed=8: ' Scylla is now in the open.'
[200] non-speech, language auto-detected: ' you'
```


tests/test_v1_stt_integration.py
```diff
@@ -703,6 +706,250 @@ def test_runner_preserves_empty_transcript(self, eos_token: int) -> None:
...
+        tokenizer = adapter.transcriber.tokenizer
+        tokenizer.eos_token_id = im_end
+        script = _script([language, asr_text, *spoken, im_end], QWEN3_ASR_VOCAB_SIZE)
+        adapter.model.prefill.side_effect = script
+        adapter.model.decode_step.side_effect = script
+        adapter._transcriber = Qwen3ASRTranscriber(adapter.model, tokenizer=tokenizer)
+
+        transcript = self._run(adapter, SamplingParams(temperature=0.0, max_tokens=1))
+
+        assert transcript == [*spoken, QWEN3_ASR_EOS]
```
LxYuan0420: Would be better to stop Qwen at the requested transcript budget instead of decoding everything and relying on downstream truncation.

tests/test_stt_sampling.py
```diff
@@ -0,0 +1,150 @@
...
+        assert sampling.next_token([], [], _step_logits(_TWO_CANDIDATES)) == LEADER
+
+    def test_allowed_token_ids_confine_the_pick(self) -> None:
+        """vLLM runs Whisper language detection as an allowed_token_ids request."""
+        params = SamplingParams(temperature=0.0, allowed_token_ids=[RUNNER_UP])
+
+        assert _pick(params, _TWO_CANDIDATES) == RUNNER_UP
+
+
+class TestSampledDecode:
```
LxYuan0420: Can we keep these tests focused on STT-owned behavior? Most cases duplicate sampler coverage already in `test_sampling.py`.

tests/test_v1_stt_integration.py
```diff
@@ -703,6 +706,250 @@ def test_runner_preserves_empty_transcript(self, eos_token: int) -> None:
...
+        leading = dict.fromkeys(self.TIED_TOKENS, 1.0)
+        leading[self.GREEDY_TOKEN] = 2.0
+
+        first = run(SamplingParams(temperature=0.0), scores=leading)
+        repeat = run(SamplingParams(temperature=0.0), scores=leading)
+
+        assert first == repeat
+        assert set(first[:-1]) == {self.GREEDY_TOKEN}
+
+    def test_runner_threads_its_own_sampler(self) -> None:
```
LxYuan0420: This test seems unnecessary. It asserts object identity rather than observable sampling behavior.

tests/test_v1_stt_integration.py
```diff
@@ -703,6 +706,250 @@ def test_runner_preserves_empty_transcript(self, eos_token: int) -> None:
...
+    def test_seeded_request_repeats_its_transcript(self, model: str) -> None:
+        run = self._run_whisper if model == "whisper" else self._run_qwen3_asr
+
+        first = run(SamplingParams(temperature=1.0, seed=11))
+        repeat = run(SamplingParams(temperature=1.0, seed=11))
+        other_seed = run(SamplingParams(temperature=1.0, seed=12))
+
+        assert len(first) == self.SAMPLED_STEPS + 1
+        assert first == repeat
+        assert first != other_seed
```
LxYuan0420: Different seeds can legitimately produce the same output, so divergence is not a deterministic contract. Same-seed repeatability is sufficient.

LxYuan0420 (approved):
Approving to unblock the progress. It would be better to stop Qwen at the requested transcript budget and keep the sampling tests focused on STT-owned behavior.

Issue #761, opened 2026-09-14
Support K-quant GGUF checkpoints (Q4_K/Q5_K/Q6_K)


ricky-chaoju:
I measured the repack path first (MLX 0.32.1, script comes with PR 1 as `tools/gguf_kquant_repack_measure.py`). Q4_K and Q5_K sub-blocks are 32 elements, so they map straight onto MLX affine quantization: scale = `d * sc`, bias = `-dmin * m`, and `mx.dequantize` comes out bit-identical to `gguf-py` as long as scales/biases stay fp32. With fp16 scales it is not exact anymore. So I want to require fp32 for the raw-block path, while the existing `mx.load` path for Q8_0/Q4_0/Q4_1 keeps fp16 as is. That is the first thing I'd like your take on.

Q6_K (and Q3_K/Q2_K) use 16-element groups and MLX has no `group_size=16` kernels. I checked the workarounds: merging adjacent groups almost never applies (<1%), requanting to 8-bit gs=32 ends up bigger than the native format, and dequanting to fp16 costs 2.4-6x memory. So Q6_K looks like exactly the case #415 reserved for a raw-block Metal kernel (raw block bytes in `qweight`, GGUF enum in `qweight_type`). It matters because `llama.cpp` puts Q6_K tensors (`output.weight`, some `attn_v`/`ffn_down`) inside every Q4_K_M file, so without Q6_K we cannot serve a typical file.

- [ ] PR 1: Q4_K raw-block parse + repack (tensor side only, loader stays closed)
- [ ] PR 2: Q5_K on the same base
- [ ] PR 3: Q6_K Metal kernel (I will spike `mx.fast.metal_kernel` vs C++ JIT first and pick by parity + decode throughput)
- [ ] PR 4: loader routing + a per-tensor qtype census tool + real Q4_K_M golden and serve evidence + docs
- [ ] PR 5: remote reference tags for whatever set ends up executable

Q2_K/Q3_K share the same 16-element layout and would reuse the PR 3 kernel shape; I left them out of this stack until the census shows which real files carry them.

ricky-chaoju:
@LxYuan0420 could you take a look at the two boundary calls here (fp32 scales for the raw-block path, and the Q6_K raw-block kernel route) before I start PR 1?

LxYuan0420:
[...]

PR #773, opened 2026-09-17
Repack Q4_K blocks into MLX affine tensors

ricky-chaoju:
`from_raw_blocks` reads Q4_K's raw 144-byte superblocks, pulls out `d`/`dmin` and the 6-bit sub-scales, and packs the codes into the same affine group-32 layout `mx.quantized_matmul` already takes. Scales and biases stay fp32 as agreed in the issue, since with fp16 the repack stops being exact. The existing Q8_0/Q4_0/Q4_1 path keeps its fp16 contract, and the two constructors reject each other's qtypes.

The loader is not touched. A test pins that a Q4_K file still fails preflight, and the routing PR will flip that later. The measurement script from the issue lands as `tools/gguf_kquant_repack_measure.py`.

 #761 PR1


tools/gguf_kquant_repack_measure.py
```diff
@@ -0,0 +1,304 @@
+#!/usr/bin/env python3
```
LxYuan0420, line 1: This is useful evidence for #761, but it looks like a one-time experiment rather than a maintained PR1 tool. Could we remove it and keep the results and a small reproduction in #761?

vllm_metal/gguf/mlx_native.py
```diff
@@ -175,9 +202,112 @@ def from_mx_load(
...
+        sub_scales[:, 4:8] = (packed[:, 8:12] & 0x0F) | ((packed[:, 0:4] >> 6) << 4)
+        sub_mins[:, 4:8] = (packed[:, 8:12] >> 4) | ((packed[:, 4:8] >> 6) << 4)
+        nibbles = blocks[:, 16:144].reshape(n_blocks, 4, 32)
+        codes = np.empty((n_blocks, 8, 32), np.uint32)
+        codes[:, 0::2, :] = nibbles & 0x0F
+        codes[:, 1::2, :] = nibbles >> 4
+        return codes, d * sub_scales, -(dmin * sub_mins)
+
+    @staticmethod
+    def _pack_codes_le(codes: np.ndarray, bits: int) -> np.ndarray:
```
LxYuan0420, line 293: This is correct but uses too much temporary memory. For a 4096 x 11008 Q4 tensor, `bit_matrix` alone uses 1.344 GiB while the result is only 21.5 MiB. Could we pack eight Q4 codes directly into each `uint32`?

```python
grouped = codes.reshape(rows, -1, 8)
shifts = np.arange(8, dtype=np.uint32) * 4
return np.bitwise_or.reduce(grouped << shifts, axis=-1)
```

LxYuan0420 (approved):
Approving to unblock progress. The Q4_K logic and tests look good. Please improve the packer in mlx_native.py before enabling loader support because it can use several GiB of temporary memory on real model weights.

PR #764, opened 2026-09-14
feat: budget GDN state independently and expand KV cache capacity

suntp:
This follows [#754](https://github.com/vllm-project/vllm-metal/pull/754), which remains a dependency and is unchanged. Compact physical slots reduce materialized GDN storage, while startup still reserves the previous worst-case state cost. This PR adds an **opt-in, independently enforced state-array budget** so startup can assign some of that reservation to KV capacity.

```text
--additional-config '{"state_cache_budget_mib": 2048}'
```

Omitting the option retains existing sizing. The MiB value bounds stable state arrays, not total GPU/RSS memory. Admission reserves completed-source and in-flight destinations; active state stays protected, and idle checkpoints must have their hashes invalidated before a retired state generation can be reused. Unsupported cache-manager geometries, partial-prefix CoW, internal checkpoints, speculation, external transfers and nonlocal layouts are rejected explicitly. Setup, supported scope, per-host budget sizing and the separate 48 GB reproduction recipe are in the [benchmark guide](https://github.com/suntp/vllm-metal/blob/24ce8e6db132237c638866cb0e9b7494eee78303/docs/state-cache-budget-benchmark.md). The `2048` MiB value is a published 27B benchmark setting, not a default for every RAM size.

## Review status and the pressure-scheduling repair

**HEAD is `24ce8e6` (docs/test + ruff format on top of `8b5e68f6`). The 64 GB formal 14-run matrix and pressure-repair engine remain `93ede07955ab7a98e12bd251daaae3ce854e2bed`; original strict output compatibility is still FAIL (six pairs PASS, one pair FAIL).** `8b5e68f6` does not change the production scheduler. It documents per-host GDN budget floors and adds official `--no-ignore-eos --apply-chat-template` coverage. This is a deliberate scheduling-policy change with the behavioral difference disclosed below, not a claim of complete old-schedule output equivalence or merge approval.

During pressure, asynchronous work still owns blocks that the scheduler wants to release. Retrying allocation immediately after a deferred preemption can preempt further requests without making any blocks available. The repair conservatively checks the running requests' additional block requirements, releases only obsolete completed state references, and drains outstanding work through normal zero-token core steps when a shortage remains. It keeps the existing lifetime fence and allows ordinary preemption after earlier results have been consumed. Continuing inside an existing decode block does not require a wait.

[...]

PR #778, opened 2026-09-18
[Refactor KV] [3/4] Use vllm upstream shared storage for hybrid caches

WindChimeRan:
## Unified KV management

**All scheduler-managed KV and state storage should use upstream allocation through `KVCacheStorage`.** vLLM owns allocation, physical layout, and global block management. Metal consumes zero-copy MLX views and runs the kernels.

## Roadmap

1. [776: Metal support](https://github.com/vllm-project/vllm-metal/pull/776): Add shared `as_strided` views and stride-aware cache operations.
2. [777: Storage adapter](https://github.com/vllm-project/vllm-metal/pull/777): Expose vLLM's allocation as zero-copy MLX views backed by per-region Metal buffers.
3. **[778: Hybrid migration](https://github.com/vllm-project/vllm-metal/pull/778) (this PR)**: Use the adapter for hybrid models, remove separate state pools, and use the full cache budget.
4. [800: Attention migration](https://github.com/vllm-project/vllm-metal/pull/800): Use the adapter for ordinary/grouped attention and remove duplicated layout/allocation code.

## Scope and remaining work

This stack migrates hybrid and ordinary/grouped attention. Two migrations remain:

- **MLA:** bind the latent cache to upstream storage and preserve sharing during writes.
- **Model-based speculative decoding:** migrate target and committed draft KV, with separate handling for lookahead scratch.

After those migrations, remove the remaining count-based initializers and allocators. The attention migration includes TODOs at both remaining initializers. The storage adapter wraps upstream memory in per-region Metal buffers, so the hybrid migration can use the full cache budget.

## This PR

Use upstream shared storage and scheduler-owned state blocks for hybrid caches. Remove separate state pools, growth, and reservation accounting. Report the full profiled cache budget; the storage adapter already exposes multiple Metal buffers.

This PR removes the cause of the GDN state over-allocation addressed by [754: compact GDN state storage](https://github.com/vllm-project/vllm-metal/pull/754). On main, separate GDN arrays grow to the highest scheduler block ID encountered, materializing unused state rows when IDs are sparse. Shared storage places GDN state and attention KV in the same upstream allocation, removing the separate state-pool growth and its extra reservation.

Keep compact decode updates in none mode; flush pending state only when affected slots are copied/reset or a request is released. Retain align-mode checkpoint materialization and the whole-step dependency guard.

## Validation

Current non-slow suite: **2,219 passed, 13 skipped, 39 deselected**. Ruff, formatting, and mypy pass.

The recorded both-mode gate at `a5d8a1c` matched all 40 same-mode main-reference outputs at default utilization 0.92. Its six cache-on/off differences were also present on main.

### Large-cache validation

Qwen3.5-0.8B on M5 Pro / 64 GiB, using the default `gpu_memory_utilization=0.92` and no block-count override:

| Metric | Capped control | This PR |
|---|---:|---:|
| Actual cache backing | 41.75 GB | 48.46 GB |
| Metal buffers | 6 × 6.96 GB | 6 × 8.08 GB |
| Scheduler blocks | 6,245 | 7,250 (+16.1%) |

The device's per-buffer limit is 41.75 GB. The new total allocation exceeds it while every imported region fits. Both cases use one upstream backing allocation. **All 20 complete outputs matched**, covering block edges, prefix reuse, short-after-long requests, and a concurrent batch. Both runs exited normally.

Capped control: `01a3a20`; this PR: `b7a4c8d`. Same model snapshot `2fc06364715b967f1860aea9cf38778875588b17`, MLX 0.32.1, vLLM 0.29.0, max model length 2048, max sequences 4, max batched tokens 2048, align-mode prefix caching, and greedy decoding.

### GDN allocation against main

Qwen3.5-0.8B, main `38d0d66` versus this PR at `a5d8a1c`, with the same `gpu_memory_utilization=0.15`.

allocation-comparison

### Serving performance

Serving benchmark at `1167abf` (below the buffer cap): Qwen3.5-0.8B, Sonnet 1024+128, 100 requests at unlimited rate, five warmups, two fresh-server runs per case, memory utilization 0.3, M5 Pro / 64 GiB. Mean inter-token latency, main → `1167abf`:

| Prefix cache | Concurrency | Mean ITL | Change |
|---|---:|---:|---:|
| Off | 8 | 12.57 → 12.74 ms | +1.3% |
| Off | 32 | 42.74 → 42.47 ms | −0.6% |
| On | 8 | 15.28 → 15.82 ms | +3.5% |
| On | 32 | 50.56 → 50.90 ms | +0.7% |

These measurements cover the changes through this PR, including the dependency guard. Related to [PR 754](https://github.com/vllm-project/vllm-metal/pull/754).


rishabhsinha17:
I read through the stack, 776 and 777 plus this one, alongside #780. Since it removes the ground #764 builds on, I went through it asking which of the properties I verified there still hold once vLLM's shared storage backs everything.

The direction is right. #764 carries generation and catalog machinery whose whole purpose is turning a scheduler and worker disagreement over a state slab into a RuntimeError, and that class of disagreement cannot exist once the engine's BlockPool is the only allocator and the worker derives its views from the scheduler's own ids. What remains is temporal, pending lazy writes racing a block handoff, and the flush points here look right to me, apply_pending_states intersected with the affected slots ahead of zero_blocks and copy_blocks, materialization at release, with #780 moving the native fallback writes into the dependency graph. Both invariants I verified line by line on #764 hold here as well. Checkpoint immutability holds, copy_slots gathers sources before any destination write and leaves them alone, and storage.copy_blocks snapshots every source region before scattering, which keeps cyclic CoW batches atomic. Flush-then-zero ordering holds too, for a subtle reason. CacheViews rebuilds each view from the depends-chained buffer at access time, and a zero issued after a flush reads a view that already depends on the flushed write. That construction is easy to miss and deserves a sentence in storage.py.

What I recorded on #764 carries over, and one part of it costs more here.

The single-buffer cap in WorkerCachePlanner changes real capacity against main on high-memory hosts. Main allocates per-layer key and value arrays plus separate state pools, and nothing there comes near max_buffer_length. Here the whole KV plus state backing is one Metal allocation. On my M1 Pro 16 GB, mx.device_info() reports max_buffer_length 9,534,832,640 against 17,179,869,184 physical, about 55 percent. If that ratio holds on the 64 GB host from the benchmark table, the cap lands near 38 GB while the budget at gpu-memory-utilization 0.92 after 0.8B weights is near 60, which would mean the 0.92 startup check in the description already ran capped, with a third of the requested cache missing and a warning log as the only trace. Serving numbers in the table used 0.3, below the cap, so they cannot show this. Accepting the cap for a first landing seems fine to me, though I would give it a louder announcement than a warning. CoW and zeroing already work region by region, and the depends chain is the main piece that assumes one backing, so a chunked import later does not look blocked by the design.

My other #764 note, the declaration gap, also carries over unchanged in kind, and it reaches further. MambaSpec shapes and dtypes and the layer name lists in initialize_from_config all derive from the hybrid plan, and nothing in the stack consults the caches mlx-lm declares through make_prompt_cache, I grepped at the #780 head. One declaration now sizes the engine's physical allocation and parameterizes the strided views into it, where before it sized only the metal-side pools. patch_model does fail loud on role mismatches in both directions, and #780's bounds validation keeps a wrong geometry inside the backing, which leaves shape-level drift showing up as wrong numerics rather than an error. An install-time check of the plan's resolved geometry against the cache the loaded model declares would close it, and initialize_from_config is a natural place since everything meets there once. Follow-up grade in my view, same as on #764, worth keeping visible while this layer is being rebuilt.

Small one to end on, the none-mode unification onto scheduler-owned blocks, one block per request with mamba_block_size pinned to max_model_len and release meaning materialize rather than free, is a cleaner lifecycle than the slot pool it replaces. Happy to rerun the #764-style native verification on this stack once it leaves draft, same M1 Pro rig.

suntp:
Thanks @rishabhsinha17 — this is a very useful read, and the two carries-over notes are actionable.

**The single-buffer cap** (`WorkerCachePlanner`) now announces itself: the warning states the requested vs. reported GB and points at `--gpu-memory-utilization`, and `VLLM_METAL_HYBRID_BUFFER_CAP=error` fails startup instead of silently serving reduced capacity. Beyond that, the long-term fix is implemented: layer-compact layouts (LBNHC) now back **one Metal buffer per layer page span** in #791, so the planner no longer caps the budget at `max_buffer_length` at all; block-outermost layouts keep the single buffer with the cap. Region dedupe is global by virtual address, so overlaid groups (shared state pools) keep the exact sharing semantics the single buffer had. Your note that "the depends chain is the main piece that assumes one backing" was exactly right — per-region anchors re-wrap the write chain on every `depend()`, which is the only structural change needed.

**The declaration gap** is closed at patch time in #780: the runtime now builds the loaded model's own declared cache (`make_cache()` + one token), aligns it to the plan's owned layers (1:1 for qwen3_next, cache-bearing-only for nemotron_h), and fails loud on shape or topology drift. As you suggested, `initialize_from_config`/`patch_model` is where plan and model meet; `patch_model` already had the model in hand and the role checks, so the check lives right beside them. We deliberately check shapes, not dtypes — dtype policy already has its own validation path (`get_mamba_state_dtype_from_config` + the mixer-dtype check).

Also added the sentence you asked for: `CacheViews`' docstring now spells out that ordering rides on the depends chain (views rebuild from the chain head at access time), not on evaluation order.

Both fixes are on #780 (commit `27e1706`); the per-region split is #791, stacked on #778 and carrying the #780 commits until that merges. Happy to take the native verification run whenever convenient.

LxYuan0420 (approved):

PR #780, opened 2026-09-18
[Bugfix] Track native GDN writes through the MLX graph

suntp:
Stacked on #778.

Native GDN fallback writes currently bypass MLX's dependency graph, so host or cross-stream consumers can observe unfinished output/state. Two adjacent planning gaps compound this: the single-buffer capacity cap is silent, and nothing cross-checks the hybrid plan's declared state geometry against what the loaded model actually declares.

This PR:
- Makes native GDN recurrence an MLX primitive returning a fresh output and an updated state alias. The existing cache owner records the dependency; no Python synchronization fence is added. Kernel math is unchanged. View construction stays on #776's shared `as_strided` helper under its minimal, `mx.as_strided`-matching guard contract — the view metadata comes from vLLM's cache layout, so no stricter validation layer is added here.
- Announces the single-buffer cap: the planner logs requested vs. capped GB, and `VLLM_METAL_HYBRID_BUFFER_CAP=error` fails startup instead of silently serving reduced capacity.
- Cross-checks the plan's declared state geometry against the loaded model's `make_cache` shapes after one probe token, failing fast on topology or shape drift.

Adds regressions for stream/compiled execution, dirty-page masking, and state-to-KV reuse after cancellation, plus planner and geometry tests.

## Verification

Serial single-request server benchmark (`vllm serve` + `vllm bench serve`, one fresh server per point, concurrency 1, Qwen3.5-0.8B hybrid) across four branches — this PR / #791 (per-region follow-up) / #778 / main — at input lengths 16K/32K/64K/128K/256K:

- **Correctness**: a needle-retrieval probe (two codes embedded at ~10% / ~90% of the prompt, greedy decode) passes at every length on every branch (20/20), including 256K. The stacked #791 stays token-identical to this PR across the stress-parity suite (S1–S5: long/short mixes, preemption via `num_gpu_blocks_override`, mid-stream cancellation/reset), and tight preemption runs are identical across all four branches.
- **Prefill / decode parity**: TTFT within <0.5% of main and #778 at every length (e.g. 128K: 50.47 vs 50.36 / 50.36 / 50.67 s; 256K: 177.6 vs 177.9 / 177.6 / 178.7 s); decode mean ITL within run-to-run noise (e.g. 128K: 42.8 vs 43.7 / 43.6 / 42.8 ms). The native-write dependency tracking and the planning cross-checks add no measurable overhead. Full table in #791.
- **Unit coverage added in this PR**: declared-geometry fail-fast checks, and the loud hybrid-buffer-cap planner modes (warn/error) with budget-capping regressions.


suntp:
@WindChimeRan This is the follow-up to our collaboration on #754, based on your #778 stack. Please take a look when you have time.

WindChimeRan (commented):
Thanks for looking into the issue! 

---

> The native cache-view helper also needs checks for overflowing or invalid view metadata.

I think this design is a bit over-defensive, I've adjusted and simplified #776, and keep minimum guards to match with mx.as_strided

suntp:
Agreed — the shared view helper belongs to your stack, so the minimal-guard contract should live in #776, not be re-hardened here. I'll drop the `cache_view` validation commit from this PR in the next push and build directly on the simplified `as_strided` (e41d9c1). This PR will then carry only the GDN write-tracking, the buffer-cap announcement, and the declared-geometry cross-check.

suntp:
Correction to my last comment: I see your follow-up commits carried the validation through the `as_strided` rename directly on both branches — thanks, that's a much cleaner path than dropping it. I've adopted your branch states as-is (#780 at 3d855cd, #791 at ad31c23); #791 is rebuilt on top and its unit suites pass. The minimal-guard contract for the shared helper in #776 with the stricter layer added here reads like a good division.

suntp:
Follow-up: per your point about the strict view checks, I've removed that layer from this PR rather than keeping the carried version. The branch is rebuilt directly on the current #778 tip, so the helper contract is exactly #776's minimal `as_strided` guards (view metadata comes from vLLM's cache layout, per your note in 136949f). What remains here is the GDN write-tracking primitive + `depend`, the loud single-buffer cap (`VLLM_METAL_HYBRID_BUFFER_CAP=warn|error`), and the declared-geometry fail-fast — retitled accordingly. #791 is rebuilt on top (per-region storage + lazy anchors) and both branches pass the unit suites plus greedy token-parity and a 64K serve probe.

WindChimeRan (approved):
Qwen3.5-0.8B, offline greedy | Main median tok/s | PR median tok/s | Δ
-- | -- | -- | --
lazy=1, batch 32 | 999 | 997 | −0.3% (noise)
lazy=1, batch 1 | 125.0 | 125.2 | ≈0%
lazy=0, batch 32 | 809 | 832 | +2.8% (runs overlap)
lazy=0, batch 1 | 92.9 | 121.9 | +31%

This fixes the native recurrent fallback (</span><code dir="ltr">lazy=0</code><span> and automatic fallback cases), where MLX could read output/state before GPU writes completed and return incorrect results; the default optimized kernels are unchanged.

**Deprecation plan**: Remove the per-request convolution after the batched replacement passes correctness and performance checks. Once optimized recurrence handles every remaining fallback case, remove the native fallback, its primitive and binding, and the switch.

Issue #782, opened 2026-09-18
[Perf] Decode kernel reads the whole context on sliding-window layers: block range is set by partitioning, window is only a mask


Lavmee:
### Summary

In `pagedattention.metal` the per-token decode kernel splits the context into `PARTITION_SIZE` slices and masks keys outside the sliding window per element (`~L939–947`, `~L1137–1138` at `589df0c`). The block range never shrinks to the window, so a decode step on a window-1024 layer reads the entire context: 32× the KV it needs at 32K. Gemma 4 has 25 such layers out of 30, so decode TPOT grows with context length although the model's attention cost should not.

This is the decode-side twin of the tiled prefill issue (#781).

### Measurement

Per layer, bf16, 16 q / 8 kv heads, head_dim 256, M5 Pro:

| context | requests | full attention | window 1024, after | speedup |
|---:|---:|---:|---:|---:|
| 8192 | 1 | 1.39 ms | 0.30 ms | 4.6× |
| 32768 | 1 | 2.12 ms | 0.23 ms | 9.3× |
| 32768 | 8 | 9.80 ms | 0.48 ms | 20.4× |
| 65536 | 8 | 21.17 ms | 0.55 ms | 38.8× |

Before the change the windowed layer cost the same as full attention (the mask saves no reads). After it, the step cost on those layers no longer depends on context length.

On the production Gemma 4 26B server (M3 Ultra) decode went from an estimated ~53 ms of sliding-layer attention per step at 32K context to 22 ms/token total TPOT at 32K (12.3 ms/token at 4K).

### Fix

`Foresko/vllm-metal@12d14a9` ([Attention] Bound the decode kernel's block range by the sliding window): the block range starts at the block containing the window start of the earliest row in the threadgroup; a partition whose blocks all lie left of the window writes the same neutral partial (max 0, sum 0, zeros) the masked loop produced and exits, so the reducer and launch grid are unchanged. Tests: partitioned and unpartitioned windowed decode against the reference (window start inside a partition and inside a block, varlen batch), the windowed rows of the spec-decode kernel, and a cost check.

Happy to open a PR from that branch.

Lavmee:
Fix: #785.

PR #784, opened 2026-09-18
[Kernel] Bound the tiled prefill tile loop by the sliding window

### LINKED ISSUE #781: [Perf] Tiled prefill kernel does full O(n²) attention on sliding-window layers: window is applied only as a per-element mask
### Summary

`pagedattention_tiled.metal` (the prefill path on every machine except M5, where the NAX kernel runs) walks every KV tile from zero for each Q tile of a sliding-window layer and only masks keys outside the window. The no-mask fast path (`tile_no_mask`) is disabled outright whenever `sliding_window >= 0`. So a window-1024 layer costs exactly as much as a full-attention layer and scales with n². On Gemma 4 26B, where 25 of 30 layers are sliding, prefill of a 32K prompt took 108 s on an M3 Ultra.

The NAX kernel already does the right thing (`kb_start = max(0, (context_len + q_pos_start + 1 - sliding_window) / TILE_KV)` in `pagedattention_nax.metal`); the tiled kernel does not.

### Where

- Tile loop starts at 0 and stops only at the causal boundary: `vllm_metal/metal/kernels_v2/pagedattention_tiled.metal` (main loop over KV tiles, ~L286–292 at `589df0c`); window applied per element only (~L424–425).
- Fast path disabled for windowed layers: `tile_no_mask` guarded by `sliding_window < 0` (~L397).
- Related, not fixed here: Q/K/V of the sliding layers (head_dim 256) are zero-padded to the model's largest head_dim (512) because the cache is allocated for the largest head (`vllm_metal/attention/impls/sdpa.py:489-535`), doubling work and traffic on those 25 layers.

### Measurement

Single layer, full prefill of one sequence, 16 q / 8 kv heads, fp16, median of 5. "Before" is the current kernel with window 1024; "after" is the patch below.

| Machine | seq | head_dim | full attention | window 1024, before | window 1024, after | speedup |
|---|---:|---:|---:|---:|---:|---:|
| M3 Ultra | 8192 | 256 | 91 ms | 93 ms | 23 ms | 4.1× |
| M3 Ultra | 16384 | 256 | 348 ms | 347 ms | 47 ms | 7.5× |
| M3 Ultra | 8192 | 512 | 580 ms | 590 ms | 147 ms | 4.1× |
| M3 Ultra | 16384 | 512 | 2681 ms | 2713 ms | 284 ms | 9.6× |
| M5 Pro | 16384 | 512 | 7400 ms | — | 672 ms | 11.0× |

The "before" column is the diagnosis: the window saves nothing today.

End to end on Gemma 4 26B (OptiQ 4-bit), M3 Ultra, cold prompt, this change plus a head_dim-512 split-D variant of the tiled kernel: 16K prefill 31 s → 11 s, 64K 408 s → 80 s.

### Fix

`Foresko/vllm-metal@74a919a` ([Attention] Bound the tiled prefill tile loop by the sliding window): the tile loop starts at the first tile that intersects the window of the earliest row in the threadgroup, and tiles that lie entirely inside the window for every row of the threadgroup take the no-mask fast path. Reducer and launch grid unchanged. Test `tests/attention/test_sliding_window_tile_skip.py`: parity against an fp32 reference for whole, chunked and tile-unaligned prefill and for a varlen batch, plus a timing check that fails on the old kernel. The fork's attention tests (232) and the full non-slow suite pass.

Happy to open a PR from that branch if you want it.

Lavmee:
Fixes #781.

### What

For a sliding-window layer the tiled prefill kernel (`pagedattention_tiled.metal`) visited every KV tile from zero and only masked keys outside the window, and the `tile_no_mask` fast path was disabled outright whenever a window was set. On Gemma 4 (25 of 30 layers with window 1024) prefill therefore cost as much as full attention and scaled with n².

- The tile loop now starts at the first tile that can hold an attended key of the threadgroup's earliest row (`tile_idx_start`), the way the NAX kernel already bounds its loop.
- A tile that lies inside the window of the threadgroup's *last* row and left of every row's causal frontier takes the unmasked fast path.
- Reducer, launch grid and the per-element mask are unchanged. Tiles left of the window contributed only `-inf` scores, so with them gone the running max is seeded by the first real tile; results agree with the reference within rounding.

### Test

`tests/attention/test_sliding_window_tile_skip.py`: parity against an fp32 reference for whole, chunked and tile-unaligned prefill and for a varlen batch with a window, plus a timing check that fails on the mask-only loop.

```
pytest tests/attention/test_sliding_window_tile_skip.py   # 14 passed
```

### Result

Single layer, full prefill of one sequence, 16 q / 8 kv heads, fp16, median of 5:

| Machine | seq | head_dim | full | window 1024 before | window 1024 after |
|---|---:|---:|---:|---:|---:|
| M3 Ultra | 8192 | 256 | 91 ms | 93 ms | 23 ms |
| M3 Ultra | 16384 | 256 | 348 ms | 347 ms | 47 ms |
| M3 Ultra | 16384 | 512 | 2681 ms | 2713 ms | 284 ms |
| M5 Pro | 16384 | 512 | 7400 ms | — | 672 ms |

End to end on Gemma 4 26B (M3 Ultra, with the decode-side change in #782 and a head_dim-512 split of the tiled kernel that is not part of this PR): cold 16K prefill 31 s → 11 s.


HanningLin:
Tested locally on **Apple M4, 24 GiB, macOS 26.6.2**, at commit `cc8d6904d214e54af80091d6a38ba878df27a982`, using MLX 0.32.1, mlx-lm 0.32.0, vLLM 0.29.0+cpu, source-built kernels, and NAX disabled.

**Kernel correctness**

- All 14 tests added by this PR passed, including the timing check.
- Supplementary tests and existing primitive/attention-sink coverage passed: **125 tests total** across two runs.
- Supplementary tiled-prefill coverage includes fp16/bf16, head sizes 64/96/128/256/512, windows 1/33/128, partial query tiles, non-contiguous page mappings, and mixed prefill/decode batches.
- Also checked a **4096-token window with contexts exceeding 8192 tokens** against the existing attention reference across both dtypes and all five head sizes.

**OLMo3 end-to-end parity**

Tested a local MLX affine 4-bit conversion (group size 64) of `allenai/Olmo-3-7B-Instruct@6e5971d9eba42665f5bd5a0fcf047f299ce1dccc`. Both native mlx-lm and vllm-metal used the same converted checkpoint.

Used the existing `tools/check_parity.py` implementation with four plain-text prompts of **5, 12, 4207, and 5012 input tokens**, generating 16 greedy tokens each. The two long inputs are observation logs followed by a summarization request, crossing OLMo3's 4096-token sliding window.

|---|---:|---:|---:|
| 1 | 4 | 0 | 0 |
| 2 | 4 | 0 | 0 |

Every generated token matched the native reference; no top-k allowance was needed. Server logs confirm **chunked prefill with a 2048-token budget**, `max_model_len=5028`, `max_num_seqs=2`, prefix caching disabled, and NAX disabled. GPU memory utilization was 0.6.

The BF16 full-model reference attempt was stopped because of memory pressure on this machine. The completed end-to-end results above apply to the 4-bit conversion, not BF16 weights. BF16 remains covered by the kernel tests. This is a small targeted corpus, not broad model qualification or a prefix-cache test.

**Kernel timing**

Single-sequence full prefill, 4 query heads / 2 KV heads, head size 128, block size 16. Median of seven measurements after two warmups. Separate processes used the parent commit's tiled shader versus this PR's shader; the harness and native C++ code were unchanged. The source builder was checked to confirm the window-skip code was absent/present respectively.

| Dtype | Tokens | Window | Before | After | Speedup |
|---|---:|---:|---:|---:|---:|
| FP16 | 8192 | 4096 | 32.74 ms | 24.37 ms | 1.34× |
| BF16 | 8192 | 4096 | 34.78 ms | 25.94 ms | 1.34× |
| FP16 | 8192 | 1024 | 32.30 ms | 8.07 ms | 4.00× |
| BF16 | 8192 | 1024 | 34.91 ms | 8.62 ms | 4.05× |

For the 2048-token control with window 4096, before/after ratios were approximately 1.02× (FP16) and 0.99× (BF16). Benchmark outputs matched exactly except for the FP16 8192/1024 case, whose maximum absolute difference was `5.96e-8`.

These timings are **kernel-level measurements**, using smaller head counts than OLMo3. The end-to-end run establishes output parity for the tested prompts; I have not measured end-to-end serving speedup.

LxYuan0420:
GitHub rejected maintainer writes to the organization-owned fork, so I pushed the signed formatter-only commit to `vllm-project:maintainer/pr-784-format`. Please apply it with:

```bash
git cherry-pick 07179fab
```

Local checks pass: Ruff, 13 non-slow kernel tests, and the slow kernel timing test. Once the PR head updates, I will watch CI and approve.

Lavmee:
Thank you for the thorough check, especially the OLMo3 parity across the 4096-token window and the 4096-window timing: that is the case I could not cover locally (my numbers are all from Gemma 4 with window 1024), and the 1.34× there matches what the loop bound should give when the window covers half the context. The FP16 8192/1024 difference of 6e-8 is the running-max seeding change described in the commit message: tiles left of the window contributed only -inf scores, so with them skipped the max is seeded by the first real tile.

LxYuan0420 (approved):
LGTM

PR #785, opened 2026-09-18
[Kernel] Bound the decode kernel's block range by the sliding window

### LINKED ISSUE #782: [Perf] Decode kernel reads the whole context on sliding-window layers: block range is set by partitioning, window is only a mask

Lavmee:
Fixes #782. Decode-side counterpart of #781.

### What

The per-token decode kernel (`pagedattention.metal`) split the context into `PARTITION_SIZE` slices and masked keys outside the sliding window per element, so a decode step on a window-1024 layer read the whole context: 32× the KV it needs at 32K, on 25 of 30 Gemma 4 layers.

- The block range now starts at the block containing the window start of the earliest row in the threadgroup.
- A partition whose blocks all lie left of the window writes the same neutral partial the masked loop produced (max 0, sum 0, zeros) and exits, so the reducer and launch grid are unchanged.

### Test

`tests/attention/test_decode_sliding_window_skip.py`: partitioned and unpartitioned windowed decode against the reference (window start inside a partition and inside a block, varlen batch), the windowed rows of the spec-decode kernel, and a cost check.

```
pytest tests/attention/test_decode_sliding_window_skip.py   # 12 passed
```

### Result

Per layer, bf16, 16 q / 8 kv heads, head_dim 256, M5 Pro:

| context | requests | full attention | window 1024, after |
|---:|---:|---:|---:|
| 8192 | 1 | 1.39 ms | 0.30 ms |
| 32768 | 1 | 2.12 ms | 0.23 ms |
| 32768 | 8 | 9.80 ms | 0.48 ms |
| 65536 | 8 | 21.17 ms | 0.55 ms |

Before the change the windowed layer cost the same as full attention. On a production Gemma 4 26B server (M3 Ultra) TPOT at 32K context is 22 ms/token after this change (12.3 ms at 4K).


LxYuan0420:
GitHub rejected maintainer writes to the organization-owned fork, so I pushed the signed minimal test fix to `vllm-project:maintainer/pr-785-test-fix`. Please apply it with:

```bash
git cherry-pick 865ca5c0
```

This derives the sequence count from `get_ops().min_decode_grid()`, asserts the non-partitioned gate, and formats the test file. Local checks pass: Ruff, 11 non-slow kernel tests, and the slow kernel timing test. Once the PR head updates, I will watch CI and approve.

PR #786, opened 2026-09-18
Per-row native top-k/top-p/min-p masks so mixed batches stay on the MLX sampler

### LINKED ISSUE #783: [Perf] One request with different top_k/top_p/min_p (or one greedy request) pushes the whole batch off the native MLX sampler, ~3× step time at 128 sequences
### Summary

`SamplingBatch.params_allow_native_random` requires every request in the batch to share the same `top_k`, `top_p` and `min_p` and all to be random; `params_allow_native_greedy` requires all to be greedy. Any other mix sends the entire step to the synchronous torch sampling path for all rows. In serving this is the common case: one client sending `temperature: 0` or its own `top_p` among requests running with the model defaults slows everyone down.

### Measurement

Gemma 4 26B (OptiQ 4-bit), M3 Ultra, 127 concurrent streams plus one "neighbour" request with different params, `max_num_seqs` 128, decode step time and aggregate tok/s. "Before" is `589df0c` + our attention patches; "after" is the fix below.

| batch | neighbour | step before | tok/s before | step after | tok/s after |
|---|---|---:|---:|---:|---:|
| greedy | none | 178 ms | 698 | 204 ms | 552 |
| greedy | temperature 1 (model default) | 540 ms | 214 | 194 ms | 552 |
| temperature 1 | none | 241 ms | 521 | 215 ms | 581 |
| temperature 1 | top_p 0.9 | 562 ms | 220 | 215 ms | 551 |
| temperature 1 | temperature 0 | 562 ms | 212 | 222 ms | 509 |
| temperature 1 | temperature 0.7, top_k 20, min_p 0.05 | — | — | 243 ms | 441 |

("After" numbers were taken under live traffic, hence the +20–30 ms on the unmixed rows.) The engine log confirms the slow mode is inside the engine (`Avg generation throughput` 243 vs 742 tok/s with the same 128 running requests). Memory, CPU, thermals, a multimodal or JSON-schema neighbour and queueing were ruled out by direct checks.

### Fix

`Foresko/vllm-metal@b637f4c` ([Sampling] Per-row native top-k/top-p/min-p masks; mixed greedy/random batches stay on the MLX path):

- the native random path builds its top-k / top-p / min-p mask per row instead of requiring shared values;
- greedy rows become a top-1 candidate at temperature 1 and are resolved by the same `mx.random.categorical`, so a greedy request no longer forces torch for the batch;
- vocabulary sorting is replaced by `argpartition` on `max(top_k)` of the batch plus a sort of the candidates when every row has a top-k: on 128 × 262 144 that is 8 ms instead of 23 ms per step.

Seed, penalties, logprobs, `allowed_token_ids` and `bad_words` still take the whole batch to torch, as before. Tests in `tests/test_native_sampling.py`: per-row masks against vLLM's `apply_top_k_top_p`, ties at the top-k boundary, mixed greedy/narrow/wide batches.

Happy to open a PR from that branch.

Lavmee:
Fixes #783.

### What

`SamplingBatch.params_allow_native_random` required every request in the batch to share `top_k`, `top_p` and `min_p` and all to be random; `params_allow_native_greedy` required all to be greedy. Any other mix sent the whole step to the synchronous torch sampler for all rows, roughly tripling the step time at 128 sequences whenever one client used `temperature: 0` or its own `top_p`.

- The native random path builds its top-k / top-p / min-p mask per row.
- Greedy rows become a top-1 candidate at temperature 1 and are resolved by the same `mx.random.categorical`, so a greedy request no longer forces torch for the batch.
- Greedy rows take the argmax (`mlx_greedy_tokens`) and ride the draw only so every row of the batch stays drawable: top-k masks by value, so an exact tie at a greedy row's maximum would otherwise be resolved by the categorical draw.

Seed, penalties, logprobs, `allowed_token_ids` and `bad_words` still take the whole batch to torch, as before. `docs/configuration.md` updated accordingly.

### Test

`tests/test_native_sampling.py`: per-row masks against vLLM's `apply_top_k_top_p`, ties at the top-k boundary for a single row and for a mixed batch, a greedy row with a tied maximum staying on the argmax over 50 draws, and mixed greedy / narrow / wide batches.

```
pytest tests/test_native_sampling.py   # 33 passed
```

### Result

Gemma 4 26B, M3 Ultra, 127 concurrent streams plus one neighbour with different params, `max_num_seqs` 128, decode step time:

| batch | neighbour | before | after |
|---|---|---:|---:|
| greedy | temperature 1 | 540 ms | 194 ms |
| temperature 1 | top_p 0.9 | 562 ms | 215 ms |
| temperature 1 | temperature 0 | 562 ms | 222 ms |
| temperature 1 | none | 241 ms | 215 ms |

---

**Update after review (b174794).** Greedy rows now take the argmax instead of a top-1 value threshold, which an exact tie at the maximum left to the draw (thanks @ricky-chaoju for the reproduction). The `max(top_k)` `argpartition` shortcut is removed: it capped the candidate set at `k_max` and dropped ties at a row's k-th largest logit that vLLM keeps, so the whole vocabulary is sorted again (candidate selection at 128 × 262144 on an M5 Pro: 33.3 ms against 25.6 ms for the partition). `_top_k_top_p_masked_logits` is gone and the mask-parity tests call the production path. The step times in the table above were measured with the partition in place; the mixed-batch win does not depend on it.


ricky-chaoju:
Nice result, the partition win reproduces here, 10.7 vs 21.5 ms at 128 x 262144 on an M1 Ultra.

One thing though. Greedy rows go in as top_k=1 so they get the value threshold, but greedy wants rank. On an exact tie at the top both ids survive and categorical draws between them. Through native_decode_tokens with [temperature=0.0, temperature=1.0 top_p=0.9] the greedy row gave both 14 and 15 over 200 calls, argmax is 14. Main raises gate desync there and falls back to the sync sampler, 200 for 200.

So temperature 0 is deterministic alone but not with a top_p neighbour. The mixed-batch test uses normal logits so it never ties. Masking greedy by position rather than value would do it. Flag is off by default so nothing changes today.

vllm_metal/v1/sampling_batch.py
```diff
@@ -303,18 +333,36 @@ def _native_random_tokens(
...
+        Greedy rows (temperature below ``GREEDY_TEMPERATURE_EPS``) become a
+        top-1 candidate set at unit temperature, so they resolve to the argmax
+        inside the same categorical draw as the random rows.
+        """
+        temperatures: list[float] = []
+        top_k: list[int] = []
+        top_p: list[float] = []
+        min_p: list[float] = []
+        for sp in sampling_params_list:
+            if sp.temperature < GREEDY_TEMPERATURE_EPS:
```
LxYuan0420, line 321: `top_k=1` does not make a row greedy when the highest logits are tied. It can randomly choose either token. Please use argmax for greedy rows, matching upstream.

vllm_metal/v1/sampling_batch.py
```diff
@@ -266,35 +259,72 @@ def _top_k_top_p_masked_logits(
...
-            scaled_logits = mx.put_along_axis(
-                mx.full(scaled_logits.shape, -mx.inf, dtype=scaled_logits.dtype),
-                order_desc,
-                sorted_desc,
-                axis=-1,
-            )
-        return scaled_logits
+        effective_k = [k if 0 < k < vocab_size else vocab_size for k in top_k]
+        k_max = max(effective_k)
+        if k_max < vocab_size:
```
LxYuan0420: Please remove this partition shortcut for now. It drops tied tokens that upstream keeps, so it changes sampling behavior.

tests/test_native_sampling.py
```diff
@@ -126,8 +142,80 @@ def test_top_p_boundary_ties_are_masked_positionally(self) -> None:
...
+        mx.eval(masked_mlx)
+        logits_torch = torch.tensor(logits.tolist())
+        k_arg = torch.tensor([k if 0 < k < VOCAB_SIZE else VOCAB_SIZE for k in top_k])
+        masked_ref = apply_top_k_top_p(logits_torch.clone(), k_arg, torch.tensor(top_p))
+
+        kept_mlx = [[v != float("-inf") for v in row] for row in masked_mlx.tolist()]
+        kept_ref = (masked_ref != float("-inf")).tolist()
+        assert kept_mlx == kept_ref
+
+    def test_partitioned_top_k_keeps_ties_at_the_kth_value(self) -> None:
```
LxYuan0420: This test misses the bug because the second row increases `k_max` to 4. Please test one `top_k=2` row against vLLM; this code keeps 2 tokens while upstream keeps 4.

vllm_metal/v1/sampling_batch.py
```diff
@@ -266,35 +259,72 @@ def _top_k_top_p_masked_logits(
         top_p: float,
         min_p: float = 0.0,
     ) -> mx.array:
-        """Mask temperature-scaled logits to the native candidate set."""
+        """Mask temperature-scaled logits to the native candidate set.
+
+        Batch-wide scalar form of :meth:`_sorted_candidate_logits`, scattered
```
LxYuan0420: This helper is now used only by tests. Please remove it and test the real native sampling path, or use it in production. We should not keep production code only for tests.

vllm_metal/v1/sampling_batch.py
Lavmee: Fixed in b174794. Greedy rows now take `mlx_greedy_tokens` (argmax) and are spliced over the drawn tokens with `mx.where`; they still enter the draw as a top-1 set only so every row of the batch has a drawable distribution. Test: a greedy row whose maximum is tied resolves to the argmax over 50 draws, and it failed before the change.

vllm_metal/v1/sampling_batch.py
Lavmee: Removed in b174794, the whole vocabulary is sorted again. Candidate selection at 128 x 262144 on an M5 Pro measures 33.3 ms against 25.6 ms for the partition, so the mixed-batch win stands on its own; I have dropped the partition claim from the PR description.

tests/test_native_sampling.py
Lavmee: You are right, the second row raised k_max to 4 and hid it. b174794 adds `test_single_row_top_k_keeps_ties_at_the_kth_value`: one row, `top_k=2`, logits `[3, 2, 2, 2, 1, 0]`, compared against `apply_top_k_top_p`. It failed with `{0, 1}` against vLLM's `{0, 1, 2, 3}` before the partition came out. The old test is kept for the mixed-batch case and renamed.

vllm_metal/v1/sampling_batch.py
Lavmee: Agreed, removed in b174794. The two mask-parity tests now call `_sorted_candidate_logits` through a test-local helper that only scatters its output back into vocab order, so they exercise the production path.

Lavmee:
@ricky-chaoju thank you, that is exactly the hole and your reproduction matches mine: a greedy row masked by value keeps both tied tokens and the draw picks between them. Fixed in b174794 by taking the argmax for greedy rows and splicing it over the drawn tokens, so a greedy request is deterministic whatever its neighbours ask for. The regression test uses `[[1, 5, 5, 0], [1, 2, 3, 4]]` with a `temperature=0` row next to a `top_p=0.9` one and asserts a single token over 50 keys; before the change it drew both. The partition shortcut is out too, for a related reason on the top-k boundary.

LxYuan0420 (approved):

ricky-chaoju (approved):
LGTM Thanks

PR #788, opened 2026-09-19
[Bugfix] Fix MiniMax full-projection Q/K norms

### LINKED ISSUE #787: [Bug] MiniMax paged attention applies full-projection Q/K norms after splitting heads
## Description

MiniMax attention in the pinned mlx-lm revision applies Q/K RMSNorm to the full projection before reshaping into heads. The paged SDPA path has no contract for `mlx_lm.models.minimax`, so it uses the default per-head placement and applies a `num_attention_heads * head_dim` norm to a `head_dim` tensor.

A minimal configuration with 4 query heads and `head_dim=4` therefore applies a 16-element Q norm after reshaping and fails with:

```text
ValueError: [rms_norm] (*weight) must have the same size as the last dimension of x but has 16 elements.
```

The same mismatch applies to K, whose norm width is `num_key_value_heads * head_dim`.

## Expected behavior

Register `mlx_lm.models.minimax` with the existing `BEFORE_HEAD_SPLIT` attention contract so paged Q/K preparation matches the native MiniMax ordering:

1. Project Q/K.
2. Apply the full-projection Q/K norms.
3. Split into heads.
4. Apply RoPE.

## Validation scope

A small numeric regression can instantiate `MiniMaxAttention`, exercise registration through `attention_contract_for()`, and compare the paged-preparation Q/K tensors exactly with the native ordering. Removing the registration reproduces the shape failure above.

The available MLX MiniMax checkpoints are too large for end-to-end validation on a 24 GiB development machine, so supported-model documentation should remain unchanged until parity is confirmed on sufficiently large hardware.

HanningLin:
Fixes #787

## Summary

MiniMax applies Q/K RMSNorm to each full projection before splitting it into heads. The paged SDPA path had no contract for `mlx_lm.models.minimax`, so the default per-head placement applied a `num_attention_heads * head_dim` norm to a `head_dim` tensor and failed with an RMSNorm width mismatch.

- Register `mlx_lm.models.minimax` with the existing `BEFORE_HEAD_SPLIT` placement.
- Add a numeric regression that exercises registration through `attention_contract_for()` and compares the prepared Q/K tensors exactly with MiniMax native ordering.
- Keep the supported-model table unchanged because a real MiniMax checkpoint was not validated end to end.

This reuses the existing preparation path; there is no attention-kernel or configuration change.

## Validation

```text
2195 passed, 14 skipped, 37 deselected

ruff check .
All checks passed

ruff format --check .
294 files already formatted

mypy vllm_metal
Success: no issues found in 140 source files

git diff --check
passed
```

`tools/check_parity.py` was also run through paged HTTP serving with a deterministic one-layer MiniMax M2 checkpoint using the official tokenizer. The fixture uses hidden size 128, 2 Q heads, 1 KV head, and 64-wide heads, producing full-projection Q/K norm widths of 128 and 64:

```text
python tools/check_parity.py --model /tmp/minimax-m2-tiny-parity \
  --top-k 5 --batch-size 1 2 --max-tokens 20 \
  --gpu-memory-utilization 0.1 --output-dir /tmp/minimax-m2-parity-results

| Prompts/request | EXACT | TOP_K_MATCH | FAIL | Exit |
|---|---:|---:|---:|---:|
| 1 | 34 | 6 | 0 | 0 |
| 2 | 34 | 6 | 0 | 0 |
```

Removing the registry entry reproduces the original error:

```text
ValueError: [rms_norm] (*weight) must have the same size as the last dimension of x but has 16 elements.
```

Real-checkpoint serving, long-context behavior, and prefix caching remain unverified. The smallest official MLX MiniMax M2 checkpoint is approximately 100.1 GB at 3-bit (the 4-bit checkpoint is 128.7 GB), which exceeds the available 24 GiB test machine and a 64 GB MBP. This PR therefore does not add MiniMax to the supported-model table.


WindChimeRan (commented):
python tools/check_parity.py --model /path/to/checkpoint --top-k 5 --batch-size 1 2 --output-dir parity-results

--- 

Please use the parity tool to check the alignment with mlx_lm reference

HanningLin:
Ran `tools/check_parity.py` against a deterministic one-layer MiniMax M2 checkpoint using the official MiniMax tokenizer and the MLX-LM `minimax` architecture:

```text
python tools/check_parity.py \
  --model /tmp/minimax-m2-tiny-parity \
  --top-k 5 \
  --batch-size 1 2 \
  --max-tokens 20 \
  --gpu-memory-utilization 0.1 \
  --output-dir /tmp/minimax-m2-parity-results
```

```text
| Prompts/request | EXACT | TOP_K_MATCH | FAIL | Exit |
|---|---:|---:|---:|---:|
| 1 | 34 | 6 | 0 | 0 |
| 2 | 34 | 6 | 0 | 0 |
```

The fixture uses 1 layer, hidden size 128, 2 Q heads, 1 KV head, and supported 64-wide heads. Its Q/K RMSNorm widths are therefore 128 and 64 respectively, so this exercises the full-projection norm placement from this PR through actual paged HTTP serving, not only the direct unit helper.

I used a small checkpoint because the smallest official MLX MiniMax M2 checkpoint I found is `mlx-community/MiniMax-M2-3bit` at 100.1 GB of safetensors; the 4-bit checkpoint is 128.7 GB. Neither fits the available 24 GiB machine (nor a 64 GB MBP). Real-weight checkpoint validation remains explicitly unclaimed in the PR description.

WindChimeRan (approved):

PR #790, opened 2026-09-19
[Performance] Enable grouped KV cache for OLMo 3

### LINKED ISSUE #789: Use grouped KV-cache planning for OLMo 3 mixed-window attention
## Problem

OLMo 3 alternates three sliding-window layers with one full-attention layer, but all layers use the same KV-head geometry. `DefaultModelAdapter.build_per_layer_kv_shapes()` therefore returns `None`, and `ModelCachePolicy._uses_deferred_layout()` currently requires non-null per-layer head arrays before enabling vLLM grouped KV-cache planning.

As a result, OLMo 3 falls back to the dense cache path even though its mixed attention windows can use grouped allocation. Sliding layers cannot release blocks outside their window independently of full-attention layers.

## Proposed change

Select deferred/grouped layout from the mixed per-layer window pattern itself, while preserving the existing exclusions for the disabled hybrid manager, TurboQuant, speculative decoding, KV sharing, MLA/hybrid models, and explicit block overrides.

Add coverage using an OLMo 3 configuration with uniform KV geometry to verify:

- 3 sliding groups plus 1 full-attention group per four-layer cycle
- sliding groups release blocks beyond their window while full groups retain context
- `--disable-hybrid-kv-cache-manager` retains the dense fallback
- the allocated grouped cache remains within the reported byte budget

## Validation target

Use `mlx-community/Olmo-3-7B-Instruct-4bit` for short parity, context longer than the 4,096-token sliding window, and repeated long-prompt prefix-cache validation.

HanningLin:
## Summary

- enable vLLM grouped KV-cache planning for mixed full/sliding-window models whose layers have uniform KV-head geometry
- keep the dense cache fallback for the existing incompatible modes and for `--disable-hybrid-kv-cache-manager`
- add focused OLMo 3 regressions for grouped policy/layout, the dense opt-out, and sliding/full block retention
- document the mixed-window behavior and limitations

Fixes #789.

## Why

OLMo 3 uses a repeating three-sliding/one-full attention pattern, but its uniform head geometry means the adapter does not return per-layer head arrays. The previous deferred-layout gate required those arrays, so OLMo 3 never reached vLLM grouped KV-cache planning and instead used dense full-context allocation for every layer.

## Validation

- `ruff check .`
- `ruff format --check .`
- `mypy vllm_metal` (140 source files)
- `pytest -m "not slow" tests/ .github/scripts/test_parity.py -q --tb=short` (2,196 passed, 14 skipped, 37 deselected)
- `pytest tests/test_per_layer_kv_cache.py -q` (22 passed)
- `mlx-community/Olmo-3-7B-Instruct-4bit` at revision `d732c91ae02e90cd5d86e810fbeb9741794b4dd9` on a 24 GB M4 MacBook Pro:
  - short HTTP parity, batch sizes 1 and 2: 57 exact, 23 top-k matches, 0 failures
  - 8,409-token prompt, beyond the 4,096-token sliding window: top-k parity match, 0 failures
  - repeated 8,409-token prompt with prefix caching: identical output and 8,400 cached-token hits
  - runtime layout: 4 groups, 8 physical slots, 2,281 blocks within a 4,783.6 MB KV budget


docs/configuration.md
```diff
@@ -52,3 +52,14 @@ model pairing, and memory considerations.
 The paged KV cache budget follows vLLM's standard `--gpu-memory-utilization`
 flag (`gpu_memory_utilization=` for `LLM()`), a fraction in `(0, 1]`. The
 former `VLLM_METAL_MEMORY_FRACTION` override has been removed.
+
+Models mixing full and sliding-window attention, including OLMo 3 with
```
LxYuan0420, line 56: This is useful but too detailed. Could we shorten it to:

Models with full and sliding-window attention use grouped KV cache, allowing sliding layers to release old blocks. This can improve long-context capacity, but not necessarily generation speed. Use `--disable-hybrid-kv-cache-manager` for dense allocation.

tests/test_per_layer_kv_cache.py
```diff
@@ -46,6 +52,108 @@ def vllm_config_for_kv_grouping() -> VllmConfig:
     return vllm_config
 
 
+@pytest.mark.parametrize("disable_hybrid_manager", [False, True])
```
LxYuan0420, line 55: This covers the right regression but tests several contracts at once. Could we separate the policy and opt-out checks from the block-manager behavior so failures are easier to diagnose?

LxYuan0420 (approved):
Approving since the cache-policy change is correct and well covered. Let’s trim the documentation and split the large test to keep the change clear and maintainable.

docs/configuration.md
HanningLin: Updated in 531b7fa using the suggested concise wording.

tests/test_per_layer_kv_cache.py
HanningLin: Split in 531b7fa into focused policy/layout, dense opt-out, and block-manager lifecycle tests. The cache test file now passes all 22 tests.

HanningLin:
@LxYuan0420 updated. Let me know how that looks

PR #791, opened 2026-09-19
[Refactor KV] Split the hybrid cache backing into per-region Metal buffers

suntp:
Stacked on #778. Includes the #780 commits (track native GDN writes, announce the buffer cap, cross-check declared cache geometry) so the diff is reviewable against `codex/shared-cache-integration`; once #780 merges, this diff shrinks to the per-region commits. Follow-up to rishabhsinha17's review on #778.

## Problem

The hybrid KV+state backing is one Metal allocation, so the planner caps the requested budget at `max_buffer_length` (~55% of RAM on an M1 Pro 16GB). On hosts where the request exceeds that limit — small models at high `--gpu-memory-utilization`, or 128GB+ hosts serving long context — real capacity silently shrinks by a third or more.

## Change

Layer-compact layouts (LBNHC) give every layer a contiguous page span, so `KVCacheStorage` now allocates **one zeroed Metal buffer per layer page span**; each sits far below the device buffer limit and the planner drops the cap entirely. Block-outermost layouts (BLNHC) keep the single-buffer path with the (now loud) cap. Concretely:

- Region strides come from upstream's `compute_layout_strides` for a single-layer layout — the `KVCacheTensor` contract stays authoritative; only the backing granularity is ours.
- Region dedupe is global by virtual address `(offset, page_stride, page_size)`, preserving overlaid-group semantics (shared state pools) exactly as the single buffer did.
- Per-region anchors lazily re-wrap the write-dependency chain on read after writes land (one node per stale region per read), so flush-then-zero, cyclic CoW, and decode reads keep their dependency edges. Kernels are untouched — all indexing was already per-view strides.
- The planner skips the cap iff the resolved layout is layer-compact.

## Tests

Region dedupe (shared state pools), cross-region CoW cycles, engine block-count changes, both planner modes (LBNHC uncapped / BLNHC capped), and the full hybrid e2e parity suites — granite/nemotron forwards still match the mlx-lm reference under per-region storage.

## Long-context verification

Serial single-request server benchmark (`vllm serve` + `vllm bench serve`, one fresh server per point, concurrency 1, Qwen3.5-0.8B hybrid, `--gpu-memory-utilization 0.85`, exact-length random dataset) across four branches — this PR / #780 / #778 / main — at input lengths 16K/32K/64K/128K/256K:

| input len | TTFT s: PR / 780 / 778 / main | decode mean ITL ms: PR / 780 / 778 / main |
|---|---|---|
| 16K | 2.00 / 2.01 / 1.99 / 1.93 | 14.1 / 13.7 / 14.0 / 13.5 |
| 32K | 4.97 / 4.95 / 4.97 / 4.98 | 18.8 / 18.2 / 18.6 / 18.0 |
| 64K | 14.82 / 14.78 / 14.78 / 14.88 | 27.2 / 26.6 / 27.3 / 26.8 |
| 128K | 50.36 / 50.47 / 50.36 / 50.67 | 43.7 / 42.8 / 43.6 / 42.8 |
| 256K | 177.9 / 177.6 / 177.6 / 178.7 | 73.3 / 73.6 / 73.5 / 74.0 |

- **Correctness**: a needle-retrieval probe (two codes embedded at ~10% / ~90% of the prompt, greedy decode) passes at every length on every branch (20/20), including 256K.
- **Prefill**: parity at every length (<0.5% spread); the shared prefill-throughput curve (8.2K → 1.44K tok/s from 16K → 256K) is identical on all four branches, so per-region storage introduces no long-context-specific pathology.
- **Decode**: within ~2–3% of #780 at 16K–128K and at parity at 256K — consistent with the interleaved in-process A/B measured after the lazy per-region anchor rebuild (−2.1%), which is the residual cost of view indirection + anchor maintenance. Single-request ITL carries ±2–3% run-to-run noise (#778 shows a similar small offset vs #780), so treat the low end of that range as noise. Eliminating the remainder would require the native `cache_view` helper to carry dependency inputs natively.


WindChimeRan:
close as absorbed by #778. Thanks for providing the testable fix!!

PR #792, opened 2026-09-19
[Bugfix] Apply the lazy GDN pipeline gate only to GDN models

LxYuan0420:
This PR is:

- To keep the decode pipeline enabled for Nemotron-H and Granite.
- To disable it only for GDN models when lazy GDN kernels are turned off.
- To stop a GDN-only setting from affecting other hybrid models.

`VLLM_METAL_GDN_LAZY_KERNELS=0` currently disables the decode pipeline for all hybrid models. This is too broad because the setting only changes how GDN models run.

This change checks the model family and disables the pipeline only when the family is `gdn`.


ricky-chaoju (approved):
LGTM

PR #793, opened 2026-09-19
Complete removal of VLLM_METAL_MEMORY_FRACTION from docs, tools, and tests

LxYuan0420:
This PR is:

- To replace ignored `VLLM_METAL_MEMORY_FRACTION` usage with vLLM's `gpu_memory_utilization`.
- To preserve the intended memory values in tests, tools, and examples.
- To remove obsolete setup without adding another memory configuration source.

PR #756 removed `VLLM_METAL_MEMORY_FRACTION` from the runtime, but several docs, tools, and tests still used it. Those values were ignored, so runs could use vLLM's default memory utilization instead.

CLI examples and subprocess tools now use `--gpu-memory-utilization`. Python entrypoints now pass `gpu_memory_utilization` to `LLM()`. Examples that used `auto` now rely directly on vLLM's default.

The only remaining reference is the historical removal note in `docs/configuration.md`.


WindChimeRan (approved):
Thanks for the great work!!!

PR #795, opened 2026-09-19
[Perf] Memoize per-layer list->array context conversions

begcdn:
## Summary

Follow-up on the per-layer block-table conversion cost observed in #482 ("~5 ms/step … small hoist, can PR it" — never tracked). Several attention/state wrappers re-converted the same per-forward Python metadata into MLX arrays **on every layer** of a step, although the values are fixed for the duration of the forward pass.

Converted sites (all previously per-layer):

| Site | Arrays |
|---|---|
| `impls/mla.py` `_kernel_fast_path_single_pass` | `context_lens`, `cu_seqlens` (block-table padding stays per-call) |
| `impls/mla.py` per-request SDPA fallback | per-request `block_tables` list |
| `impls/linear.py` `_run_recurrent_fallback` | `cu_seqlens`, `slot_ids` |
| `impls/gdn_lazy.py` `try_conv_prefill` / `try_recurrent_prefill` | `cu_seqlens`, `slot_ids` (now accepted as optional precomputed copies) |

The sdpa Metal-kernel path already memoizes through `_kernel_metadata` and is untouched.

## Change

- `PagedAttentionContext` gains an `array_memo` dict with the same per-forward lifetime rule documented for `kernel_metadata_cache`: entries are built on the first layer that needs them and die with the forward, so they cannot go stale.
- Memo helpers with explicit key semantics:
  - `memoized_slot_mapping` / `memoized_context_lens` — fixed context lists;
  - `memoized_cu_seqlens` / `memoized_block_table_arrays` — keyed by list identity (attention and GDN wrappers carry their own equivalent boundary lists);
  - `memoized_slot_ids` — keyed by value (the state manager derives fresh but equal lists per layer).
- `_GDNForwardState` carries optional memoized arrays prepared next to the ctx read; the lazy prefill request/paths accept optional precomputed arrays and fall back to converting the lists themselves, so test fakes with `(*_)` positional signatures keep working.

## Performance

Isolated host-side microbenchmark on M4 (batch 64, 512-entry block tables, i.e. ~8k ctx @ block 16):

- one `block_tables` conversion: ~281 us; `cu_seqlens[65]` ~0.9 us; `slot_ids[64]` ~0.8 us;
- removed per decode step at 16/48/64 layers: **~4.5 / ~13.6 / ~18.1 ms** of Python-side list→array work.

Directionally consistent with the ~5 ms/step block-table share measured in #482. This is a host-side removal, not an e2e serving claim; happy to run `vllm bench serve` on a batch/long-context workload if reviewers want the before/after table.

## Validation

- New `tests/attention/test_context_array_memo.py` (7 tests): identity vs value keying, dtypes, per-context isolation, and a `prepare_grouped`-fed context. All pass.
- `ruff check` + `ruff format --check` clean; `mypy vllm_metal` clean (140 files).
- Targeted suites (`test_gdn_lazy_kernels`, `test_gdn_lazy_wrapper`, `test_mla_paged_backend`, `test_mla_materialized_prefill`, `tests/attention/`): failure set is byte-identical to a stashed-baseline run of the same selection on this machine. The 89 baseline failures here are environmental (local tree lacks the full-Xcode `.metallib` artifacts that `scripts/test.sh` builds via `install.sh`; they fail identically on unmodified `main` in this venv) — CI on macos-15-arm64 is the authoritative run.

**AI-assistance disclosure:** this change was prepared with an AI coding agent (Codebuff); I reviewed the diff and the referenced code paths and take responsibility for it.

Closes the block-table item noted in #482.


vllm_metal/attention/context.py
```diff
@@ -98,6 +99,13 @@ class PagedAttentionContext:
     kernel_metadata_cache: dict[tuple[int | None, int], Any] = field(
         default_factory=dict
     )
+    # Per-forward memo for shared kernel-format conversions (slot mappings,
+    # cu_seqlens, context lens, block tables, state slot ids).  Same lifetime
+    # rule: entries are built on the first layer that needs them and die with
+    # the forward, so they can never go stale.  Helpers:
+    # ``memoized_slot_mapping`` / ``memoized_cu_seqlens`` / ``memoized_slot_ids``
+    # / ``memoized_context_lens`` / ``memoized_block_table_arrays`` below.
+    array_memo: dict[tuple[Any, ...], Any] = field(default_factory=dict)
```
LxYuan0420, line 108: This generic `dict[Any, Any]` cache with string, identity, and value keys is not an acceptable contract for shared hot-path state. Remove it and use a small typed per-forward cache for the measured MLA data only.

vllm_metal/attention/impls/linear.py
```diff
@@ -172,6 +182,8 @@ def _prepare_gdn_forward_state(
             total_tokens=x.shape[1],
             slot_ids=slot_ids,
             num_decode_requests=ctx.num_decode_requests,
+            cu_seqlens_arr=memoized_cu_seqlens(ctx, cu_seqlens),
```
LxYuan0420, line 185: Remove the GDN changes from this PR. The reported conversions save only about 0.1 ms across 64 layers, while this adds state fields, kernel API branches, and unnecessary work on pure decode.

vllm_metal/attention/context.py
```diff
@@ -112,6 +120,92 @@ def clear_context() -> None:
...
 
+# ---------------------------------------------------------------------------
+# Memoized kernel-format conversions
+# ---------------------------------------------------------------------------
+
+# Memo key for the plain (group-zero) conversions below.
+_DEFAULT_MEMO_KEY = ("default",)
+
+
+def memoized_slot_mapping(ctx: PagedAttentionContext) -> mx.array:
```
LxYuan0420, line 131: This still converts `ctx.slot_mapping` on every MLA layer, while the new memo helper is unused in production. The narrowed MLA change must reuse this array or remove the dead helper and limit the claim to block tables.

tests/attention/test_context_array_memo.py
```diff
@@ -0,0 +1,118 @@
...
+    first_ctx = PagedAttentionContext(slot_mapping=[1])
+    second_ctx = PagedAttentionContext(slot_mapping=[1])
+    first = memoized_slot_mapping(first_ctx)
+    second = memoized_slot_mapping(second_ctx)
+    assert first is not second
+    # Both contexts end up with equal-valued arrays.
+    assert first.tolist() == second.tolist()
+
+
+def test_prepare_grouped_context_feeds_memos():
```
LxYuan0420, line 107: This suite is not mergeable as written. It leaks the thread-local context and tests helpers instead of the modified runtime paths. Replace it with an MLA call-path test that cleans up context and proves two layers share one conversion.

PR #798, opened 2026-09-20
Feat/expert parallelism

bfp16:


WindChimeRan (commented):
What's the motivation of EP in apple silicon?

WindChimeRan:
close as not planned

PR #801, opened 2026-09-21
Repack Q5_K blocks into MLX affine tensors

ricky-chaoju:
Q5_K rides the Q4_K base: same header, with the 6-bit sub-scale decode now a shared helper, plus the `qh` high-bit plane merged into the nibble codes, still fp32 scales. The packer gets a second path for bits that do not divide 32, since 5-bit codes straddle word boundaries; the 4-bit fast path from #773 is untouched.

Dequantize is bit-identical to `gguf-py` for both qtypes, and the K-quant test block is parametrized over Q4_K and Q5_K instead of growing copies. The loader stays closed, pinned for both. Full suite 2208 passed, ruff and mypy clean.

On the memory point from #773: repacking a full 4096x11008 Q5_K tensor end to end adds about 220 MiB of peak RSS over the process baseline (265 to 487 MiB measured), in the same range as the Q4_K path. Real files still come with the routing PR.

PR 2  #761


vllm_metal/gguf/mlx_native.py
```diff
@@ -267,12 +271,35 @@ def _parse_q4_k(blocks: np.ndarray) -> tuple[np.ndarray, np.ndarray, np.ndarray]
...
+        A 176-byte superblock is the Q4_K header (``d``, ``dmin``, the 6-bit
+        sub-scales/mins) plus 32 high-bit bytes ``qh`` and the same 128
+        nibble bytes: element ``e`` of group ``g`` takes bit ``g`` of
+        ``qh[e]`` as its fifth bit.
         """
+        scales, biases = GGUFMLXQuantizedTensor._parse_kquant_scale_mins(blocks)
+        codes = GGUFMLXQuantizedTensor._split_nibble_chunks(blocks[:, 48:176])
+        qh = blocks[:, 16:48]
+        group_bits = np.arange(8, dtype=np.uint8)[None, :, None]
+        # Stays uint8 end to end: the in-place OR casts per buffer chunk, so no
```
LxYuan0420: Could we make the subject explicit? `codes` is already uint32; the high-bit plane stays uint8 and is merged in place, avoiding another full-size uint32 temporary.

LxYuan0420 (approved):
One nit on the comment wording.

PR #802, opened 2026-09-21
Guard unsupported MPS host-cache cleanup

TheVerdman:
Calling `torch.accelerator.empty_host_cache()` on MPS with PyTorch 2.13.0 can crash vLLM during shutdown.

This skips the unsupported cleanup call on MPS and preserves it for other accelerators. The guard is safe to apply more than once and handles PyTorch versions that do not expose the function.

All 46 compatibility tests pass, with two optional Ray cases and one opt-in FP8 case skipped. Coverage includes MPS, CUDA, no accelerator, and repeated patch application. Ruff and mypy also pass.


WindChimeRan (commented):
> crash vLLM during shutdown


I’m not sure whether this is a serious issue. Should the root cause be fixed upstream in PyTorch?

I found these: 
* https://github.com/pytorch/pytorch/issues/197561
* https://github.com/pytorch/pytorch/pull/197593

Maybe we should wait for the next PyTorch release?

PR #803, opened 2026-09-21
Honor explicit Metal cache and wired-memory limits

TheVerdman:
While validating Nano VL V2 for #743, I found that Metal accepted `--kv-cache-memory-bytes` but still sized the KV pool automatically. This makes an explicit value override `--gpu-memory-utilization`, matching vLLM. Automatic sizing stays the same when the option is unset, and cache allocations are rounded down to whole blocks.

An explicit value budgets the paged cache. Model weights, activations, and state or draft scratch space outside that pool still need additional memory.

I've also added `VLLM_METAL_DISABLE_WIRED_LIMIT=1` so a caller can preserve MLX's existing wired-memory limit during startup. The default behavior stays the same. Both settings are documented with their relationship to `--gpu-memory-utilization`.

All 64 worker, utility, and cache-layout tests pass, covering explicit values above and below the automatic budget, low-utilization cases, block rounding, hybrid-state accounting, startup diagnostics, dense and grouped caches, and the wired-limit opt-out. Tests use `VLLM_CONFIGURE_LOGGING=0` with INFO logging for pytest log capture. Ruff and mypy also pass.


WindChimeRan (commented):
Configuration | Upstream vLLM | This PR
-- | -- | --
Explicit bytes unset | Automatically calculate KV budget | Automatically calculate KV budget
Explicit bytes set | Override utilization-based KV sizing | Take the smaller of explicit bytes and the automatic budget

---

Should we preserve upstream semantics?

TheVerdman:
@WindChimeRan Yes, I agree we should preserve upstream semantics. I used the smaller of the two budgets to retain Metal’s existing memory constraint, but that changes what `--kv-cache-memory-bytes` means compared with vLLM.
I’ll update it so an explicit value overrides utilization-based KV sizing, while leaving automatic sizing unchanged when it’s unset. I’ll also update the documentation and tests to cover explicit values both below and above the automatic budget. Thanks for catching this.

WindChimeRan (commented):
The KV cache fix has a clear justification

---

Questions
*  why do we need VLLM_METAL_DISABLE_WIRED_LIMIT? Why in this PR? (if it's important, maybe we should make a separated and dedicated PR for it? so we can run some exp to see how important is it and maybe set some default values)
* vllm use --kv-cache-memory-bytes and --gpu-memory-utilization as manual mode and auto mode separately. and manual mode would skip the profiling run. Should we shortcut the profilling run as well or is it overkill?

PR #804, opened 2026-09-22
[Cleanup] Remove unused Metal shader code

WindChimeRan:
Remove three unused shaders inherited from the vendored kernel set (#136, #156), along with commented BF16 code and unused FP8 helpers. The three shader files were never included in the native build or dispatch paths.

- **`copy_blocks.metal`:** shared-cache copies snapshot source rows with MLX indexing and write through `gdn_state_scatter`; the remaining count-based caches use MLX indexed writes.
- **`gather_kv_cache.metal`:** paged attention reads through block tables directly; MLX fallback paths gather context with array indexing.
- **`kv_scale_update.metal`:** scaled-FP8 KV-cache dispatch is disabled. TurboQuant handles the supported quantized-cache path with fused encoding, scale calculation, and paged scatter.
- **Commented BF16 implementation:** delete the inactive alternative vector implementation; keep the active BF16 code unchanged.
- **Unused FP8 helpers:** delete the unreferenced E5M2 decoder, encoder wrapper, and `from_bits`; retain E4M3 and its shared encoding template.

Update the Metal README's build paths, source ordering, and active shader inventory.


ricky-chaoju (approved):
LGTM Thanks

PR #805, opened 2026-09-22
[Cleanup] Remove unused Rust packaging from vllm-metal

WindChimeRan:
- vllm-metal no longer uses Rust, and we do not expect to use it soon. Remove the unused Cargo/PyO3 scaffold and references; Git history preserves them if needed later.
- Replace Maturin with setuptools for wheel packaging. The prebuilt `_paged_ops` extension and Metal libraries still ship in the wheel.


ricky-chaoju (approved):
LGTM

PR #806, opened 2026-09-23
Bump vLLM to 0.30.0

ricky-chaoju:
0.30.0 is a quiet release for Metal. The only fallout is in tests: `get_kv_cache_config_from_groups` and `get_kv_cache_configs` now read `vllm_config.attention_config.hisparse_config` directly, so the four `SimpleNamespace` configs that go through them get a real `AttentionConfig()`, the same way #744 added `speculative_config`. `huggingface_hub` moves to 1.31.0 and `pyproject` to 0.30.0 per `RELEASE.md`. The fastapi `<0.137` cap stays because 0.30.0 still pins it, and nothing else discharges.

0.30.0 also forces Model Runner V2 for `--watermark-config` and HiSparse. On Metal both stop at upstream's config validation with the Triton error, so I left the V1 guard as it is.

`MetalPlatform` resolves, lint passes, and 2237 non-slow tests pass. `tools/check_parity.py` with the CI flags gives the same per-prompt results on 0.29.0 and 0.30.0 (Qwen3-0.6B 62 exact and 18 top-k, Qwen3.5-0.8B 61 and 19). Both spec decode e2e gates pass 7/7 and the GGUF serve smoke passes. The slow suite's failures are the known in-process memory issue plus the hybrid APC cells that fail the same way on 0.29.0.

`vllm bench serve`, Qwen3-0.6B, sonnet, 100 prompts, `--request-rate inf --temperature 0 --ignore-eos --seed 0`, three alternating sessions per arm, all at 100 requests, 54256 input and 15000 generated tokens. No regression.

| Metric | vLLM 0.29.0 | vLLM 0.30.0 | Delta |
|---|---:|---:|---:|
| Output tok/s | 1187.02 | 1186.42 | -0.1% |
| Total tok/s | 5480.53 | 5477.79 | -0.1% |
| Mean TTFT (ms) | 4289.84 | 4288.27 | -0.0% |
| P99 TTFT (ms) | 5232.15 | 5236.06 | +0.1% |
| Mean TPOT (ms) | 55.77 | 55.82 | +0.1% |
| P99 TPOT (ms) | 63.89 | 63.96 | +0.1% |
| Duration (s) | 12.64 | 12.64 | +0.1% |


WindChimeRan (approved):
Thanks!

PR #808, opened 2026-09-23
Add Q6_K raw-block tensor with Metal kernels

ricky-chaoju:
Q6_K has 16-element sub-groups and MLX has no `group_size=16` kernels, so there is nothing to repack into. This keeps the raw 210-byte superblocks in memory instead and computes on them with two `mx.fast.metal_kernel` programs, the route the spike on #761 settled (the fused qmv runs at 0.85x of MLX's affine qmv at decode shape).

For matmul, the fused kernel handles batches up to 4; above that it dequantizes once into a transient dense copy and does a normal GEMM. I measured the crossover at 4096x4096: qmv wins through B=4, the GEMM path from B=8. I did not sweep B=5..7, so 4 is the conservative pick. The dequantize kernel writes the GEMM dtype directly, and the transient tops out at 126 MiB for a 4096x11008 layer at f16 (`mx.get_peak_memory`). Embedding gathers the packed rows first and only dequantizes those; the census on #761 showed Q4_K_M files put the token embedding in Q6_K, so this path matters.

PR 3 #761


tests/test_gguf_raw_block.py
```diff
@@ -0,0 +1,255 @@
...
+
+    assert calls == {"qmv": 0, "gemm": 1}
+    ref_max = float(mx.max(mx.abs(expected)))
+    np.testing.assert_allclose(
+        np.array(out), np.array(expected), rtol=0, atol=ref_max * 4e-5
+    )
+
+
+@pytest.mark.parametrize("dtype", [mx.float16, mx.bfloat16])
+def test_matmul_output_dtype_follows_x(dtype):
```
LxYuan0420: Nit: these cases only check shape and dtype. Could we also compare float16 and bfloat16 against the dense oracle at B=4 and B=5 to cover both the QMV and GEMM paths?

LxYuan0420 (approved):

